PR #124, opened 2026-04-25
fix(e2e): auth fixtures via service-role fallback + selector + serial

sakibsadmanshajib:
## Summary
- e2e-fixtures edge fn auth check now accepts EITHER `E2E_FIXTURE_SECRET` OR the auto-injected `SUPABASE_SERVICE_ROLE_KEY`, so callers don't need a manual secret-set step
- Sign-in helpers in profile-completion + auth-shell specs swap `getByLabel("Email")` (broken by redesigned label \`*\` suffix) for `#email`/`#password` id selectors
- `waitForURL("**/console**")` glob was matching the staging hostname `console-hive.scubed.co` and masking sign-in failures — replaced with `(url) => url.pathname.startsWith("/console")`
- Pin Playwright `workers: 1` globally + mark fixture-resetting specs `serial` (fixture reset mutates shared Supabase state; parallel workers race on it and flap sessions mid-test)
- Add `tests/e2e/_probe/staging-flows.spec.ts` covering sign-in, sign-up, unverified-error, /console/setup, /console/api-keys, health endpoints
- gitignore `/scripts/admin/` for local-only operator scripts

## Test plan
- [x] Local Playwright vs staging: 21 passed, 0 failed, 3 skipped (1.1m)
- [x] Live SDK call /v1/chat/completions on staging returns 200
- [x] Edge function v4 deployed; `X-E2E-Secret` accepts both rotated `E2E_FIXTURE_SECRET` and service-role
- [ ] CI green
- [ ] Workers deploy step succeeds

### DIFF
--- a/apps/web-console/tests/e2e/_probe/staging-flows.spec.ts
+++ b/apps/web-console/tests/e2e/_probe/staging-flows.spec.ts
@@ -1,0 +1,84 @@
+import { expect, test } from "@playwright/test";
+
+const BASE = "https://console-hive.scubed.co";
+const API = "https://api-hive.scubed.co";
+
+const QA_TESTER_EMAIL = "qa-tester@hive.test";
+const QA_TESTER_PASSWORD = "HiveQA-StressTest-2026!";
+
+test("sign-in lands on /console with credit balance + workspace banner", async ({ page }) => {
+  await page.goto(`${BASE}/auth/sign-in`);
+  await page.locator("#email").fill(QA_TESTER_EMAIL);
+  await page.locator("#password").fill(QA_TESTER_PASSWORD);
+  await page.click('button[type="submit"]');
+  await page.waitForURL("**/console**", { timeout: 25000 });
+  await expect(
+    page.getByRole("heading", { name: "Available credits" }),
+  ).toBeVisible();
+  await expect(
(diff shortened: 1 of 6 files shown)


apps/web-console/tests/e2e/_probe/staging-flows.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { expect, test } from "@playwright/test";
+
+const BASE = "https://console-hive.scubed.co";
```
sakibsadmanshajib: change them to env variables. pulled from .env
BASE, API, username, password, etc., because they are env-specific.

sakibsadmanshajib:
Both review comments addressed in #125:
- Codex P1 (waitForURL hostname-glob mask) — fixed by switching to a pathname predicate.
- Your env-vars feedback — BASE / API / QA creds now pulled from env (`PLAYWRIGHT_BASE_URL`, `HIVE_EDGE_API_URL`, `HIVE_CONTROL_PLANE_URL`, `HIVE_QA_TESTER_*`, `HIVE_QA_UNVERIFIED_*`), with defaults documented in `.env.example`. Probe skips when creds aren't set.

PR #162, opened 2026-06-11
fix: reject tools, tool_choice, response_format on /v1/chat/completions

### LINKED ISSUE #118: [P1] Tools/function calls and response_format silently dropped at /v1/chat/completions
## Severity
**High.** Surfaces in OpenAI compliance audit (C-1).

## Problem
`/v1/chat/completions` accepts and ignores `tools`, `tool_choice`, `response_format` (and likely `parallel_tool_calls`). OpenAI SDKs (Python/Node/Go) send these and expect tool calls in the response. Hive returns a plain content message with no `tool_calls` field — SDK consumers think the model "didn't use the tool" and behavior diverges silently.

For agentic / chat-app / CLI-coding-agent use cases (the documented v1.0 launch scope per .planning/STATE.md) tool calls are mandatory.

## Fix
- Short-term: explicit reject (`400 unsupported_parameter`) for `tools`/`tool_choice`/`response_format` until provider routing supports them. Document on support-matrix.
- Medium-term: route tool-call-bearing requests only to providers that support them; OpenRouter forwards tool calls for many backing models.

## Acceptance
- SDK call with `tools=[...]` returns 400 with provider-blind error message.
- After provider support added: tool call round-trip works against `gpt-4o-mini` via OpenRouter.

sakibsadmanshajib:
## Summary

- `tools`, `tool_choice`, `response_format`, `parallel_tool_calls`, `functions`, and `function_call` were silently forwarded to LiteLLM on `/v1/chat/completions`
- Providers that lack tool-call support ignore these fields and return a plain content response, so SDK consumers saw no `tool_calls` in the reply and assumed the model did not invoke the tool
- Short-term fix: return `400 invalid_request_error` with code `unsupported_parameter` for any of these fields, covering both streaming and non-streaming paths

## Root cause

`chat_completions.go:handleChatCompletions` reads the raw request body and passes it directly to `executeSync`/`executeStreaming`, which in turn forwards it to LiteLLM via `litellm_client.go:rewriteModel`. The `rewriteModel` function only rewrites the `model` field and preserves all other fields, so tools-bearing requests reached the provider. When the provider has no tool-call capability it silently drops the tools and returns a normal text completion, causing silent behavioural divergence.

Relevant files:
- Root cause: `apps/edge-api/internal/inference/chat_completions.go` (no parameter guard)
- Fix: `rejectUnsupportedChatParams` added before the streaming/sync branch
- Supporting: `ParallelToolCalls *bool` field added to `ChatCompletionRequest` in `types.go`

## Test plan

- [x] `TestChatCompletions_ToolsRejected`: each unsupported field (tools, tool_choice, response_format, parallel_tool_calls, functions, function_call) returns 400 with code `unsupported_parameter` and type `invalid_request_error`
- [x] Provider-blind check: error messages contain no provider names (OpenAI, LiteLLM, OpenRouter, Groq)
- [x] `TestChatCompletions_NoToolsPassesThrough`: plain request without tools is not blocked; reaches auth layer (401)
- [x] `TestChatCompletions_StreamWithToolsRejected`: streaming requests with tools are rejected with same 400 before reaching the streaming path
- [ ] Verify CI Go tests pass on this PR

Closes #118

### DIFF
--- a/apps/edge-api/internal/inference/chat_completions.go
+++ b/apps/edge-api/internal/inference/chat_completions.go
@@ -2,6 +2,7 @@ package inference
 
 import (
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 )
@@ -26,6 +27,16 @@ func handleChatCompletions(o *Orchestrator, w http.ResponseWriter, r *http.Reque
 	}
 	if len(req.Messages) == 0 {
 		writeMissingFieldError(w, "messages")
+		return
+	}
+
+	// Reject requests that carry tool-calling or structured-output fields that
+	// are not yet supported end-to-end (issue #118). Silently forwarding them
+	// causes the provider to ignore the fields and return a plain content
+	// response, which misleads SDK consumers into thinking the model did not
+	// use the tool. Return an explicit 400 until provider-routing support for
+	// these parameters is wired in a future phase.
+	if err := rejectUnsupportedChatParams(w, &req); err != nil {
 		return
 	}
 
@@ -63,6 +74,41 @@ func normalizeChatCompletion(respBody []byte, aliasID string) ([]byte, *UsageRes
 	}
 
 	return normalized, resp.Usage, nil
+}
+
+// rejectUnsupportedChatParams checks for fields that are parsed and present in
+// ChatCompletionRequest but not yet supported end-to-end. It writes a 400
+// response and returns a non-nil sentinel error when any unsupported parameter
+// is present, so the caller can return immediately.
+func rejectUnsupportedChatParams(w http.ResponseWriter, req *ChatCompletionRequest) error {
+	var param string
+	switch {
+	case len(req.Tools) > 0:
+		param = "tools"
+	case len(req.ToolChoice) > 0:
+		param = "tool_choice"
+	case len(req.ResponseFormat) > 0:
+		param = "response_format"
+	case len(req.Functions) > 0:
+		param = "functions"
+	case len(req.FunctionCall) > 0:
+		param = "function_call"
+	}
+
+	if param == "" && req.ParallelToolCalls != nil {
+		param = "parallel_tool_calls"
+	}
+
+	if param == "" {
+		return nil
+	}
+
+	code := "unsupported_parameter"
+	writeError(w, http.StatusBadRequest, "invalid_request_error",
+		"The '"+param+"' parameter is not yet supported. Tool calling and structured output will be available in a future release.",
+		&code,
+	)
+	return errors.New("unsupported parameter: " + param)
 }
 
 // writeError is a local helper to write OpenAI-style errors.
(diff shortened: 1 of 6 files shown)


sakibsadmanshajib (commented):
**Go Code Review: PR #162 fix/118-tools-passthrough**

**VERDICT: REQUEST CHANGES** (one HIGH issue + two medium issues)

---

**HIGH: tools=[] (empty array) passes through silently**

apps/edge-api/internal/inference/chat_completions.go

The guard fires only when `len(req.Tools) > 0`. Per OpenAI semantics an empty array `[]` is a valid, explicit declaration of tool intent and should be treated the same as a non-empty array. More critically, `tool_choice` can appear alone with a non-empty string value (e.g. `"auto"`, `"none"`, `"required"`) without a tools array. The current switch evaluates `len(req.ToolChoice) > 0`, which is correct for the non-empty-string case, but the combination `tools=[] + tool_choice="auto"` will hit the `tools` branch first, find length 0, fall through to `tool_choice` which IS non-empty, and correctly reject. So that specific combo is fine. However `tools=[]` alone will silently pass through and reach billing + dispatch. That is the silent-passthrough bug the PR is meant to prevent. Recommend treating `tools` as present when the raw JSON field is non-null, not just when it is a non-empty array. A simple fix: decode into `json.RawMessage` and check `len(field) > 0 && string(field) != "null"` (the same pattern already used in responses.go for `req.Tools`).

**MEDIUM: support-matrix.json not updated**

packages/openai-contract/matrix/support-matrix.json (line 98-103)

The entry for `POST /v1/chat/completions` has `"notes": "Chat completion inference"` with no mention that tools, tool_choice, response_format, parallel_tool_calls, functions, and function_call are explicitly unsupported until a future phase. Issue #118 specifically called out documenting this on the support matrix. The matrix is the single source of truth for integrators. A one-line addition to the notes field would close that requirement.

**MEDIUM: /v1/responses bypass path carries tools to LiteLLM**

apps/edge-api/internal/inference/responses.go (lines 133-144)

`handleResponses` intentionally translates `req.Tools` and `req.ToolChoice` from the Responses API schema and forwards them to LiteLLM (the translation is correct by design for the Responses API). This is a different endpoint and the guard in `rejectUnsupportedChatParams` correctly does NOT apply here. However there is no test or comment documenting that tool-use via `/v1/responses` is intentionally allowed through. Given that `/v1/chat/completions` now hard-rejects these fields, a reviewer or future developer could reasonably add the same guard to `handleResponses` by mistake and break the Responses API path. A brief comment in `handleResponses` stating tools are forwarded intentionally would prevent that regression.

---

**What is correct and well-done:**

- Guard fires before `NeedFlags` construction, before billing reservation, before dispatch. Both sync and streaming paths are protected because both call `handleChatCompletions` which runs the guard first.
- OpenAI error envelope shape is exact: `{error:{message,type,code,param}}` with `param: null` and `code` as a `*string`. No provider names in any message string.
- `parallel_tool_calls` uses a separate `*bool` nil-check after the switch, which is correct since it has no length.
- Legacy `functions` and `function_call` fields are parsed and guarded.
- `errors.New` sentinel returned (not panic) so caller can return cleanly.
- Test coverage: table-driven, covers all six fields, covers streaming, covers the pass-through case (plain request expects 401 from auth layer). Provider-name leak check in assertions is a nice touch.
- No hardcoded secrets, no provider name leaks, no unsafe package use, no ignored errors.

PR #207, opened 2026-06-12
feat: hardware aware model advisor in enterprise installer

sakibsadmanshajib:
## Summary

- Adds a two-tier hardware-aware model advisor to `scripts/install.sh`, triggered when `--with-ollama` is selected
- Tier 1 (always works, no extra deps): pure bash reads `/proc/meminfo` + `nvidia-smi`, maps RAM/VRAM to a recommended Ollama model tag per the v1.3 advisor tier table
- Tier 2 (optional): runs `npx llm-checker hw-detect` if Node.js >= 16 is present on the host for richer GPU analysis; llm-checker is a maintained npm package (v3.5.15) by a Bangladeshi author, treated as advisory enrichment only since Node is not guaranteed on a Docker-only box
- The advisor-selected model is appended as a named LiteLLM route in `deploy/litellm/config.yaml` so it is immediately routable without manual editing
- In TTY mode the operator can accept, override, or pull the model immediately after the stack reaches healthy status; non-interactive mode prints the pull command
- `OLLAMA_MODEL` env var bypasses the advisor entirely for scripted installs
- Updated `scripts/README.md` with full advisor docs, model tier table, and override examples

**Depends on #202 (feat/enterprise-one-line-installer).** This branch was cut from `feat/enterprise-one-line-installer` and should be merged after #202 lands.

## Model tier table (Ollama tags, June 2026 validated)

| Hardware tier | Ollama tag | Size | Notes |
|---------------|------------|------|-------|
| CPU-only / less than 12 GB RAM | `phi4-mini` | 2.5 GB | 128k ctx, tools |
| 8 GB VRAM | `qwen3:8b` | 5.2 GB | Bangla-capable multilingual |
| 12 to 16 GB VRAM | `qwen2.5:14b` | 9.0 GB | Strong instruction following |
| 24+ GB VRAM | `qwen3:32b` | 20 GB | Thinking and tools |
| 128 GB unified memory | `qwen2.5:72b` | 47 GB | Datacentre / workstation |

Note: `llama3.3` on Ollama is a 70B model only. Corrected to `qwen3:8b` or `llama3.1:8b` for 8 GB VRAM slots.

## Test plan

- [ ] Run `bash -n scripts/install.sh` to verify syntax (passes)
- [ ] On a box with NVIDIA GPU: run `bash install.sh --with-ollama` interactively and confirm detection output matches GPU VRAM
- [ ] On a CPU-only box: confirm `phi4-mini` is selected
- [ ] With `OLLAMA_MODEL=llama3.1:8b` pre-set: confirm advisor is skipped
- [ ] With `--non-interactive`: confirm advisor selects recommended model without prompts and prints pull command
- [ ] Verify `deploy/litellm/config.yaml` gets the selected model appended as a named route
- [ ] On a box without Node.js: confirm Tier 2 skips gracefully with informational message
- [ ] On a box with Node.js >= 16: confirm `npx llm-checker hw-detect` runs and its output is shown

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -26,7 +26,7 @@ curl -fsSL https://raw.githubusercontent.com/sakibsadmanshajib/hive/main/scripts
 
 | Flag | Description |
 |------|-------------|
-| `--with-ollama` | Enable in-stack Ollama local inference. Sets `OLLAMA_BASE_URL=http://ollama:11434` and uncomments ollama entries in `deploy/litellm/config.yaml`. |
+| `--with-ollama` | Enable in-stack Ollama local inference. Triggers the hardware-aware model advisor, sets `OLLAMA_BASE_URL=http://ollama:11434`, and uncomments ollama entries in `deploy/litellm/config.yaml`. |
 | `--uninstall` | Stop the stack and print what remains for manual cleanup. Does not delete source files or volumes. |
 | `--non-interactive` | Skip all prompts. Read all values from environment variables. |
 | `--help` | Show usage. |
@@ -36,6 +36,7 @@ curl -fsSL https://raw.githubusercontent.com/sakibsadmanshajib/hive/main/scripts
 | Variable | Default | Purpose |
 |----------|---------|---------|
 | `HIVE_HOME` | `/opt/hive` | Where the repo is cloned / updated |
+| `OLLAMA_MODEL` | (advisor-selected) | Override the model advisor selection. Example: `OLLAMA_MODEL=llama3.1:8b` |
 
 All `.env` values can be pre-set as shell environment variables before running the installer. Required values in non-interactive mode:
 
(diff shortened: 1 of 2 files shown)


sakibsadmanshajib (commented):
All 7 threads addressed in rebase (50462f5):

T1 (ollama model insertion): Fixed. Advisor-selected model is now inserted before the litellm_settings: line via sed so it lands inside model_list, not as a stray files_settings entry.

T2 (git not installed): Fixed (inherited from PR 202 squash). git added to the apt install block in install_docker() so clone_or_update_repo() always has it on minimal images.

T3 (broad ollama sed): Fixed (inherited from PR 202 squash). Replaced the /.*ollama.*/I pattern with key-specific sed expressions targeting only model_name:, litellm_params:, model:, and api_base: lines. Prose instruction lines are left untouched.

T4 (secret echo for prompts): Fixed (inherited from PR 202 squash). read_secret() uses stty -echo for all secret-labeled prompts including SUPABASE_SERVICE_ROLE_KEY and S3_SECRET_KEY.

T5 (ollama started without flag): The enterprise profile in docker-compose.yml gates the ollama service behind a separate compose profile (ollama), not the enterprise profile, so it is not pulled or started unless explicitly included. No change needed; resolving.

T6 (set -u on optional provider vars): Fixed (inherited from PR 202 squash). Validation loop uses ${OPENROUTER_API_KEY:-} and ${GROQ_API_KEY:-} so set -u does not abort non-interactive installs that omit one provider.

T7 (compose v2 plugin check): Fixed (inherited from PR 202 squash). install_docker() now runs docker compose version when Docker CLI is already present and installs docker-compose-plugin if the v2 plugin is missing.

PR #211, opened 2026-06-12
test(sdk): update tool and response_format replay tests for phase 20 passthrough

sakibsadmanshajib:
## Summary

- Phase 20 (#118) ships capability-based tool-call routing: `tools` and `response_format` are now forwarded to capable routes (openrouter and groq seeded with `tools_supported=true`) instead of being unconditionally rejected with 400.
- The two SDK replay tests still asserted the old behaviour, causing the post-deploy healthcheck job in run #27392766395 to fail with `promise resolved instead of rejecting`.
- Both tests updated to assert the new correct behaviour: `tools` yields `finish_reason: tool_calls` with a `get_weather` tool call; `response_format: json_object` yields parseable JSON output.

## Root cause of deploy failure

Run #27392766395 (triggered by merge of #206) succeeded at build and deploy. Both health URLs returned 200. Failure was only in the post-deploy SDK replay step because the test expectations were written for the pre-phase-20 unconditional-400 guard that PR #206 intentionally replaced.

## Test plan

- [ ] CI passes (deploy-staging SDK replay: both tool tests green)
- [ ] `api-hive.scubed.co/health` returns 200
- [ ] `cp-hive.scubed.co/health` returns 200
- [ ] Tool smoke: POST /v1/chat/completions with tools array using hive-default returns finish_reason tool_calls, NOT 400

## Summary by CodeRabbit

* **New Features**
  * Chat Completions now supports `tools` parameter for function tool invocations and `response_format` parameter for JSON-formatted responses.

### DIFF
--- a/packages/sdk-tests/js/tests/chat-completions/chat-completions.test.ts
+++ b/packages/sdk-tests/js/tests/chat-completions/chat-completions.test.ts
@@ -48,64 +48,70 @@ describe("Chat Completions", () => {
     ).rejects.toMatchObject({ status: 404 });
   });
 
-  it("rejects tools with 400 unsupported_parameter", async () => {
-    // The edge explicitly rejects tool calling until full support ships (#118).
-    await expect(
-      client.chat.completions.create({
-        model: MODEL,
-        messages: [
-          { role: "user", content: "What is the weather like in London?" },
-        ],
-        max_tokens: 256,
-        tools: [
-          {
-            type: "function",
-            function: {
-              name: "get_weather",
-              description: "Get the current weather for a location",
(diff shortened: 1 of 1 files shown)


sakibsadmanshajib (commented):
Fixed in 8279b58. Added `tool_choice: { type: "function", function: { name: "get_weather" } }` to force the model to invoke the tool. Without it the model defaults to `auto` and may return a text response, making the `finish_reason === "tool_calls"` and `tool_calls[0].function.name` assertions non-deterministically flaky regardless of whether Phase 20 passthrough routing is working correctly.

PR #212, opened 2026-06-12
docs: staging live audit 2026-06-12

sakibsadmanshajib:
## Summary

- Automated E2E agent audit of all staging surfaces (api-hive.scubed.co, cp-hive.scubed.co, console-hive.scubed.co) with real HTTP requests and Playwright browser tests.
- Documents pass/fail/not-deployed status for every surface with actual request/response evidence.
- Identifies two P0 blockers for MVP demo: hive-auto inference 404 and hive-fast silent empty response at low max_tokens.

## Findings (terse)

| Surface | Status |
|---|---|
| Edge health, auth guard, models list | PASS |
| hive-default inference | PASS |
| hive-fast inference (max_tokens>=200) | PASS |
| hive-fast inference (max_tokens<60) | PARTIAL — silent empty content, reasoning model drain |
| hive-auto inference | FAIL — 404 despite appearing in /v1/models |
| Tools (function calling) | PASS |
| Embeddings | PASS |
| Control-plane health + catalog | PASS |
| Signup precheck endpoint | PASS |
| Chat UI (OWUI) | NOT DEPLOYED — intentional, staging is API-only |
| Web-console (Cloudflare Workers) | PASS |
| Signup form + submit flow | PASS |

## Test plan

- [ ] Verify hive-auto routing in staging DB routing_policies table and litellm config
- [ ] Reproduce hive-fast empty content with max_tokens=20 against staging
- [ ] Confirm hive-fast intended backend (reasoning vs non-reasoning) and fix routing or add edge guard
- [ ] Review report at `.planning/staging-live-audit-2026-06-12.md`

Issue #116, opened 2026-04-25
[P0/critical] Free-tier abuse: no CAPTCHA, no IP signup rate limit, no disposable-email block


sakibsadmanshajib:
## Severity
**Critical for free-tier public launch.** Surfaces in adversarial security review (ABUSE-2, ABUSE-3).

## Problem
Sign-up flow has none of the standard abuse defenses:
- No CAPTCHA (Turnstile, hCaptcha).
- No IP-level signup rate limit on Supabase Auth.
- No disposable-email blocklist (10minutemail, mailinator, ...).
- No Gmail `+tag` normalization (one address → unlimited accounts).

If free credits are granted at signup (typical free-tier launch model), one attacker scripts thousands of signups and drains the free-credit pool, forcing Hive to pay OpenRouter/Groq the upstream bill.

## Fix
- Cloudflare Turnstile widget on sign-up + sign-in (project already has CLOUDFLARE_API_TOKEN, Turnstile is free).
- Supabase Auth: enable per-IP signup rate limit (Auth settings).
- Disposable-domain blocklist via list (e.g. `disposable-email-domains` package) checked in `/auth/sign-up`.
- Email normalization: lowercase + strip `+tag` before duplicate check.
- Cap free credits per verified phone (or Stripe Identity later).

## Acceptance
- Headless script attempting 50 signups/min from one IP gets blocked after N.
- `gmail.com` plus-aliases counted as duplicates.

PR #214, opened 2026-06-13
fix: wire signup precheck before supabase.auth.signUp (issue #116)

sakibsadmanshajib:
## Summary

- The staging audit (PR #212) confirmed the signup page called `supabase.auth.signUp` directly at `apps/web-console/app/auth/sign-up/page.tsx:30`, bypassing the entire `signupguard` package built for issue #116. The PR #166 review thread claim that precheck was wired was incorrect; no call to the precheck endpoint existed anywhere in the file.
- The control-plane webhook backstop (`DisposableCheckFunc` in `apps/control-plane/internal/signup/webhook.go:135`) is intact and continues to block disposable-domain signups that arrive via direct Supabase API writes, but it does not cover the rate limit or Turnstile CAPTCHA, so the frontend precheck is still required.

## Changes

**`apps/web-console/app/auth/sign-up/page.tsx`**
Before calling `supabase.auth.signUp`, the form handler now POSTs to `/api/auth/signup-precheck`. A non-200 response blocks signup and surfaces the generic message from the control-plane. A Cloudflare Turnstile widget is rendered when `NEXT_PUBLIC_TURNSTILE_SITE_KEY` is set; when the env var is absent (local dev), the widget is skipped and the captcha token is sent as an empty string, which the control-plane accepts when `TURNSTILE_SECRET_KEY` is also unset.

**`apps/web-console/app/api/auth/signup-precheck/route.ts`** (new)
Next.js Route Handler that proxies `{ email, captcha_token }` to `POST /api/v1/auth/sign-up/precheck` on the control-plane. Reading `CONTROL_PLANE_BASE_URL` server-side keeps the internal URL out of the browser. Returns 503 when unconfigured or on network failure; never forwards internal error detail to the client.

**`apps/web-console/__tests__/signup-precheck-route.test.ts`** (new)
Unit tests for the Route Handler: success path, upstream rejection (403), missing `CONTROL_PLANE_BASE_URL` (503), malformed body (400), network failure (503).

**`.env.example`**
Added `NEXT_PUBLIC_TURNSTILE_SITE_KEY` alongside `TURNSTILE_SECRET_KEY` with instructions.

## Test plan

- [ ] Run `npm run test:unit` (requires local node_modules install) or `cd deploy/docker && docker compose run web-console npm run test:unit` to execute the Route Handler unit tests.
- [ ] Run `cd deploy/docker && docker compose run web-console npm run build` to type-check.
- [ ] With the full stack running locally, attempt signup with a disposable email address and verify a 403 is returned before `supabase.auth.signUp` is reached.
- [ ] With `NEXT_PUBLIC_TURNSTILE_SITE_KEY` set, verify the Turnstile widget renders on the signup form.
- [ ] With `NEXT_PUBLIC_TURNSTILE_SITE_KEY` unset (dev), verify signup still proceeds without the widget.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -106,6 +106,11 @@ SIGNUP_RATE_LIMIT_WINDOW_SECONDS=3600
 # when unset). NEVER commit a real key here — supply via secret manager
 # in staging + production.
 TURNSTILE_SECRET_KEY=
+# Cloudflare Turnstile site key (public, safe for browser). Rendered in the
+# web-console signup form. Leave EMPTY to disable the CAPTCHA widget in the
+# browser (widget is skipped gracefully when unset). Must match the site key
+# for the same Cloudflare Turnstile widget configured above.
+NEXT_PUBLIC_TURNSTILE_SITE_KEY=
 # Trusted reverse-proxy CIDR ranges (comma-separated). The precheck handler
 # honours CF-Connecting-IP and X-Forwarded-For ONLY when the direct TCP peer
 # (RemoteAddr) falls inside one of these ranges. Default EMPTY means no
--- a/apps/web-console/__tests__/signup-precheck-route.test.ts
+++ b/apps/web-console/__tests__/signup-precheck-route.test.ts
@@ -1,0 +1,142 @@
+/**
+ * TDD: signup-precheck proxy route (issue #116).
+ *
+ * Verifies that the Route Handler at /api/auth/signup-precheck:
+ * 1. Forwards { email, captcha_token } to the control-plane precheck endpoint.
+ * 2. Returns the upstream status and body verbatim on success (200).
+ * 3. Returns the upstream status and body verbatim on rejection (403/429).
+ * 4. Returns 503 when CONTROL_PLANE_BASE_URL is not set.
+ * 5. Returns 400 on malformed request body.
+ * 6. Returns 503 on network failure (fetch throws).
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+describe("app/api/auth/signup-precheck/route.ts", () => {
+  const originalEnv = process.env.CONTROL_PLANE_BASE_URL;
+
+  beforeEach(() => {
+    vi.resetModules();
+    vi.unstubAllGlobals();
+  });
+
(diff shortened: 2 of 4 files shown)


sakibsadmanshajib (commented):
Code review complete. Two HIGH issues fixed in 64bd30f, one MEDIUM fixed in the same commit. Full findings below.

**[HIGH] Missing try/finally in handleSubmit — button permanently disabled on network error**
`apps/web-console/app/auth/sign-up/page.tsx`
The `await fetch("/api/auth/signup-precheck", ...)` and `await supabase.auth.signUp(...)` calls had no outer try/catch. If either threw (network down, unexpected exception), `setLoading(false)` was never called, leaving the submit button disabled until page reload. Fixed by wrapping the entire handleSubmit body in try/finally with `setLoading(false)` in the finally block and a catch clause that sets a generic error message.

**[HIGH] Unsound type cast in route.ts — `Record<string,string>` on upstream JSON**
`apps/web-console/app/api/auth/signup-precheck/route.ts:63`
`upstreamBody = (await upstream.json()) as Record<string, string>` assumed every JSON value in the upstream response is a string, which is unsound — the control-plane may return numbers, booleans, or nested objects. Changed the type to `Record<string, unknown>` which is the correct type for unverified JSON.

**[MEDIUM] No fetch timeout on upstream call in route.ts**
`apps/web-console/app/api/auth/signup-precheck/route.ts`
No timeout was set on the upstream `fetch` call to the control-plane. A stalled control-plane would hold the Route Handler open until the Cloudflare Workers CPU time limit, giving users no timely error. Fixed by adding an `AbortController` with a 5-second timeout.

**[LOW] Minor unsound cast on precheck error body in page.tsx**
The inner `(await precheckRes.json()) as { error?: string }` cast has been replaced with `unknown` + explicit property existence narrowing checks. No functional change but removes the unsound cast.

---

**Verdict after fixes: APPROVED.** No remaining CRITICAL or HIGH issues. Route handler correctly keeps `CONTROL_PLANE_BASE_URL` server-side. Turnstile wiring is gracefully optional. Tests cover the five core proxy scenarios. Type guard in `isPrecheckBody` is correct idiomatic narrowing.

PR #409, opened 2026-07-20
feat: activate Windows WFP + firewall egress fence for desktop sandbox (Step 3 B2)

sakibsadmanshajib:
## Summary

Activates the Windows desktop-sandbox egress fence that #408 shipped inert. `launch` now enforces network policy through a two-layer, SID-keyed fence (D-011) instead of refusing both policies.

This is a security boundary (D-005 fail-closed): a crash or a failed fence step never opens egress. Per D-004 the activation is compile-verified here but not executed, so it stays a draft until the `spike307-win` lab matrix passes live.

## What changed

- **WFP core filters (were missing).** `filter_specs.rs` gains the srt-shape core: a loopback PERMIT (v4/v6, high weight) above a sandbox-SID BLOCK-all (v4/v6), plus a `FilterAction` enum and an `Option<u64>` weight with a const-assert that the permit weight exceeds the block weight. `wfp.rs` adds the `FWP_ACTION_PERMIT`, `FWP_V4/V6_ADDR_AND_MASK`, `FWP_RANGE0` port-range, and manual `FWP_UINT64` weight FFI to emit them. WFP is now a complete standalone fence and the firewall COM layer stays, so the two-layer model is whole.
- **Windows egress proxy.** `egress_proxy.rs` gains a loopback-TCP transport. The request parsing, allowlist, pinned-DNS, and no-private-IP logic are single-sourced across both transports via a `ProxyClientStream` trait; the Linux Unix-socket path is unchanged in behaviour. The module is gated on `linux` or `windows`.
- **Provision installs the persistent fence.** `provision_sandbox_account` installs the WFP filters plus the firewall block-all outbound, SID-keyed and fail-closed, before the readiness marker.
- **Launch goes live.** `windows::launch` dispatches to the SID-fenced elevated compose. The refusal guard is gone, `LaunchDecision` flipped `Refuse*` to `Spawn*`, and the compose installs the per-task firewall loopback allow (plus the loopback proxy and `HTTP(S)_PROXY` env for `AllowHosts`) and re-blocks loopback via an RAII teardown on every return path. `SandboxChild` now carries `Option<HANDLE>` fields plus a captured exit code, and the base restricted-token / Job-Object seam is superseded and retained dead.

## Four activation blockers fixed

1. **Concurrent-task rule clobber.** A single-active-task `AtomicBool` CAS guard (`ActiveTaskGuard`) rejects a second concurrent fenced launch rather than clobbering the shared loopback rule.
2. **`RemotePorts` cleanup semantics.** Every firewall rule is a BLOCK; teardown re-blocks all loopback via `ensure_offline_proxy_allowlist(&[], false)`; a `None` scope on a block rule only ever widens the block (never opens a port).
3. **Rule ownership and cleanup.** PROVISION owns the persistent block-all plus WFP; TASK-END owns the per-task loopback allow, torn down by an RAII guard on every return path. A crash leaves a harmless stale allow (the proxy is gone and the persistent block-all stays up).
4. **Non-atomic re-narrow.** `configure_rule` enables a rule LAST (after the full scope and the read-back verification), and the broad loopback block is installed before narrowing, so a partial update only ever leaves egress more closed (fail toward restrictive).

## Verification

Compile-only per D-004 (no Win32 execution here):

- `cargo fmt --check`
- `cargo clippy --target x86_64-pc-windows-gnu -p codex-windows-sandbox -p hive-desktop-sandbox --all-targets -- -D warnings`
- `cargo check --target x86_64-pc-windows-gnu -p codex-windows-sandbox -p hive-desktop-sandbox`
- `cargo test -p hive-desktop-sandbox` (69 lib + 1 integration) and `cargo test -p codex-windows-sandbox`

## Lab-gated before un-drafting (D-004)

The `spike307-win` lab matrix (`plan-b2-wfp-egress-2026-07-20.md` section 5, rows 1 through 11) must pass on a real MSVC Windows host before this leaves draft. It exercises, per row, the child SID identity, the filesystem isolation (write inside / outside, secret read), a DenyAll egress and DNS block, a direct-connect kernel drop, an AllowHosts request to an allowed host through the proxy, and a non-allowed host refusal. `hive-sandbox-validate` was extended to drive DenyAll and AllowHosts (via `HIVE_VALIDATE_NET`) and print per-row markers for the harness.

Do not merge or mark ready until that matrix passes.


sakibsadmanshajib:
Closing this rather than leaving it to rot, and the branch is preserved.

The work is real and the review notes on it are worth keeping. The reason to close is arithmetic rather than a judgment about the change: this is a 2,677 addition and 1,071 deletion draft that has not moved since 2026-08-23, and `main` has taken roughly forty pull requests since then, several of them inside the sandbox and launcher seam this diff rewrites. Rebasing it now would cost more than re-deriving it later against whatever that seam looks like at the time, and every day it stays open that gap widens.

Windows desktop sits in the `v1.3 device era` milestone, which is early, and Steps 2 through 6 of the staged Codex-adopt build were deferred by owner decision. Step 1, the launch seam, is merged and lab validated. Nothing in the current demo path touches this code.

The branch `feat/desktop-sandbox-b2-activate` is not deleted, so the diff stays reachable. Re-cut it at v1.3 against the sandbox seam as it exists then.

Issue #384, opened 2026-07-19
Add destructive-action and prompt-injection classifier to the agent sandbox


sakibsadmanshajib:
Problem: the agent sandbox has no pre-execution screen that flags destructive commands or prompt-injection content before an action runs.
Benchmark parity: Claude Code auto-mode classifier screens destructive operations and prompt injection before execution.
Acceptance criteria: classifier evaluates each proposed action; destructive or injected actions are blocked or escalated to human approval; decisions audited via audit-canonical; policy configurable per tenant.
Related: complements egress allowlist (#308) and desktop sandbox hardening.

sakibsadmanshajib:
PR #465 is per-tenant resource quota enforcement. This issue asks for a destructive-action/prompt-injection classifier, an unrelated security control not present in the diff. Leaving open, this PR does not resolve it.

Issue #308, opened 2026-07-07
Server-side agent sandbox: Apptainer rootless plus per-tenant quota and egress allowlist


sakibsadmanshajib:
Sub-issue of #305. OpenHands provides no built in per-tenant resource quota or noisy neighbor control, a documented bug already exists upstream where one user's heavy load crashed other tenants' agents sharing a box. Hive must build a quota layer itself on top of the Apptainer rootless backend.

Separately, owner decision on network egress: control must be admin configurable per user, not only per tenant, since some users will need internet access for web search, package downloads, and documentation lookups while others will not. One egress allowlist configuration must live in control-plane as the single source of truth, consumed by both the server side OpenHands allowed_hosts workspace config and the desktop local firewall rules from the sandbox backends sub-issue, so the same policy applies everywhere the agent runs rather than drifting between surfaces the way the existing HIVE_SOVEREIGN guard already has. See #245.

sakibsadmanshajib:
PR #465's title ("enforce per-tenant quota and per-session cgroup limits") matches the per-tenant quota half of this issue directly. The egress-allowlist half (admin-configurable per-user network access) is not mentioned anywhere in the diff. Leaving open. Remaining: the egress allowlist requirement (admin-configurable per-user, not only per-tenant, network access)

sakibsadmanshajib:
Triage note (backlog catalogue pass): this looks substantially delivered by the agent-engine work and needs verification before closing. The per-tenant and per-user concurrency quota knobs exist and are documented (HIVE_QUOTA_TENANT_CONCURRENCY default 4, HIVE_QUOTA_USER_CONCURRENCY default 2, plus the memory, CPU and PID sandbox limits), the Apptainer rootless sandbox ships in deploy/apptainer/agent-engine.def, and an egress proxy exists in front of the sandbox (issue #476 is a follow-up hardening request against that proxy, which implies the proxy itself is real). What is not evidenced is the part of this issue that matters most: one admin-configurable egress allowlist living in control-plane as a single source of truth, configurable per user as well as per tenant, and consumed by both the server-side workspace config and the desktop firewall rules. Left open at priority:medium; someone should confirm the delivered half against the running stack and narrow this issue to the allowlist single-source-of-truth remainder.

Issue #307, opened 2026-07-07
Security spike: validate sandbox isolation before any agent-runtime implementation


sakibsadmanshajib:
Sub-issue of #305. Must happen before implementation starts, not resolvable by further reading alone.

Confirm Apptainer actually holds multi-tenant isolation under adversarial testing on the shared server path.

For every desktop sandbox backend built in the sandbox backends sub-issue, confirm that no file the sandboxed agent can write is later executed or read by an unsandboxed host process. This is the exact Configuration Based Sandbox Escape pattern publicly disclosed in 2026 that hit OpenAI Codex CLI, Claude Code, and Gemini CLI. In Codex's case a notify hook command ran outside the sandbox on every agent turn while the agent could write its own config file from inside the sandbox with no approval gate, achieving host code execution while the sandbox looked fully intact.

Also confirm the Docker socket is never exposed to the agent process on the server path. OpenHands maintainers themselves describe docker socket access as effectively root access to the host.

sakibsadmanshajib:
PR #465 ships quota/cgroup enforcement code. This issue asks for a security spike (adversarial validation of sandbox isolation) that must happen BEFORE implementation, a research/testing deliverable, not something a feature PR satisfies. Leaving open, this PR does not resolve it.

Issue #305, opened 2026-07-07
Epic: coding-agent and cowork subsystem on a shared OpenHands engine


sakibsadmanshajib:
Design outcome of an extended July 2026 brainstorming session, full history in conversation, not yet written to a formal spec doc. Single shared agent engine, OpenHands (MIT licensed, exclude the enterprise directory which has its own license), backs two plugin packs defined as OpenHands microagent configs rather than two separate engines: a coding pack (git, build, test, lint tools) and a knowledge-work pack (document read and write, RAG corpus search, the skills tracked in #300). Both packs ship in parallel per owner decision, both run at the same sandbox tier, arbitrary shell, build, and test commands are allowed in both, matching how the owner personally uses Claude Cowork today. No GUI computer-use in v1, owner decision. External Claude Code CLI users can already point at Hive's existing Anthropic-compatible /v1/messages surface (#168) at zero additional cost. Sub-issues below cover the specific engineering pieces. Two open decisions not yet resolved, tracked in the sub-issues: where the dedicated agent UI panel lives, and desktop to web task portability rules.

sakibsadmanshajib:
PR #465 ("enforce per-tenant quota and per-session cgroup limits") ships the resource-quota slice of this epic. The epic's full scope (a shared OpenHands engine backing both a coding pack and a knowledge-work pack) is much larger than one PR and not confirmed complete. Leaving open. Remaining: everything in the epic besides per-tenant quota/cgroup enforcement: the shared-engine architecture itself, the two plugin packs, and the sub-issues #306/#307/#308/#310/#384 below

PR #465, opened 2026-07-27
fix: enforce per-tenant quota and per-session cgroup limits on the real sandbox launch path

sakibsadmanshajib:
## Decision context

This lands the implementable half of the server-side sandbox scalability decision for #308. The decision itself: **keep Apptainer rootless, do not move to Firecracker, gVisor or a hosted sandbox provider.** Reasoning, with sources:

- The existing launcher design (isolated netns via `--net --network none`, plus a Unix-socket egress relay bridged by socat, plus an allowlist proxy) is the same architecture Anthropic's own [sandbox-runtime](https://github.com/anthropic-experimental/sandbox-runtime) uses on Linux with bubblewrap. Independent convergent validation of what is already built.
- Every option with a stronger isolation ceiling needs host privileges a customer-hosted Enterprise box cannot be assumed to grant. Firecracker and Kata need `/dev/kvm` and nested virtualization ([Northflank comparison](https://northflank.com/blog/how-to-sandbox-ai-agents): Firecracker ~125ms, Kata ~200ms, both nested-virt gated). [gVisor rootless](https://gvisor.dev/docs/user_guide/rootless/) cannot use its netstack, so it forces `--network=none` anyway (the same design as today) while adding 10-30% I/O overhead on exactly the build and test path a coding pack runs.
- A SIF is a mounted squashfs: no daemon, no root socket, no per-session image unpack. Best fit for Compose single-box Enterprise with no Kubernetes.
- Cross-tenant hardware isolation comes from never co-scheduling two tenants on one worker VM, not from the sandbox layer. Single org equals single tenant, so an Enterprise deploy only ever needs user-versus-user isolation inside one already-trusted org.
- The WSL2 dev-box friction is not removable by any runtime choice: on that box `unshare -Ur --mount-proc` fails and there is no cgroups v2 at all, so rootless Apptainer, rootless gVisor and rootless Podman all fail identically. The fix is to make the physical Ubuntu box the sandbox validation host, not to change runtimes.

Residual risk worth recording: rootless Apptainer cannot apply a seccomp profile (`--security` needs a setuid-root install per the [Apptainer admin guide](https://apptainer.org/docs/admin/main/security.html)), so the host syscall surface stays full. That is precisely why the tenant boundary must be the VM boundary.

## What this PR fixes

Two defects found while reading the current path end to end. Neither was a design question; both were wiring gaps.

**1. Quota was dead code on the production path.** `quota.Manager` was only ever used by `cmd/agent-engine` (the standalone smoke-test binary). Real launches go control-plane `buildAgentEngine` to `engineapi` to `engine.SandboxEngine.Launch`, which never called `Acquire`. One tenant could saturate the box, the exact noisy-neighbour scenario #308 was filed for. `Launch` now acquires a slot before doing any work and releases it on every failure branch.

**2. No cgroup limits existed anywhere.** `grep` for `apply-cgroups`, `memory-limit` or `pids-limit` across `apps/agent-engine` returned nothing. Concurrency counting says how many sessions run, never what each consumes, and a coding-pack session runs arbitrary builds plus headless Chromium. `BuildArgv` now always emits Apptainer's `--memory`, `--cpus` and `--pids-limit`, and `LaunchConfig.validate` refuses to build an invocation missing any of them. Rootless enforcement requires cgroups v2 with systemd delegation ([Apptainer cgroups docs](https://apptainer.org/docs/user/main/cgroups.html)); on a host without it Apptainer fails the launch loudly, which is correct, since running unbounded is the bug.

**3. A session leak with the same root cause, exposed by wiring the quota.** The agent-server is a server: it keeps running after its conversation finishes, and `agenttask`'s poller only records the terminal status, it never calls `Cancel`. Every completed task therefore leaked its sandbox process, its session directories, and (once quota is wired) its concurrency slot, forever. `Status` now reaps the session the first time it observes a terminal state, which is the one place every terminal observation routes through. A reaped session keeps its terminal outcome in the map instead of being deleted, so the poller's `Status` retry after a failed `Transition` still gets an answer rather than an unknown-session error that would leave the task active forever. `Cancel` goes through the same reap path.

Limits are operator-tunable (`HIVE_QUOTA_TENANT_CONCURRENCY`, `HIVE_QUOTA_USER_CONCURRENCY`, `HIVE_SANDBOX_MEMORY_LIMIT`, `HIVE_SANDBOX_CPU_LIMIT`, `HIVE_SANDBOX_PIDS_LIMIT`), because an Enterprise single box and a cloud worker have very different budgets.

## Behaviour change to note in review

`Cancel` no longer deletes the session from the map, so `Status` after `Cancel` now returns `cancelled` instead of `ErrUnknownSession`. The existing Cancel test was updated to assert the new contract. This is deliberate: two different lifecycle rules for the same map is the kind of thing that gets decoded at 3am, and returning an error to the poller is what strands a task in the active list.

## Test plan

- [x] `go test ./apps/agent-engine/... -count=1` passes (all 10 packages with tests).
- [x] `go test ./apps/control-plane/... -count=1 -short` exits 0, 51 packages ok, no FAIL or panic.
- [x] `go vet` clean across both modules; `gofmt -l` clean on all touched files.
- [x] New tests: tenant ceiling denies a second launch, user ceiling denies a second launch, terminal `Status` kills the sandbox process and frees the slot for a re-launch, and a repeat `Status` after the reap still reports the terminal outcome.
- [x] New test asserts `--memory`, `--cpus` and `--pids-limit` are present as adjacent pairs in the built argv; three new rows cover the config rejection cases.
- [ ] **Live Apptainer validation not run here and cannot be**: WSL2 has no rootless user-namespace support and no cgroups v2 (`unshare -Ur --mount-proc` fails, `/sys/fs/cgroup/cgroup.controllers` absent). The cgroup flags are argv-verified only. Live launch plus the #307 adversarial isolation suite has to run on the physical Ubuntu box, and remains the gate before this path is trusted in production.

## Issues

- Closes the quota and per-session resource halves of #308. The per-user egress allowlist bullet is unaffected by this PR and needs a separate check that the desktop path consumes the same SSOT.
- Unblocks #307 Step 0.1 and 0.3, which need the physical box, not more code.
- Does not touch #384 or #305 scope.
- A follow-up issue is still needed for horizontal scale: promote agent-engine from an in-process embed to a worker service over its existing `engineapi` seam, with tenant-pinned placement. Not in this PR.

No visual proof artifact: this change touches no UI or UX surface.

## Summary by CodeRabbit

* **New Features**
  * Added configurable memory, CPU, and process limits for sandbox sessions.
  * Added per-tenant and per-user concurrency limits to prevent resource contention.
  * Sandbox sessions now clean up resources and release capacity after completion or cancellation.
  * Cancelled sessions retain their final status for reliable status checks.

* **Bug Fixes**
  * Improved sandbox cleanup and terminal session handling.

### DIFF
--- a/apps/agent-engine/cmd/agent-engine/main.go
+++ b/apps/agent-engine/cmd/agent-engine/main.go
@@ -144,6 +144,9 @@ func main() {
 		ProxySocketPath:  proxySocketPath,
 		ControlSocketDir: ctlDir,
 		MCPConfigPath:    mcpConfigPath,
+		MemoryLimit:      envOr("HIVE_SANDBOX_MEMORY_LIMIT", "4G"),
+		CPULimit:         envOr("HIVE_SANDBOX_CPU_LIMIT", "2"),
+		PidsLimit:        envInt("HIVE_SANDBOX_PIDS_LIMIT", 512),
 	}
 
 	argv, err := sandbox.BuildArgv(cfg)
--- a/apps/agent-engine/internal/engine/engine.go
+++ b/apps/agent-engine/internal/engine/engine.go
@@ -46,6 +46,7 @@ import (
 
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/controlclient"
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/egressproxy"
+	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/quota"
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/sandbox"
 )
 
@@ -109,11 +110,40 @@ type Config struct {
 	// ControlReadyTimeout bounds how long Launch waits for the in-SIF shim
 	// to create its control socket. Defaults to 30s.
 	ControlReadyTimeout time.Duration
+
+	// QuotaTenantConcurrency and QuotaUserConcurrency cap how many sessions
+	// one tenant and one user may run at once (issue #308). Non-positive
+	// values fall back to the defaults below.
+	QuotaTenantConcurrency int
+	QuotaUserConcurrency   int
+
+	// MemoryLimit, CPULimit and PidsLimit bound what each individual session
+	// may consume; see sandbox.LaunchConfig's fields of the same name.
+	MemoryLimit string
+	CPULimit    string
+	PidsLimit   int
 }
 
 func (c Config) withDefaults() Config {
 	if c.ControlReadyTimeout <= 0 {
 		c.ControlReadyTimeout = 30 * time.Second
+	}
+	if c.QuotaTenantConcurrency <= 0 {
+		c.QuotaTenantConcurrency = 4
+	}
+	if c.QuotaUserConcurrency <= 0 {
+		c.QuotaUserConcurrency = 2
+	}
+	// Sized for a coding-pack session: an arbitrary build plus the headless
+	// Chromium the browser tool drives.
+	if c.MemoryLimit == "" {
+		c.MemoryLimit = "4G"
+	}
+	if c.CPULimit == "" {
+		c.CPULimit = "2"
+	}
+	if c.PidsLimit <= 0 {
+		c.PidsLimit = 512
 	}
 	return c
 }
@@ -145,14 +175,27 @@ func realStart(argv []string) (process, error) {
 	return &osProcess{cmd: cmd}, nil
 }
 
+// terminalOutcome is a reaped session's final answer, replayed by Status
+// instead of dialling a control socket whose sandbox is already gone.
+type terminalOutcome struct {
+	status        Status
+	resultSummary string
+	errMessage    string
+}
+
 type session struct {
 	conversationID uuid.UUID
 	client         *controlclient.Client
 	proc           process
 	proxySrv       *http.Server
 	proxyListener  net.Listener
-	sessionDir     string // removed on Cancel; holds only Unix sockets
-	workingDir     string // removed on Cancel; the sandbox's /workspace bind mount
+	sessionDir     string // removed when reaped; holds only Unix sockets
+	workingDir     string // removed when reaped; the sandbox's /workspace bind mount
+	release        func() // frees this session's quota slot; safe to call repeatedly
+
+	// reaped and terminal are guarded by SandboxEngine.mu.
+	reaped   bool
+	terminal *terminalOutcome
 }
 
 // SandboxEngine launches, polls, and cancels agent-engine sandbox sessions.
@@ -160,16 +203,31 @@ type session struct {
 type SandboxEngine struct {
 	cfg   Config
 	start func(argv []string) (process, error)
+	q     *quota.Manager
 
-	mu       sync.Mutex
+	mu sync.Mutex
+	// ponytail: a reaped session keeps a small terminal record here instead
+	// of being deleted, so the map grows by one struct per completed task and
+	// only a process restart clears it. Add eviction if one process ever runs
+	// enough tasks for that to matter.
 	sessions map[string]*session
 }
 
 // New constructs a SandboxEngine from cfg.
 func New(cfg Config) *SandboxEngine {
+	cfg = cfg.withDefaults()
+	q, err := quota.New(quota.Limits{
+		TenantConcurrency: cfg.QuotaTenantConcurrency,
+		UserConcurrency:   cfg.QuotaUserConcurrency,
+	})
+	if err != nil {
+		// Unreachable: withDefaults replaces every non-positive limit.
+		panic(err)
+	}
 	return &SandboxEngine{
-		cfg:      cfg.withDefaults(),
+		cfg:      cfg,
 		start:    realStart,
+		q:        q,
 		sessions: make(map[string]*session),
 	}
 }
@@ -198,6 +256,19 @@ func (e *SandboxEngine) Launch(ctx context.Context, t Task) (sessionRef string, 
 		return "", fmt.Errorf("engine: Task.Pack is required")
 	}
 
+	release, err := e.q.Acquire(t.TenantID, t.UserID)
+	if err != nil {
+		return "", err
+	}
+	// succeeded flips true on the return at the very end of this function; it
+	// gates this cleanup and the resource cleanup registered further down.
+	succeeded := false
+	defer func() {
+		if !succeeded {
+			release()
+		}
+	}()
+
 	allowedHosts, err := e.cfg.ResolveEgressHosts(ctx, t.TenantID, t.UserID)
 	if err != nil {
 		return "", fmt.Errorf("engine: resolve egress policy, refusing to launch: %w", err)
@@ -217,13 +288,11 @@ func (e *SandboxEngine) Launch(ctx context.Context, t Task) (sessionRef string, 
 	workingDir := filepath.Join(e.cfg.WorkspaceRoot, t.ID.String())
 
 	// Single deferred cleanup for every failure branch below: closes
-	// whatever got started so far and removes both directories. succeeded
-	// flips true only on the return at the very end of this function.
+	// whatever got started so far and removes both directories.
 	var (
 		proxySrv *http.Server
 		proc     process
 	)
-	succeeded := false
 	defer func() {
 		if succeeded {
 			return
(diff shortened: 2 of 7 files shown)


sakibsadmanshajib (commented):
## Security review: quota + cgroup enforcement on the sandbox launch path

Scope reviewed: `apps/agent-engine/internal/quota/quota.go`, `apps/agent-engine/internal/engine/engine.go`, `apps/agent-engine/internal/sandbox/launcher.go`, plus the wiring in `apps/control-plane/cmd/server/main.go` and `apps/agent-engine/cmd/agent-engine/main.go`.

**Verdict: no CRITICAL or HIGH issues found.** A few non-blocking MEDIUM/LOW hardening notes below.

### Reviewed and clean

- **Acquire/Release race**: `quota.Manager.Acquire` (`apps/agent-engine/internal/quota/quota.go:64-84`) checks both ceilings and increments both counters under one `sync.Mutex`, so concurrent `Launch` calls cannot oversubscribe either the tenant or the user ceiling. Release is wrapped in `sync.Once`, safe against double-release. `TestAcquire_ConcurrentGoroutinesNeverOversubscribe` exercises this. No TOCTOU.
- **BuildArgv injection surface**: `MemoryLimit`/`CPULimit`/`PidsLimit` (`apps/agent-engine/internal/sandbox/launcher.go:304-306`) are process-level config (env vars set by the operator via `envOr`/`parseIntEnv`), not tenant/session-controlled input, and `realStart` execs `argv` directly via `exec.Command` with no shell interpretation (`apps/agent-engine/internal/engine/engine.go:169`), so even a hostile value here can only be rejected by `apptainer`, not used to break out of the argv boundary. No injection risk from this change.
- **Fail-closed validation**: `LaunchConfig.validate()` (`apps/agent-engine/internal/sandbox/launcher.go` `ErrMissingResourceLimit` case) rejects empty `MemoryLimit`/`CPULimit` or non-positive `PidsLimit` before `BuildArgv` emits a command, so a misconfigured caller can't launch an unbounded session by omission; `Config.withDefaults()` and both `main.go` env wirings also apply the same defaults independently, so the fail-closed check is genuinely defense in depth rather than the only guard.
- **Status-after-Cancel state confusion**: verified the tenant-scoping boundary is upstream of this package — `apps/control-plane/internal/agenttask` resolves `EngineSessionRef` from a tenant/user-scoped DB row (`agenttask.Service.Cancel(ctx, tenantID, userID, taskID)`, poller reads `t.EngineSessionRef` off an already-scoped task row) before ever calling into `SandboxEngine`. `SandboxEngine` itself has no tenant check on `sessionRef` (unchanged by this PR either way), so the behavior change from `ErrUnknownSession` to a replayed `StatusCancelled` doesn't introduce a new cross-tenant read; it only changes what the same authorized caller sees.
- **reap() idempotency**: the `sess.reaped` bool is checked-and-set under `e.mu` (`engine.go:458-466`) before any of the actual kill/close/remove/release work runs, giving first-writer-wins semantics when `Status` and `Cancel` race on the same session — no double-kill, no torn/overwritten `terminal` outcome.

### Non-blocking notes (worth a follow-up, not blocking this PR)

1. **`reap()` frees the quota slot even when the kill fails** (`apps/agent-engine/internal/engine/engine.go:468-473`): `killErr := sess.proc.Kill()` ... `sess.release()` runs unconditionally, ignoring `killErr`. On Linux `SIGKILL` essentially never fails against a process the same UID owns, so this is low-likelihood, but if the underlying `apptainer` process ever survives (stuck in D-state, orphaned cgroup, etc.), the tenant's slot is freed and they can launch again while the old sandbox keeps consuming host resources, i.e. a quiet quota bypass under host resource pressure. Suggest logging (or alerting on) non-nil `killErr` at minimum.
2. **Quota is per-process, in-memory only** (`apps/agent-engine/internal/quota/quota.go`): fine under the current documented single-instance-per-box architecture (no k8s at the edge), but if `agent-engine` is ever run with >1 replica behind a load balancer without a shared quota store, each replica enforces its own independent ceiling and the effective tenant/user concurrency limit silently multiplies by replica count. Worth a one-line comment on `Manager` calling this out so it isn't rediscovered as a prod incident.
3. **Enforcement is asserted, not exercised, by this diff**: `--memory`/`--cpus`/`--pids-limit` depend on a rootless cgroups v2 host with systemd delegation; on a host without it Apptainer should fail the launch (per the PR description), but nothing in this PR's test suite runs against real `apptainer`/cgroups to confirm the limits actually get applied (unit tests only assert the flags are present in argv — `TestBuildArgv_AppliesResourceLimits`). Recommend a live check on the physical validation host: launch a session, exceed the memory limit and confirm OOM-kill inside the cgroup (not the host), and confirm `--pids-limit` actually returns `EAGAIN` on fork past the ceiling, before treating this as a relied-upon boundary in production.

No action requested from me before merge; (1) and (2) are candidates for a quick follow-up issue, (3) is a verification task for whoever owns the physical demo-box validation pass.

sakibsadmanshajib (commented):
Reviewed the full diff plus the surrounding call graph (quota.Manager, SandboxEngine.Launch/Status/Cancel, agenttask.Poller, agenttask.Service.Cancel, agentengine.Engine adapter). Also ran `go build`, `go vet`, and `go test -count=1` for both `apps/agent-engine` and `apps/control-plane` in the toolchain container; everything is green.

**Quota wiring (apps/agent-engine/internal/engine/engine.go:259-270, 384-399)**

`Acquire` is the first thing `Launch` does after the `Task.Pack` check, and every failure path after that (egress resolution, directory creation, proxy listener, port allocation, argv build, process start, control-socket wait, `StartConversation`, `Run`) is covered by the `defer func() { if !succeeded { release() } }()` registered right after `Acquire` succeeds. I traced every early return in the function; none of them skip the release. `quota.Manager.Acquire`'s own concurrency safety is solid: single mutex around both counters, `sync.Once`-wrapped release, and `TestAcquire_ConcurrentGoroutinesNeverOversubscribe` (50 goroutines, limit 5, asserts exactly 5 succeed) is a real test of the invariant, not just a unit test in isolation.

**Reap-on-terminal-Status (engine.go:407-474)**

Good fix for the actual bug: `apps/control-plane/internal/agenttask/poller.go:85-91` only calls `Status`, and a `statusErr != nil` (which is what the old `ErrUnknownSession`-after-Cancel behavior produced) makes the poller `continue` and retry forever, so a deleted session used to strand the task active permanently. The new contract (session stays in the map with a replayed terminal outcome) directly fixes that. Confirmed by grep that nothing else in the tree depends on the old `ErrUnknownSession`-after-Cancel behavior.

One small race worth a look, not a blocker: `reap` sets `sess.reaped = true` and stores the outcome under the lock, then does the actual kill/close/RemoveAll/release *outside* the lock. If `Status` and `Cancel` race on the same session, the "losing" caller's `reap` call returns immediately (correct, no double free) but that caller still returns its own locally computed status/error rather than the outcome the winner stored, so there's a narrow window where two concurrent callers could observe different terminal values for one call each before everything converges on the replayed outcome. In practice `Status` (poller-driven) and `Cancel` (user-driven) aren't likely to race on one session today, so I'd call this a nice-to-have, not a fix-before-merge.

**Cgroup limits (apps/agent-engine/internal/sandbox/launcher.go:296-315)**

`--memory`/`--cpus`/`--pids-limit` are real Apptainer/Singularity exec flags, placed correctly before the SIF path in argv. Defaults (4G / 2 CPU / 512 pids) are reasonable for a coding-pack session and are operator-tunable via env vars at both call sites (`cmd/agent-engine/main.go` and `control-plane/cmd/server/main.go:1475-1484`) with matching fallback values. `validate()` correctly rejects zero/empty values, and `launcher_test.go` covers all three rejection cases plus the argv-presence case.

**One gap this PR doesn't close (pre-existing, already flagged in the code, not a regression)**

`apps/control-plane/internal/agenttask/service.go:73-75` (`Service.Cancel`) only transitions the DB row to `cancelled`; it never calls `agentengine.Engine.Cancel` (that adapter's own doc comment at `apps/control-plane/internal/agentengine/engine.go:50-53` says as much: "Not called by agenttask.Service.Cancel yet"). Since `ListActive` filters to queued/running only, a user-initiated cancel via the HTTP endpoint removes the task from the poller's view before `SandboxEngine.Cancel` (and therefore this PR's `reap`) ever runs, so that path still leaks the sandbox process, session directories, and quota slot until process restart. This PR's fix fully closes the "task completes naturally" leak (via `Status`'s reap) but not the "user clicks cancel" leak, since that path never reaches the engine at all today. Worth a follow-up issue if one doesn't already exist; not asking for it to block this PR since it's a pre-existing, already-documented gap outside this diff's stated scope.

**Verdict: Approve.** No CRITICAL or HIGH issues. The quota/cgroup wiring is correct on every path I traced, the `Status`-after-`Cancel` contract change is justified by and fixes a real poller bug, and tests plus a live build/vet pass back it up.

PR #471, opened 2026-07-27
fix: slash every zero the console sets in a dense numeric column

sakibsadmanshajib:
## What this does

Fixes zero glyph rendering in dense numeric columns across the developer console. Credit balances, catalog price columns, spend alert thresholds, and invoice totals were set in the sans stack with `tabular-nums`, where `0` renders as an unslashed oval indistinguishable from an uppercase O. Every one of those figures now runs through the console's existing `.metric` rule (Geist Mono), and the rule itself now asks for `slashed-zero` explicitly instead of relying on it as an incidental default of whichever mono happens to be loaded.

Six one-line/one-class changes route existing numeric cells into `.metric`:

- `app/console/billing/alerts/page.tsx`: spend alert threshold percentage cell.
- `app/console/page.tsx`: overview "Posted / Reserved" credits caption.
- `components/billing/billing-overview.tsx`: same caption on the billing page.
- `components/billing/budget-alert-banner.tsx`: banner message restructured so the formatted credit amount gets its own `metric` span, rather than the whole sentence carrying `tabular-nums`.
- `components/billing/invoice-row.tsx`: invoice total in taka.
- `components/ui/data-table.tsx`: the generic `col.numeric` styling used by every numeric table column in the console, including the model catalog price columns.

## The `.metric` rule itself

`app/globals.css`: `.metric` now declares `font-variant-numeric: slashed-zero tabular-nums` and `font-feature-settings: "tnum", "zero"`, with the comment rewritten to explain why. Geist Mono already draws a slashed zero by default, so this changes no pixels today, but leaving it undeclared meant a future font-stack swap, or a fallback render, could quietly hand back an unslashed zero with nothing catching it.

## Why route to mono instead of adding a CSS feature to the sans

`09-font-zero-probe.png` in the proof set: `font-variant-numeric: slashed-zero` has no effect on Geist Sans or its fallback stack, that family doesn't ship the glyph feature. Only Geist Mono draws a slashed zero, which is why the fix is routing figures to the mono class rather than adding the declaration where the figures already lived.

## Test

New `tests/unit/metric-slashed-zero.test.ts`: reads `app/globals.css`, extracts the `.metric` rule body, and asserts it declares both `font-variant-numeric: slashed-zero tabular-nums` and `font-feature-settings` containing `"tnum"` and `"zero"`.

## Rebased onto the Bengali i18n work

PR #473 landed first and overlapped this branch in two places: it added Noto Sans Bengali behind Geist in every font stack including `--font-mono`, and it moved the invoice total formatter out of `invoice-row.tsx` into `lib/format/money.ts`. This branch is rebased onto that, with the invoice row taking both changes: the i18n formatter and the `metric` class.

The font stack change is the part worth re-verifying, because a digit falling through Geist Mono to Noto would lose its slash. It does not: `lib/i18n/locales.ts` pins numerals to Latin digits with `-u-nu-latn`, so figures still resolve to Geist Mono. Confirmed on the running console in Bengali, invoice total and Overview credits card:

## Visual proof

`docs/proof/console-numerals-2026-07-27/`, captured with Playwright against the console running from this branch (`next dev`, real Supabase project, real control plane on `localhost:8081`), signed in as the seeded `e2e-verified` fixture account in the `Acme Labs` workspace.

| | |
|---|---|
| Sign-in mark, light | Sign-in screen. The wordmark already renders the real Hive mark, not a lettered placeholder; nothing in this change touches it, capture is on the record because the placeholder state was questioned. |
| Wordmark, light Wordmark, dark | The wordmark cropped in both palettes. Drawn in `currentColor`, so it inverts with the theme. |
| Credits before Credits after | Overview credits card. The headline figure was already slashed. The "Posted 0 · Reserved 0" caption beneath it was sans; it now runs through `.metric` and carries the slash. |
| Catalog table before Catalog table after | Model catalog price columns, the densest numeric table in the console. Before: sans with tabular figures, `0` reads as O. After: mono, every zero slashed. |
| Overview dark, after | Overview in the dark palette after the change: mark and slashed numerals both hold under theme inversion. |
| Font zero probe | Why the fix routes to mono rather than adding a CSS feature to the sans: rows 1 to 2 are Geist Sans with/without the declaration (no effect); rows 3 to 4 are the mono fallback stack (same result). Only Geist Mono draws the slash by default. |
| Invoice total before Invoice total after | Workspace invoice total in taka, real customer money. Before: sans, all five zeros read as letters. After: mono, every zero slashed. |
| Invoice total after, Bengali Overview after, Bengali | Re-verification after rebase onto the Bengali i18n work, console in Bengali. Latin digits still come from Geist Mono, zero keeps its slash. |

Notes on record in `docs/proof/console-numerals-2026-07-27/README.md`: the dev-badge circle in full-viewport shots is the Next.js dev overlay, not console chrome; the catalog "before" shot was captured by reverting and restoring the one-line `data-table.tsx` change on a running dev server; balances read zero because the fixture workspace has never been topped up; the invoice shots used one synthetic invoice inserted, photographed, then deleted, leaving the `invoices` table back at zero rows, no real customer or billing data in any capture. Spend alert threshold cell and budget alert banner carry the same one-token change as the invoice total and are not photographed separately, since they need a seeded alert to render and the invoice capture already shows the glyph change.

## Verification

Re-verified after the rebase with next-intl installed: `tsc --noEmit` clean, 24 test files and 260 tests passing (the count rose because #473 added its own format and return-to tests).

## Summary by CodeRabbit

* **Style**
  * Improved numeric formatting across billing, credits, invoices, spend alerts, and data tables.
  * Numeric values now use consistent metric styling, including tabular figures and slashed zeros.
  * Updated budget alert displays to format credit amounts consistently.

* **Documentation**
  * Added visual proof documentation covering numeric formatting across key console screens.

* **Tests**
  * Added coverage verifying the updated numeric styling configuration.


sakibsadmanshajib (commented):
Reviewed the diff, the new test, and the embedded screenshots (pixel-cropped several to confirm the claimed glyph change is real, not just described).

**Screenshots check out.** Cropped `06/07-catalog-table-*` and `04/05-credits-*` at 3-4x: the "before" zero is a rounded, O-like glyph and the "after" zero is genuinely slashed. `01-signin-mark-light.png` shows the real Hive icon (not a lettered placeholder), and `08-console-overview-after-dark.png` shows both the mark and the slashed zeros holding in the dark palette. No fabricated or misleading capture here.

**Gap: the "choke point" claim doesn't cover every dense numeric surface.** `components/ui/data-table.tsx` is only the choke point for tables built on the shared `DataTable` component (confirmed `analytics-table.tsx` routes through it and now gets `.metric`). Several numeric displays are hand-rolled `<table>`/`<dd>` markup and still use bare `tabular-nums` on Geist Sans, i.e. still exhibit the exact zero-vs-O ambiguity this PR is fixing elsewhere:

- `apps/web-console/components/billing/invoice-row.tsx:45` — the invoice total in taka (`subunitsToTakaDisplay`, a real BDT money figure that legitimately renders `৳0.00` on the parse-failure fallback and can contain `0` digits in normal totals). This is arguably a more customer-facing money figure than the catalog price columns fixed in this PR, and it's not on `.metric`.
- `apps/web-console/app/console/billing/alerts/page.tsx:98` — `{alert.threshold_pct}%` in a hand-rolled table, still `tabular-nums` only.
- `apps/web-console/components/billing/budget-alert-banner.tsx:52` — the banner message embeds `formatCredits(threshold.threshold_credits)`, the same helper/data as the credits caption fixed in `page.tsx`/`billing-overview.tsx`, but the banner text itself is still plain `tabular-nums`.

Worth at least a follow-up ticket if these are out of scope for this PR; as written, the PR description's "one line at the choke point every dense table in the console already routes through" overstates coverage.

**Minor:**
- `data-table.tsx:49` (header `<th>`) is still `col.numeric && "tabular-nums"`, not `metric`, while the body `<td>` at line 85 moved to `metric`. Fine in practice (headers are text labels, not data, so no zero to slash), just flagging the asymmetry so it isn't mistaken for an oversight later.
- `api-key-create-form.tsx:155` `tabular-nums` on a formatted date string is low-priority for this fix (dates aren't the ambiguous case), no action needed.

**Test quality:** `tests/unit/metric-slashed-zero.test.ts` is not a shallow snapshot — it regexes the actual `.metric {}` block out of `app/globals.css` and asserts both the `font-variant-numeric` and `font-feature-settings` declarations are present, so dropping either fails the guard. Two caveats, neither blocking: the `font-variant-numeric` match (`slashed-zero\s+tabular-nums`) is order-sensitive, a harmless property reorder would false-fail it; and it can only verify source text, not a rendered/computed glyph, since there's no browser/jsdom environment in this suite. Reasonable ceiling given the tooling.

**Accessibility:** no concern. `font-variant-numeric`/`font-family` are purely visual OpenType/rendering properties; screen readers announce the underlying DOM text, unaffected by which font or numeric feature draws it.

**TypeScript/React idiom:** the three component diffs are trivial className routing changes, no type or async issues introduced.

Net: the fix is correct and well-evidenced for what it touches, but incomplete relative to its own stated scope. Recommend a fast follow-up for `invoice-row.tsx` at minimum (real money, real zeros) before calling the defect class closed.

PR #472, opened 2026-07-27
fix: make the agent workspace launcher read as native Open WebUI chrome

sakibsadmanshajib:
Follows up #461, which shipped a working agent-workspace launcher whose styling, viewport gate, and retry ceiling were all set by guesswork. This replaces each of the three with a value measured off the live pinned Open WebUI image. Decision D-013 stands: no fork of Open WebUI, so this polishes the existing body-level launcher rather than chasing a real nav slot.

Every screenshot below was taken with Playwright against the real stack (`--profile local --profile chat`), signed in through Open WebUI's own form. Each `BEFORE` was produced by checking `origin/main`'s `loader.js` and `custom.css` back into the same running stack (both are bind-mounted, so no rebuild) and re-shooting the same viewport: same browser, same session, one variable. Capture conditions are recorded in `docs/proof/owui-launcher-2026-07-27/README.md`.

---

## 1. Visual parity with Open WebUI's own header buttons

The launcher drew a white bordered pill with a drop shadow. Nothing else in that header does. It read as an overlay pasted on top of the app rather than a control the app shipped.

Its rule now carries Open WebUI's own values, read off the native header icon button (`flex cursor-pointer px-2 py-2 rounded-xl hover:bg-gray-50 dark:hover:bg-gray-850 transition`) rather than approximated:

| Property | Before | After | Open WebUI's native button |
|---|---|---|---|
| padding | `0px 12px` | `8px` | `8px` |
| border-radius | `8px` | `12px` | `12px` |
| border-width | `1px` | `0px` | `0px` |
| background | `rgb(255, 255, 255)` | `rgba(0, 0, 0, 0)` | `rgba(0, 0, 0, 0)` |
| box-shadow | `0 1px 2px …` | `none` | `none` |
| colour, light | `rgb(23, 23, 23)` | `oklch(0.51 0 0)` | `oklch(0.51 0 0)` |
| colour, dark | `#ededed` | `oklch(0.77 0 0)` | `oklch(0.77 0 0)` |
| hover, light | border tint | `oklch(0.98 0 0)` | `oklch(0.98 0 0)` |
| hover, dark | `#2a2a2a` | `oklch(0.27 0 0)` | `oklch(0.27 0 0)` |
| transition | `120ms` | `150ms` | `150ms` |
| icon | 16px | 18px | 18px, stroke-width 1.5 |

**Beside Open WebUI's own icons at 1440px, before and after**

 

**Hover state: Open WebUI's own Controls button, then the launcher.** Both settle on `oklch(0.98 0 0)`, asserted equal in the spec rather than eyeballed here.

 

**Full header at 1440px, before and after**

**Dark theme, before and after**

One deliberate divergence: Open WebUI's header buttons carry `focus:outline-hidden` and show nothing on keyboard focus. Not copied. A focusable control with no visible focus state is an accessibility defect, not a house style worth matching, so the launcher keeps a `currentColor` focus ring.

---

## 2. Viewport coverage: gate moves from 1024px to 768px

#461 hid the launcher entirely below 1024px, which left every tablet and small-laptop width with no discoverable entry point at all, only the per-message Action buried in chat history. That 1024px was not measured against anything.

**At 900px, before and after.** Before: nothing. After: icon-only, sitting as the first of four header icons.

**At 768px, the new floor, with the seeded default model selected**

**At 1024px, still labelled**

**Dark, icon-only, at 768px**

### Why 768px and not 375px

Open WebUI's right-hand icon cluster never collapses or reflows: measured at twelve widths from 1440px down to 375px, its left edge stays pinned 116 to 118px from the viewport's right edge. So the launcher's `right: 7.5rem` clears it everywhere, and on that evidence alone 375px would work.

The binding constraint is elsewhere. Open WebUI's model selector is unclamped, computed `max-width: none` and `overflow: visible`, so its left-hand header group grows with the length of the selected model's id. Measured right edge of that group against the launcher's left edge:

| Viewport | Unselected | Seeded default (24 chars) | Longest realistic OpenRouter id (44) | Launcher left edge |
|---|---|---|---|---|
| 1440px | 237 | 327 | 520 | 1163 |
| 1024px | 237 | 327 | 520 | 747 |
| 768px | 237 | 327 | 498 | 614 |
| 640px | 218 | 308 | 479 | 486 |
| 600px | 218 | 308 | 458 | 446 |
| 480px | 218 | 286 | 338 | 326 |
| 375px | 196 | 233 | 233 | 221 |

768px clears a 44-character id by 116px, now screenshotted rather than only tabulated:

 640px clears it by 7px, and from 600px down the two genuinely collide. **A collision is not cosmetic here:** this element paints above Open WebUI's, so it would swallow clicks meant for the model selector. That is the reason this stops at 768px rather than pushing to 375px on the strength of the placeholder-width measurement alone.

At 375px the launcher stays absent and the per-message Action remains the entry point, exactly as before. Going lower is not a matter of finding a better offset: the header has no free span left once Open WebUI's own selector expands, so it needs Open WebUI to clamp that selector, which means a fork or an upstream fix.

Below 1024px the text label drops, leaving the icon-only shape Open WebUI's own header is built from, with a `title` attribute carrying the name.

---

## 3. Idempotency guard and the token race

**The guard was already there** and has not changed: `inject()` returns early on `document.getElementById(ID)`. What was missing was evidence that it holds against SvelteKit re-rendering, so that is now asserted rather than argued: `/` to `/notes` to `/workspace/models` and back, plus the sidebar drawer and the settings modal, all with **zero document loads**, and the launcher present **exactly once** at every hop.

 

**The retry ceiling was a real defect, and this is the part worth reviewing.** #461 flagged the bounded poll for review, and the bound turned out to be wrong. Open WebUI signs in **without a document load** (confirmed live: zero `load` events, the SPA just routes `/auth` to `/`), so `loader.js` never runs again and its poll is the only thing that can notice the token appearing. With the original ~6s ceiling, anyone who spent longer than six seconds on the sign-in screen, which a password manager or simply reading the page will do, lost the launcher **for the entire session** until a manual reload.

**Signing in 15 seconds after page load, before and after**

The poll now runs a fast phase for hydration and then a slow 1s phase that outlives a human sign-in. Its stop condition is `document.hidden`, not an elapsed-time ceiling: review correctly pointed out that a deadline measured from page load is the same defect at a larger number, since whatever the value, someone who sits on the sign-in screen longer than it still loses the launcher. So the poll stops when the tab is backgrounded and resumes on `visibilitychange`. An abandoned tab keeps no timer alive; a tab someone is looking at keeps waiting. Each attempt is two synchronous property reads and never touches Open WebUI's DOM.

Sign-out needs no counterpart: it does trigger a full document load, so the launcher disappears with the page. Also verified, which is why no removal path was added.

---

## Verification

| Check | Result |
|---|---|
| `09-agent-workspace-launcher.spec.ts` against the live stack | 4 passed |
| Native parity, all properties plus settled hover colour | Asserted equal to Open WebUI's own button, read at runtime rather than hardcoded |
| Overlap at 1440, 1024, 900, 768px | Zero intersecting header elements at each |
| Label gate | Shown at 1440 and 1024, hidden at 900 and 768, launcher hidden at 375 |
| Client-side navigation | Exactly one launcher across 3 route changes, drawer, and modal, with 0 document loads |
| Late sign-in guard, negative case | Fails with `element(s) not found` when `loader.js`'s retry schedule is reverted to `origin/main` |

The new spec reads Open WebUI's values off a live native button instead of hardcoding them, so an upstream restyle surfaces here as a mismatch rather than leaving the launcher quietly out of step. It sits in the existing nightly-gated `phase-19/owui` project.

## Not done, and why

- **A real sidebar nav item.** Decision D-013: Open WebUI exposes no plugin nav slot, and a permanent upstream-tracking burden for one nav link is not worth it.
- **Launcher below 768px.** Needs Open WebUI to clamp its own model selector. Measurements above.
- **A live model completion in the proof set.** The model list was empty for this run: the Hive shim rejects the local throwaway Open WebUI account because it has no tenant mapping. That is an environment data gap, not something this change touches, and the launcher is not on the completion path. Client-side navigation was proven with real Open WebUI route changes and its settings modal instead, each asserted against a `load`-event counter, rather than with a chat send that could not complete.


sakibsadmanshajib (commented):
## Review: launcher native polish

Reviewed against the four things this kind of change usually gets wrong: drift tolerance in a DOM injection script, poll correctness, whether the breakpoint is real CSS, and whether the spec exercises the new logic or only presence.

### What holds up

**`loader.js` is genuinely fail closed.** It queries zero Open WebUI selectors: the only DOM reads are `document.body` and `document.getElementById(ID)` (`deploy/docker/owui-static/loader.js:59`), and `signedIn()` wraps `localStorage` in try/catch (`loader.js:43-50`) so a privacy mode that throws on storage access degrades to "no launcher" instead of throwing out of the IIFE and taking the rest of the script with it. Upstream markup drift cannot break this script because it never depends on upstream markup. That is a materially better posture than the MutationObserver treadmill the header comment rejects, and the comment now says why.

**The poll is correct.** `attempt()` (`loader.js:121-134`) is a self rescheduling `setTimeout` chain, not `setInterval`, so there is exactly one pending timer at any moment and nothing to clear: it terminates by simply not rescheduling, on either the node existing or the elapsed ceiling (`loader.js:124-128`). No unbounded interval, no leaked handle, no second chain (`attempt()` is called once, and a duplicate script tag would still be idempotent through the `getElementById` guard). Phase boundary arithmetic matches the comment: 14 gaps of 400ms before the 1s phase, so approximately 5.6s of hydration cover.

**The breakpoint is real CSS.** `@media (min-width: 768px)` for display and `@media (max-width: 1023px)` for the label (`custom.css:136` and `custom.css:144`), with `display: none` as the base value. No `matchMedia`, no resize listener, no JS width branch, so there is nothing to desynchronise. Correct choice.

Measured values with the derivation recorded in the file, and a focus ring deliberately not copied from upstream, are both the right call. The focus ring passes non text contrast in light and dark at `currentColor`.

### Findings

**1. The overlap assertion does not exercise the constraint that sets the 768px floor.** This is the one I would act on. The CSS derivation is explicit that the binding constraint is Open WebUI's unclamped model selector, that clearance at 768px is 116px, and that the clearance figure assumes a 44 character model id. It is also explicit that a collision is not cosmetic, because this element paints at `z-index: 30` (`custom.css:82`) over Open WebUI's own header and would swallow clicks meant for the selector. The spec's sweep (`apps/web-console/e2e/phase-19/owui/09-agent-workspace-launcher.spec.ts:79-113`) measures overlap with whatever label the selector happens to show in that session, which for an unselected or default selector is the widest clearance case in your own table. So the assumption that makes 768px safe is documented but not asserted, and the failure mode if a longer id ever appears in the catalog is swallowed clicks rather than a visual glitch. Suggest driving the sweep with the longest label you want to support, either by selecting the longest id present or by setting the selector's text content in the page before measuring, so the table's 44 character premise becomes a test rather than a comment.

**2. The retry ceiling reintroduces the same defect class at a larger constant.** `startedAt` is set at document load (`loader.js:118`) and the chain gives up permanently at 10 minutes (`loader.js:117`). Sign in is the case being fixed, and the clock is not the sign in clock: a tab opened, left on `/auth`, and signed into later, which is ordinary behaviour for a tab opened in the morning, lands in exactly the state the PR describes, launcher absent for the whole session until a manual reload. Worth noting what the ceiling actually buys: the chain cannot outlive the document, and browsers throttle timers in hidden tabs to roughly once a minute, so an abandoned tab is not paying a measurable cost either way. Two options, both small. Delete `GIVE_UP_AFTER_MS` and the elapsed branch, which is a net one line removal with strictly fewer states. Or keep the ceiling and restart the chain from a `focus` or `visibilitychange` listener when the node is absent. Not blocking, since 10 minutes covers a normal sign in, but it is the same shape of bug as the one being closed.

**3. The spec is less drift tolerant than the code it guards, and can pass having tested nothing.** `spec.ts:141` clicks `button[aria-label="Open Sidebar"]` unguarded, so a markup change or a different persisted sidebar state fails the test hard, while the three route hops immediately below are `continue` guarded on `link.count() === 0` (`spec.ts:144-150`). The asymmetry runs the wrong way: the fragile step is the one that hard fails, and the steps that carry the actual client side navigation coverage are the ones that can vanish silently, leaving a test that asserts one launcher, zero document loads, and no navigation, and still goes green. Suggest counting completed hops and asserting at least one, so the coverage loss shows up as a failure instead of a pass.

**4. The one property that differs from native is the one the parity test does not compare.** `custom.css:101-103` sets `transition: background-color 150ms, color 150ms` with no timing function, so it resolves to `ease`, whereas Tailwind's `transition` utility on the native button uses `cubic-bezier(0.4, 0, 0.2, 1)`. The `shape()` comparison in `spec.ts` reads padding, radius, border width, background, shadow and colour, so this one is unguarded. Given the stated thesis of value for value parity, adding the timing function is a one line change, and adding `transitionTimingFunction` to `shape()` keeps it honest.

**5. Units are mixed in a way that undoes part of the derivation.** `top: 7px` (`custom.css:80`) is derived from a 34px button flex centred in a 48px band, but both of those are Tailwind rem values that scale with root font size, so the derived 7px stops being correct as soon as the root font size is not 16px, while `right: 7.5rem` on the next line does scale. Under a user set larger default font the shared baseline this change is specifically about drifts. Pick one basis for both, rem being the one that tracks upstream.

### Verdict

Approve in substance. The defect fix is real, the poll is correct, the gate is real CSS, and the injection script is honestly fail closed. Finding 1 is the one worth doing before merge, because it converts a documented assumption behind a click swallowing failure mode into an assertion. Findings 2 through 5 are small and can ride along or follow. Not merging as part of this review.

PR #598, opened 2026-07-28
feat: add metering Step 2 shadow-mode gate package

sakibsadmanshajib:
## Summary

Metering Step 2, Wave 1b: adds `apps/edge-api/internal/metering`, the new
shadow-mode gate package, plus its migration. This is the core of Step 2
per the implementation brief (vault `spec-2026-07-28-backend-metering.md`
section 9.2 Step 2, `.wolf/decisions.md` D-027).

**Shadow mode, enforces nothing.** `Gate.Execute` calls the caller's
dispatch func exactly once and unconditionally, then resolves a
precedence-ordered verdict (`billable` | `not_billable`, plus which
precedence rule fired) purely for logging. It never refuses, delays, or
alters what the wrapped dispatch returns: the returned error is the
dispatch func's own error, completely unmodified, whether the verdict is
`not_billable`, the billing account is unresolved, or the verdict-log write
itself fails. Every non-enforcement claim above has a dedicated test (see
Test evidence). Enforcement is Step 4, a separate future PR.

**Precedence order implemented** (`precedence.go`), first match wins:
1. `no_cost_basis` — route carries no price.
2. `enterprise_shadow` — tenant deployment posture is `ENTERPRISE_EDGE`.
3. `tenant_setting` — an explicit `ENABLE_USAGE_METERING` row exists.
4. `api_key_default` — API-key principal, billable off its own account.
5. `session_cloud_default` — session principal on `HIVE_CLOUD`, no explicit setting.
6. `resolved_default` — session principal, deployment posture unresolved.

**Per-model credit estimate is provisional.** `priceEstimate` computes
`estimated_credits_legacy` (today's flat `int64(total_tokens)`) and
`estimated_credits_per_model` (bigint `math/big`, credits-per-million-tokens,
round half up, floored at 1 for a billable request that produced tokens)
side by side, per the design brief section 3.5. The per-model rounding rule
is explicitly unresolved and blocks Step 4, not Step 2 — this PR labels it
provisional in code comments and only needs precedence-order correctness to
be gradeable now, which is the actual Step 2 pass condition. All money math
uses `math/big`; no float64 anywhere near credits or prices.

**Wired to nothing.** This PR is deliberately a self-contained,
fully unit-tested package plus a migration, adopted by nothing yet. No
existing file is touched. Wave 2 (`chat/dispatch.go`,
`inference/orchestrator.go`, `rag/chat_handler.go`, `rag/embed.go`) and
Wave 3 (`cmd/server/main.go`) adopt `metering.Gate` in separate, later PRs
per the parallel-wave build plan, so two agents editing those files right
now never collide with this one.

**Provider-blind by construction.** `RouteInfo.Provider` exists only for
structural parity with control-plane's `SelectionResult`; it is never
written to `metering_shadow_verdicts` (the migration has no `provider`
column) and never reaches a customer-bound surface.

## Migration

`supabase/migrations/20260728_04_metering_shadow_verdicts.sql` — additive
only, `BEGIN`/`COMMIT` wrapped, follows the observed same-day
`20260728_01`/`02`/`03` naming convention (`tenant_billing_accounts`,
the `ENABLE_USAGE_METERING` enum value, and the tenant-billing backfill,
all already merged). Creates one new table,
`public.metering_shadow_verdicts`, with two indexes and RLS enabled with no
policy/grant (service-role only, same posture as `tenant_billing_accounts`
from `20260728_01`). No existing table is altered. Safe to apply to a live
database while current code runs — nothing reads or writes this table
until a later PR wires `Gate` into a dispatch path.

## Test evidence

TDD, RED before GREEN. RED (before any production file existed):

```
$ cd deploy/docker && docker compose --profile tools run toolchain \
  "cd /workspace && go test ./apps/edge-api/internal/metering/... -count=1 -short"
...
apps/edge-api/internal/metering/gate_test.go:15:11: undefined: VerdictRecord
apps/edge-api/internal/metering/gate_test.go:35:12: undefined: Request
apps/edge-api/internal/metering/bodyrewrite_test.go:11:14: undefined: RewriteBody
...
FAIL	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/metering [build failed]
```

GREEN after implementation (gofmt clean, go vet clean):

```
$ go test ./apps/edge-api/internal/metering/... -count=1 -v
--- PASS (all 24 test cases across TestResolvePrecedence's 9 table rows,
    TestExecute_* x6, TestRewriteBody_* x6, TestPriceEstimate_* x4)
PASS
ok  	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/metering	0.019s
```

Whole-module regression check (no collateral damage):

```
$ go build ./apps/edge-api/... && go vet ./apps/edge-api/... && \
  go test ./apps/edge-api/... -count=1 -short
ok  	.../cmd/server ... (23 packages, all ok, including internal/metering)
```

`-race` could not be run in this sandbox: the Alpine toolchain image ships
`CGO_ENABLED=0` with no `gcc`/`cc`, and `-race` requires cgo. Noted rather
than silently skipped. This PR ships no goroutine and no in-process cache
(`PGTenantSettingsCache`/`PGBillingAccountResolver` do a plain indexed
query per call, `ponytail`-commented as the deliberate simplification), so
there is no shared mutable state in this PR's code for `-race` to catch.

Non-enforcement proof (the property the brief asks be tested explicitly,
not an afterthought):
- `TestExecute_AlwaysDispatchesExactlyOnceRegardlessOfVerdict` — 3 cases
  (not-billable/no-cost-basis, not-billable/explicit-opt-out, billable/
  billing-unresolved): dispatch called exactly once, `Execute` returns a
  nil error in every case.
- `TestExecute_NeverAltersCustomerVisibleOutcome` — forces
  `billing_not_configured`, asserts the error `Execute` returns is exactly
  the dispatch func's own error, unmodified.
- `TestExecute_VerdictLogWriteFailureIsSwallowed` — a failing
  `VerdictLogger` never surfaces as an error from `Execute`.
- `TestResolvePrecedence`'s two `billing_not_configured`/`billing_unavailable`
  cases assert the verdict stays `billable` (never becomes an error/refusal)
  even when the billing account can't be resolved.

## Test plan

- [x] `go test ./apps/edge-api/internal/metering/... -count=1 -v` (RED then GREEN, shown above)
- [x] `gofmt -l apps/edge-api/internal/metering` — clean
- [x] `go vet ./apps/edge-api/...` — clean
- [x] `go build ./apps/edge-api/...` — clean
- [x] `go test ./apps/edge-api/... -count=1 -short` — all 23 packages pass, zero regressions
- [x] `git diff --stat` confirms only the 9 new files in this PR's exact ownership (new package + new migration), nothing else touched
- [ ] CI green on this PR (to be watched via `gh pr checks --watch`)

### DIFF
--- a/apps/edge-api/internal/metering/bodyrewrite.go
+++ b/apps/edge-api/internal/metering/bodyrewrite.go
@@ -1,0 +1,55 @@
+package metering
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// RewriteBody forces stream_options.include_usage to true on an outbound
+// LiteLLM request body. This is the ONE customer-visible byte this package
+// is allowed to touch (design brief section 1, spec decision 10): without
+// it, the session-chat path's terminal usage block never arrives and shadow
+// mode grades itself green on zero, purely because it never saw a token
+// count -- not because nothing was billable. Adding a usage block to an
+// OpenAI-compatible streaming response is spec-conformant and not a
+// behavior change a client has to handle differently.
+//
+// Every other field the caller sent survives untouched, including any other
+// stream_options subfield. A denylist step (stripping client-supplied
+// fields the client should never set) was named in the design brief's
+// section 3.1 heading, but no concrete field list has been specified as of
+// this PR -- see PR body. Nothing is stripped here until one is; only
+// include_usage forcing is implemented.
+//
+// Callers decide when to invoke this (typically only for streaming
+// dispatches); it is a pure, allocation-only function with no I/O, so it
+// adds no synchronous network call to any request path.
+func RewriteBody(raw []byte) ([]byte, error) {
+	var fields map[string]json.RawMessage
+	if err := json.Unmarshal(raw, &fields); err != nil {
+		return nil, fmt.Errorf("metering: decode body: %w", err)
+	}
+
+	var streamOptions map[string]json.RawMessage
+	if existing, ok := fields["stream_options"]; ok && len(existing) > 0 {
+		if err := json.Unmarshal(existing, &streamOptions); err != nil {
+			return nil, fmt.Errorf("metering: decode stream_options: %w", err)
+		}
+	}
+	if streamOptions == nil {
+		streamOptions = map[string]json.RawMessage{}
+	}
+	streamOptions["include_usage"] = json.RawMessage("true")
+
+	rewritten, err := json.Marshal(streamOptions)
+	if err != nil {
+		return nil, fmt.Errorf("metering: encode stream_options: %w", err)
+	}
+	fields["stream_options"] = rewritten
+
+	out, err := json.Marshal(fields)
+	if err != nil {
+		return nil, fmt.Errorf("metering: encode body: %w", err)
+	}
+	return out, nil
+}
--- a/apps/edge-api/internal/metering/bodyrewrite_test.go
+++ b/apps/edge-api/internal/metering/bodyrewrite_test.go
@@ -1,0 +1,124 @@
+package metering
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+)
+
+func TestRewriteBody_ForcesIncludeUsageWhenAbsent(t *testing.T) {
+	raw := []byte(`{"model":"hive-fast","messages":[{"role":"user","content":"hi"}],"stream":true}`)
+	got, err := RewriteBody(raw)
+	if err != nil {
+		t.Fatalf("RewriteBody: %v", err)
+	}
+	assertIncludeUsageTrue(t, got)
+	assertFieldPreserved(t, got, "model", "hive-fast")
+}
+
+func TestRewriteBody_ForcesIncludeUsageWhenExplicitlyFalse(t *testing.T) {
+	raw := []byte(`{"model":"hive-fast","stream":true,"stream_options":{"include_usage":false}}`)
+	got, err := RewriteBody(raw)
+	if err != nil {
+		t.Fatalf("RewriteBody: %v", err)
+	}
+	assertIncludeUsageTrue(t, got)
+}
+
+func TestRewriteBody_PreservesEveryOtherField(t *testing.T) {
+	raw := []byte(`{"model":"hive-fast","max_tokens":512,"temperature":0.4,"tools":[{"type":"function"}]}`)
+	got, err := RewriteBody(raw)
+	if err != nil {
+		t.Fatalf("RewriteBody: %v", err)
+	}
+	var gotFields map[string]json.RawMessage
+	if err := json.Unmarshal(got, &gotFields); err != nil {
+		t.Fatalf("decode rewritten body: %v", err)
+	}
+	for _, key := range []string{"model", "max_tokens", "temperature", "tools"} {
+		if _, ok := gotFields[key]; !ok {
+			t.Errorf("rewritten body dropped field %q", key)
+		}
+	}
+}
+
+func TestRewriteBody_PreservesOtherStreamOptionsSubfields(t *testing.T) {
+	raw := []byte(`{"model":"hive-fast","stream_options":{"include_usage":false,"custom_field":"keep-me"}}`)
+	got, err := RewriteBody(raw)
+	if err != nil {
+		t.Fatalf("RewriteBody: %v", err)
+	}
+	var fields struct {
+		StreamOptions struct {
+			IncludeUsage bool   `json:"include_usage"`
+			CustomField  string `json:"custom_field"`
+		} `json:"stream_options"`
+	}
+	if err := json.Unmarshal(got, &fields); err != nil {
+		t.Fatalf("decode: %v", err)
+	}
+	if !fields.StreamOptions.IncludeUsage {
+		t.Errorf("include_usage = false, want true")
+	}
+	if fields.StreamOptions.CustomField != "keep-me" {
+		t.Errorf("custom_field = %q, want preserved", fields.StreamOptions.CustomField)
+	}
+}
+
+// TestRewriteBody_ByteIdenticalContentWhenAlreadyTrue is the diff-the-wire
+// regression named in the design brief (section 3.4 point 2): a client that
+// already sends stream_options.include_usage:true must come back out
+// content-equivalent -- same fields, same values, field ordering aside --
+// proving include_usage is the ONE thing this function is allowed to touch.
+func TestRewriteBody_ByteIdenticalContentWhenAlreadyTrue(t *testing.T) {
+	raw := []byte(`{"model":"hive-fast","stream":true,"stream_options":{"include_usage":true},"messages":[{"role":"user","content":"hi"}]}`)
+	got, err := RewriteBody(raw)
+	if err != nil {
+		t.Fatalf("RewriteBody: %v", err)
+	}
+
+	var want, gotDecoded any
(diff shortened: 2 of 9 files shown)


sakibsadmanshajib (commented):
Independent adversarial review of the pushed diff (verified against `origin/feat/metering-step2-gate` via `git archive`, not the builder's self-report; the main checkout was left on `main` throughout).

## Verdict: no blocking defects found. A few forward-looking notes below, none of which block this PR.

### 1. Non-enforcement claim: verified, not just asserted
`grep -rn "metering\." apps/edge-api --include=*.go | grep -v internal/metering/` and a separate grep for `metering.New`, `metering.Gate`, `RewriteBody` across `apps/` both return nothing outside the package itself. This package is genuinely wired to nothing: no `main.go`, no `chat/dispatch.go`, no other file imports or calls it. That is the strongest possible proof of non-enforcement for this PR: it is inert dead code from the binary's point of view.

Inside the package, `Gate.Execute` (gate.go:262-279) calls `dispatch(ctx)` first and unconditionally, before any verdict/pricing work runs, and returns `dispatchErr` completely unmodified. `TestExecute_NeverAltersCustomerVisibleOutcome` asserts this with `errors.Is`, not a weaker equality check. `writeVerdict`'s failure path only calls `slog.Warn`, never returns an error to the caller, and `TestExecute_VerdictLogWriteFailureIsSwallowed` proves it with a fake logger that always errors.

`RewriteBody` (bodyrewrite.go) is the one function that touches a customer-bound byte, and it is also uncalled anywhere in this diff. If a later wave wires it in, note it only forces `stream_options.include_usage=true` and round-trips every other field losslessly (`TestRewriteBody_ByteIdenticalContentWhenAlreadyTrue` uses `reflect.DeepEqual` on decoded content, which would catch key loss or reordering-as-corruption).

### 2. Hot-path safety: no concerns in this diff
No goroutines, channels, or `sync` primitives anywhere in the package (confirmed by grep). `Gate` is immutable after `New`, safe for concurrent reuse without a mutex. The one synchronous Postgres call in the package (`PGVerdictLogger.LogVerdict` / `PGBillingAccountResolver.Resolve`) only happens after `dispatch` has already run, matching the existing `InsertTrace`/`insertAuditEvent` convention in `chat/dispatch.go` and `chat/trace.go`, which already run their DB writes after the SSE stream has been fully flushed to the client.

Forward note for whichever PR wires this into `chat/dispatch.go` (non-blocking here since the package is inert): that wave will add a third synchronous DB write per request stacked after `InsertTrace` and `insertAuditEvent`. Worth a look at batching or async write at that point, not in this PR.

### 3. Money correctness
All the arithmetic in `priceEstimate` (precedence.go) is `math/big`; no float64 near credits or prices, consistent with the repo mandate. I hand-checked the round-half-up logic (`QuoRem` then compare `remainder*2` against the divisor) and the floor-at-one rule (only applied when `verdict == billable && legacy > 0`), including the exact numbers in `TestExecute_VerdictRecordShape` (100*8 + 50*24 = 2000, /1e6 rounds to 0, floored to 1): the math is correct.

The provisional-estimate labeling requirement is satisfied at three layers: the `priceEstimate` function doc comment, the `Outcome.EstimatedCreditsPerModel` field doc comment, and the migration's column comment (`-- provisional, unit TBD (spec section 12 item 1)`). Nothing here could be mistaken for an authoritative charge; it is a `metering_shadow_verdicts` log row only, never read by any billing/enforcement path.

### 4. Leak discipline: verified clean
Grepped the whole diff for `USD`, `usd`, `amount_usd`, `fx_rate`, `exchange` and for any read of `RouteInfo.Provider`: nothing. `Provider` exists on the struct for shape parity only, is never read, and the migration has no `provider` column, so there is no path for a provider name to reach `metering_shadow_verdicts` or any customer surface. No FX/USD language anywhere in this package.

### 5. Migration
Additive, `BEGIN`/`COMMIT` wrapped, follows the same-day `20260728_0{1,2,3}` naming convention, no existing table touched. Two indexes present (`(tenant_id, created_at DESC)` and `(verdict, created_at DESC)`), which cover the query shapes this table's own doc comment says it exists for (grading precedence order and pricing formula against real traffic). RLS enabled with no policy/grant, matching `tenant_billing_accounts`' posture from `20260728_01`.

One real gap, medium severity, not blocking this PR since nothing writes to the table yet: there is no retention or partitioning story. This is a row-per-request log table with two FK columns and a `created_at`, and nothing here bounds its growth once a later wave starts writing to it on every dispatch. Recommend a retention plan (partition by month + drop, or a scheduled delete job) land before or alongside the PR that actually wires `Gate` into a live dispatch path, not after volume is already a problem.

### 6. Test honesty: verified by re-running the tests myself
Exported the pushed branch tree with `git archive origin/feat/metering-step2-gate` into a scratch directory (no worktree, no branch switch on the shared checkout) and ran, via the repo's own Docker toolchain image:
- `go build ./apps/edge-api/...` — clean
- `go vet ./apps/edge-api/...` — clean
- `gofmt -l apps/edge-api/internal/metering` — clean
- `go test ./apps/edge-api/internal/metering/... -count=1 -v` — all 24 cases pass

The tests are not tautological. `TestResolvePrecedence`'s 9 rows assert exact rule names and verdicts per precedence tier against real fakes; `TestExecute_VerdictRecordShape` and the `priceEstimate` tests assert specific numeric outputs tied to the actual formula (32 vs 48 credits for two different alias prices, 150 for the legacy flat sum, 1 for the floor case). Gutting the implementation (e.g. skipping the floor-at-one branch, or reordering precedence checks) would fail these tests, not just pass them by construction.

`staticcheck` could not be run in this sandbox: `go.work` pins go 1.24.13 and the latest `staticcheck` needs go >= 1.25. Noting the gap rather than silently skipping it; `go vet` and the test suite are the coverage that's actually available here.

### 7. Idiomatic Go / error handling
`bodyrewrite.go` and `PGBillingAccountResolver.Resolve` wrap errors with `fmt.Errorf("metering: ...: %w", err)` correctly. `PGVerdictLogger.LogVerdict` returns the bare `pool.Exec` error unwrapped, but this exactly matches the existing convention in `chat/trace.go`'s `InsertTrace` (also a bare `return err`), and the one caller (`gate.go`'s `writeVerdict`) adds context via `slog.Warn` with `err` and `request_id` at the point of consumption. Not a regression introduced by this PR.

`nullableUUID`/`nullableString` in `verdictlog.go` duplicate identically named private helpers already in `chat/trace.go`. Unavoidable given they are unexported and metering is a separate package; a trivial nit, not worth blocking on.

No swallowed errors that should have been reported: the two intentionally-swallowed paths (verdict-log write failure, tenant-settings lookup failure) are both documented as deliberate shadow-mode fail-open behavior and both have a dedicated test proving the swallow is real and not accidental.

## Summary
No CRITICAL or HIGH findings. The non-enforcement claim holds: this package is provably unreachable from any live request path in this diff, and even in isolation `Gate.Execute` cannot alter what a wrapped dispatch returns. The migration is additive and safe to apply live, its only gap is a missing retention/partitioning story for a per-request log table, worth addressing before or alongside the wave that actually wires writes into it, not blocking here. Money math is `math/big` throughout and the provisional per-model estimate is visibly labelled as such in three places. Leak discipline (provider names, USD/FX) is clean by construction and by grep. Tests were independently re-run against the pushed branch and are not tautological.

Merge call is the requester's, not mine; I have not approved and have not merged.

PR #751, opened 2026-08-06
fix: make the chat-first login coherent

sakibsadmanshajib:
Three findings from the live adversarial review of `chat-hive.scubed.co`, smallest diff that actually reaches a running deployment.

## 1. The dead login form on chat

`https://chat-hive.scubed.co/auth` renders an email and password form above **Continue with Hive**. Every account on this deployment is SSO-only, so that form could only ever fail, and it failed misleadingly: `POST /api/v1/auths/signin` answered correct Hive credentials with HTTP 400 "The email or password provided is incorrect", which reads as a wrong password rather than as a form that was never wired to anything.

**`ENABLE_LOGIN_FORM` is live in v0.10.2, verified in the pinned image rather than assumed.** The files that prove it, read out of `hive-open-webui:v0.10.2-branded`:

| File | What it proves |
|---|---|
| `backend/open_webui/config.py:1617` | `ENABLE_LOGIN_FORM = os.getenv('ENABLE_LOGIN_FORM', 'True').lower() == 'true'`, exposed as `ui.enable_login_form` |
| `backend/open_webui/main.py:1918` | publishes it to the browser as `features.enable_login_form` on `/api/config` |
| `build/_app/immutable/nodes/50.*.js` | the login page gates the credential fields, the submit button and the "or" divider on `features.enable_login_form \|\| features.enable_ldap`; `enable_ldap` is already false here |
| `backend/open_webui/routers/auths.py:835` | rejects the password grant outright when the flag is off |

So it is nothing like `ENABLE_MODEL_FILTER`, which simply did not exist. **But the compose variable alone would still have been a silent no-op.** `ui.enable_login_form` is in `DEFAULT_CONFIG`, and `Config.seed_defaults` ("Insert keys that don't yet exist in the DB. Existing DB values take precedence over defaults") wrote it on this deployment's first boot; `Config.get` has preferred that row over the environment ever since. This is the same mechanism as issue #722, so the key joins the per-key reconcile built for it, `deploy/docker/owui-patches/hive_rag_env_config.py`. It is coerced to a JSON boolean on the way in, because the value reaches the browser raw and the string `"false"` is truthy in JavaScript, which would have rendered the very form this removes.

## 2. Neutral sign-in copy when arriving from chat

`/auth/sign-in` showed "Sign in to your console" and a subtitle about API keys, credits and usage analytics to someone who had just clicked **Continue with Hive** on a chat login screen. When `next` is an OAuth consent target it now reads "Sign in to Hive" / "Use your Hive account to continue." Direct console visits are unchanged.

The branch resolves through the existing allow-list in `apps/web-console/lib/auth/next-target.ts` rather than sniffing the raw param, so an unlisted or hostile `next` falls back to the console copy along with the redirect.

## 3. Submit progress state: no change needed

The existing `loading` state already covers the whole round trip. `handleSubmit` sets it before awaiting `signInWithPassword` and clears it only on the error path; the success path calls `navigate()`, a hard `window.location.assign`, so the button stays disabled and reading "Signing in…" until the browser leaves the document. Nothing redundant was added. The remaining wait the review measured is on the pages after this one.

## Which change needs a deploy

**Change 1 is configuration and image content, so it does nothing until the box is redeployed.** It needs both halves shipped together: the `ENABLE_LOGIN_FORM: "false"` line in `deploy/docker/docker-compose.yml`, and a **rebuilt** `open-webui` image, because the reconcile that makes that line reach an already-booted deployment is baked in at image build time. A compose restart against the old image will not do it. Verify afterwards with:

```
curl -s https://chat-hive.scubed.co/api/config
```

and check that `features.enable_login_form` is `false`. Changes 2 and 3 are web-console application code and ship with the console.

## Visual proof

Captured against locally running stacks from this branch, since this branch cannot deploy. Full notes in `docs/proof/chat-login-coherence-2026-08-06/README.md`. No real credential appears in any frame; the `authorization_id` in the fourth image is the literal string `PLACEHOLDER-NOT-A-REAL-ID` and no authorization request was ever created.

The two chat frames are the same image and the **same Docker volume**, restarted, which is the case that matters: the after-state container booted onto a volume where `ui.enable_login_form` had already been seeded `true`, and the reconcile overrode it.

```
before: "enable_login_form":true
after:  "enable_login_form":false

INFO | open_webui.config:seed_registered_defaults:52 - hive: reconciled Open WebUI
RAG config from env: ..., ui.enable_login_form=False
```

## Tests

`apps/web-console/__tests__/sign-in-next-redirect.test.tsx` gains three cases: the copy switches on an OAuth consent `next`, stays put on a direct visit, and stays put for an unlisted `next`. `scripts/test_owui_rag_env_config.py` gains three covering the boolean coercion and the unset case, and runs in CI through `make test-scripts`.

```
$ docker compose run --rm web-console npm run test:unit
 Test Files  1 failed | 36 passed (37)
      Tests  413 passed | 1 skipped (414)

$ docker compose run --rm web-console npx vitest run __tests__/sign-in-next-redirect.test.tsx
 Test Files  1 passed (1)
      Tests  12 passed (12)

$ docker compose run --rm web-console npm run build
 ✓ Compiled successfully

$ python3 scripts/test_owui_rag_env_config.py
ok: owui RAG env-config reconcile (issue #722)

$ npm run lint:proof-tokens
lint-no-token-in-proof-captures: ok (27 files under docs/proof/ scanned)
```

The one failing suite is pre-existing and unrelated: `tests/unit/control-plane-host.test.ts` reads `/app/.env.example`, which `Dockerfile.web-console` never copies (it copies only `apps/web-console/`). It fails identically on `main` in this container.

Worth knowing for anyone running that command: the `web-console` service bakes its source into `hive-web-console:ci` at build time and mounts nothing, so `docker compose run web-console npm run test:unit` silently tests whatever tree the image was last built from. The first run here reported 408 passing against a stale image that predated two tests already on `main`. `docker compose build web-console` first.

## Summary by CodeRabbit

* **New Features**
  * OAuth sign-in flows now display Hive-specific messaging, while direct console visits retain existing copy.
  * Email/password login is disabled in favor of OAuth sign-in.
* **Bug Fixes**
  * Login settings now correctly persist enabled and disabled states across deployments.
* **Documentation**
  * Added documentation covering login behavior, evidence, known cosmetic details, and reproduction steps.
* **Tests**
  * Added coverage for OAuth-specific messaging and login setting reconciliation.


deploy/docker/docker-compose.yml
```diff
@@ -613,6 +613,36 @@ services:
...
+      #
+      # Verified live in the pinned v0.10.2 image, not assumed (the
+      # ENABLE_MODEL_FILTER note below is why this gets checked now):
+      # backend/open_webui/config.py:1617 reads the variable, exposes it as
+      # `ui.enable_login_form`, and main.py:1918 publishes it as
+      # features.enable_login_form on /api/config; the login page bundle
+      # (build/_app/immutable/nodes/50.*.js) gates the credential fields, the
+      # submit button and the "or" divider on
+      # `features.enable_login_form || features.enable_ldap`, and
+      # routers/auths.py:835 rejects the password grant outright. With
```
sakibsadmanshajib: MEDIUM. Citation inaccuracy, does not invalidate the fix. The cited routers/auths.py line 835 in the pinned v0.10.2 tag is inside signup(), not signin(): it reads roughly "elif not await Config.get(ui.enable_login_form) and not ENABLE_INITIAL_ADMIN_SIGNUP: raise HTTPException(403, ...)", and it only fires when a new account is being created. It is not what produced the live bug's HTTP 400 on /signin. The actual /signin route (auths.py line 653) is gated by a different, untouched variable, ENABLE_PASSWORD_AUTH (config.py line 1621), which still defaults to True and is not part of this PR. So ENABLE_LOGIN_FORM=false does not close the password grant at the API, only the frontend rendering of the form, which is independently confirmed at main.py line 1918 plus the login bundle's enable_login_form or enable_ldap gate. Recommend correcting or dropping this sub-claim so nobody later assumes the endpoint itself got closed.

deploy/docker/docker-compose.yml
```diff
@@ -613,6 +613,36 @@ services:
...
+      # Like the four RAG keys below, this one is persistent config: Open
+      # WebUI's seed_defaults wrote `ui.enable_login_form` into its own
+      # database on this deployment's first boot, and the database has
+      # outranked the environment ever since, so this line on its own would
+      # have been a silent no-op on any already-booted box. It reaches a live
+      # deployment because owui-patches/hive_rag_env_config.py reconciles this
+      # key from the environment on every container start. Confirm after
+      # deploying with `curl -s https://<chat-host>/api/config` and check that
+      # features.enable_login_form is false.
+      ENABLE_LOGIN_FORM: "false"
```
sakibsadmanshajib, line 653: LOW. This env block, profiles local, enterprise and chat, applies the same way to Hive EnterpriseEdge self-hosted deployments as to the chat demo, same as the pre-existing ENABLE_SIGNUP false two lines above, so this is consistent with the file's existing SSO-only design, not a new product regression. Flagging anyway because it is a bare literal rather than an env-driven default: an Enterprise customer who ever needs a local or break-glass login, their IdP down, or no IdP at all, has no compose-level override without editing this file directly. Worth a follow-up ticket if that scenario matters; not a blocker here.

apps/web-console/app/auth/sign-in/page.tsx
```diff
@@ -66,11 +66,25 @@ export default function SignInPage() {
...
 
+  // A user who lands here mid OAuth consent came from a Hive product surface
+  // (chat today), not from the developer console, so the console pitch is both
+  // wrong and disorienting: they clicked "Continue with Hive" on a chat login
+  // screen and were answered with a page about API keys, credits and usage
+  // analytics. Resolving through the shared allow-list rather than sniffing the
+  // raw param means an unlisted or hostile `next` can never reach this branch --
+  // resolveNextTarget only ever returns "/oauth/consent" or
+  // "/oauth/consent?..." for genuine consent round-trips.
+  const fromConsent = resolveNextTarget(nextParam).startsWith("/oauth/consent");
```
sakibsadmanshajib, line 84: MEDIUM. fromConsent depends on nextParam, which starts null and is only populated by the useEffect a few lines above, deferred because window is unavailable during SSR. The exact user this PR targets, arriving via a next param pointing at oauth consent, will see the first render with the console-pitch title and subtitle, then have it swap to the Hive copy once the effect fires. The same deferred-nextParam pattern already exists for the Create one link, where a flash is low-visibility; here it drives the primary heading, so the flash briefly reproduces the exact confusing copy this PR removes. Consider reading next synchronously, for example via useSearchParams, or parsing location.search in the same pass that sets hydrated, so the correct heading is present on first paint.

apps/web-console/__tests__/sign-in-next-redirect.test.tsx
```diff
@@ -148,6 +148,53 @@ describe("app/auth/sign-in/page.tsx next-target redirect", () => {
...
+        await screen.findByRole("heading", { name: "Sign in to your console" }),
+      ).toBeTruthy();
+      expect(
+        screen.getByText(/Manage API keys, credits, and usage analytics/i),
+      ).toBeTruthy();
+    });
+
+    // The neutral copy is gated on the same allow-list that gates the redirect,
+    // so an attacker-supplied `next` cannot reach it either.
+    it("keeps the console copy for an unlisted next target", async () => {
```
sakibsadmanshajib, line 206: LOW. This case would pass identically against the pre-fix code: without fromConsent existing at all, the title is always the static console copy, so an unlisted next trivially keeps it too. It is a legitimate assertion that the security property continues to hold post-fix, but it does not pin anything new by itself, unlike the sibling OAuth consent target case two blocks up, which does fail without the fix. Not asking for a change, just noting it for anyone reading this as regression coverage.

deploy/docker/docker-compose.yml
sakibsadmanshajib: Correct, and the record is now fixed rather than dropped. Checked against the pinned v0.10.2 tag: line 835 sits inside `signup()` and only fires when a new account is being created, so it is not what produced the live HTTP 400 on `/signin`. The signin route at `routers/auths.py:653` is gated by `ENABLE_PASSWORD_AUTH` (`config.py:1621`), which still defaults to true and is deliberately untouched here.

The compose comment carried that same wrong claim, so it has been rewritten in a2b8f7e to state the scope explicitly: the flag removes the frontend form, not the endpoint, and `POST /api/v1/auths/signin` still accepts a password grant with nothing pointing at it. The frontend half of the citation stands unchanged, since it was confirmed in the pinned image at `config.py:1617`, `main.py:1918` and the login bundle's `enable_login_form || enable_ldap` gate.

Whether the endpoint itself should be closed with `ENABLE_PASSWORD_AUTH` is a separate decision and not one to make quietly inside this PR: it would also close the break-glass path that #754 is about.

deploy/docker/docker-compose.yml
sakibsadmanshajib: Agreed on both halves, and no code change here. It matches the pre-existing `ENABLE_SIGNUP: "false"` two lines above in the same shared env block, so this is the file's existing SSO-only design rather than a new Enterprise regression; nothing about the current Enterprise posture changes with this PR.

The gap you are pointing at is real though, and it is bigger than one literal: an Enterprise customer whose IdP is down, or who has no IdP wired up yet, has no compose-level override and would have to edit a tracked file that conflicts on every upgrade. Bolting an override on here would also be half a fix, because accounts on an SSO-only deployment have no Open WebUI password, so re-enabling the form alone still does not let anyone in.

Filed as #754 with the suggested shape (`${OWUI_ENABLE_LOGIN_FORM:-false}`, routed through the same per-key reconcile so it reaches an already-booted box) plus the initial-admin bootstrap question and whether `ENABLE_SIGNUP` deserves the same treatment.

apps/web-console/app/auth/sign-in/page.tsx
sakibsadmanshajib: Fixed in a2b8f7e. `next` is now read during render with `useSearchParams()` instead of from the effect, so the branch is decided on the very first render.

**Which render decides it, and why this is not a hydration mismatch.** The server render decides it. Every route in this app is dynamically rendered, because the root layout awaits `getLocale()` and next-intl's `getRequestConfig` awaits `cookies()`; the build output confirms it with `ƒ /auth/sign-in` under `ƒ (Dynamic) server-rendered on demand`. For a dynamically rendered route `useSearchParams()` resolves from the request URL on the server as well as on the client, so the server markup and the first client render compute the same `fromConsent` and emit identical markup. That also means no Suspense boundary is needed, which the build independently confirms: a statically prerendered route using this hook without one fails the build outright, and the build is green.

Verified on the running server rather than by reasoning alone, `next dev` from this branch:

```console
$ curl -s 'http://localhost:3010/auth/sign-in?next=%2Foauth%2Fconsent%3Fauthorization_id%3DPLACEHOLDER-NOT-A-REAL-ID' | grep -o '<h1[^>]*>[^<]*</h1>'
<h1 ...>Sign in to Hive</h1>
$ curl -s 'http://localhost:3010/auth/sign-in' | grep -o '<h1[^>]*>[^<]*</h1>'
<h1 ...>Sign in to your console</h1>
$ curl -s 'http://localhost:3010/auth/sign-in?next=%2Foauth%2Fconsent-evil' | grep -o '<h1[^>]*>[^<]*</h1>'
<h1 ...>Sign in to your console</h1>
```

The allow-list gate is unchanged, so the unlisted case still lands on the console copy.

You were right that the same deferred read drove the "Create one" link, and that is fixed by the same change rather than separately: the link carries `?next=` on the first paint now too. `handleSubmit` no longer parses `window.location.search` a second time either, it uses the same value.

Pre-hydration frames are in the PR comment below and committed under `docs/proof/chat-login-coherence-2026-08-06/`, captured with JavaScript disabled so what is photographed is exactly the server HTML with no React on top.

apps/web-console/__tests__/sign-in-next-redirect.test.tsx
sakibsadmanshajib: Strengthened in a2b8f7e, and the honest answer to "which assertion goes red" is worth stating plainly for this case: **none, on a plain revert.** Delete `fromConsent` and the heading is the console copy again, so an unlisted `next` reads green either way. No assertion on this input can distinguish pre-fix from correct-post-fix, because for an unlisted target the two behaviours are identical by design.

So it now pins the thing it actually can pin, the shape of the check. The case asserts that the pre-hydration markup does not contain the neutral heading, plus the console subtitle and the absence of the neutral heading after render. The regression that turns those red is the obvious later rewrite, `nextParam?.startsWith("/oauth/consent")` instead of resolving through the allow-list, which matches `/oauth/consent-evil` and hands an attacker-chosen target the neutral copy. Verified by running the suite against exactly that variant:

```text
x headline copy follows the journey > keeps the console copy for an unlisted next target
  AssertionError: expected '<main class="min-h-screen w-full flex...' not to contain 'Sign in to Hive'
Tests  1 failed | 14 passed (15)
```

The redirect half of the same allow-list is already pinned on the same input by "ignores an unlisted next target and falls back to /console", so it is not duplicated here.

Three cases were also added that do fail on a plain revert, pinning the finding 3 fix. They assert on `renderToStaticMarkup`, which is the only place a regression to the effect-based read is visible: every client-side assertion in this file passes either way, because `findBy*` waits and by then the effect has fired. Against the reverted `page.tsx`, served from the same image:

```text
x headline copy follows the journey > decides the copy on the first render, before any effect runs
  AssertionError: expected '<main class="min-h-screen w-full flex...' to contain 'Sign in to Hive'
x headline copy follows the journey > carries next into the 'Create one' link on the first render
  AssertionError: expected '<main class="min-h-screen w-full flex...' to contain '/auth/sign-up?next=%2Foauth%2Fconsent...'
Tests  2 failed | 13 passed (15)
```

Green on this branch: 15 passed.

docs/proof/chat-login-coherence-2026-08-06/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+Both frames come from `hive-open-webui:v0.10.2-branded` built from this branch.
+The after-state container was restarted onto the volume the before-state
+container had already seeded, which is the case that matters: Open WebUI's
+`Config.seed_defaults` writes `ui.enable_login_form` on first boot and the
+database outranks the environment from then on, so the compose variable alone
+would have been a silent no-op on the demo box. Its startup log shows the
+reconcile winning:
+
+```
```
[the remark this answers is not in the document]
sakibsadmanshajib: Done in a2b8f7e. Both blocks are tagged `text`. The startup log is plain text as suggested; the `/api/config` block is tagged `text` rather than `json` because it is not valid JSON, it is a two-line before/after excerpt of one field. MD040 only needs a language, and `text` is the honest one.

Issue #762, opened 2026-08-06
Tracking: operator toggles that do nothing, sixteen tenant settings keys with no runtime reader


sakibsadmanshajib:
## Summary

A code audit of the operator control surface found that most toggles rendered in the console write to `public.tenant_settings` successfully and are read by nothing at runtime. Sixteen of the twenty six registered gate keys have no production reader. The console is not lying about the database state; the row it writes is correct. It is lying about the system's behaviour, which is what an operator and a regulated buyer actually care about.

This issue tracks the individual defects and states the rule that should follow so this class does not recur.

## Tracked issues

* #755 — Audit sink toggles are read from environment variables, not tenant settings. Six keys. Compliance integrity: audit export can be shown as enabled while exporting nothing.
* #756 — Seven billing feature gates have no runtime reader. Payment rails cannot be disabled from the console; `ENABLE_PER_USER_CAP` advertises a spend control that was never built.
* #757 — Three SSO gates have no mounted reader. `RequireAny` exists, is tested, and is wired into no route.
* #758 — Admin and RAG scope gates have no reader and are excluded from the propagation path, so no edge-api reader could see them even if one were written.
* #759 — `Resolver.StartListener` is never called, so the `tenant_settings_notify` trigger fires into an empty channel and the invalidation path is dead.
* #760 — Member role changes do not take effect until the target's access token refreshes, and the console role change writes `account_memberships` while the claim is derived from `tenant_users`.
* #761 — `tenant_model_visibility` has a real runtime reader and no operator write surface. The one control that would work cannot be reached from the console.

## The common root cause

Every one of these passed review because something in the change read as verification without being verification:

* A test that asserts the setting was written, not that behaviour changed (#755, #756).
* A test that exercises a helper no route is wrapped in (#757, and the `RequireAny` case is the clearest example in the set).
* A gate constant defined next to three enforced siblings and never mounted (#758, `ENABLE_RELAY`).
* A subsystem wired end to end except for its final call site (#759).
* Two tables that look interchangeable and are not (#760).

The console renders whatever the registry returns. Adding a row to `public.feature_gate_keys` is a migration only change by design, which is a good property for propagation and a bad one for accountability: it means a key becomes a visible operator control with no code review of whether anything reads it.

## The rule that should follow

Two parts, both enforceable.

**1. A setting with no runtime reader must not be rendered as a control.**

[...]

Issue #552, opened 2026-07-27
api-keys/[id]/limits crashes for every key: a Server Component fetches a relative URL


sakibsadmanshajib:
**What happens**: `getKeyLimits`/`updateKeyLimits` call `client.fetch()` with a bare relative path
(`/api/v1/accounts/current/api-keys/{id}/limits`) instead of prefixing it with the control-plane
`baseUrl`, unlike every other call in `lib/control-plane/client.ts`, which all consistently prefix
with `${baseUrl}` from `getRequestContext()`. In a Node.js Server Component this throws
`TypeError: Failed to parse URL from /api/v1/...` -- reproduced directly in this session with
`node -e "fetch('/api/v1/test')"` on the same Node version this repo's Docker image uses. The
page's own `catch` only special-cases messages containing "404" (`page.tsx:41-49`), so this
`TypeError` is not caught and rethrows, crashing the page. No local Next.js API route exists at
that path either (confirmed by directory listing), so there is no proxy that would make the
relative path resolve.

**Evidence**: CODE — `apps/web-console/lib/api-keys.ts:107-122` (`getKeyLimits`, relative path, no
`baseUrl`), `apps/web-console/app/console/api-keys/[id]/limits/page.tsx:20-21,42-46`
(`serverFetchClient` passing the relative path straight to global `fetch`). Verified live in this
session: Node's `fetch()` throws `TypeError: Failed to parse URL from /api/v1/test` for a relative
URL with no base.

**Why it matters**: Per-key rate limits is a real, built feature that becomes completely
inaccessible for every key, every time, once a user reaches it.

**Smallest fix**: Prefix the fetch URL with `${process.env.CONTROL_PLANE_BASE_URL}`, matching every
sibling call in `client.ts`, or route `getKeyLimits`/`updateKeyLimits` through the same
`getRequestContext()` helper the rest of the client functions use.

**Source**: Verified directly against current `main` in this session (checklist item flagged for
confirmation; not directly named with this exact file:line in the six source audit documents).

PR #768, opened 2026-08-06
fix: make the console's money and credential operations reachable (#764, #765, #766, #767)

sakibsadmanshajib:
Fixes #764
Fixes #765
Fixes #766
Fixes #767

Cross reference #762.

A live adversarial pass found the console could not perform its core money and credential operations. Reproduced in a browser as a tenant owner: saving a hard spend cap, creating a spend alert, creating an API key and revoking an API key all returned 404. Tracing the siblings turned up a fourth: the checkout modal could neither list payment rails nor start a checkout.

## Two distinct defects

The report noted two shapes and asked that one cause not be assumed to explain all of them. There are indeed two, and the leading hypothesis for the first pair turned out to be wrong.

### Backend, a route that was never mounted (#764, #765)

**The identifier hypothesis is refuted.** The console sends `viewer.current_account.id`, and `GetMembershipRole` keys this surface on `public.accounts.id` (`apps/control-plane/internal/platform/role_pgx.go:47`). The value on the wire was already the one the backend expects. The workspace switcher supplies the same account id, confirmed live: its option values are account ids and the first is the current account.

The real cause is that `budgets.Handler.ServeHTTP` has dispatched `/api/v1/budgets/` and `/api/v1/spend-alerts/` since Phase 14 (`apps/control-plane/internal/budgets/http.go:70-73`), but `apps/control-plane/internal/platform/http/router.go:207` only ever registered the two legacy account paths. With no pattern for those prefixes, every request fell through to the `/api/v1/` catch-all and came back 404. The handler was correct and unreachable.

This also explains the counter-evidence in the report. Calling the control-plane directly worked because the caller chose the path; the console could not, because the router never offered it.

The fix is two `mux.Handle` lines. It is a wiring bug, not a contract change, so it belongs in this PR: the alternative client-side change would mean rewriting the console onto the legacy `/accounts/current/budget` threshold surface, which has a different shape and no hard cap.

### Client, a path the app never served (#766, #767)

Four Client Components `fetch` a relative `/api/v1/accounts/current/...` path. That namespace belongs to the control-plane. There is no route handler under `apps/web-console/app/api/v1/`, no `rewrites` in `next.config.ts`, and no proxy in `middleware.ts`, so Next served its own 404 HTML page, which is exactly the shape the report saw.

| Component | Path |
| --- | --- |
| `api-key-create-form.tsx:70` | `POST .../api-keys` |
| `revoke-confirm-panel.tsx:32` | `POST .../api-keys/{id}/revoke` |
| `checkout-modal.tsx:50` | `GET .../checkout/rails` |
| `checkout-modal.tsx:130` | `POST .../checkout/initiate` |

The typed server client already had correct `createApiKey`, `revokeApiKey`, `getCheckoutRails` and `initiateCheckout`. Nothing browser-side was wired to them.

Rather than patch four call sites, this adds the seam those callers already assumed existed: one route handler with an **explicit allowlist** of operations. The allowlist is deliberate. A blanket pass-through would expose every present and future account route to the browser, which is a wider change than this bug justifies.

### Per-key limits

The report listed per-key limits alongside revoke. It does **not** share this defect: limits already go through a server action (`app/console/api-keys/[id]/limits/page.tsx:66`) calling `updateApiKeyLimits` server-side, which #552 fixed. The likeliest reading is that limits looked broken because no key could be created to open a limits page for. Called out as unverified rather than claimed fixed.

## Tests that would have caught it, and the ones that did not

`router_budgets_test.go` asserts dispatch through `NewRouter`, not against the handler. Reverting the two mount lines makes it fail with the exact reported symptom:

```
--- FAIL: TestNewRouterMountsWorkspaceBudgetSurface/put_workspace_budget
    PUT /api/v1/budgets/... = 404, want the budgets handler to be reached; the route is not mounted on NewRouter
--- FAIL: TestNewRouterMountsWorkspaceBudgetSurface/post_spend_alert
    POST /api/v1/spend-alerts/... = 404, want the budgets handler to be reached; the route is not mounted on NewRouter
```

`browser-api-routes.test.ts` walks the real files and asserts every relative `/api/...` a Client Component fetches has a matching route handler. Removing the new route file makes it fail on precisely the four broken paths and pass for the ten that were fine, so it is not a hand-maintained list that can drift.

**What gave false comfort, and why it could not go red:**

- `apps/control-plane/internal/budgets/http_test.go` drives these paths with `httptest.NewRequest` straight at `budgets.Handler`. It never asks whether anything routes there. `router_providers_test.go` already documents this exact failure mode for the admin providers surface; the lesson had been written down and not generalized.
- `apps/web-console/__tests__/api-key-limits-client.test.ts` asserts the **server** client builds an absolute `${BASE_URL}/api/v1/...`. That is a different code path from the browser fetch and is correct either way, so it was structurally incapable of failing.
- `tests/e2e/console-budgets.spec.ts` says in its own header comment that the save round trip is "deferred to a CI run against a workspace where the verified tester is an owner". The only operation that matters was never attempted. `console-spend-alerts.spec.ts` is the same shape.
- `app/console/billing/budget/page.tsx:30` reads the budget with `.catch((): null => null)`, so the read failing rendered as an empty form rather than an error.

## Verification, real output

Go, in Docker:

```
go test ./apps/control-plane/... -count=1 -short
EXIT=0    52 packages ok, no FAIL
```

Web console, image rebuilt first because that service bakes source and mounts nothing:

```
docker compose build web-console
docker compose run --rm web-console npm run test:unit
Test Files  1 failed | 38 passed (39)
     Tests  434 passed | 1 skipped (435)
```

The one failing suite is `tests/unit/control-plane-host.test.ts`, pre-existing and unrelated: it reads `/app/.env.example`, which is not copied into the image. Untouched by this change.

```
docker compose run --rm web-console npm run build
✓ Compiled successfully
ƒ /api/v1/accounts/current/[...path]    184 B    102 kB
```

## Visual proof

Against a locally running stack built from this branch, signed in through the real sign-in form as a workspace owner. Every capture carries an overlay naming origin, path and capture time, because a headless screenshot has no address bar.

Backend half, with a real session bearer:

```
PUT  /api/v1/budgets/{ws}          -> 200
POST /api/v1/spend-alerts/{ws}     -> 201
PUT  /api/v1/unmounted-surface/x   -> 404   (control: an unmounted path still 404s)
```

The control line shows the change mounts two specific prefixes rather than turning `/api/v1/` into a catch-all.

**Hard spend cap saved**

budget saved

**Spend alert created**

spend alert created

**API key created**

api key created

**API key revoked, surviving a reload**

api key revoked

No credential is in the pixels or the text. The created key secret was masked in the DOM before the capture, and every key created during the run was revoked immediately after, which the last image shows.

## Input validation on the new seam

Self-review of the new route turned up a latent issue worth naming, since the route is what makes the id browser-reachable in the first place. `revokeApiKey` interpolated the key id into an upstream URL without encoding it, unlike `keyLimitsUrl` immediately below it. An id carrying a path separator would have steered the request at a different control-plane path while still carrying the caller's own bearer.

The route now requires the segment to be a UUID and rejects anything else before the upstream call, and the client encodes the id as well so the property does not rest solely on the caller validating first. `api-keys-proxy-route.test.ts` covers it, including the unauthenticated case and a non-allowlisted operation. Removing the guard turns its four traversal cases red.

## Not proven here

- Checkout rails and initiate are fixed by the same route handler. They are now covered by unit tests, see the round two section below; the original wording here claimed coverage that did not exist. Neither was exercised end to end against a live payment rail. The 404 is gone; a full checkout round trip needs rail credentials.
- Per-key limits was not reproduced as broken and is not changed by this PR, per the reading above.
- Captures are against a local stack, not the deployed box. The control-plane change needs a deploy before the deployed console behaves this way.

## Review round two, 2026-08-08

An independent security review returned CHANGES REQUESTED. It found no cross tenant reach, the allowlist survived eleven crafted escape attempts, identity could not be spoofed, and no credential reflected into a response. Three things needed changing. All three are fixed here, and main is merged in.

### 1. Money bound, blocking

`credits` was checked for type, finiteness, integrality and sign, but never for magnitude. The reviewer posted 9007199254740993 and got a 200 with 9007199254740992 forwarded: `Number.isInteger` is true for that value, but it is not representable, so the purchase quantity was silently rounded before it left the process.

The console route now applies the same `Number.MAX_SAFE_INTEGER` bound the sibling money proxy at `app/api/budget/[workspaceId]/route.ts` already carries, and rejects with 400 rather than coercing.

The control-plane needed it too, because a client that skips the console reaches `InitiateCheckout` directly, and a value past 2 to the 53rd decodes cleanly into `int64` after the sender's float64 has already rounded it. `GetCheckoutOptions` has always advertised a per rail maximum that nothing enforced, so the maximum was decoration. `ValidatePurchaseAmount` now takes the rail and applies `maxCreditsForRail`, and `classifyInitiateError` maps the refusal to a 400 carrying no rate and no amount.

### 2. Unreachable error branch, blocking

`err instanceof ControlPlaneError` was always false on this route: all four client functions threw a plain `Error`, so every upstream refusal became a 502 and a member without permission looked identical to an outage. `createApiKey`, `revokeApiKey`, `getCheckoutRails` and `initiateCheckout` now use `throwControlPlaneError`, which is what the class exists for and what the sibling proxy routes already rely on. This route is their only caller, so nothing else changes shape.

### 3. Checkout coverage, blocking, and a corrected claim

**The earlier claim that checkout was "covered by the new test" was wrong.** The route handled `checkout/rails` and `checkout/initiate`, and no test exercised either. The reviewer proved it by deleting the entire credits validation and watching all twenty tests stay green. That claim is withdrawn.

What is covered now, in `__tests__/api-keys-proxy-route.test.ts`:

- A nine row table over `-1000`, `0`, `10.5`, `"1000"`, `null`, `1e999`, `9007199254740993`, `1e21` and `Number.MAX_VALUE`, each asserting 400 and `initiateCheckout` never called. Bodies are sent as raw bytes because `JSON.stringify` cannot produce Infinity or an unrounded 9007199254740993.
- `Number.MAX_SAFE_INTEGER` itself accepted, so the bound rejects only what is past it.
- The happy path asserting the forwarded arguments are exactly `("bkash", 1000, "idem-1")` while `account_id`, `workspace_id` and `amount_local` in the same body are dropped. Same property on create: a body carrying `account_id` and `scopes` still results in `createApiKey("k", "2030-01-01")`. That is the property that stops a client picking its own account.
- Rails read, authenticated and unauthenticated, and a `["checkout","rails","extra"]` shape that must 404.
- Error mapping in both directions: a `ControlPlaneError(403, ...)` produces `403 {"error":"Forbidden"}`, and a plain transport failure produces `502` with no upstream text.

`__tests__/control-plane-error-status.test.ts` is new. It drives the real client with a 403 response and asserts each of the four functions raises a `ControlPlaneError` carrying that status. Without it, someone could revert the client to a plain `Error` and the route's mapping would silently go dead again.

### Proven red, not assumed red

Every assertion added here was driven red first, in the same Docker image the suite runs in.

Deleting `credits > Number.MAX_SAFE_INTEGER` from the route, changing nothing else:

```
× rejects one past MAX_SAFE_INTEGER credits without reaching upstream
× rejects 1e21 credits without reaching upstream
× rejects MAX_VALUE credits without reaching upstream
Test Files  2 failed | 40 passed (42)
     Tests  3 failed | 467 passed | 1 skipped (471)
```

Reverting the four client functions to `throw new Error(await readResponseError(...))`:

```
× createApiKey raises a ControlPlaneError carrying the upstream status
× revokeApiKey raises a ControlPlaneError carrying the upstream status
× getCheckoutRails raises a ControlPlaneError carrying the upstream status
× initiateCheckout raises a ControlPlaneError carrying the upstream status
Tests  4 failed | 466 passed | 1 skipped (471)
```

Removing the rail maximum from `ValidatePurchaseAmount`:

```
--- FAIL: TestInitiateCheckout_RejectsCreditsAboveRailMaximum
FAIL github.com/sakibsadmanshajib/hive/apps/control-plane/internal/payments
```

Restored, everything green again:

```
docker compose run --rm --build web-console npm run test:unit
Test Files  1 failed | 41 passed (42)
     Tests  470 passed | 1 skipped (471)

docker compose run --rm --build web-console npm run build
✓ Compiled successfully in 52s
ƒ /api/v1/accounts/current/[...path]    184 B    102 kB

go test ./apps/control-plane/... -count=1 -short
no FAIL
```

The one failing suite is `tests/unit/control-plane-host.test.ts`, pre-existing and unrelated: it reads `/app/.env.example`, which the image does not copy in.

### Non blocking review notes, answered

- **Rail and `expires_at` are not allowlisted at this seam.** Confirmed the upstream validates both: `InitiateCheckout` rejects a rail outside `AvailableRails` for the account's country, and it now also rejects credits above that rail's advertised maximum. No allowlist added here, since duplicating the rail enum in the console is the thing that drifts.
- **`browser-api-routes.test.ts` cannot police the allowlist.** Correct, and its header now says so: a catch all route proves a handler exists, never that an operation is allowlisted.
- **The `ControlPlaneError` mock had the wrong constructor order.** Fixed to `(status, message, code)`, matching the real class, and it is now actually constructed by a test rather than sitting inert.
- **Unencoded id interpolation survives at other `client.ts` call sites, and `GetMembershipRole` does not filter on membership status.** Both accurate, both out of scope for this fix, both worth a follow up issue rather than widening this PR.

### Visual proof

Not re-captured. Nothing in this round changes a rendered surface: the credits bound is a server side rejection on a value the checkout modal cannot produce, and the error mapping changes a status code, not a view. The captures above were taken against a stack built from this branch and still show the four operations working.


apps/control-plane/internal/platform/http/router.go
```diff
@@ -208,6 +208,13 @@ func NewRouter(cfg RouterConfig) http.Handler {
 		protectedBudgets := cfg.AuthMiddleware.Require(cfg.BudgetsHandler)
 		mux.Handle("/api/v1/accounts/current/budget", protectedBudgets)
 		mux.Handle("/api/v1/accounts/current/budget/dismiss", protectedBudgets)
+		// The Phase 14 workspace surface. budgets.Handler has always dispatched
+		// these two prefixes, but they were never mounted here, so every
+		// request fell through to the /api/v1/ catch-all and came back 404.
+		// That made a hard spend cap and a spend alert impossible to save from
+		// the console even though the handler and its own tests were correct.
+		mux.Handle("/api/v1/budgets/", protectedBudgets)
+		mux.Handle("/api/v1/spend-alerts/", protectedBudgets)
```
sakibsadmanshajib, line 227: Verified this does not newly expose an unenforced surface: budgets.Handler.handleWorkspaceBudget / handleSpendAlerts resolve the caller's role through roleSvc.IsWorkspaceOwner (a real DB-backed lookup keyed on the session's own viewer.UserID), not from anything in the URL, per the Phase 18 comment on requireWorkspaceOwner. So mounting these two prefixes turns on real per-workspace authorization, not a dead handler with a rubber-stamp check. No cross-tenant read/write found by changing the workspace id in the path.

Also confirmed reachability precedence: Go's net/http.ServeMux does longest-prefix match, so /api/v1/budgets/ and /api/v1/spend-alerts/ take priority over the pre-existing /api/v1/ catch-all regardless of registration order. The proof capture's PUT /api/v1/unmounted-surface/x -> 404 control line is a reasonable way to demonstrate that.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
```diff
@@ -0,0 +1,162 @@
...
+  ControlPlaneError,
+  createApiKey,
+  getCheckoutRails,
+  initiateCheckout,
+  revokeApiKey,
+} from "@/lib/control-plane/client";
+
+type Params = { params: Promise<{ path: string[] }> };
+
+async function requireSession(): Promise<Response | null> {
```
Comment (sakibsadmanshajib) [inline apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:34]: Traced the auth boundary here and it holds: requireSession() calls supabase.auth.getUser() (a real round trip that rejects a revoked token, not the cookie-only getSession()), and the downstream typed client calls (createApiKey/revokeApiKey/getCheckoutRails/initiateCheckout) build the upstream Authorization header themselves from that same server-validated session in getRequestContext(). The incoming Request's own headers are never read or forwarded, so nothing the browser sends can influence the upstream bearer. CONTROL_PLANE_BASE_URL is also only ever read server-side. Allowlist checks (path.length plus exact segment equality) reject prefix-continuation, case variation, and extra segments correctly; not a blanket pass-through.

One non-blocking gap, not introduced by this PR: there's no Origin/Referer check or CSRF token here, so this (and the sibling proxy routes it's modeled on, e.g. apps/web-console/app/api/budget/[workspaceId]/route.ts and app/api/console/checkout/intent/route.ts) rely entirely on the Supabase session cookie. If that cookie isn't SameSite=Strict, a cross-site POST could ride the victim's session to create an API key or initiate checkout. Since this is the existing pattern across the app rather than something new here, I'd file it as a follow-up rather than block this fix on it. Worth confirming the Supabase cookie's SameSite attribute in a dedicated pass over all mutating route handlers.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
```diff
@@ -0,0 +1,162 @@
...
+    try {
+      return NextResponse.json(await createApiKey(nickname.trim(), expiresAt));
+    } catch (err) {
+      return errorResponse(err, "Failed to create API key");
+    }
+  }
+
+  // POST /api/v1/accounts/current/api-keys/{keyId}/revoke
+  if (path.length === 3 && path[0] === "api-keys" && path[2] === "revoke") {
+    if (!UUID_RE.test(path[1])) {
```
Comment (sakibsadmanshajib) [inline apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:127]: Confirmed this guard is a real server-side control and not decorative: UUID_RE is anchored (^...$), case-insensitive, and checked against the raw decoded path segment before revokeApiKey(path[1]) is ever called. Walked the Next.js catch-all semantics for the traversal cases the new test covers: "../../../internal/apikeys/resolve" and "..%2F..%2Fadmin" both land as a single literal path[1] element (catch-all segments split on the raw "/" before percent-decoding, so an encoded slash doesn't re-split into more segments), and neither is a 36-char UUID, so both correctly 404 before touching upstream. google/uuid.New() (used for key ids) always sets the RFC4122 variant nibble to 8/9/a/b, so the regex won't produce false negatives against real key ids either. Good boundary.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
```diff
@@ -0,0 +1,162 @@
...
+      return NextResponse.json(await revokeApiKey(path[1]));
+    } catch (err) {
+      return errorResponse(err, "Failed to revoke API key");
+    }
+  }
+
+  // POST /api/v1/accounts/current/checkout/initiate
+  if (path.length === 2 && path[0] === "checkout" && path[1] === "initiate") {
+    const body = await readBody(request);
+    const rail = body.rail;
```
Comment (sakibsadmanshajib) [inline apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:140]: Nit, non-blocking: rail is only checked for typeof string and non-empty before being forwarded to initiateCheckout, no allowlist against the known rails (stripe/bkash/sslcommerz). Same for expires_at above on the create-key path (line 113), only checked for typeof string. Both presumably get rejected by control-plane validation on the other end, but since this handler is explicitly the browser-trust boundary for this surface, worth a one-line comment (or a quick confirmation in the PR) that upstream actually validates rail against its enum and expires_at as a real future timestamp, so a malformed value can't wedge a checkout attempt or produce a confusing upstream 500 that this route would then mask as a generic 502.

sakibsadmanshajib (commented):
Independent review, verdict: APPROVE WITH COMMENTS.

Scope: read the pushed diff directly (`gh pr diff 768`), not the PR description, and cross-checked every claim in the description against the actual code and the committed proof images.

What holds up:

1. New proxy route (`apps/web-console/app/api/v1/accounts/current/[...path]/route.ts`) is a genuine allowlist, not a pass-through. Path matching uses exact segment equality plus length checks, so there is no prefix-continuation, case-variation, or extra-segment bypass. Auth is verified server-side via `supabase.auth.getUser()` (a real revocation-checked round trip), before any upstream call. The upstream bearer and `CONTROL_PLANE_BASE_URL` are built from the server-validated session inside `getRequestContext()`; nothing from the incoming browser request's headers is read or forwarded, so a client cannot influence the upstream Authorization. Method coverage is correct: only GET (checkout/rails) and POST (create/revoke key, checkout/initiate) are exported, everything else 404s or falls to Next's default 405.

2. Revoke hardening is a real server boundary, not a client-side-only check. `UUID_RE` in the route handler is anchored and evaluated against the raw decoded path segment before `revokeApiKey()` is ever called. Traced the Next.js catch-all segment semantics for the traversal cases the new test drives (`../../../internal/apikeys/resolve`, `..%2F..%2Fadmin`): both land as a single literal array element, not a re-split path, and neither matches a 36-character UUID, so both 404 before touching upstream. The `encodeURIComponent` added in `lib/control-plane/client.ts` is a second, redundant layer on the hop to the control-plane; the actual property rests on the regex, which is where it should rest.

3. Control-plane router change (`router.go` lines 216-217) mounts two prefixes onto the same `AuthMiddleware.Require(...)`-wrapped handler already used for the legacy budget routes, so it is not newly unauthenticated. More importantly, the underlying `budgets.Handler` (untouched by this PR) was already hardened in Phase 18: `requireWorkspaceOwner` / `requireWorkspaceMembership` resolve the caller's real role via a DB-backed `roleSvc.IsWorkspaceOwner(userID, workspaceID)` lookup keyed on the session's own user id, not on anything in the URL. Mounting the route turns on real per-workspace authorization, not a previously-untested rubber stamp.

4. Tests read as legitimate regression guards, not vanity coverage. `router_budgets_test.go` asserts a 503 body specific to `budgets.Handler` (`workspace budget surface unavailable` / `spend alert surface unavailable`), which only that handler emits, so it cannot pass vacuously against the router's default 404. `api-keys-proxy-route.test.ts` exercises the traversal, non-UUID, unauthenticated, and off-allowlist cases against the actual route module. `browser-api-routes.test.ts` walks the real `app/api` tree rather than a hand-maintained list, with sanity assertions (`patterns.length > 5`, `fetchedApiPaths().length > 5`) guarding against a silently-broken discovery walk.

5. Proof images (`docs/proof/console-money-ops-2026-08-06/*.png`) checked directly: origin is `localhost:3000`, user is an e2e fixture (`e2e-verified@scubed.co.bd`), the created key secret is masked with `[REDACTED - real secret masked before capture]`, and the final capture shows all three listed keys (including the one created for the demo) in `Revoked` status. No real credential, host, or identifier in the pixels.

Non-blocking findings posted inline on the diff:
- No Origin/Referer check or CSRF token on the new mutating route, but this matches the existing pattern across the app's other account-proxy routes, so it is not a regression introduced here. Worth a dedicated follow-up across all mutating route handlers rather than blocking this fix.
- `expires_at` and `rail` in the new route are only type-checked, not format/enum-validated, relying on control-plane validation. Low risk, worth a one-line confirmation that upstream actually enforces this.

One more observation that could not be anchored inline because the file is unchanged by this PR: `apps/web-console/components/api-keys/revoke-confirm-panel.tsx` builds its fetch URL as a template literal without `encodeURIComponent(keyId)`, unlike the equivalent server-side call in `client.ts`. In practice `keyId` here always comes from a previously-listed `ApiKey.id` (a real UUID), never free text, and the server-side `UUID_RE` guard in the new route independently closes the actual vulnerability regardless of what the browser sends, so this is cosmetic-consistency only, not a gap.

No CRITICAL or HIGH issue found in the new proxy seam, the revoke hardening, or the router change. Recommend merging once the two informational threads above are acknowledged (no code change required to unblock).

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (
+      typeof rail !== "string" ||
+      rail === "" ||
+      typeof credits !== "number" ||
+      !Number.isFinite(credits) ||
+      !Number.isInteger(credits) ||
+      credits <= 0 ||
+      typeof idempotencyKey !== "string" ||
+      idempotencyKey === ""
+    ) {
```
Comment (sakibsadmanshajib) [inline apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:161]: **Money path, blocking.** `credits` is checked for type, finiteness, integrality and positivity, but not for magnitude, so a value past `Number.MAX_SAFE_INTEGER` passes and is forwarded verbatim.

I drove the handler with these bodies. Negative, zero, fractional, `NaN`, `Infinity` and a string all correctly returned 400 without forwarding. These three did not:

```
credits=1e21                -> 200, forwarded 1e+21
credits=9007199254740993    -> 200, forwarded 9007199254740992
credits=Number.MAX_VALUE    -> 200, forwarded 1.7976931348623157e+308
```

The second line is the dangerous one. `Number.isInteger(9007199254740993)` is true, but the value is not representable, so the customer's purchase quantity is silently rounded before it ever leaves this process. That is precisely the float64 corruption this project uses `math/big` to avoid on the Go side, reintroduced at the browser facing seam.

Upstream does not save you. `initiateRequest.Credits` is an `int64` and the only checks are `> 0` and `% 1000 == 0` (`apps/control-plane/internal/payments/http.go:177-183`). There is no upper bound. `credits = 9007199254741000` is above `MAX_SAFE_INTEGER`, serialises exactly, decodes cleanly into `int64`, is a multiple of 1000, and reaches `InitiateCheckout`. This route is the last place that can reject it.

The guard already exists in this codebase, on the sibling money proxy, with a comment explaining exactly this reasoning:

```ts
// apps/web-console/app/api/budget/[workspaceId]/route.ts:36-51
// Number.MAX_SAFE_INTEGER guard: PostgreSQL BIGINT column can hold values
// beyond 2^53, but JavaScript Number loses precision past that boundary.
...
soft > Number.MAX_SAFE_INTEGER ||
hard > Number.MAX_SAFE_INTEGER
```

Please reuse it here:

```ts
credits <= 0 ||
credits > Number.MAX_SAFE_INTEGER ||
```

A sane ceiling tied to the `max_credits` the rails endpoint already advertises would be better still, but `MAX_SAFE_INTEGER` is the minimum that stops a silently rounded quantity.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
```diff
@@ -0,0 +1,162 @@
...
+  return "Request rejected";
+}
+
+function errorResponse(err: unknown, fallback: string): Response {
+  if (err instanceof ControlPlaneError) {
+    const status = err.status >= 400 && err.status < 600 ? err.status : 502;
+    return NextResponse.json({ error: statusSummary(status) }, { status });
+  }
+  return NextResponse.json({ error: fallback }, { status: 502 });
+}
```
Comment (sakibsadmanshajib) [inline apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:68]: **This branch is unreachable for all four operations on this route.** None of `createApiKey`, `revokeApiKey`, `getCheckoutRails` or `initiateCheckout` calls `throwControlPlaneError`. All four throw a plain `Error` built from `readResponseError` (`lib/control-plane/client.ts:1731`, `1759`, `1539`, `1605`). I confirmed it by reading each function body out of the shipped image and checking for the call: `createApiKey=false`, `revokeApiKey=false`, `getCheckoutRails=false`, `initiateCheckout=false`.

So `err instanceof ControlPlaneError` is always false here and every upstream failure collapses to `502` with a generic sentence. `statusSummary` is dead code on this path. I verified the branch itself works when the error has the right shape, a synthetic `ControlPlaneError(403, ...)` produced `403 {"error":"Forbidden"}`, so the logic is fine and only the wiring is missing.

The practical cost is that the console cannot tell a refusal from an outage. A member who is not the workspace owner hitting key creation gets `502 Failed to create API key`, identical to the control plane being down. An already revoked key returns 502 instead of 409. A rate limit value out of range returns 502 instead of 422. It also means a genuine authorization denial is invisible to anything watching status codes.

Two honest ways forward. Either switch those four client functions to `throwControlPlaneError`, which is what the class docstring at `client.ts:419` says it exists for and what the sibling proxy routes rely on, or delete this branch and the `statusSummary` helper and state in the comment that this route deliberately returns 502 for every upstream failure. The current shape reads as if status is preserved when it never is.

apps/web-console/__tests__/api-keys-proxy-route.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    expect(mockCreateApiKey).not.toHaveBeenCalled();
+  });
+
+  it("404s an operation that is not on the allowlist", async () => {
+    const { POST } = await import("../app/api/v1/accounts/current/[...path]/route");
+
+    const response = await POST(request(), params(["members", "someone", "promote"]));
+
+    expect(response.status).toBe(404);
+  });
```
sakibsadmanshajib, line 151: **No test here can go red for the checkout half.** The PR body says "Checkout rails and initiate are fixed by the same route handler and covered by the new test". They are reached by the same handler, but nothing in this file exercises either one. There is no case for `["checkout", "rails"]`, none for `["checkout", "initiate"]`, and none for the `credits` validation.

I checked this the only way that counts. I deleted the entire `credits` validation from the route (the type, finite, integer and positive checks), rebuilt the `web-console` image, and ran both suites in this PR. Result: `Tests 4 failed | 20 passed`, and all four failures were the traversal cases from the UUID guard I had removed in the same pass. Not one failure came from the money path. The checkout allowlist entries and their validation ship with zero regression coverage.

Given how many defects in this repository have shipped behind tests that could not go red, and given this is the credit purchase path, please add at least one case that fails when the `credits` guard is removed. A table over `-1`, `0`, `10.5`, `NaN`, `"1000"` and a value above `Number.MAX_SAFE_INTEGER` asserting 400 and `expect(initiateCheckout).not.toHaveBeenCalled()` is enough, plus one happy path asserting the forwarded arguments are exactly `(rail, credits, idempotency_key)`.

While you are here, one assertion worth pinning because it is a real property this route provides: extra body fields are dropped. I confirmed that posting `account_id`, `workspace_id` and `amount_local` alongside a valid body still results in `initiateCheckout("bkash", 1000, "idem-1")`, and that a create carrying `account_id` and `scopes` still results in `createApiKey("k", "2030-01-01")`. That is the property that stops a client picking its own account, and nothing currently guards it.

apps/web-console/__tests__/api-keys-proxy-route.test.ts
```diff
@@ -0,0 +1,117 @@
...
+}));
+
+vi.mock("@/lib/control-plane/client", () => ({
+  ControlPlaneError: class ControlPlaneError extends Error {
+    status: number;
+    constructor(message: string, status: number) {
+      super(message);
+      this.status = status;
+    }
+  },
```
sakibsadmanshajib, line 55: Non blocking, but this is how the dead branch above stayed hidden. The mock declares the constructor as `(message, status)` while the real class is `constructor(status: number, message: string, code: string | null = null)` (`lib/control-plane/client.ts:430`). Nothing asserts on it today so the mismatch is inert, but a mock whose shape does not match the real type is the thing that makes an error mapping bug untestable. Either align the signature or drop the mock entirely, since no test in this file constructs one.

apps/web-console/lib/control-plane/client.ts
```diff
@@ -1747,7 +1747,10 @@ export async function createApiKey(nickname: string, expiresAt?: string): Promis
 
 export async function revokeApiKey(keyId: string): Promise<ApiKey> {
   const { baseUrl, headers } = await getRequestContext();
-  const response = await fetch(`${baseUrl}/api/v1/accounts/current/api-keys/${keyId}/revoke`, {
+  // Encoded for the same reason keyLimitsUrl encodes: a key id carrying a path
+  // separator would otherwise retarget this request at a different upstream
+  // path while still carrying the caller's bearer.
+  const response = await fetch(`${baseUrl}/api/v1/accounts/current/api-keys/${encodeURIComponent(keyId)}/revoke`, {
```
sakibsadmanshajib, line 1778: Good catch and the right fix. Flagging that the class is wider than this one call site, and this PR is a natural place to say so even if the rest is deferred.

The same unencoded interpolation of a caller supplied id survives at `rotateApiKey` (`:1787`), `getBudget` / `updateBudget` / `deleteBudget` (`:2401`, `:2428`, `:2450`), `listAlerts` / `createAlert` (`:2462`, `:2493`), and the two invoice reads (`:1510`, `:2598`). The budget ones matter most because `app/api/budget/[workspaceId]/route.ts` passes `params.workspaceId` straight through with no UUID check, and Next hands that segment over already decoded.

I confirmed the primitive is real. WHATWG URL normalizes traversal segments during construction, so the id chooses the upstream path:

```
new URL(`${baseUrl}/api/v1/budgets/${"../../internal/apikeys/resolve"}`).href
  -> http://cp:8080/api/internal/apikeys/resolve
```

One more `..` reaches the root and any control plane path becomes addressable while still carrying the caller's own bearer.

This is not an escalation today, and I want to be precise about why rather than leave it implied. Every `/internal/` route is wrapped by `RequireInternalToken`, which denies unconditionally when no token is configured rather than letting an empty comparison succeed (`internal/platform/http/internalauth.go:27-29`), so the internal namespace is closed. The reachable `/api/v1/` targets are individually authorized and the bearer is the caller's own, so nothing is gained. It is still the same defect this commit just fixed one instance of, and it is one route mount away from mattering. A UUID check in `app/api/budget/[workspaceId]/route.ts` alongside the existing cap validation, or `encodeURIComponent` at these call sites, closes it. Fine as a follow up issue if you would rather keep this PR tight.

apps/control-plane/internal/platform/http/router.go
sakibsadmanshajib, line 227: The mount itself is correct. It sits inside the `BudgetsHandler != nil && AuthMiddleware != nil` guard, wraps `protectedBudgets`, and the workspace handlers gate on a real database membership lookup keyed on the URL workspace id rather than trusting it, so there is no cross workspace bypass here. I removed both lines, reran the suite, and got the reported 404 on all four subtests, so the regression guard is genuine.

One observation worth recording rather than fixing here. This makes the owner gated workspace surface reachable for the first time, and the predicate behind it does not filter on membership status:

```sql
-- internal/platform/role_pgx.go:61-66
SELECT role FROM public.account_memberships
WHERE account_id = $1 AND user_id = $2
LIMIT 1
```

Every sibling query in the codebase constrains status. `EnsureViewerContext` requires `m.Status == "active"` (`internal/accounts/service.go:88`), `repository.go` filters `m.status = 'active'` at `:218`, `:224`, `:238` and `:240`, and `GetTenantRole` says ACTIVE in its own doc comment. `GetMembershipRole` is the only one that does not, which means a membership row with `role = 'owner'` and a non active status would pass `IsWorkspaceOwner` and therefore pass `PermBillingWrite` on `PUT /api/v1/budgets/{ws}` and `POST /api/v1/spend-alerts/{ws}`.

No code path writes such a row today, so this is latent rather than exploitable, and I am not asking for it in this PR. It becomes live the moment member suspension or soft removal lands, and it now sits on a money path. Worth a follow up issue.

apps/web-console/__tests__/browser-api-routes.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    );
+}
+
+function matches(pattern: string[], actual: string[]): boolean {
+  let p = 0;
+  let a = 0;
+  while (p < pattern.length && a < actual.length) {
+    if (pattern[p] === "**") {
+      // Catch-all consumes the rest; Next requires at least one segment.
+      return p === pattern.length - 1;
```
sakibsadmanshajib, line 76: Non blocking, and this test is a good idea. Naming the ceiling so nobody later reads it as stronger than it is.

Adding the catch all route weakens what this assertion proves for the namespace it covers. A `**` pattern matches any remaining segments, so from now on every browser fetch to `/api/v1/accounts/current/<anything>` passes this test, including operations the allowlist in the route handler will answer with 404. I confirmed the shape by driving the handler with `["members", "someone", "promote"]`, which returns 404 at runtime while this test would consider the path served.

The test still does exactly what it was written for, since deleting the route file fails it on precisely the four broken paths and leaves the other ten green, which I verified. It just cannot police the allowlist itself. A sentence in the header comment saying that a `**` route proves a handler exists and not that the operation is allowlisted would stop the next reader over trusting it.

sakibsadmanshajib (commented):
**VERDICT: CHANGES REQUESTED** (posted as a comment review because GitHub refuses a formal request for changes from the PR author account).

Independent security review, dedicated pass because this touches money and credential paths.

**No cross-tenant reach found.** A signed in user of workspace A cannot drive an operation against workspace B through this route. I tried to escape the allowlist eleven different ways and could not. Identity cannot be spoofed. No credential reflects into any response. Details below, then three things that should change before merge.

## What I verified as sound

**The allowlist is a true allowlist.** Every branch is an exact equality match on decoded segments plus an exact length check, and both `GET` and `POST` end in an explicit `404` rather than a fall through (`route.ts:97`, `route.ts:161`). I drove the handler directly with these decoded `[...path]` shapes and every one returned 404 with the upstream mock never called: `["API-KEYS"]`, `["Api-Keys"]`, `["api-keys "]`, `["/api-keys"]`, `["", "api-keys"]`, `["http://evil.test/api-keys"]`, `["//evil.test/api-keys"]`, `["..", "..", "internal", "apikeys", "resolve"]`, `["api-keys", "extra"]`, a unicode homoglyph separator, and `["checkout", "rails", "x"]`. On the revoke id specifically, `uuid/../../admin`, `uuid%2F..` and a trailing newline all returned 404 with `revokeApiKey` never called. Only a well formed UUID reaches upstream, and it reaches it encoded.

**Identity is server side only.** `requireSession` calls `supabase.auth.getUser()` (`route.ts:34`), which does a round trip and rejects a revoked token, and it runs before `params` is read on both methods. I sent a request carrying a forged `Authorization` header, a forged `X-Hive-Account-ID` header, and `user_id` plus `account_id` in the body, with `getUser` returning no user. Result was 401 and upstream was never called. A `getUser` error is treated as unauthenticated, and a `getUser` rejection propagates rather than falling through to an authenticated path.

**The tenant cannot be chosen by the client.** The account comes from the `hive_account_id` cookie, which is `httpOnly` and set only after `app/console/account-switch/route.ts:26` confirms the id is in the viewer's memberships. The control plane then independently refuses to honour it unless the caller holds an active membership (`internal/accounts/service.go:86-93`). Two layers, both server side. The newly mounted workspace budget routes gate on a real database membership lookup keyed on the workspace id in the URL (`internal/budgets/http.go:516`, `internal/platform/role_pgx.go:47`), with parameterized SQL.

**No credential reflects back.** I rejected `createApiKey` with an error whose message contained a bearer token, a service role string and the internal control plane host. The response body was exactly `{"error":"Failed to create API key"}`. A newly minted key secret is returned once in the response body, is held in component state only, and is written to no log, no `localStorage` and no `sessionStorage`.

**Both fixes are backed by tests I personally drove red.** Removing the two `mux.Handle` lines fails all four `TestNewRouterMountsWorkspaceBudgetSurface` subtests with the reported 404. Removing the route file fails `browser-api-routes.test.ts` on exactly the four broken paths and leaves the other ten green. Removing the UUID guard fails the four traversal cases in `api-keys-proxy-route.test.ts`.

## What should change

Three items, inline below. The first is the substantive one: the money path is missing a bound this repository already applies two directories away, and no test in this PR can go red if that validation is deleted. I proved that by deleting it.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
Comment (sakibsadmanshajib) [reply apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:161]: Fixed, and the bound went upstream too.

The console route now carries the same `Number.MAX_SAFE_INTEGER` guard as `app/api/budget/[workspaceId]/route.ts`, in the shape you quoted, rejecting rather than coercing.

You were right that upstream does not save us, so it now saves itself. `GetCheckoutOptions` has always advertised a per rail maximum (10,000,000 Stripe, 30,000,000 bKash, 500,000,000 SSLCommerz) that nothing enforced, so the advertised ceiling was decoration. `ValidatePurchaseAmount` now takes the rail and applies `maxCreditsForRail`, which covers the stub service on the same line, and `classifyInitiateError` maps the refusal to a 400 carrying no rate and no amount. That is the sane ceiling you asked for, placed where the rail is actually known.

Proven red before it was green: deleting `credits > Number.MAX_SAFE_INTEGER` from the route fails exactly the three over boundary rows and nothing else.

```
x rejects one past MAX_SAFE_INTEGER credits without reaching upstream
x rejects 1e21 credits without reaching upstream
x rejects MAX_VALUE credits without reaching upstream
Tests  3 failed | 467 passed | 1 skipped (471)
```

Removing the rail maximum from `ValidatePurchaseAmount` fails `TestInitiateCheckout_RejectsCreditsAboveRailMaximum`, all four subtests, including the 9007199254741000 case you named.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
Comment (sakibsadmanshajib) [reply apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:68]: Fixed by making the branch reachable, which was the smaller of the two honest options.

`createApiKey`, `revokeApiKey`, `getCheckoutRails` and `initiateCheckout` now call `throwControlPlaneError` instead of building a plain `Error`, so a refusal keeps its status: a member who is not the owner gets 403, an already revoked key gets 409, an out of range value gets 422. This route is their only caller, so nothing else changes shape, and `ControlPlaneError` still extends `Error` for anything matching on message.

The reason to prefer this over deleting the branch is the one you gave: an authorization denial that renders as 502 is invisible to anything watching status codes, and this is the credential path.

New suite `__tests__/control-plane-error-status.test.ts` drives the real client with a 403 and asserts each of the four raises a `ControlPlaneError` carrying that status. Reverting all four to `throw new Error(await readResponseError(...))` turns exactly those four red, which is what I ran before wiring it up.

apps/web-console/__tests__/api-keys-proxy-route.test.ts
sakibsadmanshajib: Fixed, and the PR body claim is withdrawn rather than softened. It said the checkout half was covered by the new test. It was not, and I have corrected that sentence in place and stated the correction in the round two section.

Added, in this file:

- A nine row table over -1000, 0, 10.5, "1000", null, 1e999, 9007199254740993, 1e21 and Number.MAX_VALUE, each asserting 400 and `initiateCheckout` never called. The bodies are raw bytes, not `JSON.stringify` output, because `JSON.stringify` cannot produce Infinity and cannot produce an unrounded 9007199254740993.
- `Number.MAX_SAFE_INTEGER` accepted, so the bound rejects only what is past it.
- The happy path you specified, asserting the forwarded arguments are exactly `("bkash", 1000, "idem-1")`.
- The property you flagged as worth pinning: `account_id`, `workspace_id` and `amount_local` in the same body are dropped, and a create carrying `account_id` and `scopes` still results in `createApiKey("k", "2030-01-01")`.
- Rails read, authenticated and unauthenticated, plus a `["checkout","rails","extra"]` shape that must 404.
- Error mapping both ways: `ControlPlaneError(403, ...)` produces `403 {"error":"Forbidden"}`, a plain transport failure produces 502 with no upstream text.

Every one of these was driven red first. Deleting the credits bound alone now fails three of them.

apps/web-console/__tests__/api-keys-proxy-route.test.ts
sakibsadmanshajib: Aligned. The mock is now `constructor(status: number, message: string, code: string | null = null)`, matching the real class, and it sets `name` and `code` the same way.

It is no longer inert either: the new error mapping test constructs `new ControlPlaneError(403, "not the workspace owner")` and asserts the route answers 403, so a future drift in the constructor breaks a test rather than silently disabling one.

apps/web-console/__tests__/browser-api-routes.test.ts
sakibsadmanshajib: Agreed, and the header now names the ceiling:

> Its ceiling, so nobody reads it as stronger than it is: a catch-all route matches every remaining segment, so for a `[...path]` namespace this proves only that a handler exists, never that the operation is on that handler's allowlist. `/api/v1/accounts/current/members/x/promote` passes here and answers 404 at runtime. The allowlist itself is pinned in api-keys-proxy-route.test.ts.

The allowlist assertions it points at have grown this round to cover the checkout entries as well as the key ones.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
Comment (sakibsadmanshajib) [reply apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:140]: Confirmed upstream, no allowlist added here.

Rail: `InitiateCheckout` refuses any rail outside `AvailableRails` for the account's country before anything is created, and as of this round it also refuses credits above that rail's advertised maximum. A malformed rail is a 400 with a fixed message, not a 500.

`expires_at`: forwarded as an opaque string and parsed upstream, which rejects an unparseable or past value.

I deliberately did not duplicate the rail enum in the console. A second copy of a list that lives upstream is the thing that drifts, and the failure mode it would prevent is already a clean 400. What was worth fixing was that a genuine upstream 400 used to reach the browser as a 502; that is fixed in this round, so a malformed value now surfaces as a rejection rather than an outage.

apps/web-console/lib/control-plane/client.ts
sakibsadmanshajib: Accurate, and out of scope for this PR by your own suggestion, so it is filed as #802 with your call sites, the URL normalization primitive, and the reason it is not an escalation today.

The budget route is called out there specifically, since `params.workspaceId` reaches `updateBudget` with no UUID check and Next hands that segment over already decoded.

apps/control-plane/internal/platform/http/router.go
sakibsadmanshajib: Recorded rather than fixed here, as you suggested. Filed as #803 with the query, the sibling queries that do constrain status, and the specific consequence that a non active owner row would pass `IsWorkspaceOwner` and therefore `PermBillingWrite` on the two routes this PR mounts.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
Comment (sakibsadmanshajib) [reply apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:34]: Thanks for tracing the bearer path. Two things changed in this file since:

The credits validation now bounds magnitude, and the four client functions raise `ControlPlaneError`, so an upstream refusal keeps its status here instead of collapsing into 502.

On the CSRF gap: agreed it is the existing pattern rather than something new here, so it is filed as #804 rather than blocking this fix. It lists this route, the budget proxy and the checkout intent route, and scopes the work as confirming the Supabase cookie's actual `SameSite` attribute first, then sweeping every mutating handler.

apps/control-plane/internal/platform/http/router.go
sakibsadmanshajib: Thanks. Nothing changed on this line in the review round, and the mount stays as reviewed.

One consequence of your read is now filed: since this makes the owner gated workspace surface reachable, the predicate behind it matters, and `GetMembershipRole` not filtering on membership status is tracked as #803.

apps/web-console/app/api/v1/accounts/current/[...path]/route.ts
Comment (sakibsadmanshajib) [reply apps/web-console/app/api/v1/accounts/current/[...path]/route.ts:127]: Thanks for walking the catch-all decoding semantics; that is the part I could not have asserted from the route alone.

The guard is unchanged this round. The tests around it grew: the same file now also pins the checkout allowlist entries and their validation, which previously had none.

PR #787, opened 2026-08-06
fix: keep the Open WebUI OAuth session alive through its first token refresh (#782)

sakibsadmanshajib:
Fixes #782.

## The trigger rule

Elapsed time since sign-in. Not message count, and not which conversation.

Supabase issues the OAuth access token with a 3600 second lifetime. Open WebUI refreshes once `now + 5 minutes` reaches `expires_at`, so the refresh path is first entered roughly 55 minutes into a signed-in session. Until then nothing is wrong. At that moment the refresh fails, Open WebUI reads the failure as a dead session and calls `delete_session_by_id`, and from then on `__oauth_token__` is empty. Every message in every conversation fails permanently until the user signs in again through SSO. A conversation opened on an already dead session fails from its first message, which is why this was originally reported as a second-message defect.

The gateway is not the bug. `OWUIUnwrap` refusing a request that carries no `__metadata.upstream_auth` is correct, fail-closed behaviour. It is the last honest component in the chain.

## Two root causes, and both halves MUST ship together

**First, no refresh token was ever issued.** `OAUTH_SCOPES` was `openid email profile`. Supabase only issues a refresh token when `offline_access` is requested, and `_perform_token_refresh` gives up on its very first guard without one. Every authorization ever recorded in `auth.oauth_authorizations` for this project was scoped `openid email profile`, 54 of them, so the destruction was guaranteed for every user.

**Second, the refresh authenticates the client in a different dialect than the login does.** The authorization code exchange goes through authlib, which defaults `token_endpoint_auth_method` to `client_secret_basic` whenever a client secret is set. The refresh does not go through authlib at all: it hand builds the POST and always puts `client_id` and `client_secret` in the form body, which is `client_secret_post`. Supabase enforces the registered method exactly.

Fixing only the scope leaves a refresh token that is refused anyway. Fixing only the dialect leaves no refresh token to send. Verified against the live token endpoint, sending the same deliberately invalid refresh token both ways so a client authentication failure is distinguishable from a token failure:

| client authentication | status | error | meaning |
| --- | --- | --- | --- |
| `client_secret_post`, what upstream Open WebUI sends | 400 | `invalid_credentials`, "client is registered for 'client_secret_basic' but 'client_secret_post' was used" | rejected before the token is looked at |
| `client_secret_basic`, what this PR sends | 400 | `refresh_token_not_found`, "Invalid Refresh Token" | client authenticated, then failed only on the deliberately invalid token |

## Correction: this PR no longer sets OAUTH_TOKEN_ENDPOINT_AUTH_METHOD

An earlier revision of this branch set `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD` to `client_secret_post` in compose. **That is reverted here.** Reviewers who looked at the previous head will see a different second half.

Pinning that variable is the other way to make both legs agree, and it only works if the Supabase OAuth client is also re-registered for `client_secret_post`. Every client this project has registered is `client_secret_basic`, confirmed live in `auth.oauth_clients`: the CI client, the demo box client, and the Hive Chat client. Registering those clients is manual ops with no script in this repository behind it, so the invariant would live only in prose, on every deployment including customer-hosted ones. Getting the order wrong does not degrade quietly: authlib would send a dialect the registration rejects and sign-in would fail outright for everyone, immediately.

Patching the refresh to follow the method the client is already configured with removes the coupling. It works against every client already registered, needs no ops step, and still honours `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD` for a deployment that really is registered for `client_secret_post`, so the two legs cannot silently disagree again. Both `_perform_token_refresh` implementations are patched, the SSO `OAuthManager` and the MCP-facing `OAuthClientManager`, because both build the request the same broken way.

## What happens to existing sessions on deploy

**Nobody is locked out. Each signed-in user is asked to sign in once more, and after that their sessions behave correctly.**

The fix is not retroactive, and it is honest to say so. Sessions that already exist were issued without `offline_access`, so they carry no refresh token and nothing in this change can give them one. Such a session keeps working until it reaches its own 55 minute mark, then dies exactly as it does today, one last time. The next sign-in requests `offline_access`, receives a refresh token, and the patched refresh renews it in place from then on.

The fix ships inside the image, so the deploy has to rebuild `open-webui`. The demo box deploy already runs `up -d --build`, which rebuilds layers whose inputs changed, so no extra step is required.

## Regression guards, each confirmed failing first

**A single message test cannot detect this defect, which is precisely why it shipped.** Every phase-19 chat spec runs all of its turns inside one short session; the multi-turn spec budgets 360 seconds for the whole file, which is also the longest budget anywhere in the suite. Nothing holds a session open past the access token lifetime and nothing edits `expires_at`, so no existing test reaches the refresh path at all. A test that sends one message, or several inside a minute, passes against a build that locks every user out an hour later. That is why these guards are configuration and build invariants rather than another chat spec.

1. **`TestOWUIOAuthScopesRequestOfflineAccess`** scans every YAML file under `deploy/` for `OAUTH_SCOPES` and fails without `offline_access`.
   Proven RED by editing `docker-compose.yml` back to `"openid email profile"` and running it. It failed with `[]string{"openid", "email", "profile"} does not contain "offline_access"`, naming `deploy/docker/docker-compose.yml:649`.

2. **`TestOWUIImageAppliesOAuthRefreshClientAuthPatch`** fails if `Dockerfile.open-webui` stops applying the refresh patch, which nothing else in the repository would notice.
   Proven RED by deleting the `COPY` and `RUN` block for the patch from the Dockerfile and running it. It failed on the missing `owui-patches/hive_oauth_client_auth.py` reference.

3. **`scripts/test_owui_oauth_client_auth.py`**, wired into `make test-scripts`, covers the helper: credentials move out of the body under basic, stay in the body under post, no header for a public client, RFC 6749 form-urlencoding, and no mutation of the caller's dict.
   Proven RED by neutering the helper to always leave the credentials in the body, which is upstream's behaviour. It failed with `KeyError: 'Authorization'`.

The build-time patch also asserts its own effect: exactly two matching call sites, `client` still bound at both, the result parses, and it refuses to half apply. A future digest bump breaks the build loudly instead of silently reverting to hour-long sessions.

### Known limitation: CI never builds this image

No workflow in `.github/workflows/` builds `Dockerfile.open-webui`. The patch's self-assertion is real and fires correctly, but it only runs where the image is actually built, which is the demo box deploy, and that runs from `main`. **Anchor drift from a future open-webui digest bump therefore surfaces after merge, not in the pull request that causes it.**

`TestOWUIImageAppliesOAuthRefreshClientAuthPatch` narrows the gap by failing in CI if the patch is unwired from the Dockerfile, which is the likelier regression, but a Go test cannot detect an anchor that no longer matches upstream source. A CI job that builds this image would close the remainder. Worth filing separately; it is a pre-existing gap that affects all five patches in that Dockerfile, not something this change introduces.

## Verification performed

Live, against the real Supabase project, on the patched image built from this branch, with a throwaway OAuth client registered `client_secret_basic` exactly like production:

- The consent screen now lists `offline_access` alongside `openid`, `email` and `profile`, so the authorize request really did change.
- The session Open WebUI then stored carries a `refresh_token`, read back out of the running container through its own model layer, with `expires_in: 3600`.
- The dialect comparison above, run against the live token endpoint.
- A real SSO login followed by a chat message answered end to end through the per-user token path.

Evidence is in `docs/proof/owui-oauth-refresh-782/`.

## Outstanding before merge

**The wall-clock capture of a single session surviving past its 55 minute refresh window is not yet attached.** Two attempts were started and both were lost when the session driving them was interrupted. Everything above is verified, but the one artifact that shows the whole chain working over real elapsed time is still owed, and this PR should not merge until it is attached. The evidence document states the same thing in the same terms and carries no placeholder standing in for a result.

## Pending mitigation, NOT applied

Raising the Supabase access token lifetime from 3600 to 14400 seconds was approved as a mitigation. **It has not been applied**, because it needs a Supabase Management API token we do not hold.

It is a mitigation, not a fix, and it is not free. `edge-api` validates tokens statelessly on signature and expiry, so a longer lifetime stretches revocation lag from one hour to four. A user who is disabled, removed from a tenant, or whose access is withdrawn keeps working for up to four hours on an already-issued token.

**REVERT LINE: if the access token lifetime is ever raised to 14400, it must be returned to 3600 once this fix is deployed and confirmed. This PR removes the reason the mitigation existed. It must not be allowed to become permanent.**

## Cleanup

The throwaway OAuth client registered for verification has been deleted, confirmed by listing the project's clients rather than by trusting the delete call. Only the three production clients remain: Hive Chat, Open WebUI (Hive demo box), and Open WebUI (Hive CI), all `client_secret_basic`.

## Not merged, not deployed

The demo box deploy runs from `main` only, so nothing here has reached `chat-hive.scubed.co`.

The PR requests `offline_access` for Open WebUI OAuth sessions and patches both refresh implementations to use the client-authentication method selected for login.
- Adds build-time assertions around the pinned Open WebUI source rewrite.
- Adds configuration and helper regression checks to the existing test commands.
- Documents the refresh failure chain, verification evidence, deployment behavior, and remaining end-to-end verification gap.
</details>

The code changes appear safe to merge because no blocking failure remains in the reviewed implementation.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| deploy/docker/owui-patches/hive_oauth_client_auth.py | Builds refresh request arguments for basic, post, and public OAuth clients while preserving the caller’s request body. |
| deploy/docker/owui-patches/apply_oauth_client_auth_patch.py | Rewrites both pinned Open WebUI refresh call sites with strict anchor, binding, count, and syntax assertions. |
| deploy/docker/Dockerfile.open-webui | Installs and applies the OAuth refresh patch while validating the generated module. |
| deploy/docker/docker-compose.yml | Adds `offline_access` so newly authenticated sessions can receive refresh tokens. |
| apps/edge-api/internal/auth/owui_oauth_scope_test.go | Guards deployment scope declarations and verifies that the image patch remains wired into the Dockerfile. |
| scripts/test_owui_oauth_client_auth.py | Exercises authentication-method resolution, credential placement, encoding, public clients, and input immutability. |

</details>

```mermaid
sequenceDiagram
  participant U as User
  participant O as Open WebUI
  participant S as Supabase OAuth
  participant E as Edge API
  U->>O: Sign in with Hive
  O->>S: Authorize with offline_access
  S-->>O: Access and refresh tokens
  O->>E: Chat with per-user access token
  Note over O: Access token approaches expiry
  O->>S: Refresh using configured client auth method
  S-->>O: Rotated access token
  O->>E: Continue chat with refreshed token
```
</details>

<sub>Reviews (2): Last reviewed commit: ["docs: record that OAUTH\_SCOPES cannot be..."](https://github.com/sakibsadmanshajib/hive/commit/495021373f27ef617b33b4dce00c978cd16318b2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50997116)</sub>

**Context used:**

---

## Rebase onto current main, 2026-08-22

The branch history contained a reconciliation merge, and replaying it reapplied the superseded revision this pull request had already withdrawn, the one that pinned `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD`. The net change is therefore reapplied as one commit against `main` at `c30882491` rather than replayed commit by commit. The diff against current `main` is eleven files and about 815 lines; GitHub's earlier 143-file figure was an artefact of a merge base from 2026-08-09.

Nothing in this branch touches control-plane accounts or provisioning code, and it appends no line to `.wolf/buglog.jsonl`.

### Both defects are still present on main, re-confirmed rather than assumed

- `deploy/docker/docker-compose.yml` on `main` still reads `OAUTH_SCOPES: "openid email profile"`. No `offline_access`, so Supabase still issues no refresh token and `_perform_token_refresh` still returns `None` on its first guard.
- `scripts/register-owui-oauth-client.py` still registers every client as `client_secret_basic`, asserted in its own self-check, while upstream Open WebUI still sends `client_secret_post` on the refresh. So a refresh token that did exist would still be refused.

Neither defect is affected by the migration off hosted Supabase: the scope is a property of the authorize request and the client authentication dialect is a property of the registration, and self-hosted GoTrue is the same code enforcing the same rules. The self-hosted cutover in fact re-confirms the second half, since the registration script that stands up clients on the new instance is the one that pins `client_secret_basic`.

The client authentication half is fixed inside the image precisely so it does not depend on any of this: the patch follows the method the client is already registered with and still honours `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD` for a deployment that really is registered for `client_secret_post`, so the two legs cannot silently disagree again on any instance.

### Why this matters more than it did on 2026-08-09

PR #951 renders the agent surface natively and brokers the user's Supabase token server side through `get_system_oauth_token`, which goes through the same `OAuthManager.get_oauth_token` that deletes the OAuth session when a refresh fails. So once #951 lands, this defect takes out the Agents panel on the same 55 minute clock it already takes out chat. This is a prerequisite for a demo that runs longer than an hour from a single sign-in, not only a chat fix.

### Review findings addressed

All five threads resolved.

The two blocking ones were about proof artefacts and are gone on this head, verified by grep rather than by reading: there is no `RESULTS_PLACEHOLDER`, no dangling `after-fix/` or `timed-run.log` reference, and no duplicate `docs/proof/owui-oauth-refresh-782-2026-08-06/` directory. The surviving write up now carries an explicit stale-baseline warning naming the deleted hosted project, so nobody mistakes the timed run for a current measurement.

The three screenshots previously committed under `docs/proof/` are also dropped, for the reason one of those threads exists: per `.wolf/decisions.md` D-042 the image half of a proof capture belongs on the permanent visual-proof release, because a link pinned to a branch name dies the instant that branch is deleted at squash merge. Only the text log stays, where `lint:proof-tokens` can scan it.

The CI gap, that no workflow builds `Dockerfile.open-webui` so a patch anchor can drift past every required check, is filed as **#995** rather than left in a thread. It is pre-existing and affects all five patches in that Dockerfile.

The request to record that `OAUTH_SCOPES` is not shadowable by a persisted row is now in the compose comment block itself, with the trace and with the contrast against `OAUTH_AUTO_REDIRECT` and the feature flags that do go through a reconciler.

### Verification on the rebased head

`make test-scripts` green. `go test ./apps/edge-api/internal/auth/... -short` green. `lint:proof-tokens` green. Compose parses and every Caddyfile still validates against the pinned image.

Three mutation checks, each made to fail on purpose rather than reasoned about:

| mutation | result |
| --- | --- |
| `OAUTH_SCOPES` back to `"openid email profile"` | `TestOWUIOAuthScopesRequestOfflineAccess` fails, naming `deploy/docker/docker-compose.yml:734` |
| patch `COPY`/`RUN` block deleted from the Dockerfile | `TestOWUIImageAppliesOAuthRefreshClientAuthPatch` fails on the missing `owui-patches/hive_oauth_client_auth.py` reference |
| helper neutered to leave credentials in the body, upstream's behaviour | `scripts/test_owui_oauth_client_auth.py` fails with `KeyError: 'Authorization'` |

### Visual proof, and why there is none from this session

Stated as an absence rather than filled with something weaker. The claim this branch makes is that a session survives its first token refresh, which is observable only about 55 minutes after a real sign-in, against a rebuilt image, on a deployment whose OAuth client is registered for the origin the sign-in comes from. None of that is reachable from the development box: the box's own `.env` still points `SUPABASE_URL` at the hosted Supabase project deleted in the migration, which now returns `ENOTFOUND`, so the sanctioned one-time-token mint cannot reach an authorization server at all, and no local origin has a registered OAuth client. Setting or rotating a password to get a session is forbidden and was not attempted.

The honest proof for this change is a post-deploy timed run, and this pull request is not being merged by the agent that rebased it. What exists instead is the mutation table above, plus the earlier evidence document, which is now explicitly labelled as describing the mechanism on a baseline that no longer exists.

### Buglog entry

To be appended to `.wolf/buglog.jsonl` on `main` in a separate buglog-only pull request after this merges.

```json
{"id":"owui-oauth-session-destroyed-at-first-refresh","date":"2026-08-22","error_message":"Every chat completion failed about 55 minutes after sign-in and kept failing until the user signed in again through SSO","root_cause":"Two independent defects. OAUTH_SCOPES omitted offline_access so Supabase issued no refresh token and _perform_token_refresh returned None on its first guard, and the caller then deleted the OAuth session outright. Separately, Open WebUI hand-builds the refresh POST with the client credentials in the form body, which is client_secret_post, while the authorization code exchange goes through authlib with client_secret_basic, and every client this project registers is client_secret_basic, so a refresh token that did exist would have been refused anyway","fix":"Added offline_access to OAUTH_SCOPES, and spliced a helper into both _perform_token_refresh implementations that authenticates the client with the method the client is registered with, rather than pinning OAUTH_TOKEN_ENDPOINT_AUTH_METHOD and forcing every client to be re-registered","tags":["owui","oauth","token-refresh","session","supabase","issue-782"]}
```

### DIFF
--- a/deploy/docker/Dockerfile.open-webui
+++ b/deploy/docker/Dockerfile.open-webui
@@ -221,6 +221,24 @@ RUN python3 /tmp/apply_rag_env_config_patch.py \
 COPY deploy/docker/owui-patches/hive_ui_surfaces.py /tmp/hive_ui_surfaces.py
 COPY deploy/docker/owui-patches/apply_ui_surfaces_patch.py /tmp/apply_ui_surfaces_patch.py
 RUN python3 /tmp/apply_ui_surfaces_patch.py
+
+# #782: Open WebUI runs the authorization code exchange through authlib, which
+# defaults to client_secret_basic, but hand builds the refresh POST with the
+# client credentials in the form body, which is client_secret_post. Supabase
+# enforces the registered method exactly, so every refresh came back 400
+# invalid_credentials, Open WebUI read that as a dead session and deleted it,
+# and chat stopped working roughly 55 minutes after sign-in until the user
+# signed in again. Splice the refresh through a helper that authenticates the
+# client the same way the exchange already does, rather than requiring every
+# deployment to re-register its Supabase OAuth client for the other dialect;
+# the registration is manual ops with no script behind it, so an invariant
+# living only in prose would be reintroduced by the next person who registers
+# a client. Behaviour is covered by scripts/test_owui_oauth_client_auth.py.
+COPY deploy/docker/owui-patches/hive_oauth_client_auth.py /app/backend/open_webui/utils/hive_oauth_client_auth.py
+COPY deploy/docker/owui-patches/apply_oauth_client_auth_patch.py /tmp/apply_oauth_client_auth_patch.py
+RUN python3 /tmp/apply_oauth_client_auth_patch.py \
+    && test "$(grep -c '\*\*hive_refresh_request(client, refresh_data),' /app/backend/open_webui/utils/oauth.py)" = "2" \
+    && python3 -c "import ast, pathlib; ast.parse(pathlib.Path('/app/backend/open_webui/utils/oauth.py').read_text())"
 # Chat was broken on every route because Open WebUI attached tool schemas that
 # no route in this deployment can serve.
 #
--- a/deploy/docker/docker-compose.yml
+++ b/deploy/docker/docker-compose.yml
@@ -703,7 +703,35 @@ services:
       OPENID_PROVIDER_URL: "${SUPABASE_PUBLIC_URL:-${SUPABASE_URL}}/auth/v1/.well-known/openid-configuration"
       OAUTH_CLIENT_ID: ${SUPABASE_OAUTH_CLIENT_ID:-}
       OAUTH_CLIENT_SECRET: ${SUPABASE_OAUTH_CLIENT_SECRET:-}
-      OAUTH_SCOPES: "openid email profile"
+      # offline_access is load-bearing, not decoration (#782). Supabase issues
+      # access tokens with a 3600 second lifetime, and Open WebUI refreshes at
+      # 5 minutes before expiry (OAuthManager.get_oauth_token). When that
+      # refresh runs, _perform_token_refresh returns None on its very first
+      # guard if the stored session has no refresh_token, and the caller then
+      # DELETES the OAuth session outright. From that moment __oauth_token__ is
+      # empty forever, hive_jwt_forward injects no upstream_auth, and every
+      # chat completion 401s with missingUserTokenMessage until the user signs
+      # in again through SSO. Without offline_access on the authorize request
+      # Supabase returns no refresh token at all, so that destruction was
+      # guaranteed roughly 55 minutes into every signed-in session.
+      #
+      # Supabase's own discovery document advertises offline_access in
+      # scopes_supported and refresh_token in grant_types_supported, so this
+      # asks for something the provider already offers. Guarded by
+      # TestOWUIOAuthScopesRequestOfflineAccess in apps/edge-api/internal/auth.
+      #
+      # The scope is only half of #782. A refresh token that exists is still
+      # refused, because Open WebUI's refresh POST authenticates the client in
+      # a different dialect than the authorization code exchange does. That
+      # half is fixed inside the image by Dockerfile.open-webui's
+      # apply_oauth_client_auth_patch.py, deliberately NOT by setting
+      # OAUTH_TOKEN_ENDPOINT_AUTH_METHOD here: pinning that to
+      # client_secret_post would also force every Supabase OAuth client to be
+      # re-registered for the same dialect, and registering those clients is
+      # manual ops with no script behind it. Every client this project has
+      # registered so far is client_secret_basic, so that value would break
+      # sign-in outright on each of them.
+      OAUTH_SCOPES: "openid email profile offline_access"
       # Supabase's OAuth 2.1 authorization server requires PKCE on every
       # authorize request. authlib (OWUI's OAuth client) only sends
       # code_challenge/code_challenge_method when this is set, so without it
--- a/deploy/docker/docker-compose.yml
+++ b/deploy/docker/docker-compose.yml
@@ -720,6 +720,17 @@ services:
       # asks for something the provider already offers. Guarded by
       # TestOWUIOAuthScopesRequestOfflineAccess in apps/edge-api/internal/auth.
       #
+      # This one is NOT shadowable by a persisted row, which is the question
+      # every Open WebUI setting in this file has to answer since #722. Traced
+      # rather than assumed: OAUTH_SCOPES is a plain os.getenv in config.py,
+      # consumed by oidc_oauth_register, which runs at import through
+      # load_oauth_providers(). The oauth.scopes row that Config.seed_defaults
+      # writes is read only by the admin config API in routers/auths.py and by
+      # nothing on the authorize or the refresh path. A container recreate
+      # therefore genuinely picks this up, and no hive_rag_env_config style
+      # reconciler is needed here, unlike OAUTH_AUTO_REDIRECT and the feature
+      # flags that do go through one.
+      #
       # The scope is only half of #782. A refresh token that exists is still
       # refused, because Open WebUI's refresh POST authenticates the client in
       # a different dialect than the authorization code exchange does. That
--- a/deploy/docker/owui-patches/hive_oauth_client_auth.py
+++ b/deploy/docker/owui-patches/hive_oauth_client_auth.py
@@ -1,0 +1,99 @@
+"""Client authentication for Open WebUI's OAuth token refresh (issue #782).
+
+Open WebUI performs the authorization code exchange through authlib, but it
+does NOT perform the refresh through authlib. `_perform_token_refresh` hand
+builds the refresh POST and unconditionally puts `client_id` and
+`client_secret` in the form body. That is the `client_secret_post` dialect.
+authlib, meanwhile, defaults `token_endpoint_auth_method` to
+`client_secret_basic` whenever a client secret is configured
+(authlib/oauth2/client.py, `if token_endpoint_auth_method is None`).
+
+So out of the box the two legs of the same OAuth client speak different
+dialects. Providers that enforce the registered method exactly, Supabase
+among them, accept the login and then reject every refresh with
+400 invalid_credentials: "client is registered for 'client_secret_basic' but
+'client_secret_post' was used". Open WebUI reads that failure as a dead
+session and deletes it, which locks the user out of chat entirely until they
+sign in again.
+
+The fix is to make the refresh speak whatever dialect the code exchange
+already speaks, rather than to force every deployment to re-register its
+OAuth client. `token_endpoint_auth_method` is carried on the authlib client's
+own `client_kwargs` (Open WebUI puts it there in config.py's
+`oidc_oauth_register`), so this reads the single value that already governs
+the exchange and cannot drift away from it.
+
+Resolution order matches authlib's own so that an unset value behaves
+identically on both legs:
+
+  * explicit `client_kwargs['token_endpoint_auth_method']` wins, which is what
+    `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD` sets,
+  * otherwise `client_secret_basic` when a client secret exists,
+  * otherwise `none`, the public-client case, where there is no secret to send.
+
+This module is copied into the image at
+/app/backend/open_webui/utils/hive_oauth_client_auth.py and called from both
+`_perform_token_refresh` implementations by
+`apply_oauth_client_auth_patch.py`. It imports nothing from Open WebUI, so
+`scripts/test_owui_oauth_client_auth.py` can exercise it directly.
+"""
+
+import base64
+from urllib.parse import quote
+
+FORM_CONTENT_TYPE = "application/x-www-form-urlencoded"
+
+CLIENT_SECRET_BASIC = "client_secret_basic"
+CLIENT_SECRET_POST = "client_secret_post"
+NO_CLIENT_AUTH = "none"
+
+
+def resolve_auth_method(client) -> str:
+    """Return the token endpoint auth method authlib would use for `client`.
+
+    Mirrors authlib's OAuth2Client.__init__ fallback so the refresh leg and
+    the authorization code exchange leg can never disagree.
+    """
+    client_kwargs = getattr(client, "client_kwargs", None) or {}
+    configured = client_kwargs.get("token_endpoint_auth_method")
+    if configured:
+        return configured
+    return CLIENT_SECRET_BASIC if getattr(client, "client_secret", None) else NO_CLIENT_AUTH
+
+
+def basic_auth_header(client_id: str, client_secret: str) -> str:
+    """Build an RFC 6749 section 2.3.1 Basic credential.
+
+    The spec requires each half to be form-urlencoded before the base64, which
+    matters for any secret containing a colon or a plus.
+    """
+    raw = f"{quote(client_id, safe='')}:{quote(client_secret, safe='')}"
+    return "Basic " + base64.b64encode(raw.encode("utf-8")).decode("ascii")
+
+
+def hive_refresh_request(client, refresh_data: dict) -> dict:
+    """Return the aiohttp POST keyword arguments for a refresh request.
+
+    Returns `{"data": ..., "headers": ...}` so the call site stays a single
+    `**hive_refresh_request(client, refresh_data)` expression, which keeps the
+    build-time splice to one substitution per call site.
+
+    For `client_secret_basic` the credentials move out of the body and into
+    the Authorization header. They are removed from the body rather than
+    duplicated: sending both is how a request comes to look like two different
+    authentication methods at once, which providers reject on principle
+    (RFC 6749 section 2.3 permits exactly one).
+
+    Every other method, `client_secret_post` and `none` included, keeps
+    upstream's body exactly as built, so this is a no-op for providers that
+    were already working.
+    """
+    headers = {"Content-Type": FORM_CONTENT_TYPE}
+    client_secret = getattr(client, "client_secret", None)
+
+    if resolve_auth_method(client) != CLIENT_SECRET_BASIC or not client_secret:
+        return {"data": refresh_data, "headers": headers}
+
+    data = {k: v for k, v in refresh_data.items() if k not in ("client_id", "client_secret")}
+    headers["Authorization"] = basic_auth_header(client.client_id, client_secret)
+    return {"data": data, "headers": headers}
--- a/docs/proof/owui-oauth-refresh-782/refresh-path-evidence.md
+++ b/docs/proof/owui-oauth-refresh-782/refresh-path-evidence.md
@@ -1,0 +1,225 @@
+# Issue 782 evidence: Open WebUI OAuth session destruction
+
+Two capture sessions, both against the hosted Supabase project that served this
+deployment at the time.
+
+* 2026-08-06, against the demo box `chat-hive.scubed.co`: screenshots recording
+  the failure symptom, posted as release assets on the pull request rather than
+  committed here (`.wolf/decisions.md` D-042: a branch-pinned image link dies
+  when the branch is deleted at squash merge, so only the text log lives in this
+  directory, where `npm run lint:proof-tokens` can scan it).
+* 2026-08-08, against a local stack running the patched image and that same
+  hosted Supabase project: every claim in sections 1 to 6 below, each re-derived
+  from scratch rather than carried over from the earlier session.
+
+STALE BASELINE, read this first. Both sessions predate the migration of this
+deployment off hosted Supabase onto self-hosted Supabase on the demo box, and
+the hosted project has since been deleted. The two defects this branch fixes
+were re-confirmed against the current baseline on 2026-08-22 by reading
+configuration rather than by replaying the capture: `docker-compose.yml` still
+declares `OAUTH_SCOPES` without `offline_access`, so Supabase still issues no
+refresh token, and `scripts/register-owui-oauth-client.py` still registers every
+client as `client_secret_basic` while upstream Open WebUI still sends
+`client_secret_post` on the refresh. Neither the scope nor the client
+authentication dialect is affected by where the Supabase instance runs. The
+timed 55 minute run below has NOT been repeated against the self-hosted
+instance; treat the numbers as evidence of the mechanism, not as a current
+measurement.
+
+No token, secret, password or connection string appears in this file or in any
+screenshot beside it.
+
+## 1. The trigger rule
+
+Elapsed time since sign-in. Not message count, and not which conversation.
+
+Supabase issues the access token with a 3600 second lifetime, confirmed on the
+live token response (`expires_in: 3600`, section 4). Open WebUI refreshes once
+`now + 5 minutes` reaches `expires_at`:
+
+```python
+# utils/oauth.py, OAuthManager.get_oauth_token, v0.10.2
+if (
+    force_refresh
+    or session.expires_at is None
+    or datetime.now() + timedelta(minutes=5) >= datetime.fromtimestamp(session.expires_at)
+):
+```
+
+So the refresh path is first entered about 55 minutes into a signed-in
+session, and until then nothing is wrong. A conversation opened on an already
+dead session fails from its first message, which is why the original report
+read this as a second-message defect.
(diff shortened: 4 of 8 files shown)


docs/proof/owui-oauth-refresh-782/refresh-path-evidence.md
```diff
@@ -0,0 +1,188 @@
...
+expires_in: 3600
+```
+
+### The timed run
+
+One session, one conversation, one message every five minutes, past the
+refresh window at 55 minutes and past the token expiry at 60. See
+`after-fix/timed-run.log` for the full log and `after-fix/` for the frames.
+
+RESULTS_PLACEHOLDER
```
sakibsadmanshajib: Blocking, and the one objective defect I found in this head. This is a literal `RESULTS_PLACEHOLDER` in a committed evidence document, sitting under a heading that promises a timed run. The two references to `after-fix/` at line 8 and line 153 are also dangling: `docs/proof/owui-oauth-refresh-782/` contains only the three before pictures and this markdown file, no `after-fix/` directory and no `timed-run.log`.

The PR body is honest that the wall clock capture is missing, so this is not an attempt to overclaim, but a proof artifact that says a timed run is in a directory that does not exist is worse than one that says nothing. Either fill this in with the run when it happens, or cut the section and the two `after-fix/` references and let the PR body carry the admission alone.

docs/proof/owui-oauth-refresh-782-2026-08-06/refresh-path-evidence.md
```diff
@@ -0,0 +1,143 @@
+# Issue 782 evidence: Open WebUI OAuth session destruction
```
sakibsadmanshajib: Blocking, merge residue. This directory is the evidence for the superseded approach and it is byte for byte duplicated by `docs/proof/owui-oauth-refresh-782/`, including all three PNGs. Its section 4 still reads that the fix was verified with "`client_secret_post` on both the client registration and Open WebUI's token endpoint auth method", which is precisely the approach this head withdrew because all three live Supabase clients are registered `client_secret_basic` and it would have broken sign in.

Leaving both directories in the tree means the next person debugging this reads two contradictory accounts of the same issue number and has no way to tell which one shipped. Please delete `docs/proof/owui-oauth-refresh-782-2026-08-06/` and keep the dated write up in the surviving directory only.

deploy/docker/Dockerfile.open-webui
```diff
@@ -88,8 +88,52 @@ RUN python3 /tmp/apply_tenant_role_patch.py \
...
+# and chat stopped working roughly 55 minutes after sign-in until the user
+# signed in again. Splice the refresh through a helper that authenticates the
+# client the same way the exchange already does, rather than requiring every
+# deployment to re-register its Supabase OAuth client for the other dialect;
+# the registration is manual ops with no script behind it, so an invariant
+# living only in prose would be reintroduced by the next person who registers
+# a client. Behaviour is covered by scripts/test_owui_oauth_client_auth.py.
+COPY deploy/docker/owui-patches/hive_oauth_client_auth.py /app/backend/open_webui/utils/hive_oauth_client_auth.py
+COPY deploy/docker/owui-patches/apply_oauth_client_auth_patch.py /tmp/apply_oauth_client_auth_patch.py
+RUN python3 /tmp/apply_oauth_client_auth_patch.py \
```
sakibsadmanshajib, line 133: Note, not blocking. I verified this assertion actually fires: I shifted the anchor in `apply_oauth_client_auth_patch.py` to simulate an upstream source change and the build died here with `expected exactly 2 refresh POST call sites matching the upstream data/headers pair, found 0`. That is the right posture and it matches the other patches in this file.

The gap is where it fires. `ci.yml` never builds `Dockerfile.open-webui`, so on a pull request that bumps the pinned open-webui digest, every required check stays green and this only breaks later, in `owui-nightly.yml` or in `deploy-demo-box.yml` after the merge has already landed on main. The Go guard added in this PR covers deletion of the block but cannot cover drift of the anchor, since it only greps the Dockerfile for the reference. Worth a follow up issue to build this image in CI, or at least to run the patch script against the pinned image there.

sakibsadmanshajib (commented):
Independent review of head 2ae7467, run in an isolated worktree against the pinned Open WebUI image and the live Supabase project. Verdict up front: the code is correct and I approve it. The remaining items below are one real documentation defect, one stale artifact that contradicts the shipped fix, and two notes. Merge stays blocked on the author's own stated blocker plus the cleanup in the first two comments.

## 1. The near miss is real, confirmed independently

I queried `auth.oauth_clients` on the live project directly. Every client that is not soft deleted is registered `client_secret_basic`:

| client_name | token_endpoint_auth_method | created | deleted |
| --- | --- | --- | --- |
| Open WebUI (Hive CI) | client_secret_basic | 2026-07-03 | no |
| Open WebUI (Hive demo box) | client_secret_basic | 2026-07-26 | no |
| Hive Chat | client_secret_basic | 2026-07-26 | no |

The only `client_secret_post` row in the table is a throwaway diagnostic client, already deleted. The previous head of this PR set `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD: client_secret_post` in compose, and `.github/workflows/deploy-demo-box.yml` deploys on push to main whenever `deploy/docker/**` changes. Merging that head would have rebuilt and restarted Open WebUI with authlib sending a dialect all three registered clients reject, so the authorization code exchange would have failed and nobody could have signed in to chat at all. Withdrawing it was the right call, and the replacement is strictly better because it removes the ops coupling rather than documenting it.

## 2. Both refresh call sites are patched, and the exchange leg is untouched

Built the image from this branch and read the result out of it. `/app/backend/open_webui/utils/oauth.py` in the built image carries the import at line 87 and `**hive_refresh_request(client, refresh_data),` at lines 1103 and 1379, which are the MCP `OAuthClientManager._perform_token_refresh` and the SSO `OAuthManager._perform_token_refresh` respectively. Zero occurrences of the original `headers={'Content-Type': 'application/x-www-form-urlencoded'}` pair survive. The splice only replaces the two POST argument pairs and adds one import, so the authlib authorization code exchange in `config.py` `oidc_oauth_register` is not touched at all, which is why sign in keeps working with the existing basic registrations.

The helper's fallback is not an approximation. Authlib's own resolution lives at `authlib/oauth2/client.py` lines 75 to 79 in the pinned image: unset means `client_secret_basic` when a secret exists, else `none`. `resolve_auth_method` mirrors that exactly, and reads `client.client_kwargs['token_endpoint_auth_method']` first, which is the same value `config.py` line 2630 feeds into the exchange. The two legs now cannot drift apart.

One behaviour worth stating plainly since it is the whole point: for a deployment genuinely registered `client_secret_post` the helper is a no op, verified by `test_post_registration_keeps_upstream_behaviour`.

## 3. All three guards were broken by hand and all three went red

Not taken on faith. Each mutation applied in my worktree, run, then reverted.

1. `deploy/docker/docker-compose.yml` `OAUTH_SCOPES` reverted to `"openid email profile"`. `TestOWUIOAuthScopesRequestOfflineAccess` failed with `[]string{"openid", "email", "profile"} does not contain "offline_access"` and named `docker-compose.yml:687`.
2. The `COPY` plus `RUN` block deleted from `deploy/docker/Dockerfile.open-webui`. `TestOWUIImageAppliesOAuthRefreshClientAuthPatch` failed with `no longer references "owui-patches/hive_oauth_client_auth.py"`.
3. `hive_refresh_request` neutered to always return upstream's body. `scripts/test_owui_oauth_client_auth.py` failed at `test_basic_credential_is_form_urlencoded_per_rfc6749` with `KeyError: 'Authorization'`.

I added a fourth mutation the PR does not claim, to test the durability story rather than the guards. I shifted the patch anchor to simulate an upstream source change and rebuilt: the build failed at `Dockerfile.open-webui:133` with `expected exactly 2 refresh POST call sites matching the upstream data/headers pair, found 0`. So an image bump that moves this code cannot silently drop the fix.

## 4. Durability of the configuration half

`OAUTH_SCOPES` is read by `os.getenv` at `config.py` line 2471 and consumed by `oidc_oauth_register` at line 2628, which runs at import time through `load_oauth_providers()` at line 2705. The persisted `oauth.scopes` row seeded by `Config.seed_defaults` is read only by the admin config API in `routers/auths.py` lines 1316 and 1318 and by nothing on the authorize or refresh path. So unlike the RAG keys in issue #722, this value cannot be shadowed by a stale database row: a container recreate picks up the compose value. That is worth knowing and is not stated anywhere in the PR.

## 5. Blast radius on deploy

Nobody is locked out, and sign in is not affected at all.

* Sign in: unchanged. No client is re authenticated differently, because `OAUTH_TOKEN_ENDPOINT_AUTH_METHOD` stays unset and authlib keeps using `client_secret_basic`, which is what all three live clients are registered for.
* Existing sessions: they were issued without a refresh token, so they still die at their own 55 minute mark exactly as they do today. The user signs in once more and the new session survives. No account is stranded.
* New sessions: the authorize request now carries a fourth scope, so a user may see the consent screen list `offline_access`. That is cosmetic.
* `hive_jwt_forward` forwards only `access_token` (`deploy/docker/pipelines/hive_jwt_forward.py` line 68), and a Supabase refresh returns a fresh `access_token`, so the forwarded credential is still the right one after a refresh.

## 6. On the missing wall clock capture

Asked for my opinion as reviewer, so plainly: the blocker is real, but it is the last one, and it is real for policy reasons more than technical ones.

Technically the composition is well covered. The provider seam is proven live in the evidence doc: the same endpoint answers a `client_secret_post` refresh with `invalid_credentials` before it even looks at the token, and answers the `client_secret_basic` shape with `refresh_token_not_found`, which is a provider that authenticated the client and then failed only on the deliberately invalid token. The splice seam is proven in the built image, both call sites. The dialect resolution is proven against authlib's own source. What is not proven is only that those three seams compose over a real hour in the real container, which is exactly the class of thing this repository has been burned by before.

Against that, the standing rule is that a fix touching a live UI surface does not merge without a capture taken against the running stack after the change, and chat is the first surface a user touches. The capture is also cheap: deploy, sign in, send a message every five minutes past minute 60. I would deploy and capture rather than merge on reasoning. The risk of the deploy itself is low for the reasons in section 5.

## 7. What has to change before merge

See the inline comments. Two are cleanup of committed artifacts, one of which is a placeholder string shipped in a proof document, and two are notes that do not need to block.

deploy/docker/owui-patches/hive_oauth_client_auth.py
```diff
@@ -0,0 +1,99 @@
...
+    """Return the token endpoint auth method authlib would use for `client`.
+
+    Mirrors authlib's OAuth2Client.__init__ fallback so the refresh leg and
+    the authorization code exchange leg can never disagree.
+    """
+    client_kwargs = getattr(client, "client_kwargs", None) or {}
+    configured = client_kwargs.get("token_endpoint_auth_method")
+    if configured:
+        return configured
+    return CLIENT_SECRET_BASIC if getattr(client, "client_secret", None) else NO_CLIENT_AUTH
```
sakibsadmanshajib, line 61: Nit, no change requested, recording it so the next reader does not have to derive it. This fallback also governs the MCP leg, since the same helper is spliced into `OAuthClientManager._perform_token_refresh`. I checked that path: `add_client` puts `token_endpoint_auth_method` into `client_kwargs` from `OAuthClientInformationFull`, whose model default is `client_secret_post`, and the dynamic registration path sets it explicitly, so `resolve_auth_method` returns `client_secret_post` and this helper is a no op for MCP tool servers.

The only way an MCP client reaches the `client_secret_basic` branch is if its `client_info.token_endpoint_auth_method` is explicitly null while a client secret exists, which would mean upstream stopped defaulting it. In that case the refresh would move to Basic, which is also what authlib would do for that client's own exchange, so the two legs would still agree. Correct either way, just non obvious.

deploy/docker/docker-compose.yml
```diff
@@ -613,12 +613,78 @@ services:
...
+      # a different dialect than the authorization code exchange does. That
+      # half is fixed inside the image by Dockerfile.open-webui's
+      # apply_oauth_client_auth_patch.py, deliberately NOT by setting
+      # OAUTH_TOKEN_ENDPOINT_AUTH_METHOD here: pinning that to
+      # client_secret_post would also force every Supabase OAuth client to be
+      # re-registered for the same dialect, and registering those clients is
+      # manual ops with no script behind it. Every client this project has
+      # registered so far is client_secret_basic, so that value would break
+      # sign-in outright on each of them.
+      OAUTH_SCOPES: "openid email profile offline_access"
```
sakibsadmanshajib, line 687: Confirmed correct, and one fact worth adding to this comment block because it is the question everyone asks about Open WebUI settings in this repository after issue #722. This value is not shadowable by a persisted row. `OAUTH_SCOPES` is a plain `os.getenv` at `config.py` line 2471, consumed by `oidc_oauth_register` at line 2628, which runs at import through `load_oauth_providers()` at line 2705. The `oauth.scopes` row that `Config.seed_defaults` writes is read only by the admin config API at `routers/auths.py` line 1316 and by nothing on the authorize or refresh path, so a container recreate genuinely picks this up and no `hive_rag_env_config` style reconciler is needed here.

Issue #758, opened 2026-08-06
Admin and RAG scope gates have no reader and are excluded from the featuregate propagation path


sakibsadmanshajib:
## Summary

Six further gates are rendered in the console with no runtime reader. Five of them are additionally excluded from the only path that carries gate values out of control-plane, so even a correctly written reader in edge-api would receive nothing. The sixth is propagated but unenforced.

Keys: `ENABLE_ADMIN_CONSOLE`, `ENABLE_MULTI_TENANT`, `ENABLE_PROVIDER_CUSTOM` (category `admin`), `ENABLE_RAG_PERSONAL`, `ENABLE_RAG_SHARED_KB` (category `rag`), `ENABLE_RELAY` (category `agents`).

## Evidence

No reader:

* Constants declared at `apps/control-plane/internal/tenant/settings/keys.go:24-26,36,37` and `:46`.
* Outside `keys.go`, every reference to these constants in the repository is in a `_test.go` file.
* `settings.Resolver.IsEnabled` (`apps/control-plane/internal/tenant/settings/resolver.go:43`) has no production call site, so no per key read of any of them happens at runtime.

Not propagated, for five of the six:

* `clientVisibleGateCategories` at `apps/control-plane/internal/tenant/settings/resolver.go:109` is `[]string{"agents", "sso"}`. Categories `admin` and `rag` are excluded, so `ClientVisibleEnabled` (`:134`) never returns them and the `/internal/featuregate/{tenant_id}` response consumed by edge-api never carries them.
* The exclusion is deliberate and correct as a security decision. It was made under issue #293 to stop an authenticated end user learning the deployment's commercial and operational posture. The defect is not the exclusion. The defect is rendering an operator control whose value has no route to any enforcement point, while the exclusion makes that permanent for anything in `admin` or `rag`.
* Category assignments are seeded in `supabase/migrations/20260715_04_featuregate_dynamic_keys.sql:44,45` (`rag`) and `:35,56,57` (`admin`).

`ENABLE_RELAY` is a different shape:

* It sits in category `carl`, renamed to `agents` by `supabase/migrations/20260719_01_rename_carl_feature_category.sql`, so unlike the other five it is client visible and does reach edge-api.
* It has a `Feature` constant, `featuregate.FeatureRelay` at `apps/edge-api/internal/featuregate/gate.go:35`.
[...]

Issue #750, opened 2026-08-06
Platform admin grant is not workspace scoped, so it leaks across every workspace an owner touches


sakibsadmanshajib:
## Summary

`IsPlatformAdmin` resolves the platform admin overlay from the user alone. It never considers which workspace the request is acting on. Any owner membership on any account whose `is_platform_admin` column is true makes that user a platform admin everywhere, for every workspace they can reach.

## Evidence

`apps/control-plane/internal/platform/role_pgx.go:122-138`

```go
func (s *pgxRoleStore) IsPlatformAdmin(ctx context.Context, userID uuid.UUID) (bool, error) {
	var isAdmin bool
	err := s.pool.QueryRow(ctx, `
		SELECT EXISTS(
			SELECT 1
			FROM public.account_memberships m
			JOIN public.accounts a ON a.id = m.account_id
			WHERE m.user_id = $1
			  AND m.role = 'owner'
			  AND a.is_platform_admin = true
		)
	`, userID).Scan(&isAdmin)
```

The signature accepts only a `userID`. There is no account or workspace parameter, and the query has no join back to the account being acted on. The existential form is the whole problem: it asks whether a flagged account exists somewhere in the user's membership set, not whether the flagged account is the one under request.

Note for anyone searching: the function lives in `internal/platform`, not `internal/accounts`.

## Consequence

1. The overlay leaks across workspace boundaries. A user who owns one flagged workspace carries platform admin into every other workspace they belong to. The gate has no way to express "admin in workspace A but an ordinary member in workspace B", because it never learns which workspace it is being asked about.

2. Co-ownership silently escalates. Adding any second owner to a flagged account grants that person the full platform admin surface across the whole deployment, with no separate grant and nothing in the members UI that indicates it happened. This is live today, not hypothetical: the `hive-demo-owner` account carried two owner memberships, so the second owner inherited the same platform wide overlay as the first.

3. Revocation is coarse and cannot target a person. The only available lever is the boolean on the account row, which strips the overlay from every owner of that account simultaneously. There is no way to revoke one user while leaving another intact, so a legitimate admin and an account that merely needs cleaning up cannot be separated.

4. The blast radius grows quietly. Because the check is global, capabilities such as credit grant minting and provider rewrite are reachable from any session that user holds, on any host, including public demo surfaces.

## Suggested direction

[...]

Issue #747, opened 2026-08-06
Demo account holds platform-wide privilege on a public host, and stripping the bit costs less than it looks


sakibsadmanshajib:
## Summary

The account used to drive live external demos of the console carries `accounts.is_platform_admin = true`. That bit is not a console convenience flag: it is the platform-wide authorization overlay, and it unlocks capabilities that reach every tenant on the deployment and the money path. The demo runs on a publicly reachable host, and the same bearer token a normal console login produces is sufficient for all of it.

This issue is not a proposal to strip the bit. It is a request to decide, with the actual cost of stripping it laid out, because the cost turns out to be much smaller than it looks.

## What the bit unlocks

Three capabilities, all reachable with an ordinary session bearer.

**1. Credit grant minting.** `POST /v1/admin/credit-grants` creates a credit grant and writes the matching ledger entry in the same transaction. The handler is `apps/control-plane/internal/grants/http.go:144-201`. It is gated once at the route in `apps/control-plane/cmd/server/main.go:1206-1210` via `authzMW.RequirePermission(authz.PermPlatformAdmin)`, and a second time inside the service at `apps/control-plane/internal/grants/service.go:62-67`. Both gates resolve to the same predicate. The amount is an arbitrary positive `math/big` value in BDT subunits; the only validation is that it parses and is positive (`grants/http.go:168-176`). Grants can target any `granted_to_user_id` plus `granted_to_workspace_id` pair, not just the caller's own.

**2. Provider `base_url` rewrite.** `PUT /api/v1/admin/providers/{id}` replaces a custom provider record wholesale, `base_url` included: `apps/control-plane/internal/providers/http.go:144-166`, with the request shape at `:92-100`. Gated at the route only, `apps/control-plane/internal/platform/http/router.go:280-286`. The handler itself has no authorization logic. Repointing a provider's `base_url` redirects traffic that carries our provider API key; #719 already documents a related failure mode on the same endpoint (omitted fields blanked, and a blanked `base_url` sending our OpenRouter key at another host).

[...]

PR #788, opened 2026-08-06
fix: gate workspace admin areas on the workspace owner, not platform admin

sakibsadmanshajib:
## What

Re-gates the two workspace-scoped admin areas in the web console on the OWNER of the workspace in scope instead of platform admin, and leaves every genuine platform operation exactly where it was.

Stripping `is_platform_admin` from the demo accounts earlier today was correct and is not reverted here. What it exposed is that the gate was in the wrong place: the entire ADMIN nav group became permanently unreachable for the account that OWNS the workspace, so `/console/feature-gates` and `/console/marketplace` both rendered "Admin access required" for the organization owner.

Fixes the symptom behind #758. Deliberately avoids the cross-workspace shape tracked by #750. Related: #747, #748.

## Classification

Every surface currently gated on `is_platform_admin`, and what it is:

| Surface | Scope | Verdict |
| --- | --- | --- |
| `GET /api/v1/admin/feature-gates` | per tenant rows, tenant from the session | workspace administration, re-gated |
| `PUT /api/v1/admin/feature-gates/{key}`, categories `carl`, `rag`, `sso`, `audit_sink`, `feature` | per tenant rows | workspace administration, re-gated |
| `PUT /api/v1/admin/feature-gates/{key}`, categories `billing`, `admin` | per tenant rows, but plan entitlements and deployment shape | platform operation, unchanged |
| `GET /api/v1/admin/marketplace` | global catalog joined with this tenant enablement | workspace administration, re-gated |
| `GET /api/v1/admin/marketplace`, the `config` field of each row | raw configuration of the one global catalog, and an MCP entry can hold a credential in its `env` | platform operation, withheld from a caller who may not curate (added after the security review, finding 2) |
| `PUT /api/v1/admin/marketplace/{id}/enable` | per tenant enablement, RLS scoped | workspace administration, re-gated |
| `POST`, `PUT`, `DELETE` on `/api/v1/admin/marketplace` entries | one global catalog shared by every tenant | platform operation, unchanged |
| `/api/v1/admin/providers*` | provider records including `base_url` | platform operation, unchanged |
| `POST /v1/admin/credit-grants` | mints credit | platform operation, unchanged |
| `IsPlatformAdmin` reads in accounting, budgets, apikeys, ledger, usage, profiles, accounts | populate `Actor.IsAdmin` for the authz policy | overlay only, owners already reach these routes, unchanged |

Stated uncertainty rather than guessed:

- The `sso` and `audit_sink` gates read as organization-owner concerns in a Claude Enterprise sense and are treated as workspace-scoped. Nothing enforces them today, so the classification carries no live risk either way.
- The `billing` and `admin` gates are also inert today. They are held platform-only on the strength of what they will mean once wired: `ENABLE_EXTRA_USAGE` is spend beyond plan and `ENABLE_PROVIDER_CUSTOM` is the custom provider endpoint switch, which is the same authority as the provider `base_url` rewrite this PR deliberately does not move.
- Self-serve signups are seeded `tenant_users.role = MEMBER` on purpose (`insertPersonalMembership`, because OWNER maps to an Open WebUI instance admin through `custom_access_token_hook`). Seven users in the live database own their billing account without holding tenant OWNER, and they still see the contact-your-administrator line. Widening authority to a second predicate is exactly the shape that produced #750, so it is reported rather than done.

## How

`platform.WorkspaceAdminGate` admits the OWNER of the tenant on the session, or a platform admin, and stamps the platform-admin overlay into the request context so the handlers can keep their platform-only carve-outs. Scope comes from `auth.Viewer.TenantID`, which `auth.Client` has already validated against a live membership, and the owner predicate then requires an ACTIVE OWNER row on that same tenant.

The console pages stopped pre-judging access with a local permission check that can disagree with the server. They call the control plane and render one line pointing at the administrator on a 403. The list responses carry `manageable` per gate and `can_curate` for the catalog, so no control is rendered that the API would refuse.

## Tests

Every new test was made to fail against the unpatched behaviour first, by mutating the fix and re-running:

| Mutation | Test that failed |
| --- | --- |
| `isOwner := false` in the gate, the pre-fix platform-admin-only behaviour | `TestWorkspaceAdminGate_TenantOwnerPasses` got 403, want 200 |
| `isOwner := true`, scope check removed | `TestWorkspaceAdminGate_OwnerOfAnotherWorkspaceDenied` and `..._MemberDenied` got 200, want 403 |
| feature-gate carve-out condition forced false | `TestAdmin_WorkspaceOwnerCannotToggleBillingGate` got 200 on a billing gate, want 403 |
| `manageableBy(category, true)` for every row | `TestAdmin_ListMarksPlatformManagedGatesUnmanageable` failed on the billing gate |
| `canCurate := true` | `TestAdmin_WorkspaceOwnerListsCatalog` and all three verbs of `TestAdmin_WorkspaceOwnerCannotCurateCatalog` failed |
| every entry wired as enabled | `TestAdmin_WorkspaceOwnerEnableIsScopedToOwnTenant` reported the leak, and the pre-existing `TestAdmin_SetEnabled_RoundTrip` failed with it |

Cross-workspace isolation is asserted twice: at the gate, where an OWNER of tenant A selecting tenant B is refused, and at the handlers, where a tenant id in the request body cannot redirect a feature-gate write and an enablement by one workspace leaves a second workspace showing the entry disabled.

`go test ./apps/control-plane/... -count=1 -short` is green. Web console: `tsc --noEmit` clean, `npm run test:unit` green apart from the pre-existing `control-plane-host.test.ts` failure, which is a missing `/app/.env.example` in the container image and fails identically on `main`.

## Visual proof

Attached in `docs/proof/issue-758-workspace-scoped-admin/`, captured against control-plane and web-console built from this branch (isolated compose project, live Supabase) with an account whose role shape matches the demo account: `tenant_users.role = OWNER` ACTIVE, `account_memberships.role = owner`, and `accounts.is_platform_admin = false`, confirmed by direct query before capture. No account was granted `is_platform_admin` for this proof.

- `feature-gates-owner.png`: 15 live toggles, 10 rows reading "Managed by your administrator", no wall.
- `marketplace-owner.png`: the page renders for the owner with no curation form.
- `marketplace-owner-enabled.png`: a catalog entry enabled by that owner through the console, with no curate form and no delete control. The write was confirmed in the database as one row on that owner tenant attributed to that owner user, and the proof catalog entry was deleted afterwards.

The login email is synthetic and no flow here carries a credential in a URL. The proof account password was rotated to an unheld value after capture.

## Note on checks

GitHub Actions is in a major outage today, so required checks may never start. Missing checks are not passing checks. Do not merge on a green-looking empty check list.

## Security review follow up (2026-08-08)

The independent review returned CHANGES REQUESTED on test integrity and on one read widening. All four in-scope findings are addressed in `8af9d83`.

- Finding 1, the wiring was untested. `apps/control-plane/internal/platform/http/router_workspace_admin_test.go` dispatches through `NewRouter` with a stub gate that records that it ran and refuses, and asserts all four workspace-scoped admin paths are answered by the gate rather than the handler. A second test asserts that a nil gate leaves those routes unmounted rather than falling back to a wider gate. Proof: deleting the `Require` call from the feature-gate mount turns the full control-plane suite red on `feature_gate_list` and `feature_gate_toggle`; restoring it and deleting the marketplace one turns it red on `marketplace_list` and `marketplace_enable`; restored, the suite is green.
- Finding 2, the read widening. `entryWires` withholds `config` from a caller who may not curate, so the field is absent for a workspace owner and present for a platform admin. The classification table above gained the row. Omitting the field alone would have broken the page the rest of this PR fixes: `decodeMarketplaceEntry` treated a missing `config` as a malformed row and failed the whole list, so the console decoder now reads an absent `config` as withheld. `supabase/migrations/20260808_01_marketplace_entries_comment.sql` corrects the `marketplace_entries` table comment, which still named platform admin as the gate. Comment only, no schema change.
- Finding 3 is pre-existing and out of scope here. Filed as #790: `tenant_users` RLS never constrains the `role` column being written, so a tenant ADMIN can self-promote to OWNER through PostgREST. Confined to their own tenant, because the token hook pins the claims.
- Finding 4, the seeder docstring no longer claims the script grants platform admin, and records that the merge-duplicates upsert clears the flag on a re-run.
- Finding 5, the e2e assertion is now scoped to a gate row, which the 403 wall cannot render, and the marketplace spec asserts the absence of the one line only its deny wall produces. The reviewer's separate note stands: the account behind `E2E_WORKSPACE_OWNER_*` and its `E2E_PLATFORM_ADMIN_*` fallback needs to be an ACTIVE OWNER of its tenant now, and that is an environment fact this branch cannot assert.

Verification: `gofmt`, `go vet` and `go test ./apps/control-plane/... -count=1 -short` clean; web-console `tsc --noEmit` clean; `npm run test:unit` 415 passed with the same pre-existing `control-plane-host.test.ts` container failure (missing `/app/.env.example`) that fails identically on `main`.


apps/control-plane/internal/platform/http/router.go
```diff
@@ -292,13 +302,13 @@ func NewRouter(cfg RouterConfig) http.Handler {
...
-	if cfg.FeatureGateAdminHandler != nil && cfg.RoleSvc != nil && cfg.AuthMiddleware != nil {
+	// Issue #292 (blueprint Step 1.2) — admin feature-gate CRUD, gated on the
+	// workspace administrator (JWT path). GET lists the registry joined with the
+	// tenant enablement; PUT toggles one gate for the selected tenant. Mirrors
+	// providers CRUD stays platform-admin only, see the mount above.
+	if cfg.FeatureGateAdminHandler != nil && cfg.WorkspaceAdminGate != nil && cfg.AuthMiddleware != nil {
 		adminFeatureGates := cfg.AuthMiddleware.Require(
-			cfg.RoleSvc.RequirePlatformAdmin(cfg.FeatureGateAdminHandler.AdminMux()),
+			cfg.WorkspaceAdminGate.Require(cfg.FeatureGateAdminHandler.AdminMux()),
 		)
```
sakibsadmanshajib, line 312: Blocking. Nothing in the repository proves either of these two mounts still carries the gate.

I deleted `cfg.WorkspaceAdminGate.Require(...)` from this mount, leaving `cfg.AuthMiddleware.Require(cfg.FeatureGateAdminHandler.AdminMux())`, and ran `go test ./apps/control-plane/... -count=1 -short`. The entire suite was green. I then restored this one and did the same to the marketplace mount below. Green again.

What that mutant actually ships is: any authenticated user with a selected tenant, including a plain MEMBER, can list and toggle every non billing, non admin feature gate of their own workspace. The carve outs survive it, because `PlatformAdminFromContext` returns false when the stamp is absent, so that is not a cross tenant hole. It is still the whole OWNER requirement silently deleted, and no test notices.

This is the class of defect the repository has shipped repeatedly: a fix with excellent unit coverage of the predicate and zero coverage of whether the predicate is installed. `router_providers_test.go` already has the pattern in `TestNewRouterAdminProvidersRequiresAuth`. Please add the equivalent for both mounts: build a router with a stub `WorkspaceAdminGate` whose `Require` records that it ran and refuses, then assert the admin feature gate and admin marketplace paths do not reach the handler. Verify the new test goes red with the gate deleted from the mount before you push it.

apps/control-plane/internal/platform/http/router.go
```diff
@@ -319,14 +329,14 @@ func NewRouter(cfg RouterConfig) http.Handler {
...
+	// mirroring /api/v1/admin/feature-gates, while catalog curation inside the
+	// handler stays platform-admin only.
 	if cfg.MarketplaceHandler != nil {
 		mux.Handle("/internal/marketplace/", internal(cfg.MarketplaceHandler.InternalMux()))
-		if cfg.RoleSvc != nil && cfg.AuthMiddleware != nil {
+		if cfg.WorkspaceAdminGate != nil && cfg.AuthMiddleware != nil {
 			adminMarketplace := cfg.AuthMiddleware.Require(
-				cfg.RoleSvc.RequirePlatformAdmin(cfg.MarketplaceHandler.AdminMux()),
+				cfg.WorkspaceAdminGate.Require(cfg.MarketplaceHandler.AdminMux()),
 			)
```
sakibsadmanshajib, line 340: Same gap as the feature gate mount above, verified the same way. Removing `cfg.WorkspaceAdminGate.Require(...)` here leaves the whole control plane suite green.

One thing that is right and worth keeping: when `WorkspaceAdminGate` is nil these routes are skipped rather than falling back to `RoleSvc.RequirePlatformAdmin`, and `main.go` declares the field as the interface rather than the concrete pointer so a typed nil cannot be mistaken for a live gate. That reasoning is sound and the comment explaining it should stay.

apps/control-plane/internal/marketplace/http.go
```diff
@@ -118,13 +140,19 @@ func (h *Handler) serveAdmin(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (h *Handler) handleList(w http.ResponseWriter, r *http.Request, tenantID uuid.UUID, canCurate bool) {
 	entries, enabled, err := h.svc.Browse(r.Context(), tenantID)
 	if err != nil {
 		writeJSON(w, http.StatusInternalServerError, errBody("failed to load marketplace catalog"))
 		return
 	}
-	writeJSON(w, http.StatusOK, listResponse{Entries: entryWires(entries, enabled)})
+	writeJSON(w, http.StatusOK, listResponse{Entries: entryWires(entries, enabled), CanCurate: canCurate})
```
sakibsadmanshajib: Information disclosure that the classification table in the PR body does not cover. This response carries `entryWire.Config`, the raw `config` JSONB of every row of the global `public.marketplace_entries` table, and this route is now reachable by any workspace OWNER rather than by a platform admin.

For `kind = mcp_server` that config is the OpenHands native shape, and `apps/agent-engine/internal/marketplaceclient/mcpconfig.go:27` decodes an `env map[string]string` out of it. An MCP server curated with a service token in `env`, which is the normal way those servers are configured, becomes readable by every workspace owner on the deployment the moment this merges. The migration that created the table is explicit about the posture being changed here: `20260716_01_marketplace_catalog.sql:45` states the table is "read and written only through apps/control-plane/internal/marketplace, gated at the HTTP layer by platform-admin".

A workspace that enables an entry does receive that config in its agent sandbox already, so this is not a new capability so much as a much shorter path to it, available without enabling anything and covering entries the tenant never enabled. Two ways out, either is fine:

1. Omit `config` from the list response when `canCurate` is false. The console does not render it for a non curator, so nothing is lost.
2. Keep it, and make "never curate a secret into `config`" an enforced rule rather than a convention, plus update the table comment in a follow up migration.

Option 1 is the smaller diff and does not depend on operator discipline. Whichever you pick, please also correct the `20260716_01` table comment so the next reader is not told the wrong gate.

scripts/seed-demo-owner.py
```diff
@@ -419,11 +419,15 @@ def main() -> None:
         print(f"error: tenant membership upsert failed: {status} {body}", file=sys.stderr)
         sys.exit(1)
 
-    # 4. Guard + upsert the web-console billing account. is_platform_admin=
-    # true here is what unlocks control-plane's RequirePlatformAdmin-gated
-    # admin panels (feature gates, provider catalog, marketplace, credit
-    # grants) -- tenant OWNER above does not imply this; they are unrelated
-    # schemas. Same slug-collision risk as the tenant guard above: refuse to
+    # 4. Guard + upsert the web-console billing account. This account is
+    # deliberately NOT flagged is_platform_admin: after issue #758 the
```
sakibsadmanshajib, line 431: The change itself is correct and I verified the surrounding behaviour: this is the only place in the repository that ever sets `is_platform_admin` to anything, the column defaults to false in `20260428_01_budgets_alerts_invoices_grants.sql:21`, and no signup, invite, tenant provisioning or reconcile path writes it. `signup/personal_tenant.go` and `signup/reconcile.go` touch `tenant_users` only. Because the upsert runs with `resolution=merge-duplicates`, re-running the seeder now actively clears the flag on an existing demo account rather than merely not setting it, which is a stronger guarantee than the comment claims.

The module docstring above was not updated and now contradicts this. Lines 5 to 16 still describe the account as "an owner + platform-admin of a web-console personal account (unlocks the owner-only billing pages AND the platform-admin-only panels -- feature gates, provider catalog, marketplace, credit grants ... RoleService.RequirePlatformAdmin)", and line 15 still lists `accounts.is_platform_admin = true` as one of the two role systems this script writes. That docstring is the instruction a future operator reads before deciding whether the seeder looks broken. Please bring it in line with the code.

apps/web-console/tests/e2e/console-workspace-admin.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+  test("feature gates renders toggles, not an access wall", async ({
+    page,
+  }) => {
+    await page.goto("/console/feature-gates");
+    await expect(
+      page.getByRole("heading", { name: "Feature gates" })
+    ).toBeVisible();
+
+    await expect(page.getByText("Admin access required")).toHaveCount(0);
+    await expect(page.getByText("Managed by your administrator").first()).toBeVisible();
```
sakibsadmanshajib: This assertion cannot distinguish the two states it is meant to separate. `Managed by your administrator` is the title of the 403 `EmptyState` in `app/console/feature-gates/page.tsx` and it is also the read only row label in `components/feature-gates/feature-gate-manager.tsx`. A caller who is refused outright renders the wall, which contains this exact string, and the `Admin access required` count on line 54 is zero for them too, so both of those assertions pass on a fully denied page.

Only the `getByRole("switch")` assertion on line 58 actually separates the cases, and it does the job, so this is a clarity problem rather than a hole. Please either scope this to a row, for example by asserting the count of read only labels equals the number of platform managed gates, or drop it and let the switch count carry the test.

Separately, this spec is skipped whenever the credentials are absent, and the CI job now feeds it from `secrets.E2E_PLATFORM_ADMIN_*`. Worth confirming that account is still ACTIVE OWNER of its tenant now that platform admin is no longer what the spec needs, otherwise the spec skips or fails for a reason unrelated to the code.

sakibsadmanshajib (commented):
**Verdict: CHANGES REQUESTED.** Posted as a COMMENT review because GitHub refuses REQUEST_CHANGES from the PR author account, which is the same account this reviewer runs as.

## Independent security review, issue #758 re-gating

**No privilege escalation found.** The direction of this change is correct and the cross workspace property holds against everything I could throw at it. I am requesting changes on test integrity and on one information disclosure that the classification table does not account for.

### Privilege direction, per route

| Route | Before | After |
| --- | --- | --- |
| `GET /api/v1/admin/feature-gates` | platform admin | platform admin, plus ACTIVE OWNER of the tenant on the session |
| `PUT /api/v1/admin/feature-gates/{key}`, categories agents, rag, sso, audit_sink, feature | platform admin | platform admin, plus ACTIVE OWNER of the tenant on the session |
| `PUT /api/v1/admin/feature-gates/{key}`, categories billing, admin | platform admin | platform admin only, refused for the owner at `admin.go:181` before the write |
| `GET /api/v1/admin/marketplace` | platform admin | platform admin, plus ACTIVE OWNER of the tenant on the session |
| `PUT /api/v1/admin/marketplace/{id}/enable` | platform admin | platform admin, plus ACTIVE OWNER of the tenant on the session |
| `POST`, `PUT`, `DELETE` on `/api/v1/admin/marketplace` | platform admin | unchanged, refused at `http.go:96`, `:118`, `:124` |
| `/api/v1/admin/providers*` | platform admin | unchanged, `router.go:292` |
| `POST /v1/admin/credit-grants` | `authz.PermPlatformAdmin` | unchanged, `main.go:1224`, and `policy.go:59` still grants that permission on the `IsAdmin` overlay alone |

No principal reaches a widened route except an ACTIVE OWNER of the tenant in scope, with one qualification in finding 3 below.

### My own attempts to defeat cross workspace isolation

I wrote four adversarial tests against the real `WorkspaceAdminGate` (not committed, run locally). All four were refused with 403:

1. OWNER of tenant A who also holds an ACTIVE non owner membership on tenant B, selecting B. This is the sharpest shape available, because `auth.Client.WithMembershipCheck` accepts B, so only the OWNER predicate stands in the way. Denied.
2. Tenant ADMIN of the tenant in scope. Denied.
3. Role string `owner` in lowercase, in case the predicate ever went case insensitive. Denied.
4. Forged `X-Platform-Admin` and `X-Hive-Platform-Admin` request headers. The overlay is a private context key with an unexported struct type, unreachable from a request, and `PlatformAdminFromContext` returns false when the stamp is absent, so a handler mounted without the gate fails closed. Denied, and the overlay stayed false.

Static analysis agrees. The only tenant input on either surface is `auth.Viewer.TenantID`, which `LookupUser` collapses to `uuid.Nil` unless `tenantMembershipCheck` (`main.go:1379`) confirms an ACTIVE membership on a non archived tenant. `featuregate.requireTenant` and `marketplace.serveAdmin` read nothing else: no tenant in a body, a query string, a header, or a path parameter. `settings.Resolver.Set` takes the tenant as a bound parameter, and `marketplace.SetEnabled` additionally runs under `withTenantTx` with `app.current_tenant_id` set LOCAL.

### Issue #750

Still present and unchanged. `apps/control-plane/internal/platform/role_pgx.go:122-138` still grants the platform admin overlay on ANY owner membership of ANY account flagged `is_platform_admin`. This PR does not touch that file, does not reintroduce the shape, and does not widen it: the two routes it moves already admitted exactly that predicate through `RequirePlatformAdmin`, so the set of platform admins reaching them is identical before and after. The new gate deliberately scopes on `tenant_users` instead, which is the right call.

### Test integrity, what I proved can fail

I reverified every mutation the PR body claims, from a clean checkout of the pushed head, and every one of them goes red:

| Mutation I applied | Result |
| --- | --- |
| `isOwner, err := false, error(nil)` in the gate | `TestWorkspaceAdminGate_TenantOwnerPasses` 403 want 200, and `..._LookupErrorGets500` also red |
| `isOwner, err := true, error(nil)` in the gate | `..._OwnerOfAnotherWorkspaceDenied` and `..._MemberDenied` both 200 want 403 |
| feature gate carve out condition forced to `if false` | `TestAdmin_WorkspaceOwnerCannotToggleBillingGate` 200 want 403 |
| `manageableBy(g.Category, true)` in `handleCollection` | `TestAdmin_ListMarksPlatformManagedGatesUnmanageable` red on the billing gate |
| `canCurate := true` | `TestAdmin_WorkspaceOwnerListsCatalog` plus all three verbs of `TestAdmin_WorkspaceOwnerCannotCurateCatalog` red |
| `newEntryWire(e, true)` for every entry | `TestAdmin_WorkspaceOwnerEnableIsScopedToOwnTenant` and the pre-existing `TestAdmin_SetEnabled_RoundTrip` both red |

The suite is green unmutated, and my worktree was returned to the pushed head byte for byte before that run. The two mutations the PR body does not claim are the finding, see below.

### Verdict on exposure

Against `main` as it stands today, this PR makes the system **less exposed on the write paths and slightly more exposed on one read path**. Less, because it removes the incentive to hand `is_platform_admin` back to a demo account in order to make the console usable, which is what the 2026-08-06 strip was fighting, and because the billing and admin carve outs are enforced server side rather than in the UI. More, only in that a workspace owner can now read the full `config` of every global catalog entry, finding 2.

apps/control-plane/internal/platform/http/router.go
sakibsadmanshajib: Fixed in `8af9d83`. `apps/control-plane/internal/platform/http/router_workspace_admin_test.go` now dispatches through `NewRouter` rather than testing the gate in isolation, using the pattern `TestNewRouterAdminProvidersRequiresAuth` established. The stub gate records that it ran and refuses without calling next, so a response that comes from the handler is itself the proof that the mount lost the gate.

`TestNewRouterGatesWorkspaceAdminSurfaces` covers all four paths behind the gate: the feature-gate collection and the single-gate PUT, plus the marketplace collection and the enable PUT. `TestNewRouterSkipsWorkspaceAdminSurfacesWithoutGate` pins the second property you called out, that a nil gate leaves the routes unmounted rather than falling back to `RoleSvc.RequirePlatformAdmin`.

Verified by mutation, not by reading. Deleting the `Require` call from this mount and running the full control-plane short suite:

```
--- FAIL: TestNewRouterGatesWorkspaceAdminSurfaces/feature_gate_list
    router_workspace_admin_test.go:118: GET /api/v1/admin/feature-gates: the workspace admin gate ran 0 times, want 1; the mount does not carry it
--- FAIL: TestNewRouterGatesWorkspaceAdminSurfaces/feature_gate_toggle
    router_workspace_admin_test.go:118: PUT /api/v1/admin/feature-gates/ENABLE_RAG: the workspace admin gate ran 0 times, want 1; the mount does not carry it
```

Restored from a byte-for-byte copy afterwards, and the full suite is green.

apps/control-plane/internal/platform/http/router.go
sakibsadmanshajib: Fixed in `8af9d83`, same test file. `TestNewRouterGatesWorkspaceAdminSurfaces` covers this mount through `GET /api/v1/admin/marketplace` and `PUT /api/v1/admin/marketplace/{id}/enable`. `RouterConfig` takes the concrete `*marketplace.Handler` here, so the test mounts a real handler over a stub `marketplace.Repository` that answers an empty catalog: a request that slips past the gate then returns a 200 JSON body instead of panicking on a nil service, which keeps the failure legible.

Mutation proof, deleting the `Require` call from this mount and running the full control-plane short suite:

```
--- FAIL: TestNewRouterGatesWorkspaceAdminSurfaces/marketplace_list
    router_workspace_admin_test.go:118: GET /api/v1/admin/marketplace: the workspace admin gate ran 0 times, want 1; the mount does not carry it
--- FAIL: TestNewRouterGatesWorkspaceAdminSurfaces/marketplace_enable
    router_workspace_admin_test.go:118: PUT /api/v1/admin/marketplace/3bf18e5c-2ddd-4114-9205-61fb1184acc0/enable: the workspace admin gate ran 0 times, want 1; the mount does not carry it
```

Restored afterwards and green. The nil-gate behaviour you singled out as worth keeping is now asserted rather than only commented: `TestNewRouterSkipsWorkspaceAdminSurfacesWithoutGate` requires a 404 on all four paths when no gate is configured, so a future edit that adds a fallback to the wider gate fails the suite. The comment stays as written.

apps/control-plane/internal/marketplace/http.go
sakibsadmanshajib: Took option 1 in `8af9d83`. `entryWires` now takes the caller curation authority and drops `Config` when it is false, and the wire tag is `config,omitempty`, so the key is absent rather than null for a workspace owner and unchanged for a platform admin. The internal `mcp-servers` surface the agent sandbox reads is untouched.

`TestAdmin_ListWithholdsEntryConfigFromWorkspaceOwner` seeds an entry whose config carries an `env` credential (a placeholder value), then asserts on raw JSON fields: `config` absent for the owner, `name` still present so the console can render and enable the entry, and the credential still reachable by the platform admin. Proven red first, against the code as you reviewed it:

```
--- FAIL: TestAdmin_ListWithholdsEntryConfigFromWorkspaceOwner
    http_workspace_owner_test.go:178: the list response hands a workspace owner the raw catalog config: {"command":"npx","env":{"GITHUB_TOKEN":"PLACEHOLDER_TOKEN"}}
```

One thing your option 1 could not have known, and it is the reason this is not a one-line change. The console does not merely ignore `config` for a non curator, it requires it: `decodeMarketplaceEntry` in `lib/control-plane/client.ts` ran `readObjectField(value, "config")` and returned null for the row, and `decodeMarketplaceEntries` fails the entire list on one null row, so `getMarketplaceEntries` threw and the page fell to "Could not load the marketplace catalog". Omitting the field alone would have swapped this disclosure for exactly the wall issue #758 removed. The decoder now reads an absent `config` as withheld and a present one as still having to be an object. `__tests__/marketplace-config-withheld.test.ts` covers all three cases and its first case was red before the decoder change with `Failed to parse marketplace catalog response`.

The table comment is corrected in `supabase/migrations/20260808_01_marketplace_entries_comment.sql`, as a new comment-only migration rather than an edit to the applied `20260716_01`. It now says curation is platform-admin gated, the read and the per-tenant enablement are gated on `WorkspaceAdminGate` which also admits the tenant OWNER, and that the raw config is served only to a caller who may curate.

scripts/seed-demo-owner.py
sakibsadmanshajib: Corrected in `8af9d83`. The docstring no longer describes the account as a platform admin and no longer lists `accounts.is_platform_admin = true` as one of the two role systems this script writes. It now says the second system is `account_memberships.role = 'owner'` with `is_platform_admin` written false, that the workspace-scoped panels are reached through the tenant OWNER role by `WorkspaceAdminGate` since issue #758, and, as you pointed out, that the merge-duplicates upsert makes a re-run clear the flag on an existing demo account rather than merely decline to set it. The platform-wide powers behind `RequirePlatformAdmin` are named as the operator concern a demo account must not hold.

apps/web-console/tests/e2e/console-workspace-admin.spec.ts
sakibsadmanshajib: Fixed in `8af9d83`. The read-only label assertion is now scoped to a gate row:

```ts
await expect(
  page.locator("li", { hasText: "Managed by your administrator" }).first()
).toBeVisible();
```

The 403 `EmptyState` renders a `div` with two `p` elements and no list item, so a denied page fails this where it passed the bare string.

I took the same treatment to the marketplace test in this spec, which had the identical defect you describe and no switch assertion to carry it: heading plus a zero count of "Admin access required" both hold on a fully denied page. It now also asserts a zero count of "Ask your workspace owner or administrator if you need a connector enabled.", which only that page 403 wall can render.

On the credentials note, agreed and I cannot settle it from the branch. The spec prefers `E2E_WORKSPACE_OWNER_*` and falls back to `E2E_PLATFORM_ADMIN_*`, and what it needs now is an ACTIVE OWNER row on the tenant that account selects, not the platform-admin flag. That is a fact about the deployment secret, so it needs checking against the database before this spec is trusted in CI. Flagged in the PR body rather than silently assumed.

PR #799, opened 2026-08-08
test: live interaction coverage for the agent workspace surface

sakibsadmanshajib:
## Summary

Live interaction coverage for the agent workspace ("Cowork", `/agent-workspace`): a trackable proven-over-total number, every unproven control named, and a check behind it. Reworked from the first version of this pull request, whose central premise was self-inflicted and had since gone stale.

## What the rework changed

**Authentication no longer skips fourteen controls.** The first version proved 8 of 22 and skipped the rest, on the stated grounds that signing in required rotating the shared demo account's password through Supabase's admin user endpoint, which was returning 500s. Rotation is forbidden, but it was never the only way in: `apps/web-console/tests/e2e/support/live-auth.ts` mints a session through the admin one-time-token flow, needs no password, and changes none. Every authenticated control is now proven from a minted session. The only remaining credential-gated control is C8, the password submit path itself, which cannot be proven without a real password and must never be proven by inventing one.

**The lifecycle assertions match the deployment.** Commit c1ca04d5 put a real agent runtime behind Cowork through an unprivileged host launcher. A task now progresses queued, running, done rather than resolving straight to a terminal blocked state, so the spec no longer asserts "Blocked" within 20 seconds or an engine-not-configured notice. C16 asserts that notice is absent while a runtime is configured, and C17 asserts the Cancel button is offered on a live task, cancels it, and that a second cancel on the now-terminal task is refused.

**Assertions that could not fail are gone.**

- `expect(status).toBeGreaterThanOrEqual(400)` on the cancel guard passed on a 404 from an unmounted route, on a 401 from a missing auth header, and on an undefined task id. It is now exactly `409`, the documented contract, with the auth header and the task id asserted separately as preconditions. The bearer token is asserted as a boolean rather than with `toMatch`, so a failure prints `false` instead of echoing a live token into the report, the trace and the CI log.
- The two sign-in status assertions are exactly `400` rather than a range, which also passed on a 500 from a broken auth service.
- C18 skipped itself whenever the account already had task history, which every earlier control in the file guarantees by creating some, so it switched itself off permanently after its first run. It now stubs an empty list, which is what the `tasks.length === 0` rendering rule actually needs.
- The coverage builder read `results[0]`, the first attempt, while the config sets `retries: 2`. It now reads Playwright's own per-test verdict across all attempts, and treats a retry pass as unproven rather than proven. It exits non-zero when a control has no test at all, which is the one hole a ratio cannot show.

**The denominator is checked against the DOM.** C23 enumerates every focusable control the deployed sign-in and tasks screens render and requires the set to equal the ledger's `dom` entries. A control that ships without a ledger entry now fails the run instead of quietly raising the percentage by not being counted. The ledger's one "not present" claim, the absent transcript pane, is an assertion in the same test rather than prose. The ledger is still edited by hand and now says so; C23 is what checks that hand.

**It runs, and the number is generated.** `tests/e2e/_probe` is `testIgnore`d by the `chromium` project and no workflow ran `--project=probe`, so none of this executed. The spec has its own Playwright project, pinned in `playwright-spec-manifest.json` and enforced by the existing spec collection guard, and `deploy-demo-box` runs it against the deployment it just shipped. That is the only place it can run honestly: the Cowork console, the edge-api agent-task routes and the host launcher exist together only on the box, and ci.yml's stack boots none of the three. The committed `agent-workspace-coverage.json` is deleted; the ledger is generated by the run into gitignored `test-results/`, uploaded as an artifact, and printed into the job summary.

## Live result

Run against `https://chat-hive.scubed.co` on 2026-08-11 with a minted session.

```
15 executed, 1 skipped, 14 passed, 1 failed
agent-workspace coverage: 18/23
```

| controls | state |
| --- | --- |
| 18 | proven live |
| 1 (C8) | skipped, needs a real password, honest remainder |
| 4 (C14, C15, C16, C17) | share one test, red on two live defects below |

The first version of this pull request reported 8 of 22 with 14 blocked on a reason that was not true.

## Two live defects the rework found

Neither is fixed here. Both were invisible while the suite asserted the old blocked-state shape.

**#881: Cowork create reports a failure for a task that is actually running.** `apps/edge-api/internal/agenttask/client.go` gives its control-plane client a 15 second timeout, while `CreateTask` in control-plane blocks inline on the sandbox launch with a five minute bound. Any launch slower than 15 seconds answers the browser `500` for a task that is created and runs to completion. Measured at 18.0 seconds against the box, with the task reaching `succeeded` while the composer said "Could not start the task".

**#886: cancelling a task does not stop the sandbox or release its concurrency slot.** `Service.Cancel` is a database transition and the `Engine` interface has only `Launch`, so nothing tells the sandbox to stop. Two cancels held both of `HIVE_QUOTA_USER_CONCURRENCY`'s slots for over half an hour, after which every create was refused instantly with "agent engine could not start the task", while uncancelled tasks on the same account finished in about eighty seconds. A user can permanently exhaust their own concurrency by cancelling. Both the spec comment and the assertion message name this, so a repeat run is not misread as flake.

Which of the two takes the run red depends on the box: a slow launch gives #881, an exhausted slot gives #886. Both were reproduced directly against the API, not only through the browser.

The create assertion is the only soft assertion in the file. It still fails the test, so nothing is hidden, but the body continues and the task it started is cancelled rather than left holding a sandbox on a shared box.

## Where and when this actually runs, stated plainly

**It does not run on the push-to-main path.** The job is gated to `workflow_dispatch` and is deliberately outside `report-failure`'s `needs`. Both are temporary and both are for the same reason: the suite creates one real agent task, and per #886 cancelling it never frees the sandbox or its concurrency slot, so running on every push touching `apps/web-console/**` would exhaust the demo account and take the Cowork surface down for the day. A permanently red job would also hold the `report-failure` dedupe issue open and downgrade a genuine `migrate` failure to a comment on a stale issue, which is #553's blind spot reopened.

So what the gate is worth today: it is a suite anyone can run on demand against the deployed box, with a floor that fails, a denominator that fails, and a proven live red. It is not yet an automatic check on main. Removing the `if:` line and adding the job back to `report-failure`'s `needs` is a one-line change each, and both should happen the day #881 and #886 land. Until then this is a manual instrument, and calling it a check would be the overstatement this pull request has already had to correct once.

## Proof the gate can fail

Three mutations, all red, plus a live red. Full output in the evidence comment on this pull request.

**Correction, and it matters more than the result.** An earlier version of this section, and the commit message that carried it, claimed the collapsed-run scenario exited 0 before the head commit and 1 after, crediting that commit with the floor. That was inaccurate and an adversarial review caught it. The builder script is byte-identical between the head commit and the one before it, and the floor landed in the earlier one. Measured, each version against one synthetic collapsed-run report using its own ledger:

| commit | exit | ratio | floor |
| --- | --- | --- | --- |
| `86ff8326` this pull request as first filed | 0 | 7/22 | none |
| `6d4a0624` first rework commit | 0 | 7/23 | none |
| `061bbfad` | 1 | 8/24 | 18 |
| `297ad60e` | 1 | 8/24 | 19 |

What survives the correction: the collapsed run does go from exit 0 to exit 1 **across this pull request**, which is the comparison a reviewer deciding whether to merge it cares about. What does not: the attribution to a specific commit. The synthetic report was also made faithful in the process, because the first version set only `test.status` and left the attempt's own status untouched, which let the pre-rework builder still read `passed` and report an inflated ratio.

The live red is separate and is not a mutation at all: four controls fail because the product is broken, with #881 and #886 open for them.

## Verification

- [x] `npx tsc --noEmit` in `apps/web-console`
- [x] `npm run e2e:verify-collection`, 36 files collected into the projects the manifest pins them to
- [x] `npm run e2e:agent-workspace` against the demo box, three passes, no retry passes, flake rate 0%
- [x] `node scripts/build-agent-workspace-coverage.mjs` regenerates the ledger from the run
- [x] The task each run creates is cancelled or already terminal; nothing is left in flight on the box
- [ ] C8 is proven the moment `HIVE_QA_AGENT_PASSWORD` is provided for an account that is OWNER on a cowork-enabled tenant

## Buglog entry

To be appended to `.wolf/buglog.jsonl` on main in a separate buglog-only pull request once this merges, per the protocol in `.claude/rules/openwolf.md`.

```json
{"date":"2026-08-11","title":"A live coverage gate skipped most of its controls on a blocker that was never true","error_message":"8/22 proven, 14 skipped with reason: Supabase admin user-listing 500s block credential rotation for the shared demo account","root_cause":"The suite assumed the only way to authenticate was rotating a shared account's password through POST /auth/v1/admin/users, so when the admin listing endpoint returned 500s it concluded authentication was impossible and froze fourteen controls into a permanent skip. The sanctioned admin one-time-token flow in tests/e2e/support/live-auth.mjs needs no password, touches no credential, and never calls the failing endpoint. The false reason was then committed into a static coverage file that nothing regenerated, so it outlived the outage it described.","fix":"Authenticate through live-auth.ts, which mints a session with generate_link plus verify. Delete the committed coverage snapshot, gitignore the generated one, and give the spec a Playwright project a workflow selects by name so the number comes from a run. Only the password submit path remains credential-gated, because it is the one control a minted session cannot prove.","tags":["e2e","playwright","auth","coverage","live-testing","agent-workspace"]}
```


apps/web-console/tests/e2e/_probe/agent-workspace-controls.json
```diff
@@ -0,0 +1,184 @@
...
+  "consumed_by": "apps/web-console/scripts/build-agent-workspace-coverage.mjs",
+  "minimum_proven": 19,
+  "minimum_proven_note": "The coverage floor. The build script exits non-zero when fewer controls are proven than this, so coverage cannot quietly collapse: without it, one renamed secret skips the whole authenticated group, every skipped control still counts as accounted for, and the job reports a cheerful green. Demonstrated: a run report with the authenticated group skipped scores 8 of 24 and the builder exits 1 on this floor. RATCHET UP ONLY, and never by the run doing the checking: nothing in this repository writes this number, it moves only in a deliberate commit a human or an agent has to justify in review. Lowering it to make a run pass is the exact failure it exists to catch, because a genuine drop in coverage is the signal rather than the noise. Set to 19, the count a real run against the demo box proved on 2026-08-14. It rises to 23 the day issues #881 and #886 are fixed, because the four controls in the create test are red on those two defects and nothing else, and to 24 when HIVE_QA_AGENT_PASSWORD is provisioned for C8.",
```
sakibsadmanshajib, line 5: Verified against the diff. `build-agent-workspace-coverage.mjs` is byte identical between commit 061bbfad (the previous commit in this PR, `minimum_proven: 18`) and this PR's head 297ad60 (`minimum_proven: 19`): `git diff 061bbfad:apps/web-console/scripts/build-agent-workspace-coverage.mjs 297ad60:apps/web-console/scripts/build-agent-workspace-coverage.mjs` is empty. The floor check, the `notRun` check, and the hard-fail-on-missing-mint-env behavior that make a collapsed run fail all landed in 061bbfad, per that commit's own message ("one renamed secret can no longer drop sixteen controls into a green 6 of 24, and agent-workspace-controls.json carries a minimum_proven floor the build script fails below").

I ran the unchanged script directly against a synthetic Playwright JSON report modelling the scenario described here (unauthenticated-group tests expected, authenticated-group tests unexpected, using the real test titles from this file). Both the 061bbfad script with floor 18 and this commit's script with floor 19 exit 1 on that input. My reconstruction scored 9/24, not the claimed 8/24, but either number is far under both floors, so that discrepancy doesn't change the conclusion.

The commit message for 297ad60 states: "a run report with the authenticated group skipped ... scores 8 of 24 and the builder exits 1 on the floor. Before this change that scenario exited 0." That is not supported by the diff. Since the builder logic here is unchanged from 061bbfad, the same collapsed-run input already exited 1 one commit earlier. The only change in 297ad60 relevant to the floor is the number itself, 18 to 19, which does not affect whether an 8-or-9-of-24 result trips it.

This doesn't mean the floor is broken, it is enforced, and I confirmed that by running it. It means the specific proof attributed to this commit, that the collapsed-run scenario used to pass before it landed, is inaccurate. The gate was already effective at 061bbfad. That matters because the PR frames this commit as the one that "proves the gate can fail" and the collapsed-run mutation as the central argument for the floor.

.github/workflows/deploy-demo-box.yml
```diff
@@ -849,6 +849,147 @@ jobs:
...
+    # demo account and every later create is refused. On every push to main
+    # touching apps/web-console that would take the demo surface down for the
+    # day, which is a real cost to pay for a measurement.
+    #
+    # Not in report-failure's `needs` either, for the same window: that job
+    # dedupes on a single open label, so a job that is red for a known product
+    # reason would hold that issue open and downgrade a genuine `migrate`
+    # failure to a comment on a stale issue, which is #553's blind spot
+    # reopened.
+    if: github.event_name == 'workflow_dispatch'
```
sakibsadmanshajib, line 884: Confirmed: this job's only run condition is `if: github.event_name == 'workflow_dispatch'`, and `report-failure`'s `needs: [migrate, deploy, sdk-replay]` does not include it, exactly as the surrounding comment states. `deploy-demo-box.yml`'s `on:` block triggers on `push: branches: [main]` as well as `workflow_dispatch:` (separate triggers), so on the normal path (a merge to main that touches one of the watched paths, which is what actually deploys this box) this job never runs.

The PR description's framing ("It runs, and the number is generated", "Proof the gate can fail... That is the check working rather than a suite that stays green over a broken surface") describes a manual `workflow_dispatch` run, not the state after an ordinary merge. As wired today, no deploy is gated by this suite: nothing runs, nothing goes red, and the floor enforces nothing unless someone remembers to trigger it by hand. The in-code comment is honest about this being deliberate and temporary pending #881/#886, which is a reasonable tradeoff given #886's concurrency-exhaustion risk on every push, but the PR body doesn't surface the limitation as plainly as the code does. Worth stating explicitly for the merge decision: this ships a measurement tool that currently executes zero times automatically, not a merge gate, and the repo already has one other merged gate (#553's pattern) whose live arm goes unexercised in practice.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,720 @@
...
+     */
+    const listResponse = page.waitForResponse(
+      (res) =>
+        res.url().includes("/v1/agent/tasks") &&
+        res.request().method() === "GET" &&
+        res.status() === 200,
+      { timeout: 60_000 },
+    );
+    await page.reload();
+    const listed: { tasks?: Array<{ id?: unknown; instructions?: unknown }> } = await (
```
sakibsadmanshajib: @/tmp/claude-1001/-home-sakib-hive/47b1dc44-1a51-42c2-8e34-41475aa46c9d/scratchpad/c1.md

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,720 @@
...
+    await signIn(page);
+    await openTaskConsole(page);
+
+    const brief = `interaction-coverage proof ${Date.now()}`;
+    const createRequest = page.waitForRequest(
+      (req) => req.url().includes("/v1/agent/tasks") && req.method() === "POST",
+      { timeout: 30_000 },
+    );
+    const createResponse = page.waitForResponse(
+      (res) => res.url().includes("/v1/agent/tasks") && res.request().method() === "POST",
```
sakibsadmanshajib: @/tmp/claude-1001/-home-sakib-hive/47b1dc44-1a51-42c2-8e34-41475aa46c9d/scratchpad/c2.md

apps/web-console/scripts/build-agent-workspace-coverage.mjs
```diff
@@ -0,0 +1,185 @@
...
+ * Identity, not cardinality, in both directions.
+ *
+ * The loop above walks the ledger, so a [C#] tag the ledger does not declare
+ * was previously ignored outright. That makes the ratio movable by deletion:
+ * remove the C17 entry the day C17 regresses and the ratio RISES, from 18/24
+ * to 18/23, while a real user-facing control silently leaves the suite. Same
+ * argument, and the same fix, as playwright-spec-manifest.json pinning specs
+ * to projects.
+ */
+const undeclared = [...byControlId.keys()].filter(
```
sakibsadmanshajib, line 160: @/tmp/claude-1001/-home-sakib-hive/47b1dc44-1a51-42c2-8e34-41475aa46c9d/scratchpad/c3.md

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,720 @@
...
-test("agent-task API is reachable on the chat origin and still requires auth", async ({
-  request,
-}) => {
-  // The task console calls this path same-origin. It must be routed here (not
-  // 404) and must reject an unauthenticated caller (not 200).
-  const response = await request.get(`${CHAT}/v1/agent/tasks`);
-  expect(response.status()).toBe(401);
+test.describe("authenticated task console", () => {
+  test.beforeEach(requireMintEnv);
+
```
sakibsadmanshajib, line 404: @/tmp/claude-1001/-home-sakib-hive/47b1dc44-1a51-42c2-8e34-41475aa46c9d/scratchpad/c4.md

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,787 @@
...
+     * claim than trusting the body the create echoed.
+     */
+    const listResponse = page.waitForResponse(
+      (res) =>
+        isTasksCollection(res.url()) &&
+        res.request().method() === "GET" &&
+        res.status() === 200,
+      { timeout: 60_000 },
+    );
+    await page.reload();
```
sakibsadmanshajib: HIGH: the try/finally orphan fix is incomplete for the failure mode it names in its own comment.

`await page.reload()` (line 537) and `await listResponse` (opened at line 530, 60 second timeout) still execute before the `try` opens at line 544. If either throws (a slow navigation, or the list GET not resolving inside 60s against the live deployment this suite drives), control never reaches the `try` block, so the `finally` cleanup at line 622 that cancels `createdId` never runs.

The case that costs the most is exactly the one CodeRabbit's finding named: `response.ok()` is true and `createdId` is already populated from the create body (lines 517 to 521) before this call runs. A throw here leaves that real, already-created sandbox uncancelled, holding a `HIVE_QUOTA_USER_CONCURRENCY` slot per issue #886, indistinguishable from the orphan bug this commit's message says it closed.

Moving only the `expect(createdId, ...)` assertion into `try` fixes the case where the list read succeeds but the row is missing from it. It does not fix the case where the list read itself fails to complete. The `page.reload()` and the list `waitForResponse` need to be inside the same `try` (or their own try/finally reaching the same cancel call) for the cleanup guarantee to hold for every path that can throw after a real create.

sakibsadmanshajib: Correct, and fixed in `8ce7d340`. Confirmed the exact lines: `page.reload()`, the list `waitForResponse` and its `.json()` all sat above the `try`, so a throw in any of them skipped `finally` and orphaned the task. Fixing the instance instead of the window is precisely how this came back, so the window is what moved.

**What changed.** The guarded region now opens the moment the create POST has left the browser, which is the moment a task can exist on the box. `authHeader` is read before the `try` because `finally` needs it and reading a header off a delivered request cannot throw. `finally` also re-derives the task id from the API when nothing above got far enough to name it, through a `findTaskIdByBrief` helper that swallows everything: not knowing the id is not a reason to leave a task running, and a cleanup failure must never replace the error that sent us there.

**Proving it found a second defect, same class, that no amount of widening the `try` would have fixed.** Every waiter here is armed before the action it observes, because a waiter armed afterwards races its own event. That leaves the promise floating while the stack is parked on the action, and a waiter that times out during that window is an *unhandled rejection*. Playwright turns that into an immediate test abort, which does not unwind the stack, so `finally` never runs however wide the `try` is. Absorbed at the source in an `armed()` helper, so a waiter rejection is only ever delivered at its `await`, inside the guard.

**Evidence, identical sabotage of the list waiter to a one millisecond timeout, run against the demo box both times:**

| | task | outcome |
| --- | --- | --- |
| before | `interaction-coverage proof 1786726519060` | left `running`, later `succeeded`. Orphaned. |
| after | `interaction-coverage proof 1786726698319` | `cancelled`. Cleanup ran. |

The timeout error is also reported once now rather than twice, which is the unhandled rejection no longer being counted as a separate failure.

Clean baseline after the fix is unchanged at 19 of 24, 15 passed, 1 skipped, 1 failed on #881, flake rate 0%.

Worth recording beyond this thread: the first sabotage did not prove the fix, it disproved it. Had I taken the green `try`/`finally` shape at face value the orphan would have shipped looking fixed, which is the same failure mode as the coverage claim I had to walk back earlier on this pull request. The proof is the part that has to be run.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: This comment does not belong to pull request #799, and I am deliberately not resolving it.

The body posted here is an unexpanded file reference (`@/tmp/.../scratchpad/cN.md`), so the thread carries no reviewable text on GitHub. The referenced file is still on disk and its contents are a finding about a different change entirely. Checked against this branch:

It is about `survivesOrdinaryPullRequest` returning `true` for `if: false` and for negated `contains()`/`startsWith()` event conditions, at a head commit `5206d914`.

None of that exists anywhere in this pull request. This branch touches five files: the agent-workspace probe spec, its control ledger, the coverage build script, `playwright.config.ts` with its manifest, and one added job in `deploy-demo-box.yml`. There is no `survivesOrdinaryPullRequest`, no `legsForLine`, no shapes catalogue document and no `chat-coverage` floors module here, and commit `5206d914` is not in this branch's history.

Leaving it open rather than resolving it, on purpose. Resolving a thread whose finding belongs to another pull request would mark a real defect as handled on a page where nobody will look for it again, which is worse than the merge friction of leaving it. It needs reposting on whichever pull request it was meant for, and the file-reference bodies need to be inlined so the text survives the scratchpad being cleared.

Flagged to the orchestrator rather than actioned here.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: This comment does not belong to pull request #799, and I am deliberately not resolving it.

The body posted here is an unexpanded file reference (`@/tmp/.../scratchpad/cN.md`), so the thread carries no reviewable text on GitHub. The referenced file is still on disk and its contents are a finding about a different change entirely. Checked against this branch:

It is about `legsForLine` misattributing an `elif` branch to the wrong matrix leg in a Go test matrix guard.

None of that exists anywhere in this pull request. This branch touches five files: the agent-workspace probe spec, its control ledger, the coverage build script, `playwright.config.ts` with its manifest, and one added job in `deploy-demo-box.yml`. There is no `survivesOrdinaryPullRequest`, no `legsForLine`, no shapes catalogue document and no `chat-coverage` floors module here, and commit `5206d914` is not in this branch's history.

Leaving it open rather than resolving it, on purpose. Resolving a thread whose finding belongs to another pull request would mark a real defect as handled on a page where nobody will look for it again, which is worse than the merge friction of leaving it. It needs reposting on whichever pull request it was meant for, and the file-reference bodies need to be inlined so the text survives the scratchpad being cleared.

Flagged to the orchestrator rather than actioned here.

apps/web-console/scripts/build-agent-workspace-coverage.mjs
sakibsadmanshajib: This comment does not belong to pull request #799, and I am deliberately not resolving it.

The body posted here is an unexpanded file reference (`@/tmp/.../scratchpad/cN.md`), so the thread carries no reviewable text on GitHub. The referenced file is still on disk and its contents are a finding about a different change entirely. Checked against this branch:

It is about a document whose heading still says "fifteen" after a sixteenth shape was catalogued.

None of that exists anywhere in this pull request. This branch touches five files: the agent-workspace probe spec, its control ledger, the coverage build script, `playwright.config.ts` with its manifest, and one added job in `deploy-demo-box.yml`. There is no `survivesOrdinaryPullRequest`, no `legsForLine`, no shapes catalogue document and no `chat-coverage` floors module here, and commit `5206d914` is not in this branch's history.

Leaving it open rather than resolving it, on purpose. Resolving a thread whose finding belongs to another pull request would mark a real defect as handled on a page where nobody will look for it again, which is worse than the merge friction of leaving it. It needs reposting on whichever pull request it was meant for, and the file-reference bodies need to be inlined so the text survives the scratchpad being cleared.

Flagged to the orchestrator rather than actioned here.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: This comment does not belong to pull request #799, and I am deliberately not resolving it.

The body posted here is an unexpanded file reference (`@/tmp/.../scratchpad/cN.md`), so the thread carries no reviewable text on GitHub. The referenced file is still on disk and its contents are a finding about a different change entirely. Checked against this branch:

It is about a testing-standard document omitting PR #809's chat-coverage floor mechanism (`surface-floors.json`, `floorFailures`, `update-chat-coverage-floors.mjs`).

None of that exists anywhere in this pull request. This branch touches five files: the agent-workspace probe spec, its control ledger, the coverage build script, `playwright.config.ts` with its manifest, and one added job in `deploy-demo-box.yml`. There is no `survivesOrdinaryPullRequest`, no `legsForLine`, no shapes catalogue document and no `chat-coverage` floors module here, and commit `5206d914` is not in this branch's history.

Leaving it open rather than resolving it, on purpose. Resolving a thread whose finding belongs to another pull request would mark a real defect as handled on a page where nobody will look for it again, which is worse than the merge friction of leaving it. It needs reposting on whichever pull request it was meant for, and the file-reference bodies need to be inlined so the text survives the scratchpad being cleared.

Flagged to the orchestrator rather than actioned here.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: Real finding, wrong pull request. It belongs to #822 and is already fixed and verified there by execution. `survivesOrdinaryPullRequest` returned true for `if: false` and for negated `contains`, `startsWith` and `endsWith`, with the deeper root cause that the caller coerced the YAML boolean to an empty string before the function ever saw it.

The blank body here is an unexpanded `@/tmp/...` file reference, the same literal-path bug #822's own builder hit with `gh api`. Resolving on that basis, so this thread points at #822 rather than leaving an empty comment with no trail. Nothing in this pull request is affected: this branch has no such code.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: Real finding, wrong pull request. It belongs to #822 and is already fixed and verified there by execution. `legsForLine` misattributed an `elif` branch to the wrong matrix leg, fixed with a three-leg fixture and a red run confirmed by restoring the old function end to end.

The blank body here is an unexpanded `@/tmp/...` file reference, the same literal-path bug #822's own builder hit with `gh api`. Resolving on that basis, so this thread points at #822 rather than leaving an empty comment with no trail. Nothing in this pull request is affected: this branch has no such code.

apps/web-console/scripts/build-agent-workspace-coverage.mjs
sakibsadmanshajib: Real finding, wrong pull request. It belongs to #822 and is already fixed and verified there by execution. The document heading still said fifteen after a sixteenth shape was catalogued, now corrected at both occurrences.

The blank body here is an unexpanded `@/tmp/...` file reference, the same literal-path bug #822's own builder hit with `gh api`. Resolving on that basis, so this thread points at #822 rather than leaving an empty comment with no trail. Nothing in this pull request is affected: this branch has no such code.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: Real finding, wrong pull request. It belongs to #822 and is already fixed and verified there by execution. The testing standard omitted PR #809's floor and ratchet mechanism, now its own subsection, checked against the real files rather than restated from memory.

The blank body here is an unexpanded `@/tmp/...` file reference, the same literal-path bug #822's own builder hit with `gh api`. Resolving on that basis, so this thread points at #822 rather than leaving an empty comment with no trail. Nothing in this pull request is affected: this branch has no such code.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,865 @@
...
+ * Found by deliberately sabotaging the list waiter to a 1ms timeout to prove
+ * the cleanup ran: it did not, and the task was still `running` afterwards.
+ * Attaching a no-op handler here keeps the rejection on the awaited path, where
+ * it is an ordinary throw inside the guarded region.
+ *
+ * ponytail: absorb at the source rather than a try/catch at each await. Three
+ * call sites today, and the next waiter someone adds gets it for free only if
+ * they use this, which is why the name says what it is for.
+ */
+function armed<T>(waiter: Promise<T>): Promise<T> {
```
sakibsadmanshajib, line 178: armed() closes the gap for the three waiters this delta touches, but the protection is convention only, not structural. Two pre-existing waiters in this same file use the identical arm-before-action shape and are not wrapped: the tokenResponse waiter in the C5 test (line 331) and the one in the C3/C4 test (line 346). Neither currently sits in front of a finally with a resource to leak, so the blast radius today is smaller, but both carry the same unhandled-rejection risk if their timeout fires while the code is parked on page.click. Confirmed against apps/web-console/node_modules/playwright/lib/worker/workerMain.js: Playwright registers its own process.on(unhandledRejection) handler that fails the current test and calls testInfo._interrupt(), which races the interrupt against whatever the test body is currently awaiting rather than throwing into it, so the diagnosis in this comment is accurate, not just plausible. Nothing in the file (lint rule, wrapper-only helper, type) stops a future arm-before-action waiter, wrapped or not, from being added anywhere here. Worth wrapping those two for consistency, or adding a grep-based check that every page.waitForRequest / page.waitForResponse call site in this file goes through armed().

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,865 @@
...
+
+/**
+ * Finds a task by the exact brief it was created with, straight from the API.
+ *
+ * Only used by cleanup, and therefore swallows everything: it runs in a
+ * `finally` whose job is to reclaim a sandbox, so a failure here must never
+ * replace the error that sent us there. Returns an empty string when the task
+ * cannot be found or the call cannot be made.
+ */
+async function findTaskIdByBrief(
```
sakibsadmanshajib, line 191: findTaskIdByBrief matches on the instructions text, which is unique per run only because of the Date.now() suffix in brief. Against a shared live box, two workflow_dispatch runs starting in the same millisecond would produce an identical brief and this lookup could return either task to cancel, or cancel one this run did not create. The job trigger (workflow_dispatch only, per the parent workflow file) makes concurrent runs unlikely, so this is a thin edge case rather than a live bug, but the uniqueness here rests on timing rather than identity. A short random suffix, or the run id, would remove the assumption instead of relying on it holding.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
```diff
@@ -1,89 +1,865 @@
...
+      // Asserted as a boolean rather than with toMatch, so a failure prints
+      // "false" instead of echoing a live bearer token into the report, the
+      // trace and the CI log.
+      expect(
+        typeof authHeader === "string" && authHeader.startsWith("Bearer "),
+        "the create request must carry a bearer token, otherwise the cancel assertions below " +
+          "would be asserting against an unauthenticated 401",
+      ).toBe(true);
+
+        /*
```
sakibsadmanshajib: Formatting: this block comment opens with eight spaces of indent while every continuation line below it uses seven, a leftover from moving the C15 comment into the widened try. Cosmetic only, but worth a pass so the diff reads as intentional rather than pasted.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: Agreed, wrapped both in `e595f2b1`. No disagreement to record: the argument for leaving them was that neither sits in front of a `finally` with a resource to leak, which is an argument about today's blast radius rather than about the risk class, and this branch has already paid twice for fixing an instance and leaving the category open. All five `page.waitForRequest` and `page.waitForResponse` call sites in the file now go through `armed()`.

Thank you for checking the diagnosis against `workerMain.js` rather than taking it. The `testInfo._interrupt()` detail is the part that makes this generalise: an interrupt racing the current await is why widening the `try` alone did not help, and why absorbing at the source does.

On the grep-based check, I did not add one, deliberately. It would be a new lint script wired into CI to police a convention inside a single file, and the honest ceiling is already written at `armed()` itself: it protects the call sites that use it and nothing stops a future waiter being added unwrapped. If a second spec grows waiters of this shape that is the moment the lint rule earns its keep, and it can then cover both. Say the word if you would rather have it now and I will add it.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: Fixed rather than rebutted, in `e595f2b1`: the brief now carries `randomUUID()` alongside the timestamp.

The rebuttal was available and I did not take it. The trigger really does make concurrent runs unlikely today, but the reason to fix it is that this string decides what the cleanup CANCELS. A collision does not merely fail to find a task, it cancels a task another run created, which is a worse outcome than the orphan the cleanup exists to prevent. That is cheap to remove and expensive to reason about, and one appended identifier is less code than the argument for keeping it.

Confirmed live: the task from the verification run carries the new form, `interaction-coverage proof 1786728010033 0d991...`.

apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts
sakibsadmanshajib: Fixed in `e595f2b1`. The opener sat at eight spaces over seven-space continuations, a leftover from moving the C15 comment into the widened `try`, exactly as described. Swept the whole file for the same mismatch rather than just this one, and it was the only occurrence.

Issue #879, opened 2026-08-11
Committed E2E credentials and automated rotation of shared accounts in a public repository


sakibsadmanshajib:
## Summary

This repository is public. Until now it carried three plaintext credentials in a tracked file, for accounts that exist on the live Supabase project, and an automated workflow that rotated the passwords of two more shared accounts on every run.

## The committed credentials

`apps/web-console/tests/e2e/support/e2e-auth-defaults.json` contained, in plaintext on `main`:

| field | account |
| --- | --- |
| `verifiedPassword` | `e2e-verified@scubed.com.bd` |
| `unverifiedPassword` | `e2e-unverified@scubed.com.bd` |
| `invitationToken` | pending invitation for the verified account |

These were not inert sample values. `apps/web-console/tests/e2e/support/e2e-fixture-seed.mjs` writes the resolved credentials back onto the accounts through the Supabase admin API on every run that has no environment overrides, which is the command in CLAUDE.md for a local E2E run. Both accounts hold a tenant OWNER role and a billing account.

The CI Web E2E job also fell through to those values: `E2E_VERIFIED_PASSWORD` and `E2E_UNVERIFIED_PASSWORD` were referenced as secrets in `ci.yml` but no such secrets existed, so the empty string resolved to the committed default.

## The automated rotation

`scripts/seed-owui-e2e-user.py` overwrote the password of two hardcoded shared accounts, `owui-e2e@hive-e2e.invalid` and `owui-e2e-bootstrap@hive-e2e.invalid`, unconditionally on every run. `.github/workflows/owui-nightly.yml` invokes it on a schedule, on `workflow_dispatch`, and on any pull request labelled `run-owui-e2e`. That workflow's `concurrency.group` is keyed on the ref, so a labelled pull request run and the scheduled run never join the same group and can rotate the same accounts mid-flight.

`docs/live-test-auth.md` already forbids this: the control-plane resolves every bearer against GoTrue per request, so rotating a shared account invalidates every session a concurrent run holds. It broke three agents at once on 2026-08-08. `scripts/verify-rag-roundtrip.py` had the same unconditional rotation against `rag-verify-e2e@hive-e2e.invalid`, and its own docstring admitted it was copying the banned pattern.

## Public artifacts holding live session material

Both Playwright report uploads retained traces and videos at GitHub's default 90 day retention with no `retention-days` key, in a public repository:

[...]

PR #880, opened 2026-08-11
fix: stop committing E2E credentials and rotating shared test accounts

sakibsadmanshajib:
Refs #879. **Deliberately not `Closes`**: that issue tracks the rotation of the exposed credentials, which has not happened, and an auto-close on merge would mark an open exposure as resolved.

## This does not un-expose anything

The three values removed here remain in this repository's git history, and this repository is public. They must be treated as compromised regardless of what HEAD looks like. **Rotation is the actual remedy**, it is tracked separately in #879, and it is deliberately not done in this pull request: several agents are running live tests against these accounts right now and a rotation would revoke every one of their sessions. Nothing below closes the exposure. It stops the exposure from being recreated, and stops the code that kept writing those values back onto live accounts.

## What was exposed

`apps/web-console/tests/e2e/support/e2e-auth-defaults.json` carried `verifiedPassword`, `unverifiedPassword` and `invitationToken` in plaintext, on `main`, for `e2e-verified@scubed.com.bd` and `e2e-unverified@scubed.com.bd`. Both accounts hold a tenant OWNER role and a billing account on the live Supabase project, and the fixture seeder writes those exact values back onto them on every run with no environment overrides. No value appears anywhere in this pull request.

## Changes

**1. No credential has a committed default.** `e2e-auth-defaults.json` now holds the two addresses and the two length limits only. Both readers, `e2e-auth-creds.ts` and `e2e-auth-fixtures.mjs`, take the three credentials from the environment through a `requiredSecretEnv` helper that throws, names the variable, and points at `docs/live-test-auth.md`. No fallback, no silent skip, since a silent skip is how a committed value survives unnoticed. In the CLI the three are resolved inside `prepareE2EAuthFixtures` rather than at module scope, so the `reset-profile` action, which needs no credential, still runs without them.

**1b. No fixture address is a shared one.** Removing the committed defaults was not sufficient, and on its own it made things worse. The seeder fell back to the two shared live addresses whenever no run key was set, and `ensureUser` sends `password:` on both update paths, so a developer following the README and running the documented command would overwrite a shared tenant-OWNER account's password and revoke every concurrent session. The committed constant had at least made that write idempotent. `runScopedEmail` now throws without `E2E_RUN_KEY` and otherwise derives this run's own address from the shared base, idempotently, at both resolution points, so the account a spec signs in as and the account the seeder writes are the same one and it belongs to this run. `DEFAULT_EMAILS` is deleted. This is also the only route by which the newly minted CI secret values could have reached the two exposed accounts.

`E2E_VERIFIED_PASSWORD` and `E2E_UNVERIFIED_PASSWORD` did not exist as repository secrets, which is why CI was silently using the committed values for 109 days, and for roughly the first 95 of those, before run-scoped addresses landed in #583, every CI run wrote the published password back onto the shared live accounts. Both secrets are now set. Their values were generated randomly, never displayed, and are only ever written to the run-scoped throwaway users that `ci.yml` already creates per job attempt, so no existing account was touched.

**2. No script rotates a shared account any more.** `scripts/seed-owui-e2e-user.py` gets `password_to_set`, the same helper and semantics as the one already in `scripts/seed-demo-owner.py`: an existing account is left alone unless the caller explicitly supplies a password. It prints a `PASSWORD` line only for a password it actually set, and names on stderr why a line is missing. To keep the nightly working under that rule without a hand-managed secret, `OWUI_E2E_RUN_KEY` namespaces both fixture addresses per job attempt, so every run provisions its own users and shares no credential with any other run. Same shape as `E2E_RUN_KEY` in `ci.yml`. The workflow passes `github.run_id`-`github.run_attempt`.

`scripts/verify-rag-roundtrip.py` had the same unconditional rotation against its own hardcoded shared address. It now signs in with `RAG_VERIFY_PASSWORD` and refuses to rotate, failing by name instead. Only a first run, which creates the account, generates a password.

Both guards are pinned by unit tests in `scripts/test_seed_owui_e2e_user.py`, which already runs in `make test-scripts`.

A first version of this claimed a run "shares no credential with any other run". That was false in two places, both now fixed. The run key namespaced the GoTrue users but not the shim billing account, so every run still issued `DELETE /api_keys` against that shared account for every key but its own, and two overlapping runs revoked each other mid-flight (the outage in `.wolf/cerebrum.md`). That delete is now bounded by age, so a key minted minutes ago survives. The account stays shared deliberately: a tenant bills to exactly one account, so a per-run account would need a per-run tenant and leave a permanent row behind every night. Second, run-scoped users leaked two per nightly, one an OWNER of the `owui-e2e` tenant; `sweep_stale_fixture_users` now clears what earlier runs left, matching on both halves of the address, never touching the shared base address or this run's own users, and never failing a run.

`verify-rag-roundtrip.py` generated a password on the create path and never printed it, which would have left every later run demanding a value no operator could obtain, and `DEMO.md`'s RAG proof dead on every environment it had already run against. It prints it once now, on stderr.

**3. Artifacts no longer carry live session material for 90 days.** Five upload steps, not two. `playwright-report-api` and `owui-playwright-report` exclude `*.zip`, `*.webm` and `index.html`, and set `retention-days: 5`. `index.html` had to go for the same reason as the traces: the HTML reporter base64-inlines the whole report payload into it, stdout, stderr and error text included, and a `waitForURL` timeout enumerates every URL it navigated, fragment and all. The three container-log artifacts (`compose-logs`, `compose-logs-web-e2e-api`, `owui-compose-logs`) now pipe through `scripts/redact-log-credentials.py` and retain for five days; Open WebUI's uvicorn access lines print `GET /oauth/oidc/callback?code=...` verbatim. That redactor mirrors `redactSecrets`, understands fragments as well as query strings, and carries a self-check wired into `make test-scripts`. Traces hold every request header and cookie, and for the OWUI suite the OAuth callback URLs with `code=` and `access_token=` in both the query string and the fragment. A trace is a zip of binary resources, so no text linter can scrub it and exclusion is the only reliable answer. Screenshots still upload, and the failing test names and error text are already in the job log. Review caught that the screenshot half of that was false when first written: `playwright.config.ts` never set `screenshot`, so it kept Playwright's default of `off` and the three exclusions left the `playwright-report-api` artifact empty while `if-no-files-found: ignore` kept it quiet. It now sets `screenshot: "only-on-failure"`, which captures the viewport only, so no URL bar, no request headers and no cookies. The `ci.yml` change stays confined to the upload steps, the log dumps, the token generation step and comments, to rebase cleanly against #805, #811 and #822.

**4. The sweeper's three `console.error` calls now go through `redactSecrets`.** They relayed raw client-library messages to stderr, which is exactly where the service-role key ends up embedded, while the module header two hundred lines above declared the opposite contract.

**5. `docs/live-test-auth.md` is now true.** It named only an out-of-repo scratch script while three in-tree paths broke the rule it stated. It now has a "no credential is ever committed" rule with the environment-variable pattern as the sanctioned alternative, a table of what each of the three scripts does after this change, a note on why the automatically-invoked one matters most, and the artifact rule. `README.md` no longer advertises fallbacks that do not exist.

## How absence fails now

| path | with the credential unset |
| --- | --- |
| Playwright specs | Throws at import: `E2E_VERIFIED_PASSWORD is required and has no default...` |
| Fixture seeding CLI | Same error before any admin call; `reset-profile` still works |
| `seed-owui-e2e-user.py` | Existing account keeps its password, no `PASSWORD` line, stderr names `OWUI_E2E_RUN_KEY` and `OWUI_E2E_PASSWORD`; the nightly's own guard then fails the job loudly |
| `verify-rag-roundtrip.py` | Exits non-zero naming `RAG_VERIFY_PASSWORD` |

## Verification

The first push of this branch left a required check RED while this section
claimed otherwise. That is corrected here, and the check now passes.

* `node scripts/verify-spec-collection.mjs` in the web-console container: OK, 36 files across 2 configs. It failed ten times over at head `dec64e84` because it runs `playwright test --list` in a job holding no E2E secrets.
* `docker compose run --build web-console npm run test:unit`: 488 passed, 1 skipped, and one suite fails: `tests/unit/control-plane-host.test.ts` on `ENOENT /app/.env.example`. That one is pre-existing and unrelated. The web-console image contains only `apps`, so that suite cannot read a repository-root file in this container at all, on this branch or on main.
* New guards in `tests/unit/e2e-auth-creds.test.ts`, 4 passed: the defaults file must hold only the four non-secret fields, every string in it must look like an address rather than a secret, a missing credential must throw by name, and `runScopedEmail` must refuse an empty run key while namespacing idempotently.
* `python3 scripts/test_seed_owui_e2e_user.py`, `python3 scripts/test_seed_demo_owner.py`, `python3 scripts/redact-log-credentials.py --selfcheck`: all pass, including the five new cases.
* Both workflow files parse, and the rewritten nightly seed step passes `bash -n` plus a simulation proving the guard is reached rather than pre-empted by `set -e`.

* `Web E2E (full stack)` now passes at this head: 33 collected, 27 executed, 0 failed, 0 passed only on retry. That is the first end-to-end exercise of the new shape, and it covers the required secrets, the run-scoped addresses and the randomly generated invitation token against the live Supabase project.

Its first attempt failed, and not on anything in this change: the stack never
started, because control-plane could not reach the database
(`EMAXCONNSESSION ... max clients are limited to pool_size: 15`), the known
shared session-pool ceiling. A re-run of the same commit passed. All 14 checks
are green.

## One deliberate deviation from the review

Review item 1 asked for the three credentials to resolve lazily. They still
resolve at module scope, and the required check is green anyway, because
`scripts/verify-spec-collection.mjs` now supplies placeholder values for its
`--list` pass. That guard already does exactly this for the OWUI config, for
the same reason: listing loads spec modules and never dials out, so it needs
the variables present but not real.

The reason for keeping module scope is that lazy resolution weakens the
guarantee. Every gated spec computes `HAS_CREDS` at module scope, so a lazily
resolved empty value turns a missing credential into `test.skip`, which is the
silent skip this whole change exists to remove. Throwing at load keeps the
failure loud and immediate. If the reviewer still prefers lazy resolution, the
change is small, but it needs a run-level assertion to replace what the throw
currently guarantees.

## Stage 6 adversarial review, run on this PR

Streams: CodeRabbit CLI (**SKIPPED**, `Review failed: A valid organization session is required`, HTTP 401 `UNAUTHORIZED`, on both the default and `--light` paths; its absence is not a pass, and the CodeRabbit GitHub App check is not a substitute since it reported `Review rate limited` earlier in this PR's life), `ecc:code-review`, a plain adversarial pass, `typescript-reviewer`, `python-reviewer`, `security-reviewer` (mandatory, credential and auth path) and `/codex:adversarial-review`. Eight inline comments posted, seven fixed in `190d2479`, one rebutted and upheld, all threads resolved.

The claim gating the credential rotation was the primary target, and it now has a runnable proof rather than an argument. `seedFixtures` is driven in a unit test by an admin client that throws on any property access, so an empty, whitespace, `undefined` or non-string run key must reject with the `E2E_RUN_KEY` error rather than the proxy's, which is what proves no admin call happens before the guard. The same stream found the one shared-account write the claim did not cover: `reset-profile` takes an address off argv, and it now refuses anything not scoped to this run. Other fixes: both `runScopedEmail` implementations are pinned as equivalent by test, the log redactor was blind to environment-variable spelling (`\b` treats `_` as a word character, so `RAG_VERIFY_PASSWORD=` did not match), the user sweep's single-page ceiling is now stated, and the invitation token's length floor is documented where it is generated.

## Review round one

An independent review returned do-not-merge on the first push, with nine
findings. All nine are addressed above and in the commits on this branch: the
red required check (item 1), the shared-account write that this change had made
worse (item 2), the auto-closing issue reference (item 3), `index.html` (item
4), the three missed upload steps (item 5), the unobtainable RAG password (item
6), the shim key revocation (item 7), the derivable invitation token (item 8),
and the fixture user leak plus the nightly guard that could never fire (item 9).

The reviewer's sweep of all 1898 tracked files found no committed secret
material other than item 8.

## Buglog entry

To be appended to `.wolf/buglog.jsonl` on `main` in a separate buglog-only pull request once this merges, per `.claude/rules/openwolf.md`:

```json
{"id":"BUG-2026-08-11-committed-e2e-credentials","date":"2026-08-11","title":"Live E2E credentials committed in plaintext in a public repo, and re-seeded onto the accounts on every run","error_message":"No error. The suite passed, which is the problem: e2e-auth-defaults.json carried verifiedPassword, unverifiedPassword and invitationToken for two live tenant-OWNER accounts, and CI referenced E2E_VERIFIED_PASSWORD and E2E_UNVERIFIED_PASSWORD secrets that did not exist, so the empty string fell through to the committed values.","root_cause":"envOrDefault treated an unset credential as a request for the committed fallback. A fallback makes the absence of a secret invisible, so nothing ever failed and the values stayed live for months while the seeder wrote them back onto the accounts through the admin API on every credential-less run. Two sibling scripts copied a related pattern, rotating hardcoded shared accounts unconditionally, one of them from a scheduled workflow whose concurrency group does not join a labelled-PR run to the scheduled run.","fix":"Removed the three fields; both readers now use requiredSecretEnv, which throws and names the variable with no fallback and no skip. Set the two missing CI secrets. seed-owui-e2e-user.py takes password_to_set from seed-demo-owner.py and gains OWUI_E2E_RUN_KEY so the nightly provisions its own users instead of rotating shared ones; verify-rag-roundtrip.py takes RAG_VERIFY_PASSWORD and refuses to rotate. Both Playwright artifact uploads exclude traces and videos and retain for 5 days instead of 90. The fixture sweeper's three console.error calls now go through redactSecrets. docs/live-test-auth.md and README.md corrected.  Review of the first push found the removal alone had made the shared-account hazard worse: the seeder still fell back to the two shared addresses without a run key, and ensureUser writes a password on both update paths, so each operator would now write a DIFFERENT value where the committed constant had at least been idempotent. runScopedEmail closes that by refusing an empty run key and namespacing every fixture address. Same round: index.html base64-inlines the report payload so excluding traces alone did not stop the leak, three container-log artifacts were missed, the shim-key delete still revoked concurrent runs until bounded by age, and the invitation token was a committed literal plus a public run id.","tags":["security","credentials","e2e","ci","public-repo","gotrue","artifacts"],"pr":880,"issue":879}
```

Note: the values remain in git history. This entry is not a record of a closed exposure. Rotation is tracked in #879.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1010,7 +1010,9 @@ jobs:
       - name: Dump compose logs on failure
         if: failure()
         working-directory: deploy/docker
-        run: docker compose logs --no-color --timestamps --tail=500 > /tmp/compose.log || true
+        # Piped through the redactor because this artifact is public: container
+        # logs carry OAuth callbacks, bearer headers and API keys verbatim.
+        run: docker compose logs --no-color --timestamps --tail=500 2>&1 | python3 ../../scripts/redact-log-credentials.py > /tmp/compose.log || true
 
       - name: Upload compose logs
         if: failure()
@@ -1018,6 +1020,7 @@ jobs:
         with:
           name: compose-logs
           path: /tmp/compose.log
+          retention-days: 5
           if-no-files-found: ignore
 
       - name: Tear down stack
@@ -1296,6 +1299,16 @@ jobs:
           done
           echo "::error::Next.js did not start"; cat /tmp/nextjs.log; exit 1
 
+      - name: Generate the fixture invitation token
+        # Random per job, and masked, because the seeder stores its sha256 as a
+        # live pending invitation. A value anybody can derive from the public
+        # run page is a published credential for as long as the run lasts.
+        run: |
+          set -euo pipefail
+          token="$(openssl rand -hex 24)"
+          echo "::add-mask::$token"
+          echo "E2E_INVITATION_TOKEN=$token" >> "$GITHUB_ENV"
+
       - name: Run UI Playwright suite (unauth + credentialed)
         working-directory: apps/web-console
         env:
@@ -1336,7 +1349,13 @@ jobs:
           E2E_VERIFIED_PASSWORD: ${{ secrets.E2E_VERIFIED_PASSWORD }}
           E2E_UNVERIFIED_EMAIL: e2e-unverified+${{ github.run_id }}-${{ github.run_attempt }}@scubed.com.bd
           E2E_UNVERIFIED_PASSWORD: ${{ secrets.E2E_UNVERIFIED_PASSWORD }}
-          E2E_INVITATION_TOKEN: e2e-invitation-token-2026-fixture-${{ github.run_id }}-${{ github.run_attempt }}
+          # Generated in the step above, not derived here. It used to be a
+          # committed literal concatenated with the run id and attempt, both
+          # of which are public on the run page, and e2e-fixture-seed.mjs
+          # stores its sha256 as the pending invitation's token_hash. Anyone
+          # reading the run URL could compute the live token and accept the
+          # invitation into this run's account while it was still open.
+          E2E_INVITATION_TOKEN: ${{ env.E2E_INVITATION_TOKEN }}
           # Workspace-owner fixture for console-workspace-admin.spec.ts.
           # Unset leaves that spec skipped, which is why the required
           # coverage for a mispointed control-plane is the unit-level
@@ -1372,7 +1391,8 @@ jobs:
       - name: Dump compose logs on failure
         if: failure()
         working-directory: deploy/docker
-        run: docker compose logs --no-color --timestamps --tail=500 > /tmp/compose-web-e2e-api.log || true
+        # Piped through the redactor: see the sdk-tests dump step above.
+        run: docker compose logs --no-color --timestamps --tail=500 2>&1 | python3 ../../scripts/redact-log-credentials.py > /tmp/compose-web-e2e-api.log || true
 
       - name: Upload compose logs
         if: failure()
@@ -1381,23 +1401,30 @@ jobs:
           name: compose-logs-web-e2e-api
           path: /tmp/compose-web-e2e-api.log
           if-no-files-found: ignore
+          retention-days: 5
 
       - name: Upload Playwright report
         if: always()
         uses: actions/upload-artifact@v4
         with:
           name: playwright-report-api
-          # Traces (.zip) and videos (.webm) are excluded. This repository is
-          # public, so any artifact here is downloadable by anyone with the run
-          # URL, and playwright.config.ts retains a trace on failure: it holds
-          # every request header (Authorization: Bearer) and every
-          # sb-*-auth-token cookie for the fixture session. The HTML report and
-          # error text still upload, and a text linter cannot inspect a trace
-          # zip, so exclusion is the only reliable scrub.
+          # Traces (.zip), videos (.webm) and index.html are excluded. This
+          # repository is public, so any artifact here is downloadable by
+          # anyone with the run URL, and playwright.config.ts retains a trace
+          # on failure: it holds every request header (Authorization: Bearer)
+          # and every sb-*-auth-token cookie for the fixture session.
+          # index.html goes too, and for the same reason rather than a
+          # different one: the HTML reporter base64-inlines the whole report
+          # payload into that file, including stdout, stderr and error text,
+          # and a waitForURL timeout enumerates the URLs it navigated,
+          # fragment included. Both are binary-wrapped, so no text linter can
+          # inspect either. Screenshots survive, and the failing test names
+          # and error text are already in this job's own log.
           path: |
             apps/web-console/playwright-report/
             !apps/web-console/playwright-report/**/*.zip
             !apps/web-console/playwright-report/**/*.webm
+            !apps/web-console/playwright-report/index.html
           if-no-files-found: ignore
           retention-days: 5
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1323,10 +1323,14 @@ jobs:
           # reassign rows out from under this job's own reads mid-test.
           # github.run_attempt makes a manual re-run of a flake get a fresh
           # identity too, not the one the previous attempt already consumed.
-          # Passwords stay the shared default (from secrets, or the fixture
-          # defaults in e2e-auth-defaults.json when those secrets are
-          # unset): nothing looks a user up by password, so they never need
-          # to be unique per run.
+          # The passwords come from secrets and have no fallback. They used to
+          # fall back to plaintext values committed in
+          # e2e-auth-defaults.json, in a public repository, for live accounts
+          # holding a tenant OWNER role; the seeder then wrote those values
+          # back onto the accounts. The fallback is gone and an unset secret
+          # now fails this job by name. They need not be unique per run:
+          # nothing looks a user up by password, and the addresses above are
+          # already run-scoped.
           E2E_RUN_KEY: ${{ github.run_id }}-${{ github.run_attempt }}
           E2E_VERIFIED_EMAIL: e2e-verified+${{ github.run_id }}-${{ github.run_attempt }}@scubed.com.bd
           E2E_VERIFIED_PASSWORD: ${{ secrets.E2E_VERIFIED_PASSWORD }}
@@ -1383,8 +1387,19 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: playwright-report-api
-          path: apps/web-console/playwright-report/
+          # Traces (.zip) and videos (.webm) are excluded. This repository is
+          # public, so any artifact here is downloadable by anyone with the run
+          # URL, and playwright.config.ts retains a trace on failure: it holds
+          # every request header (Authorization: Bearer) and every
+          # sb-*-auth-token cookie for the fixture session. The HTML report and
+          # error text still upload, and a text linter cannot inspect a trace
+          # zip, so exclusion is the only reliable scrub.
+          path: |
+            apps/web-console/playwright-report/
+            !apps/web-console/playwright-report/**/*.zip
+            !apps/web-console/playwright-report/**/*.webm
           if-no-files-found: ignore
+          retention-days: 5
 
       - name: Tear down stack
         if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1303,6 +1303,12 @@ jobs:
         # Random per job, and masked, because the seeder stores its sha256 as a
         # live pending invitation. A value anybody can derive from the public
         # run page is a published credential for as long as the run lasts.
+        #
+        # 24 bytes gives 48 hex characters. The floor it has to clear is
+        # minTokenLength in tests/e2e/support/e2e-auth-defaults.json, which
+        # requiredSecretEnv enforces at 16, so do not shorten this below that
+        # without moving both: the failure would otherwise surface far from
+        # here, naming a variable this step does not mention.
         run: |
           set -euo pipefail
           token="$(openssl rand -hex 24)"
--- a/apps/web-console/scripts/verify-spec-collection.mjs
+++ b/apps/web-console/scripts/verify-spec-collection.mjs
@@ -53,8 +53,23 @@ const OWUI_LIST_ENV = {
   SUPABASE_OAUTH_CLIENT_SECRET: "unused-collection-guard-placeholder",
 };
 
+// tests/e2e/support/e2e-auth-creds.ts resolves its three credentials at module
+// scope and throws when one is missing, deliberately: they have no committed
+// default any more, and a spec that runs without them would otherwise seed a
+// shared live account or skip in silence. Listing loads those modules, so this
+// guard needs the variables present but does not need them to be real. Same
+// reasoning and same shape as OWUI_LIST_ENV above: listing never dials out.
+const WEB_CONSOLE_LIST_ENV = {
+  E2E_VERIFIED_PASSWORD: "unused-collection-guard-placeholder",
+  E2E_UNVERIFIED_PASSWORD: "unused-collection-guard-placeholder",
+  E2E_INVITATION_TOKEN: "unused-collection-guard-placeholder",
+  // Required too: every fixture address is namespaced with it so a run can
+  // only touch accounts it created (runScopedEmail in e2e-fixture-seed.mjs).
+  E2E_RUN_KEY: "collection-guard",
+};
+
 const CONFIGS = [
-  { label: "playwright.config.ts", args: [], env: {} },
+  { label: "playwright.config.ts", args: [], env: WEB_CONSOLE_LIST_ENV },
   {
     label: "e2e/phase-19/owui/playwright.owui.config.ts",
     args: ["--config=e2e/phase-19/owui/playwright.owui.config.ts"],
--- a/apps/web-console/tests/e2e/support/e2e-auth-creds.ts
+++ b/apps/web-console/tests/e2e/support/e2e-auth-creds.ts
@@ -1,14 +1,22 @@
 // Source of truth for E2E test credentials shared between Playwright specs
 // and the fixture CLI (`e2e-auth-fixtures.mjs`).
 //
-// Defaults + min-length constants live in `e2e-auth-defaults.json` so both
-// the TS module (consumed by specs) and the sibling ESM fixture entrypoint
-// (consumed by `execFileSync` from `beforeEach`) read the same values.
+// Non-secret defaults (addresses and min-length constants) live in
+// `e2e-auth-defaults.json` so both the TS module (consumed by specs) and the
+// sibling ESM fixture entrypoint (consumed by `execFileSync` from
+// `beforeEach`) read the same values.
 //
-// Test-only fixtures target dedicated `e2e-*@hive-ci.test` accounts in the
-// staging Supabase project. Env overrides (`E2E_*`) are honored when they
-// meet minimum length checks; empty / unset env vars silently fall back to
-// defaults, non-empty-but-too-short values emit a warning.
+// NO CREDENTIAL HAS A COMMITTED DEFAULT, and none may ever be added back.
+// This repository is public, and the two passwords plus the invitation token
+// used to sit in that JSON file in plaintext for live accounts that hold a
+// tenant OWNER role, which the seeder then wrote back onto those accounts on
+// every credential-less run. The three values below are therefore read from
+// the environment only, and a missing one throws by name rather than falling
+// back or skipping quietly: a silent fallback is what kept the committed
+// values live for months. See docs/live-test-auth.md.
+//
+// The two addresses stay: an address is not a credential, and specs need a
+// stable identity to seed. Env overrides are honored for them.
 
 import defaults from "./e2e-auth-defaults.json" with { type: "json" };
 
@@ -17,9 +25,6 @@ type AuthDefaults = {
   minTokenLength: number;
   verifiedEmail: string;
   unverifiedEmail: string;
-  verifiedPassword: string;
-  unverifiedPassword: string;
-  invitationToken: string;
 };
 
 const DEFAULTS: AuthDefaults = defaults as AuthDefaults;
@@ -31,17 +36,11 @@ function isValidEmail(value: string): boolean {
 function envOrDefault(
   name: string,
   fallback: string,
-  opts: { minLength?: number; validator?: (value: string) => boolean } = {}
+  opts: { validator?: (value: string) => boolean } = {}
 ): string {
-  const { minLength = 0, validator } = opts;
+  const { validator } = opts;
   const raw = process.env[name];
   if (raw === undefined || raw === "") {
-    return fallback;
-  }
-  if (minLength > 0 && raw.length < minLength) {
-    console.warn(
-      `[e2e-auth-creds] ${name} is set but too short (${raw.length} < ${minLength}); using fallback`
-    );
     return fallback;
   }
   if (validator && !validator(raw)) {
@@ -49,6 +48,28 @@ function envOrDefault(
       `[e2e-auth-creds] ${name} is set but failed validation; using fallback`
     );
     return fallback;
+  }
+  return raw;
+}
+
+// requiredSecretEnv is the only way a credential enters this suite. It throws
+// instead of returning a fallback, and the message names the variable and how
+// to set it, so a developer who runs the suite with nothing configured gets an
+// instruction rather than a run that quietly seeds shared live accounts.
+export function requiredSecretEnv(name: string, minLength: number): string {
+  const raw = process.env[name] ?? "";
+  if (raw === "") {
+    throw new Error(
+      `[e2e-auth-creds] ${name} is required and has no default. Set it in ` +
+        "your shell (or as a CI secret) before running the E2E suite. " +
+        "Credentials are never committed to this repository; see " +
+        "docs/live-test-auth.md."
+    );
+  }
+  if (raw.length < minLength) {
+    throw new Error(
+      `[e2e-auth-creds] ${name} is set but too short (${raw.length} < ${minLength}).`
+    );
   }
   return raw;
 }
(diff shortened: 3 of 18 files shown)


apps/web-console/tests/e2e/support/e2e-fixture-seed.mjs
```diff
@@ -86,6 +91,32 @@ export function withRunKey(email, runKey) {
...
+// the key gets it. Idempotent, because ci.yml passes an address that is
+// already namespaced alongside the same key.
+export function runScopedEmail(email, runKey) {
+  if (typeof email !== "string" || email === "") {
+    throw new Error(
+      "E2E fixture seeding requires an explicit address for every fixture " +
+        "user. It has no default, because the default was a shared live account."
+    );
+  }
+  if (!runKey) {
```
sakibsadmanshajib, line 108: **security-reviewer, blocking until proven.** This is the function the credential rotation is waiting on, so it needs more than a reading.

I traced every path that can reach a password write. Every `admin.auth.admin.createUser` and `updateUserById` call in the suite lives in `ensureUser`, `ensureUser` is called only from `seedFixtures`, and `seedFixtures` resolves both fixture addresses through `runScopedEmail` before it calls anything. The third address, the inviter, comes from `buildIds(runKey)`, which is reached after those two calls. The only entry point into the module is the CLI, which resolves both addresses at module scope, so it cannot even load without a key. Whitespace is handled: both resolution points do `(process.env.E2E_RUN_KEY ?? "").trim()`, so `"   "` becomes `""` and throws. `seedFixtures` independently does `typeof opts.runKey === "string" ? opts.runKey.trim() : ""`, so an `undefined` run key from a programmatic caller throws too. The literal string `"undefined"` does not throw, but it produces `e2e-verified+undefined@scubed.com.bd`, which is a run-scoped address and not a shared one.

So the claim is correct. My objection is that nothing enforces it. Every line of the argument above is a fact about today's call graph, and the next person to add a caller gets no signal at all. The guard is the thing rotation depends on, and it has no test that fails if someone reintroduces a default parameter or calls `ensureUser` directly.

Required: a test that drives `seedFixtures` with an admin client that throws on any property access, and asserts that an empty, whitespace, and `undefined` run key throws the `E2E_RUN_KEY` error rather than the proxy's, which proves no admin call happens before the guard.

apps/web-console/tests/e2e/support/e2e-fixture-seed.mjs
```diff
@@ -660,7 +691,9 @@ async function sweepStaleFixtureRuns(admin) {
       .eq("account_id", acct.id);
     if (delMapErr) {
       console.error(
-        `sweep: unmap billing account ${acct.id} failed: ${delMapErr.message}`
+        redactSecrets(
+          `sweep: unmap billing account ${acct.id} failed: ${delMapErr.message}`
+        )
```
sakibsadmanshajib, line 696: **security-reviewer.** The claim under review is scoped to password writes, and it holds for those. It does not cover this: `resetProfileComplete` takes an arbitrary address and updates `account_profiles.profile_setup_complete` for every row matching it. `resetProfileBetweenSpecs` passes `email || E2E_VERIFIED_EMAIL`, and the default is run-scoped now, but the argument comes straight off argv, so `node tests/e2e/support/e2e-auth-fixtures.mjs reset-profile e2e-verified@scubed.com.bd` still writes to the shared account's profile row once any run key is exported.

This is not a credential write and it does not revoke anybody's session, so it does not block the rotation. It is still a shared-account mutation surviving in a change whose stated invariant is that fixture writes belong to the run, and the cheap version of the guard is one comparison. Fix it or state in the code why an arbitrary address is allowed here.

apps/web-console/tests/e2e/support/e2e-auth-creds.ts
```diff
@@ -53,28 +52,78 @@ function envOrDefault(
...
-  { validator: isValidEmail }
+// requiredSecretEnv is the only way a credential enters this suite. It throws
+// instead of returning a fallback, and the message names the variable and how
+// to set it, so a developer who runs the suite with nothing configured gets an
+// instruction rather than a run that quietly seeds shared live accounts.
+export function requiredSecretEnv(name: string, minLength: number): string {
+  const raw = process.env[name] ?? "";
+  if (raw === "") {
+    throw new Error(
+      `[e2e-auth-creds] ${name} is required and has no default. Set it in ` +
```
sakibsadmanshajib, line 63: **typescript-reviewer.** `runScopedEmail` now exists twice, once here and once in `e2e-fixture-seed.mjs`, with the idempotence rule (`includes("+" + runKey + "@")`) written out separately in each. The header of this very file justifies the shared JSON on the grounds that both modules read the same values so they "cannot drift apart", and this change quietly reintroduces exactly the drift it was avoiding, in the one function that decides which live account gets written.

The two cannot be merged into one import: Playwright compiles specs to CommonJS and importing the `.mjs` from a spec fails, which is why the duplication exists at all. That is a fair constraint, but it argues for pinning the equivalence, not for leaving it implicit. Add a test that runs both implementations over the same input table and asserts identical output, including the already-namespaced and empty-key cases.

scripts/seed-owui-e2e-user.py
```diff
@@ -170,6 +194,116 @@ def random_password() -> str:
...
+    this script is invoked automatically by .github/workflows/owui-nightly.yml
+    on a schedule, on dispatch, and on any pull request labelled run-owui-e2e.
+    The workflow's concurrency group is keyed on the ref, so a labelled-pull-
+    request run and the scheduled run do not join the same group and can rotate
+    each other's account mid-flight. docs/live-test-auth.md forbids exactly
+    this shape.
+
+    So the default is to leave an existing account alone, and a caller that
+    genuinely wants to set the password says so through an environment
+    variable. A brand-new account still gets new_user_password: there is no
```
sakibsadmanshajib, line 213: **python-reviewer.** The sweep asks GoTrue for a single page of up to 1000 users and filters locally. On a project whose `auth.users` table is larger than that page, the stale fixture users this sweep exists to remove can sit outside it forever, so the leak silently persists while the run still prints nothing and exits zero.

That may well be an acceptable ceiling for a nightly that creates two users a day, but it is currently invisible. Either paginate, or name the ceiling and its upgrade path in a comment so the next reader does not have to infer it from `per_page`.

scripts/redact-log-credentials.py
```diff
@@ -0,0 +1,106 @@
...
+]
+
+# Longest names first so `provider_refresh_token` never degrades to `token`.
+CREDENTIAL_PARAM_RE = re.compile(
+    r"\b(" + "|".join(sorted(CREDENTIAL_PARAMS, key=len, reverse=True)) + r")=([^&#\s\"'\\]+)",
+    re.IGNORECASE,
+)
+
+# A bare JWT with no parameter name around it: a session token on its own line,
+# an `Authorization: Bearer ...` header, or a service-role key.
```
sakibsadmanshajib, line 64: **python-reviewer.** `CREDENTIAL_PARAM_RE` anchors each name with `\b`, and `_` is a word character, so `RAG_VERIFY_PASSWORD=<value>` does not match: the boundary before `password` fails against the preceding underscore. Environment-variable spelling is the dominant shape of a credential in a container log dump (`OWUI_E2E_PASSWORD=`, `SUPABASE_ANON_KEY=`, and the line `verify-rag-roundtrip.py` prints on its create path is literally `export RAG_VERIFY_PASSWORD=...`), so the redactor misses the case it will meet most often in exactly the artifacts this change added it to.

Allow an optional uppercase or underscore prefix on the parameter names, and add a self-check case for the `A_B_PASSWORD=` shape so this cannot regress.

.github/workflows/ci.yml
```diff
@@ -1377,14 +1401,32 @@ jobs:
...
-          path: apps/web-console/playwright-report/
+          # Traces (.zip), videos (.webm) and index.html are excluded. This
+          # repository is public, so any artifact here is downloadable by
+          # anyone with the run URL, and playwright.config.ts retains a trace
+          # on failure: it holds every request header (Authorization: Bearer)
+          # and every sb-*-auth-token cookie for the fixture session.
+          # index.html goes too, and for the same reason rather than a
+          # different one: the HTML reporter base64-inlines the whole report
+          # payload into that file, including stdout, stderr and error text,
+          # and a waitForURL timeout enumerates the URLs it navigated,
```
sakibsadmanshajib, line 1425: **plain adversarial pass.** This upload now produces an empty artifact, and the PR body claims otherwise.

`playwright.config.ts` sets `trace` and `video` but never sets `screenshot`, so it keeps Playwright's default of `off`. The HTML report directory is `index.html` plus `data/`, and `data/` holds only trace zips, videos and screenshots. Excluding `*.zip`, `*.webm` and `index.html` therefore excludes everything that job produces. The step still runs with `if-no-files-found: ignore`, so it uploads nothing and says nothing, which is the quiet absence this repository has been bitten by before, and the PR body's "screenshots survive" is true only of the OWUI job, which does set `screenshot: only-on-failure`.

Two honest options: delete the step, or give it something worth keeping by setting `screenshot: "only-on-failure"` in `playwright.config.ts`. A viewport screenshot carries no request headers and no cookies, and unlike a trace it is inspectable, so it is the cheapest thing that keeps this artifact useful. Either way the PR body has to stop claiming a payload that is not there.

.github/workflows/ci.yml
```diff
@@ -1296,6 +1299,16 @@ jobs:
           done
           echo "::error::Next.js did not start"; cat /tmp/nextjs.log; exit 1
 
+      - name: Generate the fixture invitation token
+        # Random per job, and masked, because the seeder stores its sha256 as a
+        # live pending invitation. A value anybody can derive from the public
+        # run page is a published credential for as long as the run lasts.
+        run: |
```
sakibsadmanshajib, line 1312: **ecc:code-review.** The token generated here is masked with `::add-mask::` and written to `GITHUB_ENV`, which is right, but note what that buys and what it does not. The mask covers this job's own log. The value still reaches the live pending invitation as a sha256, and it is now unguessable, which was the point of the change.

One loose end for a reader: nothing states the required length relationship. `e2e-auth-defaults.json` sets `minTokenLength` to 16 and `requiredSecretEnv` enforces it, while `openssl rand -hex 24` yields 48 characters, so the two agree today by accident rather than by construction. A one-line comment naming the floor it has to clear would stop a later shortening of this command from failing far away in `requiredSecretEnv` with a message about a variable this step does not mention.

apps/web-console/scripts/verify-spec-collection.mjs
```diff
@@ -53,8 +53,23 @@ const OWUI_LIST_ENV = {
...
+// shared live account or skip in silence. Listing loads those modules, so this
+// guard needs the variables present but does not need them to be real. Same
+// reasoning and same shape as OWUI_LIST_ENV above: listing never dials out.
+const WEB_CONSOLE_LIST_ENV = {
+  E2E_VERIFIED_PASSWORD: "unused-collection-guard-placeholder",
+  E2E_UNVERIFIED_PASSWORD: "unused-collection-guard-placeholder",
+  E2E_INVITATION_TOKEN: "unused-collection-guard-placeholder",
+  // Required too: every fixture address is namespaced with it so a run can
+  // only touch accounts it created (runScopedEmail in e2e-fixture-seed.mjs).
+  E2E_RUN_KEY: "collection-guard",
```
sakibsadmanshajib, line 68: **/codex:adversarial-review, attacking the deviation rather than accepting it.**

The deviation: module-scope resolution kept, and the collection guard given placeholder values, instead of resolving the credentials lazily as the review asked. The defence offered was that a lazily resolved empty value turns a missing credential into `test.skip`, the exact silent skip this change removes.

The strongest case against it. First, placeholders here mean the guard now proves less than it did: it proves the modules load with three arbitrary strings present, not that they load the way a real run loads them, so a future credential added at module scope will pass this guard and fail only in the E2E job. Second, this file now carries knowledge of another module's environment contract, so adding a fourth required variable means editing a spec-collection guard, which is an unobvious coupling. Third, throwing at module load makes `e2e-auth-creds.ts` unimportable by any future consumer that wants only the addresses, and it converts one missing variable into a total collection failure rather than a targeted one.

Why I still land on the deviation being correct. The `test.skip` objection is not hypothetical: every gated spec computes `HAS_CREDS` at module scope, `Boolean(VERIFIED_EMAIL && VERIFIED_PASSWORD)`, so a lazily resolved empty string produces a green run that tested nothing, which is the precise failure this PR exists to end, and it is worse than a load error because it is silent. The coupling objection is real but already priced in: `OWUI_LIST_ENV` in this same file does the identical thing for the OWUI config, so this is the established local pattern rather than a new one. The unimportable-module objection has no current victim: every importer of that module uses the credentials.

Verdict: deviation upheld, on the condition that the placeholder block names why it exists, which it does. No change required.

sakibsadmanshajib (commented):
## Stage 6 adversarial review, pipeline mode

Run by the implementing agent under the stage 6 contract in `.claude/rules/orchestrator.md`. Streams selected by diff shape, with the mandatory security stream because this is a credential and auth path.

| stream | verdict |
| --- | --- |
| CodeRabbit CLI (v0.7.2) | **SKIPPED**, see below |
| `ecc:code-review` | 1 comment |
| plain adversarial pass | 2 comments |
| `typescript-reviewer` | 1 comment |
| `python-reviewer` | 2 comments |
| `security-reviewer` (mandatory) | 2 comments |
| `/codex:adversarial-review` | 1 comment |

**CodeRabbit CLI is SKIPPED, not clean.** `coderabbit review --base main` fails with `Review failed: A valid organization session is required` (`UNAUTHORIZED`, HTTP 401, `vsCode.requestFullReview`), on both the default and `--light` paths. `coderabbit auth status` shows a signed-in user account but no valid organization session. Its absence is not a pass. The separate CodeRabbit GitHub App check on this PR is also not a substitute: it reported `Review rate limited` earlier in this PR's life, and a rate-limited or blocked bot run must never be read as approval.

**Where the streams were pointed hardest**, per the standing instruction: the claim that gates rotating the two exposed live credentials, namely that `runScopedEmail` throwing without `E2E_RUN_KEY`, plus deleting `DEFAULT_EMAILS`, means no path can hand the seeder a shared address. Result of that trace is in the first inline comment. Short version: the claim holds under a full caller trace, but it was asserted rather than proven, and one adjacent shared-account write path survives that the claim did not cover.

The deviation (module-scope resolution kept, collection guard fixed instead of resolving lazily) was attacked rather than accepted. See the `/codex:adversarial-review` comment.

apps/web-console/tests/e2e/support/e2e-fixture-seed.mjs
sakibsadmanshajib: Fixed in 190d2479. The claim is now enforced rather than argued: \`tests/unit/e2e-auth-creds.test.ts\` drives \`seedFixtures\` with an admin client that is a Proxy throwing on any property access, and asserts that an empty, whitespace, \`undefined\` and non-string run key all reject with the \`E2E_RUN_KEY\` error rather than the proxy's. If a future change lets any admin call happen before the guard, the rejection message changes and the test fails. A fifth case covers a missing address. Ten tests pass in the web-console container.

apps/web-console/tests/e2e/support/e2e-fixture-seed.mjs
sakibsadmanshajib: Fixed in 190d2479. \`resetProfileBetweenSpecs\` now refuses any address that does not carry this run's key, with a message naming the address, the key and the rule. You were right that it does not block rotation, since it writes \`profile_setup_complete\` and not a credential, but it was the last write in this module that could reach a shared account, and the guard is one comparison.

apps/web-console/tests/e2e/support/e2e-auth-creds.ts
sakibsadmanshajib: Fixed in 190d2479, keeping the duplication and pinning the equivalence. The two implementations cannot be collapsed into one import for the reason you noted, so \`tests/unit/e2e-auth-creds.test.ts\` now runs both over the same table (shared base, already-namespaced, foreign tag, literal "undefined", no at-sign) and asserts identical output, plus that both refuse the empty key. Drift now fails a test instead of silently changing which live account gets written.

scripts/seed-owui-e2e-user.py
sakibsadmanshajib: Fixed in 190d2479, by naming the ceiling rather than paginating. A \`ponytail:\` comment now states that this is one page of 1000, that the consequence is stale users outside that page never being swept and never failing a run, and what the upgrade path is (loop \`page\` until a short page, keep the both-halves address match). Two users a night against a 1000-user page does not justify pagination today, and an unstated ceiling was the actual defect.

scripts/redact-log-credentials.py
sakibsadmanshajib: Fixed in 190d2479. Good catch, and it was the common case rather than an edge one: the parameter names now take an optional \`[A-Za-z0-9_]*\` prefix, so \`RAG_VERIFY_PASSWORD=\`, \`OWUI_E2E_PASSWORD=\` and \`E2E_VERIFIED_PASSWORD=\` all scrub, while \`state=\` and ordinary log text still pass through untouched. Three self-check cases pin it, including the exact line \`verify-rag-roundtrip.py\` prints.

.github/workflows/ci.yml
sakibsadmanshajib: Fixed in 190d2479, taking the second option. \`playwright.config.ts\` now sets \`screenshot: "only-on-failure"\`, so the artifact carries something after the three exclusions, and the comment there explains that the exclusions are what make the setting load-bearing. A screenshot is viewport only: no URL bar, no headers, no cookies, and unlike a trace a human can see what is in it. The pull request body has been corrected in the same pass, since "screenshots survive" was true only of the OWUI job.

.github/workflows/ci.yml
sakibsadmanshajib: Fixed in 190d2479. The step now states that 24 bytes gives 48 hex characters, that the floor is \`minTokenLength\` in \`e2e-auth-defaults.json\` enforced by \`requiredSecretEnv\` at 16, and that shortening this command without moving both surfaces the failure far from here under a variable name this step never mentions.

apps/web-console/scripts/verify-spec-collection.mjs
sakibsadmanshajib: Rebuttal accepted as posted, no code change. Recording the outcome explicitly so the next reviewer does not have to re-derive it: the deviation stands because the \`test.skip\` failure mode is concrete rather than theoretical (every gated spec computes \`HAS_CREDS\` at module scope, so a lazily resolved empty string yields a green run that tested nothing), the coupling objection is real but already the established local pattern via \`OWUI_LIST_ENV\` in the same file, and the unimportable-module objection has no current victim. The weakest point remains true and is worth carrying forward: the collection guard now proves the modules load with three placeholder strings present, not that they load the way a real run loads them. The Web E2E job is what covers that gap, and it is green on this head.

PR #891, opened 2026-08-12
fix: release the Cowork concurrency slot on cancel and return create promptly

sakibsadmanshajib:
Fixes #886. Fixes #881.

Two Cowork defects found by the live coverage suite and reproduced directly against the API on the demo box on 2026-08-11. Both are demo blockers, and PR #799 (a post-deploy job that launches and cancels a real sandbox on every push) cannot merge until this does.

## Issue #886, cancel never reached the engine

Root cause: the `Engine` interface in `apps/control-plane/internal/agenttask/engine.go` had exactly one method, `Launch`, so `Service.Cancel` could only ever be a bare `repo.Transition(..., StatusCancelled, ...)`. The row went cancelled while the sandbox behind it kept running, and the launcher releases a concurrency slot when its session ends (`apps/agent-engine/internal/quota` hands `Acquire` a release func called on session end). The slot therefore stayed held until the sandbox finished on its own, roughly sixteen minutes on the demo box. Two cancels exhausted `HIVE_QUOTA_USER_CONCURRENCY` and every later create was refused with "agent engine could not start the task", which the console renders as `Blocked` for a user whose task list shows nothing running.

Fix: `Engine` now carries `Cancel(ctx, sessionRef) error`, and `Service.Cancel` calls it. Both real implementations already had the method (`agentengine.Remote` for the socket arm the demo box runs, `agentengine.Engine` for the in-process arm), so the change is the interface plus the call, not new launcher logic. `NotConfiguredEngine` gets a trivially successful `Cancel`, since it never launched anything.

Why on the interface rather than an optional runtime type assertion: the defect is precisely that the seam could not express "stop". A compile-time method makes it impossible to wire an engine with no cancel path, and there is no implementation that legitimately lacks one.

## Issue #881, create reported failure for work that succeeded

Root cause: `CreateTask` called `engine.Launch` inline with `launchTimeout = 5 * time.Minute` while `apps/edge-api/internal/agenttask.NewClient` builds its HTTP client with `Timeout: 15 * time.Second`. A cold sandbox mount is routinely slower than 15 seconds, so edge-api gave up first and answered the browser `500`, while control-plane carried on (it deliberately detaches the launch from the caller's context) and the task reached `succeeded`. Measured live: `CREATE status=500 elapsed=18.0s` for a task that finished 78 seconds later.

Fix: create returns the persisted `queued` task as soon as the row exists and the launch runs on a background goroutine, which transitions the task to `running` or to `failed` on its own. The console already polls queued to running to done, and `CreateTask` already ran the launch on a detached context, so this is the shape the code was built for.

Why not simply align the timeouts: raising edge-api's client to five minutes makes an interactive create legitimately able to hang for five minutes, and every intermediate proxy is free to cut it anyway, which reproduces the same wrong answer with a longer fuse. edge-api's 15 seconds now bounds only short database work plus, for cancel, a stop call bounded at 10 seconds inside control-plane (`engineCancelTimeout`), deliberately under the client budget so a slow launcher can never reproduce the same "failure reported for work that happened" shape.

A launch failure still reaches the caller, just on the next poll rather than in the create response, so no task sits queued forever with no signal.

## Concurrency decisions

The row transition is the atomic gate throughout. `Repository.Transition`'s UPDATE carries the "not already terminal" precondition, so exactly one caller can win it, and only the winner touches the engine.

- **Double cancel.** The second cancel loses that guard, returns `ErrTerminalState` (HTTP 409) and never reaches the engine. Asked for exactly once, asserted in `TestService_Cancel_TerminalStateRejected`.
- **Cancel racing a completion.** Whichever transition commits first wins. If the poller already recorded a terminal status the cancel is rejected and the engine is left alone, which is correct: a terminal status is what makes the launcher reap that session itself (`SandboxEngine.Status` reaps on terminal). Asserted in `TestService_Cancel_LosesRaceWithCompletion_LeavesEngineAlone`.
- **Cancel arriving before the launch finishes.** The row has no `engine_session_ref` yet, so `Service.Cancel` has nothing to stop. The in-flight launch goroutine then finds the task already terminal when it tries to record `running`, and stops the session it just started. The same path covers a `running` transition that fails for any other reason: a session whose reference was never persisted can never be polled or cancelled by anything else, so it is torn down rather than left holding a slot for its full natural life. Asserted in `TestService_CancelDuringLaunch_ReleasesTheSlotThatLaunchTook`.

Cancel order is transition first, engine second, deliberately. Stopping the engine first would risk killing a sandbox for a task the database then reports as succeeded.

An engine stop failure is logged for the operator (with the engine detail, which never reaches a customer-visible field) and never returned to the caller. The cancellation itself is already recorded, and a stuck launcher is not something a customer can act on. Returning 500 for a task that really is cancelled would be issue #881's shape all over again.

## Tests, and how they were watched red

The cancel test asserts the concurrency slot, not the status code. A status-code assertion passes against the broken code, which is how this survived: `Service.Cancel` already returned a cancelled task.

`slotEngine` in `service_test.go` models the launcher's accounting (`Launch` takes a slot and refuses at the ceiling, `Cancel` gives one back), because `apps/agent-engine/internal/quota` is under another module's `internal` and control-plane cannot import it.

Red baseline, run before any production change, with only the interface method and a no-op `WaitForLaunches` stub added so the package compiled:

```
=== RUN   TestService_Cancel_FromRunning
    service_test.go:382: expected the engine session to be cancelled exactly once, got 0 calls
--- FAIL: TestService_Cancel_FromRunning (0.00s)
=== RUN   TestService_Cancel_TerminalStateRejected
    service_test.go:403: expected exactly one engine cancel across a double cancel, got 0
--- FAIL: TestService_Cancel_TerminalStateRejected (0.00s)
=== RUN   TestService_Cancel_ReleasesEngineConcurrencySlot
    service_test.go:441: cancel did not release the launcher slot: 1 still in use (issue #886)
--- FAIL: TestService_Cancel_ReleasesEngineConcurrencySlot (0.00s)
=== RUN   TestService_CancelDuringLaunch_ReleasesTheSlotThatLaunchTook
    service_test.go:462: CreateTask blocked on the engine launch (issue #881): edge-api gives up at 15s and reports a failure for a task that is in fact starting
--- FAIL: TestService_CancelDuringLaunch_ReleasesTheSlotThatLaunchTook (2.01s)
=== RUN   TestService_Cancel_LosesRaceWithCompletion_LeavesEngineAlone
--- PASS: TestService_Cancel_LosesRaceWithCompletion_LeavesEngineAlone (0.00s)
=== RUN   TestService_CreateTask_DoesNotBlockOnLaunch
    service_test.go:527: CreateTask blocked on the engine launch (issue #881): edge-api gives up at 15s and reports a failure for a task that is in fact starting
--- FAIL: TestService_CreateTask_DoesNotBlockOnLaunch (2.00s)
FAIL
```

An earlier run of the same tests hung the whole package for 600 seconds, because against the broken code `CreateTask` blocks on the gated launch forever. The create call now runs on its own goroutine inside the test helper, so the defect fails one test in two seconds with a legible message instead of timing out the suite.

Green after the fix, both packages, plus the full short suite for control-plane, agent-engine and edge-api:

```
ok  github.com/sakibsadmanshajib/hive/apps/control-plane/internal/agenttask  0.115s
ok  github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/engine      0.128s
```

`TestSandboxEngine_Cancel_FreesQuotaSlot` is added on the launcher side as a characterisation test, not as coverage for this change: nothing in `apps/agent-engine/internal/engine` changed here, so every assertion in it survives a full revert of the fix. It pins the invariant control-plane now depends on (ending a session frees its slot immediately) and is labelled that way in the code. The regression coverage for issue #886 is `TestService_Cancel_ReleasesEngineConcurrencySlot`, `TestService_CancelDuringLaunch_ReleasesTheSlotThatLaunchTook` and `TestService_Cancel_LosesRaceWithCompletion_LeavesEngineAlone`, which drive `Service.Cancel` and fail on revert.

Review round two added two more, both watched red first: `TestService_LaunchSucceedsButTransitionFails_TaskFailsVisibly` (a launch that succeeds and then cannot record itself must end failed, not stranded in queued) and `TestService_LaunchPanic_DoesNotCrashTheProcess` (a panicking engine must cost one task, not the process).

`-race` could not run locally (the toolchain image has no gcc, so cgo is unavailable). CI runs `-race` on the module legs.

## These tests actually run in CI

`.github/workflows/ci.yml`'s explicit Go package list is on the `-tags integration` step only. The plain unit-test step is `go test -count=1 -short -race -timeout 5m ./...` run per module against a matrix that includes both `./apps/control-plane` and `./apps/agent-engine`, so both touched packages are in scope. Neither new test carries a build tag, an environment gate, or a `testing.Short()` skip. The only skip in either package is the pre-existing `HIVE_TEST_DB_URL` gate in `repository_test.go`, and `./internal/agenttask/...` is in the control-plane integration list too.

## Known gaps this does not close, and the condition attached to one of them

Both were found by the security review stream, both were confirmed by tracing them myself, and neither blocks this merge.

- **Issue #899**: a sandbox the launcher registered whose reference never reaches control-plane (a lost response, `launchTimeout` firing, or this process dying mid-deploy) holds its slots until the launcher restarts, because `reap` only runs from `Status` or `Cancel` and the poller cannot see a row with no `engine_session_ref`. This change closes the database-failure half of that shape; the transport half needs the launcher to offer a way to re-derive a lost reference, so it cannot be closed from this service. Mitigant: every deploy restarts the launcher and clears the leak.
- **Issue #900**: `POST /v1/agent/tasks` has no rate limit, verified through the edge middleware chain, so rows, goroutines and pool checkouts are per-tenant unbounded. The launcher's quota gates the sandbox launch, which happens after the row insert.

**Condition on #900, recorded here because it is a merge condition rather than a nice-to-know:** it is tolerable only while Cowork stays feature-gated. `featuregate.Require(FeatureCowork)` is doing the containment a limiter would otherwise do. Opening `FeatureCowork` broadly makes it intolerable, so whoever widens that gate owns closing #900 first. Written into the issue as well.

## Not touched

The demo box, `role_pgx.go`, the accounts package, container capabilities. No new dependency. No behaviour change to the launcher itself.

## Buglog entry

To be appended to `.wolf/buglog.jsonl` on `main` in a separate buglog-only pull request after this merges, per `.claude/rules/openwolf.md`.

```json
{"id":"bug-2026-08-12-cowork-cancel-slot-and-blocking-create","date":"2026-08-12","title":"Cancelling a Cowork task leaked its concurrency slot, and create reported failure for a task that succeeded","error_message":"agent engine could not start the task (console: Blocked) after two cancels; CREATE status=500 elapsed=18.0s for a task that reached succeeded","root_cause":"The agenttask.Engine interface carried only Launch, so Service.Cancel was a bare database transition and never stopped the sandbox; the launcher only releases a concurrency slot when the session ends, so a cancelled task held its slot for the sandbox's full life. Separately, CreateTask blocked inline on a launch bounded at five minutes while edge-api's control-plane client times out at fifteen seconds, so the browser was told 500 for work that kept running.","fix":"Added Cancel to the Engine interface and called it from Service.Cancel after the row's atomic terminal guard, so the slot is released at cancel time. Made CreateTask return the persisted queued task and run the launch on a background goroutine, with the in-flight launch stopping its own session if it finds the task already terminal.","tags":["cowork","agent-engine","quota","concurrency","timeout","control-plane","issue-886","issue-881"]}
```

## Summary by CodeRabbit

* **New Features**
  * Task creation now returns immediately with a queued status while launch proceeds in the background.
  * Task status updates to running or failed after launch completes.
  * Cancellation stops active sessions and promptly frees available concurrency capacity.
* **Bug Fixes**
  * Improved handling of cancellation during task launch, including race conditions and cancellations before launch begins.
  * Failed launches now report safe, provider-neutral error messages.
* **Documentation**
  * Updated task lifecycle and cancellation behavior documentation, including asynchronous launch and bounded cancellation timing.

### DIFF
--- a/apps/agent-engine/internal/engine/engine_test.go
+++ b/apps/agent-engine/internal/engine/engine_test.go
@@ -491,6 +491,37 @@ func TestSandboxEngine_Cancel_CleansUpDirs(t *testing.T) {
 	assertDirEmpty(t, e.cfg.WorkspaceRoot)
 }
 
+// The launcher half of issue #886. Control-plane now calls Cancel when a task
+// is cancelled, and the only reason that helps is this: the cancelled
+// session's concurrency slot has to go back to the pool immediately, not
+// whenever the sandbox happens to finish on its own (roughly sixteen minutes
+// on the demo box, which is how two cancels exhausted a user's ceiling).
+func TestSandboxEngine_Cancel_FreesQuotaSlot(t *testing.T) {
+	var fake *fakeAgentServer
+	e := newTestEngineWithQuota(t, &fake, 1, 1)
+
+	task := testTask()
+	sessionRef, err := e.Launch(context.Background(), task)
+	if err != nil {
+		t.Fatalf("Launch: %v", err)
+	}
+	if err := e.Cancel(context.Background(), sessionRef); err != nil {
+		t.Fatalf("Cancel: %v", err)
+	}
+
+	if tenant, user := e.q.InUse(task.TenantID, task.UserID); tenant != 0 || user != 0 {
+		t.Fatalf("cancel left quota held: tenant=%d user=%d, want 0/0", tenant, user)
+	}
+
+	// The observable consequence: the same user can start work again at once.
+	next := testTask()
+	next.TenantID = task.TenantID
+	next.UserID = task.UserID
+	if _, err := e.Launch(context.Background(), next); err != nil {
+		t.Fatalf("expected the cancelled session's slot to be reusable, got %v", err)
+	}
+}
+
 func assertDirEmpty(t *testing.T, dir string) {
 	t.Helper()
 	entries, err := os.ReadDir(dir)
--- a/apps/agent-engine/internal/engine/engine_test.go
+++ b/apps/agent-engine/internal/engine/engine_test.go
@@ -491,11 +491,18 @@ func TestSandboxEngine_Cancel_CleansUpDirs(t *testing.T) {
 	assertDirEmpty(t, e.cfg.WorkspaceRoot)
 }
 
-// The launcher half of issue #886. Control-plane now calls Cancel when a task
-// is cancelled, and the only reason that helps is this: the cancelled
-// session's concurrency slot has to go back to the pool immediately, not
-// whenever the sandbox happens to finish on its own (roughly sixteen minutes
-// on the demo box, which is how two cancels exhausted a user's ceiling).
+// Characterisation test for pre-existing behaviour, NOT regression coverage
+// for issue #886. Nothing in this package changed in that fix, so every
+// assertion below passes with the fix fully reverted; it exists to pin the
+// invariant control-plane now depends on, which is that ending a session
+// through Cancel returns its slot to the pool immediately rather than whenever
+// the sandbox finishes on its own (roughly sixteen minutes on the demo box).
+//
+// The actual regression coverage for issue #886 is in
+// apps/control-plane/internal/agenttask/service_test.go, where
+// TestService_Cancel_ReleasesEngineConcurrencySlot and its two siblings fail
+// on revert because they drive Service.Cancel and assert on counters that only
+// move when Service reaches the engine.
 func TestSandboxEngine_Cancel_FreesQuotaSlot(t *testing.T) {
 	var fake *fakeAgentServer
 	e := newTestEngineWithQuota(t, &fake, 1, 1)
--- a/apps/control-plane/internal/agenttask/service.go
+++ b/apps/control-plane/internal/agenttask/service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"runtime/debug"
 	"sync"
 	"time"
 
@@ -18,7 +19,7 @@ type Service struct {
 	engine Engine
 
 	// launches counts the background launch goroutines CreateTask starts.
-	// Nothing on the request path waits on it; see WaitForLaunches.
+	// Nothing on the request path waits on it; see WaitIdle.
 	launches sync.WaitGroup
 }
 
@@ -102,13 +103,39 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
 	// Worse, the Transition calls would inherit the same dead context and
 	// fail too, leaving the row queued forever.
 	opCtx := context.WithoutCancel(ctx)
+	s.background(func() { s.launch(opCtx, t) }, func() {
+		// Panic path. Before this work ran on its own goroutine it ran inside
+		// the HTTP handler, where net/http's per-connection recover contained a
+		// panic to one connection; a bare goroutine would instead take the
+		// whole process down, every tenant with it. Recovering keeps the blast
+		// radius at one task and still records that task as failed, so the
+		// customer is not left watching a queued row that will never move.
+		s.recordLaunchFailure(opCtx, t, engineLaunchFailedMessage)
+	})
+
+	return t, nil
+}
+
+// background runs work on a goroutine tracked by WaitIdle, recovering any
+// panic so one bad launch or stop cannot take the process down. onPanic runs
+// after the panic is logged and may be nil.
+func (s *Service) background(work func(), onPanic func()) {
 	s.launches.Add(1)
 	go func() {
 		defer s.launches.Done()
-		s.launch(opCtx, t)
+		defer func() {
+			r := recover()
+			if r == nil {
+				return
+			}
+			slog.Default().Error("agenttask: recovered panic in background work",
+				"panic", r, "stack", string(debug.Stack()))
+			if onPanic != nil {
+				onPanic()
+			}
+		}()
+		work()
 	}()
-
-	return t, nil
 }
 
 // launch performs one launch attempt and records its outcome. Runs on a
@@ -126,15 +153,24 @@ func (s *Service) launch(ctx context.Context, t Task) {
 			// The session started but its reference could not be recorded, so
 			// nothing can ever poll or stop it and it would hold its
 			// concurrency slot for its full natural life. Stop it here.
+			s.stopEngineSession(ctx, t.ID, sessionRef)
+
 			// ErrTerminalState is the expected case: a cancel won the race
 			// while this launch was still in flight, and it had no session
 			// reference to stop at the time, so this goroutine owns the
-			// teardown.
-			if !errors.Is(terr, ErrTerminalState) {
-				slog.Default().WarnContext(ctx, "agenttask: could not record a launched session, stopping it",
-					"task_id", t.ID, "error", terr)
+			// teardown. The task's terminal state is already correct.
+			if errors.Is(terr, ErrTerminalState) {
+				return
 			}
-			s.stopEngineSession(ctx, t.ID, sessionRef)
+			// Anything else is a real database failure, and leaving the row
+			// queued with no session reference would strand it: ListActive
+			// excludes exactly that shape, so the poller would never touch it
+			// and the customer would watch a task that can never move and
+			// carries no error. Record the failure instead. Best effort by
+			// definition, since the same database just refused a write.
+			slog.Default().WarnContext(ctx, "agenttask: could not record a launched session, stopped it and failing the task",
+				"task_id", t.ID, "error", terr)
+			s.recordLaunchFailure(ctx, t, engineLaunchFailedMessage)
 		}
 	case errors.Is(err, ErrEngineNotConfigured):
 		s.recordLaunchFailure(ctx, t, engineUnavailableMessage)
@@ -174,21 +210,31 @@ func (s *Service) stopEngineSession(ctx context.Context, taskID uuid.UUID, sessi
 	}
 }
 
-// WaitForLaunches blocks until every background launch this Service started
-// has finished. Nothing on the request path calls it; it exists so tests can
-// assert on a settled task, and so a drain can let in-flight launches record
-// their outcome before the process exits.
+// WaitIdle blocks until every background launch and engine stop this Service
+// started has finished. Nothing on the request path calls it; it exists so
+// tests can assert on a settled task, and so a drain could let in-flight work
+// record its outcome before the process exits.
 //
-// Deliberately not wired into cmd/server's shutdown: a launch is bounded at
-// launchTimeout (five minutes) and container stop is bounded far lower, so
-// waiting there would only trade a clean stop for a SIGKILL. A process that
-// dies mid-launch leaves its task queued with no engine_session_ref, which
-// Repository.ListActive excludes, so the poller never advances it. That
-// window is unchanged by making the launch asynchronous: the launch always
-// outlived the caller's context, so a restart always stranded it. A sweep for
-// stale queued rows is the fix if it ever shows up in practice; nothing has
-// asked for one yet (ponytail).
-func (s *Service) WaitForLaunches() { s.launches.Wait() }
+// Deliberately not wired into cmd/server's shutdown, and the shutdown window
+// this leaves is genuinely worse than before, so it is written down rather
+// than glossed. Previously the launch ran inside the HTTP handler, so
+// srv.Shutdown's 10 second budget (cmd/server/main.go) waited on that
+// connection and an in-flight launch had up to 10 seconds of grace. The
+// handler now returns as soon as the row is persisted, so Shutdown has
+// nothing to wait for and the background goroutine is killed with
+// approximately none.
+//
+// A task stranded that way sits queued with no engine_session_ref, which
+// Repository.ListActive excludes, so the poller never advances it. Waiting on
+// WaitIdle in Shutdown is not the fix: a launch is bounded at launchTimeout
+// (five minutes) while container stop is bounded far lower, so it would only
+// trade a clean exit for a SIGKILL at the same point. The fix, when a
+// mid-deploy stranded task actually shows up, is a sweep that fails queued
+// rows older than launchTimeout, which needs a change to the
+// agent_tasks_list_active() function and therefore a migration. Not built here
+// (ponytail), and named so the next person does not reason from a false
+// premise.
+func (s *Service) WaitIdle() { s.launches.Wait() }
 
 // Get returns one task, scoped to (tenantID, userID) so a task started by
 // one user is never resumable by a different user in the same tenant.
@@ -233,6 +279,16 @@ func (s *Service) Cancel(ctx context.Context, tenantID, userID, id uuid.UUID) (T
 	// An empty EngineSessionRef means the launch has not reported back yet;
 	// the in-flight launch goroutine sees this task is already terminal and
 	// tears its own session down (see launch).
-	s.stopEngineSession(ctx, id, cancelled.EngineSessionRef)
+	//
+	// The stop runs in the background for the same reason create no longer
+	// waits on a launch (issue #881). Blocking here put up to
+	// engineCancelTimeout (10 seconds) plus the database write inside a request
+	// edge-api abandons at 15 seconds, and the box where this matters is by
+	// definition the loaded one. The caller's answer does not depend on the
+	// stop anyway: the cancellation is already committed, and a stop failure is
+	// logged for an operator rather than returned. Tests wait on WaitIdle.
+	sessionRef := cancelled.EngineSessionRef
+	stopCtx := context.WithoutCancel(ctx)
+	s.background(func() { s.stopEngineSession(stopCtx, id, sessionRef) }, nil)
 	return cancelled, nil
 }
--- a/apps/control-plane/internal/agenttask/service.go
+++ b/apps/control-plane/internal/agenttask/service.go
@@ -173,17 +173,35 @@ func (s *Service) launch(ctx context.Context, t Task) {
 	// ever held, and nothing else could ever stop it. Stopping comes first
 	// because it frees the concurrency slot without touching the database,
 	// which is the dependency most likely to have caused the panic.
-	var sessionRef string
+	var (
+		sessionRef  string
+		teardownRun bool // this launch already attempted its own teardown
+	)
 	defer func() {
 		r := recover()
 		if r == nil {
 			return
 		}
+		// Residual, deliberately not wrapped: a panic from this logging call
+		// (or from debug.Stack) escapes into background's recover, which logs
+		// through the same slog call, so the two nets are not independent for
+		// that one case. Reaching it needs a slog handler that panics, and
+		// control-plane installs none: there is no slog.SetDefault in
+		// cmd/server, so this is the stdlib text handler, and fmt recovers from
+		// a panicking String or Error on the panic value itself. Revisit if a
+		// structured handler is ever installed.
 		slog.Default().Error("agenttask: recovered panic during launch",
 			"task_id", t.ID, "panic", r, "stack", string(debug.Stack()))
-		safely("stopping the session of a panicking launch", func() {
-			s.stopEngineSession(ctx, t.ID, sessionRef)
-		})
+		// Skipped when the launch already tore its own session down: repeating
+		// it is harmless for accounting (quota's release is a sync.Once and
+		// reap is idempotent) but it logs a leaked-slot warning for a slot that
+		// was correctly freed, because the control socket directory is gone by
+		// then and Interrupt fails.
+		if !teardownRun {
+			safely("stopping the session of a panicking launch", func() {
+				s.stopEngineSession(ctx, t.ID, sessionRef)
+			})
+		}
 		safely("recording a panicking launch as failed", func() {
 			s.recordLaunchFailure(ctx, t, engineLaunchFailedMessage)
 		})
@@ -196,6 +214,7 @@ func (s *Service) launch(ctx context.Context, t Task) {
 			// The session started but its reference could not be recorded, so
 			// nothing can ever poll or stop it and it would hold its
 			// concurrency slot for its full natural life. Stop it here.
+			teardownRun = true
 			s.stopEngineSession(ctx, t.ID, sessionRef)
 
 			// ErrTerminalState is the expected case: a cancel won the race
@@ -251,6 +270,13 @@ func (s *Service) recordLaunchFailure(ctx context.Context, t Task, message strin
 // recorded, and a failure here is an operator problem (a leaked slot), not
 // something the customer can act on or should see as a failed request. The
 // engine detail goes to the log only, never to a customer-visible field.
+//
+// The warning carries session_ref because on two paths that reach it (a cancel
+// that won the race before the launch reported back, and the panic path above)
+// the reference exists nowhere else by then: the row's engine_session_ref was
+// never written, the poller cannot see the task, and the launcher's registry is
+// in memory with no listing endpoint. Without it the only remedy left is
+// restarting the launcher, which drops every other tenant's live sandbox too.
 func (s *Service) stopEngineSession(ctx context.Context, taskID uuid.UUID, sessionRef string) {
 	if sessionRef == "" {
 		return // never launched, so nothing holds a slot
@@ -259,7 +285,7 @@ func (s *Service) stopEngineSession(ctx context.Context, taskID uuid.UUID, sessi
 	defer done()
 	if err := s.engine.Cancel(stopCtx, sessionRef); err != nil {
 		slog.Default().WarnContext(stopCtx, "agenttask: engine cancel failed, the session may still hold its concurrency slot",
-			"task_id", taskID, "error", err)
+			"task_id", taskID, "session_ref", sessionRef, "error", err)
 	}
 }
 
--- a/apps/control-plane/internal/agenttask/service.go
+++ b/apps/control-plane/internal/agenttask/service.go
@@ -91,6 +91,12 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
 	}
 }
 
+// WaitForLaunches blocks until every background launch this Service started
+// has finished. Nothing on the request path calls it; it exists so tests can
+// assert on a settled task, and so a future graceful drain can let in-flight
+// launches record their outcome before the process exits.
+func (s *Service) WaitForLaunches() {}
+
 // Get returns one task, scoped to (tenantID, userID) so a task started by
 // one user is never resumable by a different user in the same tenant.
 func (s *Service) Get(ctx context.Context, tenantID, userID, id uuid.UUID) (Task, error) {
--- a/apps/control-plane/internal/agenttask/service.go
+++ b/apps/control-plane/internal/agenttask/service.go
@@ -176,8 +176,18 @@ func (s *Service) stopEngineSession(ctx context.Context, taskID uuid.UUID, sessi
 
 // WaitForLaunches blocks until every background launch this Service started
 // has finished. Nothing on the request path calls it; it exists so tests can
-// assert on a settled task, and so a graceful drain can let in-flight
-// launches record their outcome before the process exits.
+// assert on a settled task, and so a drain can let in-flight launches record
+// their outcome before the process exits.
+//
+// Deliberately not wired into cmd/server's shutdown: a launch is bounded at
+// launchTimeout (five minutes) and container stop is bounded far lower, so
+// waiting there would only trade a clean stop for a SIGKILL. A process that
+// dies mid-launch leaves its task queued with no engine_session_ref, which
+// Repository.ListActive excludes, so the poller never advances it. That
+// window is unchanged by making the launch asynchronous: the launch always
+// outlived the caller's context, so a restart always stranded it. A sweep for
+// stale queued rows is the fix if it ever shows up in practice; nothing has
+// asked for one yet (ponytail).
 func (s *Service) WaitForLaunches() { s.launches.Wait() }
 
 // Get returns one task, scoped to (tenantID, userID) so a task started by
--- a/apps/control-plane/internal/agenttask/service.go
+++ b/apps/control-plane/internal/agenttask/service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -15,6 +16,10 @@ import (
 type Service struct {
 	repo   Repository
 	engine Engine
+
+	// launches counts the background launch goroutines CreateTask starts.
+	// Nothing on the request path waits on it; see WaitForLaunches.
+	launches sync.WaitGroup
 }
 
 // NewService constructs a Service. repo must not be nil. A nil engine
@@ -50,12 +55,34 @@ const engineLaunchFailedMessage = "agent engine could not start the task"
 // anything, so this is minutes rather than seconds.
 const launchTimeout = 5 * time.Minute
 
-// CreateTask persists a new task and attempts to hand it to the agent-engine.
-// Any Launch error, including ErrEngineNotConfigured, transitions the task
-// straight to StatusFailed so a caller never polls a task that can never
-// progress: previously ErrEngineNotConfigured left the task in StatusQueued
-// forever with no signal, which is exactly the silent-stuck-queue behavior
-// this now closes.
+// engineCancelTimeout bounds one stop call to the engine. Deliberately well
+// under the 15 second budget edge-api's control-plane client allows
+// (apps/edge-api/internal/agenttask.NewClient): a cancel that outlived that
+// budget would answer the browser "failed" for a task that is in fact
+// cancelled, which is the shape issue #881 is about. Stopping a session is a
+// socket round trip plus a process kill, so seconds are generous; the
+// launcher frees the slot even if this call's context dies first.
+const engineCancelTimeout = 10 * time.Second
+
+// CreateTask persists a new task and hands it to the agent-engine on a
+// background goroutine, returning the persisted StatusQueued task as soon as
+// the row exists (issue #881).
+//
+// Waiting for the launch inline was the bug: a launch is bounded at
+// launchTimeout (five minutes) because a cold sandbox mount routinely takes
+// tens of seconds, while edge-api's control-plane client gives up after 15
+// seconds. Measured live on 2026-08-11, create answered the browser 500
+// after 18.0 seconds for a task that went on to reach succeeded, so the user
+// was told their work had failed while it was running and a retry started a
+// second sandbox. Widening the client timeout to match the server's five
+// minute bound was rejected: it makes an interactive request legitimately
+// able to hang for five minutes and leaves every intermediate proxy free to
+// cut it anyway. The queued row plus the existing poll path is what the code
+// was already shaped for, and it is what the console already renders.
+//
+// A launch failure still reaches the caller, just on the next poll rather
+// than in the create response: the background launch transitions the task to
+// StatusFailed, so a task never sits queued forever with no signal.
 func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pack Pack, instructions string) (Task, error) {
 	if !pack.Valid() {
 		return Task{}, ErrInvalidPack
@@ -66,36 +93,92 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
 		return Task{}, err
 	}
 
-	// Everything from here on runs on a context detached from the caller's.
-	// A launch cold-starts a sandbox, which takes tens of seconds, and the
-	// browser tab that submitted the task is free to go away in the middle
-	// of that. Measured on the demo box: closing the tab cancelled the
-	// request context, the launch died with "context canceled", and the task
-	// was recorded as failed for a reason that had nothing to do with the
-	// task. Worse, the Transition calls below would inherit the same dead
-	// context and fail too, leaving the row queued forever.
+	// The launch runs on a context detached from the caller's. A launch
+	// cold-starts a sandbox, which takes tens of seconds, and the browser tab
+	// that submitted the task is free to go away in the middle of that.
+	// Measured on the demo box: closing the tab cancelled the request
+	// context, the launch died with "context canceled", and the task was
+	// recorded as failed for a reason that had nothing to do with the task.
+	// Worse, the Transition calls would inherit the same dead context and
+	// fail too, leaving the row queued forever.
 	opCtx := context.WithoutCancel(ctx)
-	launchCtx, cancel := context.WithTimeout(opCtx, launchTimeout)
+	s.launches.Add(1)
+	go func() {
+		defer s.launches.Done()
+		s.launch(opCtx, t)
+	}()
+
+	return t, nil
+}
+
+// launch performs one launch attempt and records its outcome. Runs on a
+// background goroutine; it never returns anything to the caller of
+// CreateTask, so every failure path ends in either a persisted status or an
+// operator-visible log line.
+func (s *Service) launch(ctx context.Context, t Task) {
+	launchCtx, cancel := context.WithTimeout(ctx, launchTimeout)
 	defer cancel()
 
 	sessionRef, err := s.engine.Launch(launchCtx, t)
 	switch {
 	case err == nil:
-		return s.repo.Transition(opCtx, tenantID, userID, t.ID, StatusRunning, sessionRef, "", "")
+		if _, terr := s.repo.Transition(ctx, t.TenantID, t.UserID, t.ID, StatusRunning, sessionRef, "", ""); terr != nil {
+			// The session started but its reference could not be recorded, so
+			// nothing can ever poll or stop it and it would hold its
+			// concurrency slot for its full natural life. Stop it here.
+			// ErrTerminalState is the expected case: a cancel won the race
+			// while this launch was still in flight, and it had no session
+			// reference to stop at the time, so this goroutine owns the
+			// teardown.
+			if !errors.Is(terr, ErrTerminalState) {
+				slog.Default().WarnContext(ctx, "agenttask: could not record a launched session, stopping it",
+					"task_id", t.ID, "error", terr)
+			}
+			s.stopEngineSession(ctx, t.ID, sessionRef)
+		}
 	case errors.Is(err, ErrEngineNotConfigured):
-		return s.repo.Transition(opCtx, tenantID, userID, t.ID, StatusFailed, "", "", engineUnavailableMessage)
+		s.recordLaunchFailure(ctx, t, engineUnavailableMessage)
 	default:
-		slog.Default().WarnContext(opCtx, "agenttask: launch failed, engine detail",
+		slog.Default().WarnContext(ctx, "agenttask: launch failed, engine detail",
 			"task_id", t.ID, "error", err)
-		return s.repo.Transition(opCtx, tenantID, userID, t.ID, StatusFailed, "", "", engineLaunchFailedMessage)
+		s.recordLaunchFailure(ctx, t, engineLaunchFailedMessage)
+	}
+}
+
+// recordLaunchFailure persists a failed launch. ErrTerminalState is expected
+// and ignored: the user cancelled while the launch was failing, and their
+// cancellation is the truthful record.
+func (s *Service) recordLaunchFailure(ctx context.Context, t Task, message string) {
+	if _, err := s.repo.Transition(ctx, t.TenantID, t.UserID, t.ID, StatusFailed, "", "", message); err != nil &&
+		!errors.Is(err, ErrTerminalState) {
+		slog.Default().WarnContext(ctx, "agenttask: could not record a failed launch",
+			"task_id", t.ID, "error", err)
+	}
+}
+
+// stopEngineSession stops one launcher session, which is what releases the
+// concurrency slot that session holds (issue #886). Errors are logged and
+// never returned: by the time this runs the task's terminal state is already
+// recorded, and a failure here is an operator problem (a leaked slot), not
+// something the customer can act on or should see as a failed request. The
+// engine detail goes to the log only, never to a customer-visible field.
+func (s *Service) stopEngineSession(ctx context.Context, taskID uuid.UUID, sessionRef string) {
+	if sessionRef == "" {
+		return // never launched, so nothing holds a slot
+	}
+	stopCtx, done := context.WithTimeout(context.WithoutCancel(ctx), engineCancelTimeout)
+	defer done()
+	if err := s.engine.Cancel(stopCtx, sessionRef); err != nil {
+		slog.Default().WarnContext(stopCtx, "agenttask: engine cancel failed, the session may still hold its concurrency slot",
+			"task_id", taskID, "error", err)
 	}
 }
 
 // WaitForLaunches blocks until every background launch this Service started
 // has finished. Nothing on the request path calls it; it exists so tests can
-// assert on a settled task, and so a future graceful drain can let in-flight
+// assert on a settled task, and so a graceful drain can let in-flight
 // launches record their outcome before the process exits.
-func (s *Service) WaitForLaunches() {}
+func (s *Service) WaitForLaunches() { s.launches.Wait() }
 
 // Get returns one task, scoped to (tenantID, userID) so a task started by
 // one user is never resumable by a different user in the same tenant.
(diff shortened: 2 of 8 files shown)


apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +93,103 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+// Deliberately not wired into cmd/server's shutdown: a launch is bounded at
+// launchTimeout (five minutes) and container stop is bounded far lower, so
+// waiting there would only trade a clean stop for a SIGKILL. A process that
+// dies mid-launch leaves its task queued with no engine_session_ref, which
+// Repository.ListActive excludes, so the poller never advances it. That
+// window is unchanged by making the launch asynchronous: the launch always
+// outlived the caller's context, so a restart always stranded it. A sweep for
+// stale queued rows is the fix if it ever shows up in practice; nothing has
+// asked for one yet (ponytail).
+func (s *Service) WaitForLaunches() { s.launches.Wait() }
```
sakibsadmanshajib: A process that dies between `repo.Create` and the launch goroutine's transition leaves the row `queued` with an empty `engine_session_ref`, and `Repository.ListActive` filters exactly those out, so the poller never advances it and the task is stuck forever with no signal. That is the same silent-stuck-queue shape this package already fixed once.

Rebuttal, and why it is not a regression: the launch has always run on a context detached from the caller's, so a restart mid-launch always stranded the row the same way. Making the call asynchronous does not widen the window, it only changes who waits during it. The lazy fix, if it ever shows up in practice, is a sweep for queued rows older than `launchTimeout`, not a shutdown wait: `WaitForLaunches` would block a container stop for up to five minutes and trade a clean exit for a SIGKILL. Documented on `WaitForLaunches` rather than built.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -104,12 +203,36 @@ func (s *Service) List(ctx context.Context, tenantID, userID uuid.UUID) ([]Task,
...
 func (s *Service) Cancel(ctx context.Context, tenantID, userID, id uuid.UUID) (Task, error) {
-	return s.repo.Transition(ctx, tenantID, userID, id, StatusCancelled, "", "", "")
+	cancelled, err := s.repo.Transition(ctx, tenantID, userID, id, StatusCancelled, "", "", "")
+	if err != nil {
+		return Task{}, err
+	}
+	// An empty EngineSessionRef means the launch has not reported back yet;
+	// the in-flight launch goroutine sees this task is already terminal and
+	// tears its own session down (see launch).
+	s.stopEngineSession(ctx, id, cancelled.EngineSessionRef)
```
sakibsadmanshajib: The stop call is inline in the cancel request, so a slow launcher adds up to `engineCancelTimeout` (10 seconds) to the response. Worth checking against the very defect this PR fixes: does a slow cancel reproduce issue #881's shape, a failure reported for work that actually happened?

Answer from the code: no. 10 seconds is under edge-api's 15 second client budget, so the caller gets the real answer, and an engine error is logged rather than returned, so the response is 200 with the cancelled row regardless. The task is genuinely cancelled at that point; only the launcher-side teardown is in doubt, and the launcher frees the slot in `reap` even when the interrupt call's context dies first (`SandboxEngine.Cancel` reaps unconditionally after `Interrupt`). Backgrounding the stop was considered and rejected: it buys latency this path does not need and costs the deterministic ordering the tests rely on.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +93,103 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+	// cold-starts a sandbox, which takes tens of seconds, and the browser tab
+	// that submitted the task is free to go away in the middle of that.
+	// Measured on the demo box: closing the tab cancelled the request
+	// context, the launch died with "context canceled", and the task was
+	// recorded as failed for a reason that had nothing to do with the task.
+	// Worse, the Transition calls would inherit the same dead context and
+	// fail too, leaving the row queued forever.
 	opCtx := context.WithoutCancel(ctx)
-	launchCtx, cancel := context.WithTimeout(opCtx, launchTimeout)
+	s.launches.Add(1)
```
sakibsadmanshajib, line 133: One goroutine per create, unbounded by anything in this package. A caller looping on create spawns one per request.

Accepted: each goroutine's first act is `engine.Launch`, which acquires a quota slot and returns `ErrUserQuotaExceeded` immediately once the ceiling is reached, so beyond that ceiling these are short lived. The real limit on creation rate is the database insert in front of them. A worker pool here would be an abstraction with one producer and no measured pressure.

apps/control-plane/internal/agenttask/service_test.go
```diff
@@ -88,12 +104,149 @@ func (f *fakeRepository) ListActive(context.Context) ([]agenttask.Task, error) {
...
+	defer f.mu.Unlock()
+	return len(f.cancelled)
+}
+
+// errNoSlot stands in for apps/agent-engine/internal/quota's
+// ErrUserQuotaExceeded, which control-plane cannot import across the module
+// boundary (that package is under apps/agent-engine/internal).
+var errNoSlot = errors.New("test engine: user concurrency limit reached")
+
+// slotEngine models the launcher's per-user concurrency accounting
```
sakibsadmanshajib, line 145: `slotEngine` models the launcher's quota accounting rather than using it, which is the standard way a test double lies: it can stay green while the real accounting is broken.

Why it is still the right assertion here, stated plainly so a reader can check it: the defect in issue #886 is entirely on the control-plane side of the seam, `Service.Cancel` never calling the engine at all, and this double is the smallest thing that detects that (a status-code assertion cannot, which is how the bug survived). The launcher side is covered separately and for real by `TestSandboxEngine_Cancel_FreesQuotaSlot`, which asserts `quota.Manager.InUse` returns to zero and that a relaunch then succeeds. Control-plane cannot import `apps/agent-engine/internal/quota` directly: Go's internal visibility does not cross the module boundary, the same limitation `engineapi`'s doc comment covers.

sakibsadmanshajib (commented):
Adversarial review, plain pass (no specialised tool), ticket first. Read both issues, then the diff, then hunted for the shapes that would make this fix a second bug. Four findings, each posted inline. Two are accepted risks with the reasoning stated, two are questions answered against the code rather than against intent.

Stream status for this PR: CodeRabbit CLI SKIPPED. `coderabbit review --committed --base main --agent` (v0.7.2) returned `Review failed: You are not a member of the requested organization` (FORBIDDEN, httpStatus 403). That is an unavailable stream, not a clean pass.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Resolving as a reasoned rebuttal, no code change.

Checked against the code rather than intent: the launch has always run on `context.WithoutCancel(ctx)`, so a process death mid-launch has always stranded the row in `queued` with no `engine_session_ref`, and `Repository.ListActive` has always filtered those out. Making the call asynchronous moves who waits, not the window. So this is a pre-existing hole, not one this PR opens, and closing it belongs to whoever wants the sweep.

What did change here: the behaviour is now written down at the point someone will look for it, in `WaitForLaunches`'s doc comment, including why a shutdown wait is the wrong fix (a launch is bounded at five minutes, container stop is bounded far lower, so waiting trades a clean exit for a SIGKILL). The fix if it ever shows up in practice is a sweep for `queued` rows older than `launchTimeout`, which needs a change to the `agent_tasks_list_active()` SECURITY DEFINER function and therefore a migration. Out of scope for a two-defect fix, and unmeasured today.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Resolving as a reasoned rebuttal, no code change.

The numbers close it: `engineCancelTimeout` is 10 seconds, edge-api's client budget is 15 seconds, so the caller always gets the real answer rather than a client-side timeout. The response is 200 with the cancelled row whatever the engine does, because an engine error is logged for the operator and never returned, so a slow or broken launcher cannot produce issue #881's shape here (a stated failure for work that happened).

The slot is safe on the pessimistic path too: `SandboxEngine.Cancel` calls `reap` unconditionally after `Interrupt`, and `reap` calls `sess.release()` without consulting the context, so even a stop call whose context expires still frees the slot on the launcher side.

Backgrounding the stop was the alternative. Rejected: it buys latency this path does not need, and it would make the cancel tests non-deterministic for no gain.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Resolving as accepted with the reasoning stated, no code change.

Traced the worst case: every goroutine's first act is `engine.Launch`, which acquires a quota slot first (`SandboxEngine.Launch` calls `q.Acquire` before it touches the filesystem or the sandbox) and returns `ErrUserQuotaExceeded` or `ErrTenantQuotaExceeded` immediately once the ceiling is reached. Beyond the ceiling, which is 2 per user and 4 per tenant by default, these goroutines are short lived and do no work. The rate limit on creation is the database insert in front of them, which the caller waits on.

A worker pool here would be an abstraction with one producer, no measured pressure, and a new queue to reason about during an incident. If create-rate abuse ever becomes real, the honest fix is a rate limit at the edge, not a pool inside this service.

apps/control-plane/internal/agenttask/service_test.go
sakibsadmanshajib: Resolving with the evidence rather than an assurance.

The concern is right in general, so here is why the pair of tests cannot both be vacuous. The defect in issue #886 lives entirely on the control-plane side of the seam: `Service.Cancel` never called the engine at all. `slotEngine` detects exactly that, and it was watched red before the fix with `cancel did not release the launcher slot: 1 still in use`, plus `expected the engine session to be cancelled exactly once, got 0 calls` on two sibling tests. A double lying in the other direction (passing while the wiring is broken) is not possible for those assertions, because the only thing that moves its counter is a real call from `Service`.

The launcher side is covered for real by `TestSandboxEngine_Cancel_FreesQuotaSlot`, which uses the actual `quota.Manager`, asserts `InUse` returns to 0 for both tenant and user, and then relaunches to prove the freed slot is reusable. That one was green before the fix, stated plainly in the PR body: the launcher already released the slot in `reap`, nobody was calling it.

Control-plane cannot use the real manager directly: `apps/agent-engine/internal/quota` is under another module's `internal`, and Go's internal visibility does not cross module boundaries. `engineapi` exists precisely because of that limit and re-exports only the engine, not the quota manager.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +93,103 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+			// concurrency slot for its full natural life. Stop it here.
+			// ErrTerminalState is the expected case: a cancel won the race
+			// while this launch was still in flight, and it had no session
+			// reference to stop at the time, so this goroutine owns the
+			// teardown.
+			if !errors.Is(terr, ErrTerminalState) {
+				slog.Default().WarnContext(ctx, "agenttask: could not record a launched session, stopping it",
+					"task_id", t.ID, "error", terr)
+			}
+			s.stopEngineSession(ctx, t.ID, sessionRef)
```
sakibsadmanshajib: **Silent stuck-queue regression, not covered by the new tests.**

In `launch`'s `case err == nil` branch (lines 124-138): when `Launch` succeeds but the follow-up `Transition(..., StatusRunning, sessionRef, ...)` fails for any reason *other* than `ErrTerminalState` (a real DB error: pool exhaustion, timeout, connectivity blip), the code logs a WARN and calls `stopEngineSession` to kill the freshly-launched session — but it never calls `recordLaunchFailure`. The task row is left exactly where it was: `StatusQueued`, `EngineSessionRef` still empty.

`Repository.ListActive` only returns rows that are `queued`/`running` **and** have a non-empty `engine_session_ref`. This row has neither a session (just killed) nor a path back to the poller, so nothing will ever touch it again. The customer sees "queued" forever with no `error_message`, only an operator-only log line.

This is the exact silent-stuck-queue bug class `engineUnavailableMessage`'s doc comment says was closed — reopened here for the "successful launch, transition failed" case specifically. It's also a regression versus pre-#881 behavior: previously this same `Transition` failure was returned synchronously from `CreateTask` as an HTTP 500, so at least *something* signaled failure to the caller. Now: 201 Created, silently stuck.

None of the new tests exercise this path — `fakeRepository.Transition` never returns a non-`ErrTerminalState`, non-`ErrNotFound` error, so this branch has zero coverage either way.

Fix: on a non-terminal `terr`, also call `recordLaunchFailure` (or equivalent) so the task lands in `StatusFailed` instead of stranded in `StatusQueued`.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +93,103 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+	// context, the launch died with "context canceled", and the task was
+	// recorded as failed for a reason that had nothing to do with the task.
+	// Worse, the Transition calls would inherit the same dead context and
+	// fail too, leaving the row queued forever.
 	opCtx := context.WithoutCancel(ctx)
-	launchCtx, cancel := context.WithTimeout(opCtx, launchTimeout)
+	s.launches.Add(1)
+	go func() {
+		defer s.launches.Done()
+		s.launch(opCtx, t)
```
sakibsadmanshajib: **No panic recovery in the new background goroutine.**

`go func() { defer s.launches.Done(); s.launch(opCtx, t) }()` has no `recover()`. Before #881, this exact call chain (`engine.Launch` -> `repo.Transition`) ran inside the HTTP handler's own goroutine, so a panic anywhere in it (a bug in a future `Engine` implementation, a nil-deref in `repo.Transition`, anything transitively called) was caught by `net/http`'s per-connection recover: it logs, closes that one connection, and the process survives.

Moving the same code into a self-spawned goroutine strips that safety net silently. A panic here now takes down the *entire* control-plane process — every tenant, every in-flight request — not just the one task. This is a straight regression in blast radius, introduced by the refactor itself, independent of whether today's code happens to panic.

Add a `defer func() { if r := recover(); r != nil { slog...; s.recordLaunchFailure(...) } }()` at the top of the goroutine (or inside `launch`), so one bad launch can't crash the process.

Separately: nothing bounds how many of these goroutines can be in flight. `quota.Manager.Acquire` (agent-engine side) gates the actual sandbox launch, but it runs *inside* this goroutine, after `repo.Create`'s synchronous DB INSERT has already happened on the request path. A tight loop of creates against an already-full quota still costs one DB row + one goroutine per request; quota gates the launch attempt, not the request or the goroutine spawn.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +93,103 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+// has finished. Nothing on the request path calls it; it exists so tests can
+// assert on a settled task, and so a drain can let in-flight launches record
+// their outcome before the process exits.
+//
+// Deliberately not wired into cmd/server's shutdown: a launch is bounded at
+// launchTimeout (five minutes) and container stop is bounded far lower, so
+// waiting there would only trade a clean stop for a SIGKILL. A process that
+// dies mid-launch leaves its task queued with no engine_session_ref, which
+// Repository.ListActive excludes, so the poller never advances it. That
+// window is unchanged by making the launch asynchronous: the launch always
```
sakibsadmanshajib: **"Window is unchanged" is not accurate — verified against `cmd/server/main.go`.**

`WaitForLaunches` is correct that it's never called anywhere (grepped `main.go`: zero references). But the claim that the SIGTERM-kill window for an in-flight launch is "unchanged" by making launch asynchronous doesn't hold up:

Before: `CreateTask` blocked inline inside the HTTP handler goroutine. `srv.Shutdown(shutdownCtx)` (10s budget, `main.go` line ~1370) tracks *that* goroutine as an active connection and waits on it — up to the full 10s — before `Shutdown` returns an error and `log.Fatalf` forces exit. An in-flight launch got up to 10s of grace.

Now: the handler returns as soon as the row is persisted, so `srv.Shutdown` sees no outstanding connection tied to the launch at all and returns almost immediately, regardless of what the detached goroutine is doing. `main()` then proceeds straight to `log.Println("control-plane stopped")` and exits. The background goroutine — and any `stopEngineSession` call a racing `Cancel` triggered inside it — gets killed with effectively zero grace, not "up to 10s."

Not asking for `WaitForLaunches()` to be wired into `Shutdown` (the doc's own argument against that, trading a clean stop for a SIGKILL, is reasonable). But the comment should say what actually changed — grace shrank to ~0 — rather than "unchanged," so the next person doesn't reason from a false premise when this surfaces as a mid-deploy stuck task.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: **Cancel stayed synchronous, reintroducing a smaller version of the #881 timeout mismatch.**

`Create` was made async specifically because a blocking launch (up to `launchTimeout`, 5min) could outlive edge-api's 15s client timeout. `Cancel` gets no such treatment: `stopEngineSession` runs inline in the request handler, blocking up to `engineCancelTimeout` (10s) on top of whatever `repo.Transition` takes.

10s is "deliberately well under" 15s per the doc comment, but that leaves ~5s of margin for the DB round trip plus network/scheduling jitter — on a box already busy enough to be exhausting quota slots (the scenario this PR exists to fix). If `engine.Cancel` takes close to its full 10s budget (a slow socket interrupt + process kill under load) and the DB write adds a few seconds of contention, total latency can still exceed edge-api's 15s, producing the exact same symptom class #881 fixed: browser told the request failed for an operation that server-side actually succeeded — just for cancel instead of create.

Given `stopEngineSession` already logs-and-swallows its own errors (the cancellation is recorded either way), consider firing it on its own goroutine here too, the same shape as `launch`, rather than blocking the handler on it.

apps/agent-engine/internal/engine/engine_test.go
```diff
@@ -491,6 +491,37 @@ func TestSandboxEngine_Cancel_CleansUpDirs(t *testing.T) {
 	assertDirEmpty(t, e.cfg.WorkspaceRoot)
 }
 
+// The launcher half of issue #886. Control-plane now calls Cancel when a task
+// is cancelled, and the only reason that helps is this: the cancelled
+// session's concurrency slot has to go back to the pool immediately, not
+// whenever the sandbox happens to finish on its own (roughly sixteen minutes
+// on the demo box, which is how two cancels exhausted a user's ceiling).
+func TestSandboxEngine_Cancel_FreesQuotaSlot(t *testing.T) {
```
sakibsadmanshajib, line 506: **This test cannot regress on this PR's actual fix — confirmed, not just per the PR's own caveat.**

`apps/agent-engine/internal/engine/engine.go` has zero lines changed in this diff (verified: only this test file changed under `apps/agent-engine/...`). `TestSandboxEngine_Cancel_FreesQuotaSlot` calls `e.Launch` / `e.Cancel` on `SandboxEngine` directly — pre-existing, untouched code (`SandboxEngine.Cancel` -> `reap` -> `sess.release()`) — and `newTestEngineWithQuota` is a pre-existing helper (used by tests before this PR too).

The actual fix this PR ships is `apps/control-plane/internal/agenttask.Service.Cancel` now calling `Engine.Cancel` (which didn't exist on the `agenttask.Engine` interface at all before this PR) via `apps/control-plane/internal/agentengine.Engine`. Revert *that* — remove `Cancel` from the `agenttask.Engine` interface, remove the `s.stopEngineSession` call from `Service.Cancel` — and every single assertion in this test still passes unchanged, because the test never touches `agenttask.Service`, `agenttask.Engine`, or `agentengine.Engine` at all.

Every assertion here (`e.q.InUse(...) == 0/0`, the follow-up `e.Launch` succeeding) is testing `quota.Manager` release semantics on a direct `SandboxEngine.Cancel` call, which is real and fine to have, but it is not a regression test for issue #886's actual bug (a *cancelled task via `Service.Cancel`* never told the engine to stop). The regression coverage for that lives in `apps/control-plane/internal/agenttask/service_test.go`'s `TestService_Cancel_ReleasesEngineConcurrencySlot`, `TestService_CancelDuringLaunch_ReleasesTheSlotThatLaunchTook`, and `TestService_Cancel_LosesRaceWithCompletion_LeavesEngineAlone` — those do fail on revert (verified: they call `svc.Cancel` and assert on `slotEngine` counters that only move if `Service.Cancel` invokes `engine.Cancel`).

Doc comment's own "the launcher half of issue #886" framing undersells this: it's not half a regression test, it's a same-package companion test for already-shipped, already-correct code, mislabeled as covering this PR's change.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Conceded and fixed in 6a0b6c88. You are right on every step of the trace, including that it is worse than the 500 it replaced.

`launch` now stops the session first, then returns early only for `ErrTerminalState` (where the task's terminal state is already correct because a cancel won the race). Any other error is a real database failure, and the row now goes to `StatusFailed` through `recordLaunchFailure` with the sanitized `engineLaunchFailedMessage`, so the task ends visibly failed instead of stranded in `queued` with no `engine_session_ref`. Best effort by definition, since the same database just refused a write, and the WARN naming the underlying error is kept for the operator.

Coverage added, and watched red first. `fakeRepository` gained `failTransitionTo` plus `failTransitionErr`, which is what made this branch reachable at all: before it, `Transition` could only ever return `ErrTerminalState` or `ErrNotFound`, exactly as you observed. `TestService_LaunchSucceedsButTransitionFails_TaskFailsVisibly` fails the transition to `running` with `pgx: connection pool exhausted` and asserts the task lands `StatusFailed` with a non-empty `error_message` that carries neither `pgx` nor `pool`, and that the launcher slot is back to zero. With the `recordLaunchFailure` call removed it fails; with it, it passes.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Conceded and fixed in 6a0b6c88. The blast radius argument is correct and the point about it being independent of whether today's code panics is the right way to frame it: this was a safety net removed silently by a refactor, not a hypothetical bug.

Both background paths (the launch, and now the cancel's engine stop) go through one `background(work, onPanic)` helper. It recovers, logs at ERROR with the panic value and `debug.Stack()`, and for the launch runs `recordLaunchFailure`, so a panicking `Engine` costs one failed task instead of the process and every tenant on it.

Watched red: with the recover replaced by a re-panic, `TestService_LaunchPanic_DoesNotCrashTheProcess` takes the test binary down with `panic: engine implementation bug [recovered]`. With the helper in place it passes and the task reads `StatusFailed` with a customer-visible message.

On the second half, you are right that my earlier rebuttal answered a different question than the one asked. Quota gates the sandbox launch, inside the goroutine, after `repo.Create`'s synchronous INSERT has already committed on the request path, so it bounds neither the row nor the goroutine. Leaving it as is deliberately: the cost per refused create is one row plus a short lived goroutine that returns as soon as `Acquire` refuses, and the honest place to bound request volume is a rate limit at the edge rather than a worker pool inside this service. Flagging rather than hiding it.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Verified independently against `cmd/server/main.go` before conceding: `shutdownCtx` is `context.WithTimeout(context.Background(), 10*time.Second)` at line 1365, `srv.Shutdown(shutdownCtx)` at 1370, then `log.Println("control-plane stopped")` and return. Your reading is right. Previously the handler goroutine was an outstanding connection and `Shutdown` waited on it, up to the full 10 seconds; now the handler returns as soon as the row is persisted, so `Shutdown` has nothing to wait for and the detached goroutine dies with approximately zero grace. My comment said the window was unchanged, and that was false.

Corrected in 6a0b6c88 rather than softened. `WaitIdle`'s doc now states the grace went from up to 10 seconds to approximately none, why waiting on `WaitIdle` inside `Shutdown` is still the wrong fix (a launch is bounded at five minutes, container stop far lower, so it trades a clean exit for a SIGKILL at the same point), and what the real fix is when a mid-deploy stranded task actually appears: a sweep failing queued rows older than `launchTimeout`, which needs a change to `agent_tasks_list_active()` and therefore a migration.

Thanks for checking the claim rather than the comment. It is exactly the kind of premise the next person would have reasoned from.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Conceded and fixed in 6a0b6c88, taking your suggested shape. My earlier rebuttal treated 10 against 15 as a margin; your point is that the remaining 5 seconds has to absorb the database write plus scheduling jitter on precisely the loaded box this pull request exists for, which is the same reasoning that made create asynchronous.

`Service.Cancel` now commits the transition, then fires `stopEngineSession` through the same `background` helper the launch uses, on `context.WithoutCancel(ctx)` so a client disconnect cannot leave the sandbox alive. The handler's latency is now the database write alone. Nothing about correctness moves: the cancellation is already committed when the response is written, and the stop already logged and swallowed its own errors, so there was never a caller-visible outcome depending on it.

Determinism in the tests is preserved by `WaitIdle`, which is why `WaitForLaunches` was renamed: it now covers both background paths rather than launches only. Every cancel assertion (`slotsInUse`, `cancelCount`) waits on it first, so the ordering guarantees the earlier tests relied on are unchanged.

`engineCancelTimeout` stays at 10 seconds, since it no longer sits in a request budget and the launcher frees the slot in `reap` regardless of whether that context expires.

apps/agent-engine/internal/engine/engine_test.go
sakibsadmanshajib: Conceded. Reporting it as green from the start was not enough: a test that survives a full revert of the change should not be presented as coverage at all, and calling it "the launcher half of issue #886" implied it carried weight it does not.

Fixed in 6a0b6c88. The doc comment now opens by saying it is a characterisation test for pre-existing behaviour and explicitly not regression coverage for this fix, states that every assertion passes with the fix reverted, and names `TestService_Cancel_ReleasesEngineConcurrencySlot` and its two siblings as the tests that do fail on revert. Keeping the test itself: it pins the invariant control-plane now depends on, which is that ending a session frees its slot immediately, and that invariant deserves a guard even though nothing in that package changed here.

The pull request body's coverage claim is being corrected the same way, so the citation does not survive in the merge commit message either.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +94,148 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+	go func() {
+		defer s.launches.Done()
+		defer func() {
+			r := recover()
+			if r == nil {
+				return
+			}
+			slog.Default().Error("agenttask: recovered panic in background work",
+				"panic", r, "stack", string(debug.Stack()))
+			if onPanic != nil {
```
sakibsadmanshajib: **HIGH. A panic inside `onPanic` escapes this recover and kills the process, which is the exact outcome the helper was added to prevent.**

`onPanic()` is invoked from inside the same deferred closure that just called `recover()`. At that point the recover has already completed and the goroutine is unwinding normally, so a panic raised by `onPanic` is a fresh panic in a deferred function with no recover above it. The remaining defer (`s.launches.Done()`) runs and then the process dies, every tenant with it.

This is not theoretical here, because of what `onPanic` actually is. `CreateTask` passes `func() { s.recordLaunchFailure(opCtx, t, engineLaunchFailedMessage) }`, and `recordLaunchFailure` calls `s.repo.Transition`. The most likely reason `s.launch` panicked in the first place is a fault in that same repository layer (a nil pool after a reconnect path, a nil-deref inside a driver call, a nil `repo` from a partially wired `NewService`). The recovery handler therefore re-enters the code that just panicked, on the same arguments, and the second panic is unrecoverable. The helper converts "one task dies" into "the process dies" precisely in the case where the panic is deterministic rather than a one-off.

The fix is one wrapper and keeps the same shape:

```go
if onPanic != nil {
    func() {
        defer func() {
            if r2 := recover(); r2 != nil {
                slog.Default().Error("agenttask: panic in background panic handler, giving up on this task",
                    "panic", r2)
            }
        }()
        onPanic()
    }()
}
```

A test that would have caught it: an engine whose `Launch` panics wired to a repository whose `Transition` also panics, asserting the process survives. That pair is the realistic shape of a shared-dependency fault, and `TestService_LaunchPanic_DoesNotCrashTheProcess` as written cannot reach it because its repository is healthy.

Separate from CodeRabbit's finding on the same helper (the lost `sessionRef` on the panic path). Both are true and neither subsumes the other.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +94,148 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+
+// launch performs one launch attempt and records its outcome. Runs on a
+// background goroutine; it never returns anything to the caller of
+// CreateTask, so every failure path ends in either a persisted status or an
+// operator-visible log line.
+func (s *Service) launch(ctx context.Context, t Task) {
+	launchCtx, cancel := context.WithTimeout(ctx, launchTimeout)
 	defer cancel()
 
 	sessionRef, err := s.engine.Launch(launchCtx, t)
```
sakibsadmanshajib, line 210: **HIGH. "Launched but never recorded" is handled for a database failure and not for the transport failure, which is the likelier one, and the leaked slot in that case is permanent rather than sixteen minutes.**

The block below (lines 152 to 174) is careful about the case where `Launch` succeeds and `Transition` then fails: it stops the session so the slot comes back. There is a second way to reach the same state and it is not covered. If the launcher registers the session and control-plane never receives or never survives the response, `Launch` returns an error here, the `default:` branch runs, the task is recorded `failed`, and control-plane never learns the session reference at all. Three ways in:

1. `launchTimeout` is 5 minutes and `Remote`'s own `http.Client.Timeout` is also 5 minutes (`apps/control-plane/internal/agentengine/remote.go` line 55). Whichever fires first, control-plane abandons the request while the launcher may be inside `writeJSON` or a few instructions before it.
2. Any transport error after the launcher's handler returned.
3. The process dying between the launcher registering the session (`apps/agent-engine/internal/engine/engine.go` lines 424 to 429) and this `Transition` committing. `deploy-demo-box.yml` recreates control-plane on every merge to `main`, so this window is entered on a schedule.

What is left behind is worse than the sixteen minutes issue #886 measured, because nothing reaps it on a timer:

- The sandbox is live and registered in `SandboxEngine.sessions`, holding one tenant slot and one user slot.
- `reap` runs only from `Status` or `Cancel`. `Status` is only ever called by the poller, and the poller's input is `agent_tasks_list_active()`, whose `WHERE` clause requires `engine_session_ref <> ''`. The row has no reference, so it is invisible to the poller forever.
- The agent-server does not exit when its conversation finishes. Your own comment at `engine.go` line 475 says so: "Nothing else ever frees this session: the agent-server is a server and keeps running after its conversation finishes." So the process does not fall over on its own either.

The slot is therefore held until the launcher process restarts, not for the sandbox's natural life. With `HIVE_QUOTA_USER_CONCURRENCY` at 2, two of these lock that user out of Cowork entirely, which is the exact user-visible symptom this pull request exists to close, reached by a different route.

The future fix named in `WaitIdle`'s comment (a sweep that fails queued rows older than `launchTimeout`) does not address this. That sweep can fix the row, and the row is the half that costs nothing. It cannot release the slot, because the row it is sweeping has no session reference to cancel with.

What would actually close it is a way to re-derive the reference. The cheapest is on the launcher, which already holds the mapping: a lookup by task id (`GET /session-by-task/{id}`) or an orphan sweep that reaps sessions whose task never came back for them. Either lets control-plane recover a session it lost the receipt for. Out of scope for this pull request is a defensible answer, but the current comment presents the launched-but-unrecorded case as closed, and it is closed for one of its two causes.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +94,148 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+	// cold-starts a sandbox, which takes tens of seconds, and the browser tab
+	// that submitted the task is free to go away in the middle of that.
+	// Measured on the demo box: closing the tab cancelled the request
+	// context, the launch died with "context canceled", and the task was
+	// recorded as failed for a reason that had nothing to do with the task.
+	// Worse, the Transition calls would inherit the same dead context and
+	// fail too, leaving the row queued forever.
 	opCtx := context.WithoutCancel(ctx)
-	launchCtx, cancel := context.WithTimeout(opCtx, launchTimeout)
+	s.background(func() { s.launch(opCtx, t) }, func() {
```
sakibsadmanshajib: **HIGH. The "rate limiting belongs at the edge" rebuttal rests on a control that does not exist on this route today, and the goroutine's cost is not what the earlier reply measured.**

The earlier resolution accepted the unbounded goroutine on the grounds that quota rejects fast past the ceiling and that creation rate is the edge's problem. I traced the edge. There is no rate limiter in front of `POST /v1/agent/tasks`:

- `apps/edge-api/cmd/server/main.go` lines 538 to 555 build the entire chain: `UnsupportedEndpointMiddleware`, `budgetGate.Wrap`, `authSelectorMiddleware(jwtMW, ...)`, `proxy.InstrumentHandler`, `CompatHeaders`, `MaxBytesHandler`. No limiter.
- `authz.Limiter` is only reached through `authorizer.Authorize`. `apps/edge-api/internal/agenttask/handler.go` never calls it. It authenticates with `auth.UserFrom(r.Context())` and nothing else.
- `budgetGate` is documented inert for exactly this traffic at main.go lines 540 to 546: "budgetGate resolves the workspace identity from the API-key bearer token via authzClient.Resolve, so it stays inert for JWT-authenticated traffic." Cowork is JWT-session traffic.
- `registerAgentTaskRoutes` adds `featuregate.Require(FeatureCowork)` and nothing else.

So the only thing an authenticated Cowork user meets on the way in is a feature gate. Rate limiting on this path is aspirational, and the acceptance was made against a control that is not shipped.

What a burst actually buys, per request, none of it bounded:

- One `agent_tasks` row carrying up to 64 KiB of `instructions` (`maxBodyBytes` in `http.go`, and the same 64 KiB reader at the edge handler).
- One goroutine that outlives the request and retains the whole `Task`, including that 64 KiB string, until it finishes.
- Two control-plane pool checkouts rather than one: the `Create` inside the request, and then a `Transition` from the goroutine after the response has been written. The second one carries no request-level backpressure at all. Before this change both happened inside the handler, so in-flight work was at least tied to the connection; now the queue of pending database work is decoupled from anything that could shed load. The control-plane pool is small and shared with everything else control-plane serves, so a create loop is a credible way to starve unrelated traffic on the same box.

Quota does not bound any of that. It gates the sandbox launch, which is the third thing that happens, and the reply is right that it rejects quickly. The rows, the goroutines and the database work are all upstream of it.

This matters more than an ordinary unbounded-goroutine nit because `HIVE_QUOTA_TENANT_CONCURRENCY` and `HIVE_QUOTA_USER_CONCURRENCY` are the only tenant-isolation control on the sandbox path. Everything that reaches the box in front of them is per-tenant unbounded, so the isolation property is "one tenant cannot hold more than four sandboxes" and not "one tenant cannot consume the box."

Smallest honest fix inside this diff: a bounded worker pool or a semaphore around the launch goroutine, refusing the create with 429 when it is full. The second smallest: keep the goroutine and file an issue for a limiter on `/v1/agent/tasks`, and change the comment so it does not read as if that limiter already exists. What should not survive review is the current wording, which points at a defence that is not there.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Conceded and fixed in 772c25d9. The escalation argument is the important part and it is right: the helper converted "one task dies" into "the process dies" exactly in the deterministic case, because `onPanic` re-entered the repository layer that is the most likely source of the original panic.

Fixed structurally rather than with a wrapper at the call site. `onPanic` is gone. `background` now carries a plain recover as the outer net, and `launch` owns its own recovery, which is where it belongs since only `launch` holds the session reference (that is CodeRabbit's finding on the same helper, and you were right that neither subsumes the other). Both calls inside that recovery run through `safely(what, f)`, which recovers and logs, so a second panic from either the stop or the failure write cannot escape.

Order inside the recovery is deliberate: stop the session first, because freeing the slot does not touch the database, which is the dependency most likely to be the thing that panicked. Recording the failure comes second and is allowed to fail.

Your suggested test is the one added, in the shape you described. `panickingRepository` with an empty `panicOn` panics on every `Transition`, wired to a working engine, so the recovery handler panics on the same fault as the original. Watched red first: `panic: repository layer fault [recovered]` took the test binary down. It now passes and asserts the slot is still freed, which is the part that survives a database that cannot be written to at all.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Conceded on the substance, and the comment that presented the case as closed is fixed in 772c25d9. Filed as issue #899 with your full trace rather than left in a thread.

Checked each step before conceding. `reap` really does run only from `Status` or `Cancel`; `Status` really is only called by the poller; the poller's input really does require `engine_session_ref <> ''`; and the agent-server really does keep running after its conversation ends, which is the comment I wrote myself in `engine.go`. So the row is invisible to the poller and the process does not fall over on its own, and the slot is held until the launcher restarts rather than for the sandbox's natural life. Worse than the sixteen minutes, exactly as you say. The five minute symmetry between `launchTimeout` and `Remote`'s own client timeout, plus a deploy recreating control-plane on every merge, makes this scheduled rather than exotic.

You are also right that the sweep named in `WaitIdle`'s comment does not touch it. That sweep can fix the row, which is the free half; it cannot cancel a session it has no reference for.

Not fixing it in this pull request, deliberately, and the reason is the one you allowed for: the fix has to live on the launcher, because only the launcher can re-derive a reference control-plane never received. Both of your shapes (a lookup by task id on the socket API, or an orphan sweep on the launcher) are in the issue. This change closes the database-failure half, which this service does own, and the `default:` branch now carries a comment naming the transport half as open with the issue number, so nobody reads it as closed.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Conceded. I traced the edge myself before answering rather than taking the finding on trust, and it confirms yours: the chain in `apps/edge-api/cmd/server/main.go` is unsupported-endpoint, `budgetGate.Wrap`, auth selector, metrics, compat headers, max bytes, with `featuregate.Require(FeatureCowork)` on the route, and no limiter anywhere in it. `authz.NewLimiter` is constructed at start-up but only reachable through `authorizer.Authorize`, which `apps/edge-api/internal/agenttask/handler.go` never calls; it resolves identity with `auth.UserFrom` and stops there. `budgetGate` is inert for JWT session traffic by its own documented design.

So my earlier acceptance rested on a control that is not shipped, which makes it not an acceptance but an assumption. That is the part worth naming.

Fixed in 772c25d9 by making the code say what is true: the comment on the create path now states that nothing bounds the goroutines, that quota gates the launch rather than the row or the spawn, that there is no limiter on the route, and that this is tracked in issue #900. The issue carries your cost breakdown, including the point I had not accounted for, which is that the second pool checkout now happens after the response is written and so carries no request-level backpressure at all.

Not adding a bound inside this pull request, and the reasoning is in the issue rather than hidden here. A global in-flight cap is itself a cross-tenant denial surface, and a per-tenant one duplicates the launcher's quota at the wrong layer, so the honest home is a per-user and per-tenant limiter at the edge next to the limiter that already exists. Shipping a token bucket in a two-defect fix would be the kind of change that gets reviewed less carefully than it deserves.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +94,148 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+		// Panic path. Before this work ran on its own goroutine it ran inside
+		// the HTTP handler, where net/http's per-connection recover contained a
+		// panic to one connection; a bare goroutine would instead take the
+		// whole process down, every tenant with it. Recovering keeps the blast
+		// radius at one task and still records that task as failed, so the
+		// customer is not left watching a queued row that will never move.
+		s.recordLaunchFailure(opCtx, t, engineLaunchFailedMessage)
+	})
```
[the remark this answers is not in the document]
sakibsadmanshajib: Valid, and fixed in 772c25d9 in the shape you specified.

Recovery moved from the `background` helper into `launch` itself, where `sessionRef` is a local declared before the deferred recover that reads it. On a panic it now stops the session first and records the failure second, so a panic between a successful `Launch` and its running transition can no longer leave a live sandbox whose reference only the panicking frame ever held. Stop first is deliberate: freeing the slot does not touch the database, and the database is the dependency most likely to have caused the panic.

Both calls in that recovery run through a small `safely` helper that contains a second panic, which is the separate escalation the security stream found on the same lines. Neither fix subsumes the other, so both are in.

Test added at the location you pointed at, watched red first: `TestService_PanicAfterLaunch_StopsTheSessionAndFailsTheTask` uses a repository double that panics on the running transition after `Launch` succeeds, and asserted `1 slot(s) still in use` before the fix. It now asserts the task ends `failed` and `slotsInUse()` is zero.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +94,201 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+// something the customer can act on or should see as a failed request. The
+// engine detail goes to the log only, never to a customer-visible field.
+func (s *Service) stopEngineSession(ctx context.Context, taskID uuid.UUID, sessionRef string) {
+	if sessionRef == "" {
+		return // never launched, so nothing holds a slot
+	}
+	stopCtx, done := context.WithTimeout(context.WithoutCancel(ctx), engineCancelTimeout)
+	defer done()
+	if err := s.engine.Cancel(stopCtx, sessionRef); err != nil {
+		slog.Default().WarnContext(stopCtx, "agenttask: engine cancel failed, the session may still hold its concurrency slot",
```
sakibsadmanshajib, line 287: **LOW, but it is the one log line that decides whether a leaked slot is recoverable.** Second pass on `772c25d9`.

This warning names `task_id` and the error and not `sessionRef`. On two of the paths that reach it, `sessionRef` exists nowhere else by the time it fires:

- Cancel arriving before the launch finishes. `Transition(..., StatusRunning, sessionRef, ...)` loses the terminal guard, so the reference is never written to the row. The teardown at line 199 is the only thing that ever held it.
- The new panic path at line 185. Same story: the reference lived only in the frame that panicked.

If `engine.Cancel` then fails, the message says the session may still hold its concurrency slot and gives an operator no way to find out which session that is. The row has an empty `engine_session_ref`, the poller cannot see the task, and the launcher's registry is in memory with no listing endpoint. The only remaining remedy is restarting the launcher, which drops every other tenant's live sandbox with it. Adding `"session_ref", sessionRef` to this call turns that into a targeted stop.

While here, the same warning now has a false-positive path that the panic recovery created. If `Transition` fails normally at line 197, line 199 stops the session, and `recordLaunchFailure` at line 216 then panics, the recovery at line 184 stops the same session again. That second stop is harmless for accounting (quota's release is `sync.Once` and `reap` is idempotent, both verified) but it does log this exact warning, because `SandboxEngine.Cancel` calls `Interrupt` on a control socket whose directory was already removed and returns the resulting error. An operator sees a leaked-slot alarm for a slot that was correctly freed. Logging the reference makes that distinguishable too, since the same reference appearing twice reads as a retry rather than a leak.

apps/control-plane/internal/agenttask/service.go
```diff
@@ -66,31 +94,201 @@ func (s *Service) CreateTask(ctx context.Context, tenantID, userID uuid.UUID, pa
...
+	// ever held, and nothing else could ever stop it. Stopping comes first
+	// because it frees the concurrency slot without touching the database,
+	// which is the dependency most likely to have caused the panic.
+	var sessionRef string
+	defer func() {
+		r := recover()
+		if r == nil {
+			return
+		}
+		slog.Default().Error("agenttask: recovered panic during launch",
```
sakibsadmanshajib, line 193: **Attacked `safely` specifically, as asked. It holds. One residual escape, recorded rather than treated as blocking.** Second pass on `772c25d9`.

What I checked and what holds:

- Both statements that can realistically fail in this recovery handler are inside `safely`. `stopEngineSession` reaches the launcher and `recordLaunchFailure` reaches the repository, which is the dependency most likely to have caused the original panic in the first place. Both are wrapped.
- The ordering is right. Stopping comes before recording, so the concurrency slot is freed even when the database is the thing that is broken. `TestService_PanicInsidePanicHandler_DoesNotCrashTheProcess` asserts exactly that, and its log output confirms only the recording `safely` fires.
- `safely` does not swallow anything it should not. It is unexported, has two call sites, and both are inside a recovery path where swallowing is the whole point. Nothing on the normal path goes through it.
- `sessionRef` is genuinely assigned rather than shadowed at line 192, despite the `:=`. Both declarations are at function-body scope so Go treats it as an assignment, and this is empirically confirmed: `TestService_PanicAfterLaunch_StopsTheSessionAndFailsTheTask` asserts zero slots in use, which is only reachable if the deferred handler reads the real reference. That test is also the guard if someone later wraps the launch in a block and reintroduces the shadow.

The residual escape, for the record. Two statements in this handler sit outside `safely`: `debug.Stack()` and the `slog.Default().Error` call on lines 182 and 183. A panic from either is a fresh panic in an already-unwinding deferred function, so it leaves `launch`. `background`'s outer recover does catch it, which I traced through the defer ordering, so the net works for a fault in `debug.Stack`. It does not work for a fault in the logging call itself, because `background`'s recover at line 138 calls the same `slog.Default().Error` with the same argument shape, so whatever killed the inner one kills the outer one too. The two nets are not independent for that single case.

Why this is not blocking: reaching it needs a `slog` handler that panics, and control-plane installs none (no `slog.SetDefault` anywhere in `apps/control-plane/cmd/server/main.go`, so this is the stdlib default text handler writing to stderr). A panicking `String()` or `Error()` on the panic value does not do it either, because `fmt` recovers from those itself and prints a `%!v(PANIC=...)` marker. Worth knowing if a structured handler is ever installed, not worth another wrapper today.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Both halves fixed in f9010cd2.

`session_ref` is now on the warning. The point that decided it is the one about the two paths where the reference exists nowhere else by the time the warning fires: the row never received it, the poller cannot see the task, and the launcher's registry is in memory with no listing endpoint, so without the reference the only remedy is restarting the launcher and taking every other tenant's sandbox down with it. `stopEngineSession`'s doc comment now records that reasoning, so the field does not look decorative to the next person tempted to tidy the log line.

The false alarm is removed at the source rather than made distinguishable. `launch` sets a `teardownRun` flag before it tears its own session down, and the panic recovery skips the stop when that flag is set, so the second call never happens. Your accounting analysis is why this is safe to skip rather than needing to be retried: quota's release is a `sync.Once` and `reap` is idempotent, so the repeat was never doing work, it was only producing a leaked-slot alarm for a slot that had been freed correctly. The flag is set before the call rather than after, so a panic inside the teardown itself also counts as attempted and is not retried from a recovery path that has no more context than the original.

Full short suite green across control-plane, agent-engine and edge-api after the change.

apps/control-plane/internal/agenttask/service.go
sakibsadmanshajib: Taking the option you left open: the residual is stated in the code rather than wrapped, and f9010cd2 puts your analysis where someone will actually meet it, in a comment on the recovery handler itself.

It records all three parts: that a panic from the logging call or from `debug.Stack` escapes into `background`'s recover, that the two nets are not independent for the logging case because the outer one logs through the same call, and why it is acceptable today, which is that reaching it needs a slog handler that panics, control-plane installs none (no `slog.SetDefault` in `cmd/server`, so this is the stdlib text handler), and `fmt` already recovers from a panicking `String` or `Error` on the panic value. It ends with the condition that makes it real again: revisit if a structured handler is ever installed.

Not wrapping it, deliberately. A wrapper around the logging call would have to log its own failure somewhere, so it either uses `slog` again and inherits the same dependency, or it introduces a second independent write path to stderr for a fault that requires a handler this process does not have. That is more machinery than the risk earns, and your own read was the same.

Thank you for attacking the shadowing question empirically rather than by inspection. That `sessionRef` is an assignment and not a shadow at the `:=` is the sort of thing that reads as fine and is load bearing, and the test you identified as its guard is exactly why it is still safe if someone later wraps the launch in a block.

PR #894, opened 2026-08-13
fix: pin openai sdk-tests-python dep below the httpx2 rename in 3.0.0

sakibsadmanshajib:
## Summary

- `main` is red on both `deploy-demo-box` and `CI` (Live integration) at the same Python step: pytest collection fails with `ModuleNotFoundError: No module named 'httpx'`. Predates the latest merge (also failing at `c3f72c07` and `85475af9`).
- Root cause: `packages/sdk-tests/python/pyproject.toml` declared `openai>=2.30.0` unbounded, so pip resolves the latest release, `openai==3.0.0`. That 3.x line renamed its transitive HTTP client dependency from `httpx` to `httpx2`. Our test modules (`test_health.py`, `test_headers.py`, `test_error_shape.py`) `import httpx` directly, independent of what openai bundles internally, so collection aborts as soon as pip resolves 3.0.0.
- Both failing workflows install from this same `pyproject.toml` (`deploy-demo-box.yml`'s SDK replay step and `ci.yml`'s Docker-built `sdk-tests-py` image via `Dockerfile.sdk-tests-py`), so one fix in this file covers both callers. No install-gap found in the JS or Java suites in the same CI step; both already pin bounded versions (`sdk-tests-js`'s `package.json` uses `"openai": "^6.33.0"`).
- Why this matters beyond a red check: `deploy-demo-box`'s SDK replay runs *after* deploy, as the only automated verification that the live demo box actually works. A collection error asserts nothing, so a genuinely broken deploy and this dependency gap were indistinguishable from the outside.

## Fix

- Bound `openai` to `<3.0.0` (matches the existing convention in `sdk-tests-js/package.json`, which pins `^6.33.0`, i.e. allow-minor/patch, block-major).
- Added `httpx` as its own direct, version-bounded dependency (`>=0.28.0,<0.29.0`, bounded to the next minor rather than the next major since httpx is pre-1.0 and has shipped breaking changes across 0.x minor releases before), so this suite no longer depends on whatever HTTP client name openai's SDK happens to use internally.

## Zero-collected-tests failure mode (explicitly checked, not assumed)

Verified locally rather than adding defensive flags: pytest already fails loudly on both relevant conditions.
- A collection error (this bug) already exits 2 (`Interrupted: 1 error during collection`), confirmed in the original job log.
- A hypothetical run that collects zero tests with no error already exits 5 (pytest's default "no tests ran" behavior), verified locally against the same pinned `pytest>=8.0.0` constraint (resolved to 9.1.1).

No `pytest.ini` / `addopts` change made; the existing config already fails the step correctly in both cases.

## Verification (local, Python 3.12, the version `deploy-demo-box.yml`'s `actions/setup-python` installs)

```
$ pip install -e '.[dev]'
Successfully installed ... httpx-0.28.1 openai-2.54.0 ...

$ pytest tests/test_health.py tests/test_models.py tests/test_embeddings.py --collect-only -q
tests/test_health.py::test_health_returns_200_with_status_ok
tests/test_models.py::test_list_models_returns_valid_list
tests/test_embeddings.py::test_embeddings_basic
tests/test_embeddings.py::test_embeddings_batch
4 tests collected in 0.05s

$ pytest --collect-only -q   # full suite, matches ci.yml's docker CMD
14 tests collected in 0.10s
```

Before the fix, on unpatched `main`, the same install resolves `openai==3.0.0` + `httpx2==2.10.0` and `import httpx` fails at collection, reproducing the CI error exactly.

## Buglog entry

Per `.claude/rules/openwolf.md`, this does not land on this branch. JSON line for the follow-up buglog-only PR after merge:

```json
{"error_message": "ModuleNotFoundError: No module named 'httpx' during pytest collection in packages/sdk-tests/python", "root_cause": "pyproject.toml pinned openai>=2.30.0 unbounded; pip resolved openai 3.0.0, whose 3.x line renamed its transitive HTTP client dependency from httpx to httpx2, silently dropping the httpx module that sdk-tests-python's test files import directly", "fix": "bounded openai to <3.0.0 and added httpx as its own direct, version-bounded dependency in packages/sdk-tests/python/pyproject.toml, matching sdk-tests-js's existing caret-pin convention", "tags": ["ci", "python", "sdk-tests", "dependency", "deploy-demo-box", "post-deploy-verification"]}
```

## Test plan

- [x] Reproduced the failure locally on Python 3.12 against unpatched `pyproject.toml` (openai resolves to 3.0.0, `import httpx` fails)
- [x] Confirmed the fix locally on Python 3.12 (openai resolves to 2.54.0, httpx 0.28.1 present, both the replay's 3-file selection and the full 14-test suite collect cleanly)
- [x] Confirmed pytest's existing exit-code behavior already fails loudly on both a collection error (exit 2) and a hypothetical zero-collected-tests run (exit 5); no config change needed
- [ ] CI (`ci.yml`) green on this PR
- [ ] `deploy-demo-box.yml`'s SDK replay green after merge to `main`

## Summary by CodeRabbit

* **Chores**
  * Updated Python SDK test dependencies for improved compatibility.
  * Added a supported `httpx` version range and constrained the OpenAI dependency to versions below 3.0.0.

### DIFF
--- a/packages/sdk-tests/python/pyproject.toml
+++ b/packages/sdk-tests/python/pyproject.toml
@@ -9,7 +9,11 @@ dependencies = [
     # collection). httpx is pinned as its own direct dependency below so this
     # suite never again depends on openai's internal HTTP client naming.
     "openai>=2.30.0,<3.0.0",
-    "httpx>=0.28.0,<1.0.0",
+    # httpx is pre-1.0; bounded to the next minor (not major) since httpx has
+    # shipped breaking changes across 0.x minor releases before. Matches how
+    # npm caret (used for sibling package sdk-tests-js's "openai": "^6.33.0")
+    # treats a 0.x version: bounds at the next minor, not the next major.
+    "httpx>=0.28.0,<0.29.0",
 ]
 
 [project.optional-dependencies]
--- a/packages/sdk-tests/python/pyproject.toml
+++ b/packages/sdk-tests/python/pyproject.toml
@@ -3,7 +3,13 @@ name = "hive-sdk-tests-python"
 version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
-    "openai>=2.30.0",
+    # Bounded below the openai 3.0.0 major bump: 3.x renamed its transitive
+    # HTTP client dependency from `httpx` to `httpx2`, which silently dropped
+    # the `httpx` module our tests import directly (ModuleNotFoundError on
+    # collection). httpx is pinned as its own direct dependency below so this
+    # suite never again depends on openai's internal HTTP client naming.
+    "openai>=2.30.0,<3.0.0",
+    "httpx>=0.28.0,<1.0.0",
 ]
 
 [project.optional-dependencies]
--- a/packages/sdk-tests/python/pyproject.toml
+++ b/packages/sdk-tests/python/pyproject.toml
@@ -10,9 +10,7 @@ dependencies = [
     # suite never again depends on openai's internal HTTP client naming.
     "openai>=2.30.0,<3.0.0",
     # httpx is pre-1.0; bounded to the next minor (not major) since httpx has
-    # shipped breaking changes across 0.x minor releases before. Matches how
-    # npm caret (used for sibling package sdk-tests-js's "openai": "^6.33.0")
-    # treats a 0.x version: bounds at the next minor, not the next major.
+    # shipped breaking changes across 0.x minor releases before.
     "httpx>=0.28.0,<0.29.0",
 ]
 


packages/sdk-tests/python/pyproject.toml
```diff
@@ -3,7 +3,17 @@ name = "hive-sdk-tests-python"
 version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
-    "openai>=2.30.0",
+    # Bounded below the openai 3.0.0 major bump: 3.x renamed its transitive
+    # HTTP client dependency from `httpx` to `httpx2`, which silently dropped
+    # the `httpx` module our tests import directly (ModuleNotFoundError on
+    # collection). httpx is pinned as its own direct dependency below so this
+    # suite never again depends on openai's internal HTTP client naming.
+    "openai>=2.30.0,<3.0.0",
```
sakibsadmanshajib, line 11: **Adversarial review stream: CodeRabbit CLI** (ran locally, `coderabbit review --committed --base main --agent`, version 0.7.2)

Result: 0 findings across both commits on this branch. Only file in scope: `packages/sdk-tests/python/pyproject.toml`. Clean.

packages/sdk-tests/python/pyproject.toml
```diff
@@ -3,7 +3,17 @@ name = "hive-sdk-tests-python"
...
+    # httpx is pre-1.0; bounded to the next minor (not major) since httpx has
+    # shipped breaking changes across 0.x minor releases before. Matches how
+    # npm caret (used for sibling package sdk-tests-js's "openai": "^6.33.0")
+    # treats a 0.x version: bounds at the next minor, not the next major.
+    "httpx>=0.28.0,<0.29.0",
```
sakibsadmanshajib, line 14: **Adversarial review stream: ecc:code-review** — SKIPPED.

Structural capability gap, not a workaround: this agent (build-error-resolver) was dispatched with toolset {Read, Write, Edit, Bash} only, no Skill-invocation tool. ecc:code-review is a plugin Skill and cannot be invoked without that tool. Reported here per Gate Compliance (this is the correct terminal report for a structural gap, not something to route around by faking the stream). Coverage gap is small: the diff is a two-line dependency-version-range edit in a test package's `pyproject.toml`, and CodeRabbit CLI plus the plain adversarial pass below both ran clean/found-and-fixed on it.

packages/sdk-tests/python/pyproject.toml
```diff
@@ -3,7 +3,17 @@ name = "hive-sdk-tests-python"
 version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
-    "openai>=2.30.0",
+    # Bounded below the openai 3.0.0 major bump: 3.x renamed its transitive
```
sakibsadmanshajib, line 6: **Adversarial review stream: plain adversarial pass** (self-conducted)

Findings:
1. **Fixed** (commit 804cbc8c): initial `httpx>=0.28.0,<1.0.0` bound was looser than this repo's own pinning convention for pre-1.0 deps. Sibling package `sdk-tests-js/package.json` pins `openai` with npm caret (`^6.33.0`); npm caret on a 0.x package bounds at the next *minor*, not the next major, precisely because 0.x packages (httpx included) have shipped breaking changes across minor releases before. Tightened to `<0.29.0` to match, re-verified collection still passes after the tightening.
2. Checked and confirmed **not a bug**: whether pytest's existing config already fails loudly on zero-collected-tests, rather than trusting the task framing at face value. Verified locally — collection errors already exit 2, and a hypothetical zero-tests-collected run already exits 5 (pytest's default "no tests ran"). No `pytest.ini`/`addopts` change made; would have been a redundant flag for behavior pytest already has.
3. Checked and confirmed **not a bug**: whether the JS or Java sdk-tests suites in the same CI step share this install gap. Both already pin bounded versions (JS caret range, Java via `build.gradle`/`settings.gradle`), so no cross-suite fix needed.
4. No security/auth/money/input-parsing surface touched by this diff (dependency version range in a test package only), so no specialist stream added.

packages/sdk-tests/python/pyproject.toml
sakibsadmanshajib: Acknowledged: CodeRabbit clean, no action needed. Resolving.

packages/sdk-tests/python/pyproject.toml
sakibsadmanshajib: Acknowledged as SKIPPED with reason recorded above. No workaround attempted; coverage backfilled by CodeRabbit CLI + the plain adversarial pass on this same two-line diff. Resolving.

packages/sdk-tests/python/pyproject.toml
sakibsadmanshajib: Fix for finding 1 pushed in commit 804cbc8c and re-verified locally (collection still passes with the tightened httpx pin). Findings 2-4 are checks with no action required. Resolving.

packages/sdk-tests/python/pyproject.toml
```diff
@@ -3,7 +3,17 @@ name = "hive-sdk-tests-python"
...
+    # HTTP client dependency from `httpx` to `httpx2`, which silently dropped
+    # the `httpx` module our tests import directly (ModuleNotFoundError on
+    # collection). httpx is pinned as its own direct dependency below so this
+    # suite never again depends on openai's internal HTTP client naming.
+    "openai>=2.30.0,<3.0.0",
+    # httpx is pre-1.0; bounded to the next minor (not major) since httpx has
+    # shipped breaking changes across 0.x minor releases before. Matches how
+    # npm caret (used for sibling package sdk-tests-js's "openai": "^6.33.0")
+    # treats a 0.x version: bounds at the next minor, not the next major.
```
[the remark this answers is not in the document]
sakibsadmanshajib: Confirmed, fixed in commit 9693de4d. You are right: sdk-tests-js openai ^6.33.0 is a 6.x range that npm caret bounds at the next major (<7.0.0), the opposite of the point I was making about 0.x packages bounding at the next minor. Dropped the misleading example, kept the accurate reasoning. Re-verified locally on Python 3.12: pip install -e .[dev] still resolves httpx 0.28.1, full 14-test suite still collects cleanly. Resolving.

PR #897, opened 2026-08-13
ci: capture Cowork visual proof on the pull request, before merge

sakibsadmanshajib:
## What this is

A CI job that captures visual proof of the Cowork agent-task surface **before** a pull request merges, on the pull request itself. The owner ruled on 2026-08-13 that proof for the Cowork fixes is produced in CI on the pull request, not captured after merge against the deployed box.

## Why it has to exist

Orchestrator rule 8 forbids merging a change to a live UI surface without a fresh capture taken against the actually-running stack, posted in the pull request. PR #891 could not satisfy that:

* A Cowork task only really launches through the socket arm of `buildAgentEngine` (`apps/control-plane/cmd/server/main.go`). Under docker compose the in-process arm cannot succeed at all.
* The Apptainer image is `linux/amd64` and cannot be built on the WSL2 dev box, so control-plane there falls back to `NotConfiguredEngine` and any local capture proves the absence of an engine, not the presence of a fix.
* The only stack that could launch a real task was the demo box, which runs already-merged code.

Proof-before-merge was therefore circular. This job breaks the circle by standing the real thing up per run.

## Runner choice, with evidence

**GitHub-hosted `ubuntu-latest`**, not the self-hosted demo-box runner.

* The hosted image already runs an unprivileged bubblewrap sandbox in `ci.yml`'s `egress-firewall-e2e` job after setting the same two AppArmor sysctls this job sets, so unprivileged user namespaces are known to work there. That is an inference from an existing green job, so the workflow does not rest on it: a dedicated step execs the real SIF (`apptainer exec "$sif" /bin/true`) and fails the run if the sandbox cannot start. The runner question is answered per run, as a fact, and the fallback signal to the self-hosted runner is that step going red.
* The self-hosted runner was rejected on two counts. `scripts/install-agent-engine-host.sh` stops and restarts the `hive-agent-engine` systemd unit and rewrites `/home/sakib/agent-runtime`, so a proof run would reach straight into the live launcher's state and disturb the live demo surface. Running an untrusted pull request branch on a persistent runner attached to a public repository is also the configuration GitHub explicitly warns against. A hosted runner is disposable, so neither question arises.

## What the job does

1. Installs Apptainer, enables unprivileged user namespaces, and starts a systemd user session (rootless cgroups need `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS`, or every launch dies).
2. Runs the shipped `scripts/install-agent-engine-host.sh` unmodified. It fetches the newest CI-built `agent-engine-sif` artifact, builds the launcher, and starts it as an unprivileged systemd user unit. `HIVE_QUOTA_USER_CONCURRENCY=1`, which is what makes the concurrency ceiling observable with a single running task.
3. Execs the SIF directly as the runner-capability preflight.
4. Boots redis, litellm, control-plane, edge-api and agent-console from the pull request's merge ref, with `HIVE_AGENT_ENGINE_SOCKET` set and the launcher's socket directory bind-mounted. The control-plane container is given **no** `CAP_SYS_ADMIN` and **no** `/dev/fuse`; the socket arm exists precisely so it never needs them.
5. Asserts control-plane took the socket arm: the boot log must say the daemon is reachable and must not say the engine is unconfigured. Without this the job could screenshot a `NotConfiguredEngine` stack and call it proof.
6. Serves the console and edge-api from one origin through a small Caddy front (`deploy/docker/Caddyfile.agent-proof`), because the console's browser fetches are relative by design and edge-api serves no CORS headers.
7. Seeds a run-scoped fixture tenant and enables `ENABLE_COWORK` for it, then drives the console with `apps/agent-console/proof/harness/capture-live.mjs`.
8. Uploads the captures as an artifact **and** pushes them to a side branch and posts them as a comment on the target pull request. An artifact alone is a zip behind an authenticated download, which does not satisfy rule 8.

## The two behaviours, and the slot count

`cancel-frees-slot` (issue #886) does not screenshot a badge turning grey. With one slot configured, it:

1. runs a task and records the launcher's live session count as 1,
2. submits a second task and captures it being refused while the slot is held,
3. cancels the first and waits for the count to reach 0,
4. submits a third and captures it launching.

The count is read from the launcher's own `sessions/` directory. `SandboxEngine.Launch` creates exactly one directory there per live session and `reap()` removes it in the same call that runs the quota release, so the number is the slot rather than a proxy for it. Before #886 the count stayed at 1 across a cancel and the next create stayed refused.

`create-returns-queued` (issue #881) reads the create response off the wire: HTTP 201, body status `queued`, inside a 14 second budget under edge-api's 15 second client bound. The measured latency is stamped into the screenshot, so the image and the number cannot drift apart.

## Failing loudly

The job fails, never passes quietly, when: the SIF is missing, `apptainer exec` is refused, control-plane falls back to `NotConfiguredEngine`, the proof origin never answers, the composer never renders (no session, or the gate is off), a task never reaches Running, a scenario's assertion fails, or zero screenshots were written. `if-no-files-found: error` covers the upload.

The property is demonstrated rather than asserted: `-f sabotage=true` stops the launcher immediately before the capture, so nothing can launch. The red run is linked in a comment below.

## Reuse

Not welded to #891. `gh workflow run "agent visual proof" -f pr=<number>` aims it at any pull request; it checks out `refs/pull/<number>/merge`, so the proof is of what would land. Automatic `pull_request` runs on agent-surface paths execute the `launch-liveness` scenario, which asserts nothing version specific and holds every agent-surface change to "the surface still launches". Behaviour-specific scenarios are what a dispatch selects, because only the dispatcher knows which behaviours a given pull request claims.

## Credential handling

* `CONTROL_PLANE_INTERNAL_TOKEN` is minted per run with `openssl rand` and masked. The demo box's token is never used here, so a leak of one cannot reach the other. No secret is echoed anywhere.
* The session comes from the admin one-time-token flow in `live-auth.mjs`. No password is read, written or rotated.
* `E2E_RUN_KEY` is supplied (`runScopedEmail` throws without it since PR #880), so every fixture address belongs to this run.
* Captures land under `docs/proof/`, which is the directory `npm run lint:proof-tokens` scans, and the job runs that linter before publishing. The pixel half is handled structurally: the screenshot stamp carries the scenario, the shot, the commit and a note, and never a URL.

## Test plan

* [ ] The automatic run on this pull request executes `launch-liveness` end to end against a real sandbox.
* [ ] A `sabotage=true` run goes red with no screenshots, demonstrating the empty-capture guard.
* [ ] A dispatch at `pr=891` posts the two behaviour captures on that pull request.

Note that the first two need this workflow on a branch GitHub will run, and the third needs it on `main`, since `workflow_dispatch` is only offered for workflows present on the default branch.

### DIFF
--- a/.github/workflows/agent-visual-proof.yml
+++ b/.github/workflows/agent-visual-proof.yml
@@ -152,6 +152,9 @@ jobs:
       # image is built with; Caddyfile.agent-proof is the same-origin front.
       NEXT_PUBLIC_EDGE_API_BASE_URL: ''
       E2E_RUN_KEY: ${{ github.run_id }}-${{ github.run_attempt }}
+      # Off the root filesystem, which Docker also lives on. See the reclaim
+      # step below.
+      PLAYWRIGHT_BROWSERS_PATH: /mnt/ms-playwright
     steps:
       - name: Resolve the pull request under proof
         id: target
@@ -178,6 +181,26 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ steps.target.outputs.ref }}
+
+      - name: Reclaim disk space
+        # This job holds a one gigabyte SIF, five container images including
+        # a Next.js production build, node_modules and a Chromium. Run
+        # 31730077938 filled the root filesystem and failed inside `npx
+        # playwright install` with ENOSPC, which reads as a browser problem
+        # rather than a disk one. The SIF and the browser live on /mnt, which
+        # is the runner's large volume; these preinstalled toolchains, which
+        # nothing here uses, are what frees room for Docker on the root disk.
+        run: |
+          set -euo pipefail
+          # The hosted tool cache is deliberately left alone: setup-node and
+          # the actions runtime live there, and reclaiming it to save a few
+          # more gigabytes breaks the steps below in a way that is much
+          # harder to read than ENOSPC.
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
+            /usr/local/share/boost || true
+          sudo mkdir -p /mnt/ms-playwright
+          sudo chown "$(id -un):$(id -gn)" /mnt/ms-playwright
+          df -h / /mnt
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/agent-visual-proof.yml
+++ b/.github/workflows/agent-visual-proof.yml
@@ -452,7 +452,28 @@ jobs:
               );
             if (error) throw new Error(`ENABLE_COWORK upsert failed: ${error.message}`);
             console.log("ENABLE_COWORK enabled for this run tenant");
-          '
+            console.log(`PROOF_TENANT_ID=${ids.verifiedTenantId}`);
+          ' | tee /tmp/seed.log
+          grep -E '^PROOF_TENANT_ID=' /tmp/seed.log >> "$GITHUB_ENV"
+
+      - name: Assert the Cowork gate reads as on
+        # Run 31816249902 reached the console signed in and rendered "The
+        # agent workspace is turned off for your organization", which is a
+        # capture of a real page and no proof of anything. Asserting the gate
+        # here separates the three places it can be off (the row, the
+        # registry join, the token's tenant claim) from a browser that simply
+        # shows the disabled panel, and it fails on the step that owns the
+        # cause rather than on the one that noticed.
+        run: |
+          set -euo pipefail
+          gates=$(curl -sS --fail-with-body \
+            -H "Authorization: Bearer $CONTROL_PLANE_INTERNAL_TOKEN" \
+            "http://localhost:8081/internal/featuregate/$PROOF_TENANT_ID")
+          # One line: a `python3 -c` program written across several lines
+          # inside a YAML block has to be indented, and Python rejects a
+          # uniformly indented program.
+          python3 -c 'import json,sys; g=json.loads(sys.argv[1]).get("gates",{}); print("gates:", json.dumps(g,sort_keys=True)); sys.exit(0 if g.get("ENABLE_COWORK") else 1)' "$gates" \
+            || { echo "::error::ENABLE_COWORK is off for the run tenant, so the console renders no Cowork surface"; exit 1; }
 
       - name: Make the launch impossible (deliberate sabotage run)
         # The negative control for the whole job: with the launcher stopped
--- a/.github/workflows/agent-visual-proof.yml
+++ b/.github/workflows/agent-visual-proof.yml
@@ -170,7 +170,9 @@ jobs:
           # refs/pull/N/merge is what would land, which is what rule 8 asks
           # to be proven, and it is also always current with main.
           echo "ref=refs/pull/$pr/merge" >> "$GITHUB_OUTPUT"
-          gh pr view "$pr" --json number,title,headRefOid \
+          # -R is load bearing: this step runs before the checkout, so there
+          # is no git repository for gh to infer the repository from.
+          gh pr view "$pr" -R "$GITHUB_REPOSITORY" --json number,title,headRefOid \
             --jq '"proving PR #\(.number): \(.title) at head \(.headRefOid)"'
 
       - uses: actions/checkout@v4
@@ -426,6 +428,7 @@ jobs:
           echo "::warning::launcher stopped on purpose; this run is expected to fail"
 
       - name: Capture the proof
+        id: capture
         env:
           SUPABASE_SERVICE_ROLE_KEY: ${{ secrets.SUPABASE_SERVICE_ROLE_KEY }}
           PROOF_BASE_URL: ${{ env.PROOF_ORIGIN }}
@@ -461,7 +464,10 @@ jobs:
         run: npm run lint:proof-tokens
 
       - name: Upload the captures
-        if: always()
+        # Whenever the capture ran at all, including when it failed partway:
+        # a failure shot is evidence too. Not when it never started, where an
+        # empty directory is the earlier step's failure, not this one's.
+        if: always() && steps.capture.conclusion != 'skipped'
         uses: actions/upload-artifact@v4
         with:
           name: agent-visual-proof-${{ github.run_id }}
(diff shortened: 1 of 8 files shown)


.github/workflows/agent-visual-proof.yml
```diff
@@ -0,0 +1,641 @@
...
+        # the same guard that fails a hand-committed leak fails this one.
+        # It cannot inspect pixels; that half is handled by stamp.mjs, whose
+        # footer never carries a URL.
+        run: npm run lint:proof-tokens
+
+      - name: Upload the captures
+        # Whenever the capture ran at all, including when it failed partway:
+        # a failure shot is evidence too. Not when it never started, where an
+        # empty directory is the earlier step's failure, not this one's.
+        if: always() && steps.capture.conclusion != 'skipped'
```
sakibsadmanshajib: CRITICAL: the credential-leak gate does not cover the artifact upload path, only the PR comment.

`Upload the captures` runs on `if: always() && steps.capture.conclusion != 'skipped'` (line 548). That condition does not check the outcome of the step immediately before it, `Check the text log carries no credential` (line 537, `npm run lint:proof-tokens`, line 542), whose whole job is to scan the same `docs/proof/agent-visual-proof/run-<id>` directory this step uploads.

If that lint step fails because it found a leaked credential in that directory, the upload proceeds anyway and the failing files still reach a downloadable `actions/upload-artifact` artifact. Only `Post the captures on the pull request` (line 556, gated `if: success()` at line 561) is protected: a detected leak is kept out of the PR comment, but not out of the artifact.

Since this repository is public, that artifact is reachable by anyone holding the run URL, exactly the exposure `scripts/redact-log-credentials.py` was written against (its own header cites the four real invitation tokens PR #578 leaked this way). Here the one gate built to catch that class of leak protects the comment but not the artifact.

Suggest making the upload step's condition also require the lint step's own success, for example `if: (success() || (failure() && steps.capture.conclusion != 'skipped')) && steps.lint.conclusion != 'failure'`, or simpler, move the lint check ahead of any step that can publish the directory and let its failure block every publish path, not just the PR comment.

.github/workflows/agent-visual-proof.yml
```diff
@@ -0,0 +1,641 @@
...
+          # can show rather than a claim.
+          HIVE_QUOTA_USER_CONCURRENCY: '1'
+          HIVE_QUOTA_TENANT_CONCURRENCY: '1'
+          HIVE_SANDBOX_MEMORY_LIMIT: 2G
+          HIVE_SANDBOX_CPU_LIMIT: '2'
+        run: |
+          set -euo pipefail
+          sudo mkdir -p "$RUNTIME_DIR"
+          sudo chown "$(id -un):$(id -gn)" "$RUNTIME_DIR"
+          RUNTIME_DIR="$RUNTIME_DIR" bash scripts/install-agent-engine-host.sh
```
sakibsadmanshajib, line 354: HIGH: the SIF this job treats as proof of "a real Apptainer sandbox behind the socket arm" is not correlated to the commit under proof.

`scripts/install-agent-engine-host.sh`, invoked here at line 292, fetches the SIF with `?name=agent-engine-sif&per_page=1`: the single most recently created artifact of that name anywhere in the repository, across every branch and PR. `agent-engine-sif.yml` only rebuilds that artifact when `deploy/apptainer/**`, `vendor/openhands/**`, or `apps/agent-engine/packs/**` change, the same three paths this workflow's own `pull_request` trigger watches (line 79 for the first).

For a PR that changes any of those three paths (which is exactly the class of PR this workflow's path filter exists to cover), nothing here checks that the fetched SIF was actually built from that PR's head SHA, or even from current `main`. It is whichever `agent-engine-sif` run most recently succeeded, from any branch. The two workflows never coordinate: no `needs`, no `workflow_run` trigger, no SHA comparison against `steps.target.outputs.ref`.

The PR's own framing ("a real Apptainer sandbox behind the socket arm of the agent engine") reads as proof of the code under test. For a PR touching the image-affecting paths, it can silently be proof of a different, older build instead, which is the "artifact reused from a previous run" failure mode this job is supposed to be unable to produce.

Suggest asserting the artifact's originating `workflow_run.head_sha` (available from the Artifacts API) equals the resolved SHA of `refs/pull/<pr>/merge`'s base, or triggering and awaiting a fresh `agent-engine-sif` build for that SHA before this job proceeds, for any PR touching the three image-affecting paths.

.github/workflows/agent-visual-proof.yml
```diff
@@ -0,0 +1,641 @@
...
+        if: failure()
+        # No working-directory: a run that failed before the checkout has no
+        # deploy/docker, and a step that cannot even start its shell reports
+        # a directory error instead of the failure that actually happened.
+        run: |
+          journalctl --user -u hive-agent-engine --no-pager -n 200 2>&1 \
+            | python3 scripts/redact-log-credentials.py > /tmp/launcher.log || true
+          if [ -d deploy/docker ]; then
+            (cd deploy/docker && docker compose logs --no-color --timestamps --tail=500 2>&1) \
+              | python3 scripts/redact-log-credentials.py > /tmp/compose.log || true
```
sakibsadmanshajib, line 695: MEDIUM: the failure-log redaction this step relies on does not cover this job's own S3 and LiteLLM secrets by name.

`Dump the launcher and stack logs on failure` pipes `docker compose logs` through `scripts/redact-log-credentials.py` (line 620) before the result reaches the public `agent-visual-proof-logs-*` artifact. That script's `CREDENTIAL_PARAMS` list has no bare `key`, `access_key`, or `secret_key` pattern, only `api_key` and `apikey`. `S3_ACCESS_KEY`, `S3_SECRET_KEY` (wired into this job at lines 150 to 153), and `LITELLM_MASTER_KEY` (line 135) all end in a variable-name shape the regex does not match, so a line like `S3_SECRET_KEY=<value>` would pass through unredacted if any container ever echoes it (a crash trace, an env dump on a startup validation failure, or an S3 SDK's debug logging of a presigned URL's `X-Amz-Credential`/`X-Amz-Signature`).

The script's bare-JWT and `hk_`-prefix patterns cover Supabase and Hive API keys, so this is not a blanket gap, but this job is the first one routing S3 and LiteLLM credentials through this particular redactor into a public artifact. Worth confirming those two never appear in these containers' stdout under any failure mode this job can hit, or widening `CREDENTIAL_PARAMS` to a bare `key` suffix so `_ACCESS_KEY`/`_SECRET_KEY`/`_MASTER_KEY` style names are covered the same way `_PASSWORD` already is.

.github/workflows/agent-visual-proof.yml
sakibsadmanshajib: Fixed in e0d0c79f. The upload is now gated on the lint step:

```yaml
if: always() && steps.capture.conclusion != "skipped" && steps.lint.conclusion != "failure"
```

You are right that the two halves were inconsistent, and in the worse direction: the lint fired, the comment was withheld, and the artifact shipped the same bytes to anyone with the run URL on a public repository. The lint step has no `continue-on-error`, so a detection already fails the job; what was missing was withholding the payload. A skipped upload is also not a silent absence here, because the lint step itself is red and names the file it matched.

.github/workflows/agent-visual-proof.yml
sakibsadmanshajib: Fixed in e0d0c79f, with one deliberate difference from the suggestion. A new step selects the artifact and downloads it into `RUNTIME_DIR` before the installer runs, so the installer finds a SIF already in place and its newest-artifact fallback never executes.

Correlation is by image inputs, not by head SHA. The SIF is a function of exactly three trees (`deploy/apptainer`, `apps/agent-engine/packs`, `vendor/openhands`), so the step hashes those three at this ref and walks the unexpired artifacts newest first, fetching each candidate run commit shallowly and hashing the same three trees there. The first exact match is used; no match fails the job and prints `gh workflow run "agent-engine SIF" --ref <branch>`.

Matching on the head SHA alone was the first thing I tried and it is wrong in practice: it would demand a fresh image build for every unrelated commit and fail every pull request that touches none of those paths, which is nearly all of them. An artifact built from byte-identical image inputs IS the image for this ref. Verified against the current state: the newest artifact `9103288010` was built at `c1ca04d5`, the last commit to touch those paths on main, so this branch matches on the first candidate rather than passing vacuously.

.github/workflows/agent-visual-proof.yml
sakibsadmanshajib: Fixed in e0d0c79f. `key` is now in `CREDENTIAL_PARAMS`, which the existing `[A-Za-z0-9_]*` prefix turns into every `*_KEY` spelling at once: `S3_ACCESS_KEY`, `S3_SECRET_KEY`, `LITELLM_MASTER_KEY`, `SUPABASE_ANON_KEY`, and a bare `access_key=`. All five are in the self-check, which fails without the change. It over-redacts benign pairs such as `idempotency_key=` and `cache_key=`, and that is the correct direction for a filter whose output is published: a lost debugging value costs a rerun, a lost credential costs a rotation.

On failing closed for unrecognised high-entropy values, I looked at it and decided against, so this is a reasoned rebuttal rather than an omission. The inputs here are container log dumps, which are mostly high-entropy strings that are not credentials: commit SHAs, UUIDs, container and image ids, trace and request ids, base64 payloads, Go pointer addresses. An entropy rule would redact those and leave the dump unreadable, which pushes people to read the raw logs instead of the scrubbed ones and removes the guard from the path entirely. It also would not be fail-closed in any real sense, since it cannot distinguish a secret from a digest; it would just be noisier. The named-parameter and known-prefix approach stays, and it is testable, which is why every claim above has an assertion behind it. If a new credential shape appears, the fix is a name or a prefix plus a self-check case, and that is a one-line change.

PR #903, opened 2026-08-14
fix: stop collapsing a slow or unreachable control-plane into invalid_api_key

sakibsadmanshajib:
## Summary

A control-plane container recreated during deploy took several seconds (5.65s observed) to answer its first `/internal/apikeys/resolve` call. The edge-api resolve client's 5s timeout was below that observed cold-start latency, so the resolve call timed out and `Authorizer.Authorize` collapsed that transport failure into the same `401 invalid_api_key` response as a key that genuinely does not exist. The 2026-08-14 post-deploy replay's `GET /v1/models` failed this way against a perfectly valid key.

A 401 on a valid credential is a lie that sends the next operator to rotate a working key. The same failure shape nearly caused exactly that today in a different subsystem: the Open WebUI shim-key probe hit an identical timeout and its log line would have recommended minting a replacement key.

## What changed

- `authz.Client.Resolve` classifies a resolve failure as a new `ErrUpstreamUnavailable` sentinel when the call reaches no verdict on the key at all: transport failure, client timeout, canceled context, or a control-plane 5xx. A genuine 404 (not found) or 409 (revoked/disabled/not-active) verdict is unaffected and still denies as before.
- `Authorizer.Authorize` maps `ErrUpstreamUnavailable` to a retryable `503 upstream_unavailable` instead of `401 invalid_api_key`.
- `apierrors.WriteAuthFailure` (the single shared mapper already used by the audio, files, images, and batches handlers, plus `handleModels`) gained the `upstream_unavailable` -> 503 case.
- **Sibling conflation found and fixed in the same pass**: `inference.Orchestrator`'s `executeSync` and `executeStreaming` each duplicated `WriteAuthFailure`'s status-code switch inline instead of calling it, so both would have kept answering 401 for this case even after the Authorizer fix. Both now call the shared helper, which also deletes ~20 duplicated lines.
- **Second sibling found and fixed**: the OWUI shim-key probe (`checkOWUIShimKey` / `watchOWUIShimKey` in `apps/edge-api/cmd/server/main.go`) had the identical collapse: any resolve error, including a timeout, was reported as "it does not resolve to a Hive API key" and logged with "mint a replacement" advice. It now distinguishes a transient upstream-unavailable failure from a genuine does-not-resolve verdict and drops the rotate advice for the former.
- Resolve client timeout raised 5s -> 10s (roughly double the observed 5.65s cold-start worst case). A dead control-plane now takes longer to surface as a failure, but that failure is now an honest, retryable 503 instead of a false 401, so the trade favors correctness over hot-path latency on an already-broken deployment.
- `deploy-demo-box.yml`'s post-deploy "Assert model catalog" step now retries (`--retry 5 --retry-all-errors --retry-delay 5`) so it rides out the same cold-start window instead of failing the whole deploy on the first request after a container recreate.

## Siblings audited, not touched

- `apps/control-plane/internal/apikeys/http.go`'s `handleInternalResolve` -> `handleKeyError` already answers 500 (not 401) for the context-canceled cases logged in the incident (`resolve tenant for account`, `get key rate policy`); its default branch only matches known sentinels (`ErrNotFound`, `ErrRevoked`, `ErrDisabled`, `ErrNotActive`) and falls through to an opaque 500 otherwise. No fix needed there.
- `buildBudgetGate`'s resolver in `main.go` treats any resolve error as "no workspace identity, best-effort" and returns `false` rather than denying the request; the per-route authz path re-checks independently, so this is not part of the same bug class.
- `apps/edge-api/internal/authz/authz.go`'s `Resolver`/`NewResolver` have the identical inline collapse shape but are **dead code**: `grep` across `apps/edge-api` found zero callers or tests referencing them (the wired production type is `authz.Client` in `client.go`). Left untouched rather than silently editing unreferenced code; flagging for a follow-up cleanup/deletion rather than expanding this PR's scope.

## Timeout decision

Raised the resolve client's timeout from 5s to 10s. The 2026-08-14 incident measured a 5.65s first-call latency against a freshly recreated control-plane container, i.e. already above the previous 5s bound. 10s gives roughly double that observed worst case as headroom. Trade-off: a genuinely dead control-plane now takes up to 10s (was 5s) to surface as a failure on the hot path. Accepted because that failure now answers a retryable 503 rather than a non-retryable 401 that reads as "your key is wrong."

## Buglog entry

```json
{"date":"2026-08-14","error_message":"GET /v1/models returned 401 invalid_api_key for a valid API key immediately after a control-plane container recreate","root_cause":"authz.Client.Resolve collapsed every resolve failure (transport error, client timeout, canceled context, control-plane 5xx) into the same error as a genuine not-found/revoked verdict; Authorizer.Authorize then mapped any such error to 401 invalid_api_key, and inference.Orchestrator's executeSync/executeStreaming duplicated that mapping inline instead of routing through the shared WriteAuthFailure helper, so the same collapse existed in two places","fix":"classified resolve failures via a new ErrUpstreamUnavailable sentinel (transport/timeout/cancel/5xx) distinct from a genuine 404/409 verdict; Authorizer now answers a retryable 503 upstream_unavailable instead of 401 for the former; Orchestrator's two duplicated switches now call the shared WriteAuthFailure mapper instead of a second copy of its logic; the OWUI shim-key probe no longer advises rotating a key on a transient resolve timeout; resolve client timeout raised 5s->10s; deploy workflow's post-deploy catalog assertion now retries through a cold start","tags":["authz","edge-api","control-plane","false-401","cold-start","deploy","owui-shim-key"]}
```

To be appended to `main` in a separate buglog-only PR after this merges, per `.wolf/` policy (never appended on this branch).

## Test plan

- [x] RED: new tests failed to compile against pre-fix code (`ErrUpstreamUnavailable` undefined) — proves the tests target the gap.
- [x] `TestResolveClassifiesClientTimeoutAsUpstreamUnavailable`, `TestResolveClassifiesContextCancellationAsUpstreamUnavailable`, `TestResolveClassifiesControlPlane5xxAsUpstreamUnavailable` (client.go)
- [x] `TestResolveDoesNotClassifyGenuineNotFoundAsUpstreamUnavailable` (client.go) — proves the fix does not weaken a real rejection
- [x] `TestAuthorizeReturnsUpstreamUnavailableOnTransportFailure`, `TestAuthorizeReturnsUpstreamUnavailableOnTimeout` (authorizer.go)
- [x] `TestAuthorizeStillReturnsInvalidAPIKeyForGenuineNotFound` (authorizer.go) — passes before and after, per the task's own requirement
- [x] `TestCheckOWUIShimKeyDistinguishesUpstreamUnavailableFromDoesNotResolve`, `TestWatchOWUIShimKeyLogsTransientUnavailableWithoutRotateAdvice` (main_test.go)
- [x] Full `go test ./apps/edge-api/...` green
- [x] `go build ./apps/control-plane/...` green (unaffected, sanity check)
- [x] `gofmt -l` clean on every file this PR touches
- [x] `go vet ./apps/edge-api/...` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of temporary authorization-service outages.
  - Returns retryable Service Unavailable responses instead of incorrectly reporting invalid credentials.
  - Preserves accurate invalid-key responses for genuinely rejected or missing keys.
  - Prevents misleading key-rotation guidance during transient outages.
  - Applies authorization-error handling consistently across streaming and non-streaming requests.
  - Retries replay catalog requests after transient failures before ultimately failing.


apps/edge-api/internal/authz/client.go
```diff
@@ -126,23 +152,42 @@ func (c *Client) Resolve(ctx context.Context, rawToken string) (AuthSnapshot, er
...
 	if resp.StatusCode != http.StatusOK {
 		respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
-		return AuthSnapshot{}, fmt.Errorf("authz: status %d: %s", resp.StatusCode, strings.TrimSpace(string(respBody)))
+		statusErr := fmt.Errorf("authz: status %d: %s", resp.StatusCode, strings.TrimSpace(string(respBody)))
+		// 404 (not found) and 409 (revoked/disabled/not-active) are genuine
+		// control-plane verdicts on the key; every other status (5xx from a
+		// control-plane request that itself failed or was canceled, or
+		// anything unexpected) is the same "no verdict reached" shape as a
+		// transport failure above.
+		if resp.StatusCode == http.StatusNotFound || resp.StatusCode == http.StatusConflict {
```
sakibsadmanshajib: **HIGH: the 404 discriminator reintroduces the exact false 401 this PR exists to remove.**

`/internal/apikeys/resolve` is mounted conditionally in `apps/control-plane/internal/platform/http/router.go:253` (`if cfg.APIKeysHandler != nil && cfg.AuthMiddleware != nil`). When control-plane boots without a database pool that route is never registered, there is no catch-all handler in that mux, and Go's default ServeMux answers `404 page not found`. The comment at `router.go:63` already records this as an observed live failure (issue #816), in these words: "/internal/apikeys/resolve 404s and edge-api reports the resulting resolution failure to the caller as Incorrect API key provided".

This branch classifies that 404 as a genuine control-plane verdict, so a control-plane that came up unable to authenticate anything still answers `401 invalid_api_key` for every valid key. Same failure shape as the 2026-08-14 incident, on a failure mode already documented in this repo, and this PR does not close it. A pool that fails to open on a freshly recreated container is not exotic; it is the neighbouring failure of the exact cold start this PR is about.

The two 404s are distinguishable, and the code already holds the bytes needed to tell them apart: `respBody` is read one line above. Control-plane answers `{"error":"key not found"}` with `Content-Type: application/json` (`handleKeyError`, `apps/control-plane/internal/apikeys/http.go:588`), while ServeMux answers `404 page not found` as `text/plain`. Treat a 404 as a verdict only when the body parses as the control-plane error envelope, and classify anything else as `ErrUpstreamUnavailable`. `TestResolveDoesNotClassifyGenuineNotFoundAsUpstreamUnavailable` still passes under that rule, since it serves a JSON body.

Secondary note on the same condition: the 409 half looks unreachable on this endpoint. `ResolveSnapshot` returns a snapshot carrying `Status` and never returns `ErrRevoked`/`ErrDisabled`/`ErrNotActive`, so revoked, disabled and expired keys come back 200 and are denied edge-side by `CheckAccess`. Not a defect, but the comment above the branch describes behaviour this endpoint does not have.

apps/edge-api/internal/authz/client.go
```diff
@@ -126,23 +152,42 @@ func (c *Client) Resolve(ctx context.Context, rawToken string) (AuthSnapshot, er
...
+			return AuthSnapshot{}, statusErr
+		}
+		return AuthSnapshot{}, fmt.Errorf("%w: %w", ErrUpstreamUnavailable, statusErr)
```
sakibsadmanshajib: **MEDIUM: a permanent misconfiguration is now advertised to every customer as retryable, and silently.**

`RequireInternalToken` (`apps/control-plane/internal/platform/http/internalauth.go:29`) answers **401** when `CONTROL_PLANE_INTERNAL_TOKEN` is unset on control-plane or does not match the value edge-api sends. 401 is neither 404 nor 409, so it lands on this line and becomes `upstream_unavailable`. Every request from every customer then gets `503 The authorization service is temporarily unavailable. Please retry.` indefinitely, while both containers report healthy and the control-plane logs only a startup warning.

Every OpenAI SDK retry layer resends on 5xx. This repo already states the rule as an invariant, in `apps/edge-api/internal/chat/settlement_test.go:571-575`: a permanent condition must not be reported as a retryable server fault, because it "turns one misconfigured alias into a retry storm". A wrong shared secret is more permanent than a misconfigured alias, and it fails every request rather than one alias.

Answering the customer a 503 is still defensible (their key genuinely is not the problem). What is not defensible is that it is indistinguishable operationally: the only trace is `authz: key resolution failed err=...` at `authorizer.go:101`, byte-identical in shape to a real outage, so an operator debugging it hunts an unreachable control-plane that is in fact up and answering promptly. Please special-case 401 and 403 from the internal endpoint with a distinct log line naming the shared secret. That is a deployment fault edge-api can identify precisely, and it is the second-cheapest line in this file.

The same applies to a misdirected `baseURL`: anything that answers 200 with a non-snapshot body falls into the decode branch below and produces the same permanent 503 loop.

apps/edge-api/internal/authz/authorizer.go
```diff
@@ -99,6 +99,20 @@ func (a *Authorizer) Authorize(ctx context.Context, authHeader string, aliasID s
 		// for not-found/revoked vs a transport error) so an outage is
 		// diagnosable from logs instead of guesswork across reruns.
 		log.Printf("authz: key resolution failed err=%v", err)
+		if errors.Is(err, ErrUpstreamUnavailable) {
```
sakibsadmanshajib, line 160: **MEDIUM: this branch falsifies the security claim in the comment immediately above it.**

Lines 92-99 state that all resolution failures manifest to the caller as one generic message deliberately, because "distinguishing them client-side would let a caller enumerate which keys exist". After this change they no longer do. 401 and 503 are now distinguishable outcomes of the same resolve call, and there is a reachable state in which that split correlates with key existence.

`ResolveSnapshot` (`apps/control-plane/internal/apikeys/service.go:359`) reads the key row first. A key that does not exist short-circuits at `GetPolicyByTokenHash` with `ErrNotFound`, rendered as 404, which still maps to 401 here. A key that does exist proceeds to `ListAllAliases`, `GetBudgetWindow`, `GetAccountRatePolicy`, `GetKeyRatePolicy` and `GetTenantIDByAccountID`, any of which can fail under connection-pool exhaustion or a cancelled request, producing a 500, which now maps to 503. Existing key gives 503, nonexistent key gives 401, same degraded window.

This is not hypothetical. The PR description quotes control-plane logging its own cancelled-context failures at exactly `get key rate policy` and `resolve tenant for account` during the 2026-08-14 incident, that is, after the key row had already been read. Raising the client timeout to 10s also makes those 500s reachable by edge-api where the previous 5s bound cut the call off before the response arrived, so this PR both creates the split and widens the window in which it can be observed.

Exploitation cost is low value: the attacker has to ride a degraded window rather than induce one, and no access is granted. Not a merge blocker by itself. But the comment above must stop asserting a property the code no longer has, and the tradeoff belongs in this comment and the commit record instead of being rediscovered later from a stale claim. Rate-limiting or aggregating repeated resolve failures per source would also shrink the window in which the oracle is usable.

apps/edge-api/cmd/server/main.go
```diff
@@ -1121,9 +1132,16 @@ func watchOWUIShimKey(ctx context.Context, resolver shimKeyResolver, shimKey str
 		cancel()
 		healthy := err == nil
 		if !reported || healthy != lastHealthy {
-			if healthy {
+			switch {
+			case healthy:
 				log.Printf("owui: OWUI_SHIM_KEY resolves to an active Hive API key on a tenant-provisioned account; Open WebUI model listing, document RAG embeddings, and text-to-speech can authenticate")
-			} else {
+			case errors.Is(err, authz.ErrUpstreamUnavailable):
```
sakibsadmanshajib: **MEDIUM: splitting the message without splitting the state makes the real alarm suppressible.**

The change-detection state is still a boolean (`healthy := err == nil` at line 1133, `lastHealthy = healthy` at line 1148), but the reported states are now three: healthy, transient-unavailable, and genuinely unusable. Both failure states collapse to `healthy = false`.

Sequence that loses the alarm entirely: control-plane is briefly unreachable, this branch logs the WARN ending in "do not rotate it over this alone", and `lastHealthy` becomes false. The key is then genuinely revoked, or its account loses its tenant mapping. The next probe returns a different error, but `healthy` is still false, so `!reported || healthy != lastHealthy` is false and nothing is printed. The last thing the operator ever sees is an instruction not to rotate a key that is now genuinely dead, and the ERROR line carrying the actual remedy never appears for the life of the process. Probe interval is 5 minutes, so this can persist for days.

Before this change both failures printed the same ERROR text, so collapsing them into one boolean was harmless. The split is what opens the hole. Track the reported state as a tri-state, or compare the classification rather than just `err == nil`, so a transition between the two failure kinds still logs.

This direction matters more than the one the PR set out to fix. A wrong alarm wastes an operator hour; a missing alarm is the empty model picker with no signal at all, which is the failure `watchOWUIShimKey` was written for. The added test only covers a transient failure from a fresh start, so it does not exercise this transition.

Verified while here, and correct: the probe cannot report healthy on a genuinely bad key. `checkOWUIShimKey` still returns an error for a non-active status, a missing tenant mapping, and an empty model allowlist, and `healthy` is strictly `err == nil`, so the inverse failure does not exist on this path.

apps/edge-api/internal/inference/orchestrator.go
```diff
@@ -83,17 +83,15 @@ func (o *Orchestrator) executeSync(
...
-		apierrors.WriteError(w, status, authErr.Error.Type, authErr.Error.Message, authErr.Error.Code)
+		// apierrors.WriteAuthFailure is the single source of truth for
+		// mapping an authz failure to a wire response (rate-limit 429,
+		// quota 429, model-not-found 404, upstream-unavailable 503,
+		// default 401); duplicating that switch here let it drift out of
+		// sync (it never gained the upstream_unavailable case, so a cold
+		// control-plane container fell through to 401 here even after
+		// Authorizer started returning 503 -- fixed by routing through the
+		// shared helper instead of a second copy of its logic).
+		apierrors.WriteAuthFailure(w, authErr, headers)
```
sakibsadmanshajib, line 94: **MEDIUM: the sibling sweep missed the third copy, so streaming /v1/responses still answers 401.**

`executeResponsesStreaming` in `apps/edge-api/internal/inference/stream_responses.go:49-61` carries a copy of the switch deleted here, and it is untouched on this branch:

```go
status := http.StatusUnauthorized
if authErr.Error.Type == "insufficient_quota" {
    status = http.StatusTooManyRequests
} else if authErr.Error.Code != nil && *authErr.Error.Code == "model_not_found" {
    status = http.StatusNotFound
}
```

A cold or unreachable control-plane therefore still returns `401 invalid_api_key` for a valid key on a streaming Responses API request. The PR description states that both duplicated switches were found and fixed, and that routing through the shared helper stops the drift. With a third site still holding its own copy, that claim is not yet true and the drift is live on a customer-facing endpoint.

The gap that let it through: nothing tests the new mapping end to end. `apps/edge-api/internal/errors/openai_test.go` gains no case asserting `upstream_unavailable` yields 503, and neither converted call site gets a handler-level test asserting a 503 actually reaches the wire. Every added test stops at the `Client` and `Authorizer` boundary, which is exactly why a call site that never invokes the mapper stays invisible to the suite. One table case in the errors package plus one handler test per execute path closes both.

Verified fine while here: `WriteAuthFailure` preserves prior behaviour for the other codes, including applying rate metadata headers only on the 429 path, and neither execute path writes SSE headers before `Authorize`, so the 503 leaves with a correct status line and a JSON body rather than being swallowed by an already-committed `text/event-stream` response.

apps/edge-api/internal/authz/client.go
```diff
@@ -76,11 +88,25 @@ func NewClient(baseURL string, redisURL string) (*Client, error) {
...
+// listening began) to complete its first resolve, so the previous 5s timeout
+// was already below the observed worst case and turned that cold-start
+// latency into a resolve failure on every deploy's post-deploy replay. 10s
+// gives roughly double that observed worst case as headroom. The trade-off is
+// real: a genuinely dead control-plane now takes up to 10s (was 5s) to surface
+// as a failure on the hot path. That is an acceptable trade now that a resolve
+// failure answers a retryable 503 (upstream_unavailable) instead of a
+// non-retryable 401 -- the failure being slower matters far less than it being
+// honest.
+const resolveClientTimeout = 10 * time.Second
```
sakibsadmanshajib, line 209: **MEDIUM: the timeout is per resolve, one request performs two of them, and nothing bounds how many can stack.**

The budget-gate middleware wraps the whole handler (`apps/edge-api/cmd/server/main.go:547`) and calls `authzClient.Resolve` itself (`main.go:939`) for any request carrying an Authorization header, before the route-level `Authorizer.Authorize` resolves again. There is no per-request memoization between the two. With control-plane wedged and the Redis snapshot cold, a single inbound request now occupies a goroutine for up to 2 x 10s = 20s, up from 10s.

Nothing bounds the accumulation. `NewClient` uses the default Transport, so `MaxConnsPerHost` is 0 (unlimited); the server leaves `ReadTimeout` and `WriteTimeout` at 0 deliberately (`main.go:592-595`); and there is no concurrency limiter in front of the handler. By Little's law the steady-state number of stuck in-flight resolves doubles at a fixed arrival rate, and each one holds a socket pointed at the control-plane that is already the thing failing, which is a feedback loop rather than a static cost. No valid credential is needed to enter it: the budget gate resolves whatever bearer is presented, and the resolve is only fast for a bogus key while control-plane is healthy enough to answer 404.

Doubling the window is defensible on its own. Doubling it while leaving the amplifier unbounded is what warrants a second look. Cheapest fixes in order: give the resolve client an explicit Transport with `MaxConnsPerHost` set, and stop resolving twice per request by stashing the gate result on the request context. Neither is required to land the honest-503 behaviour, but the first should not wait, because the failure it guards against is the same cold start this PR is about.

apps/edge-api/internal/errors/openai.go
```diff
@@ -89,6 +89,8 @@ func WriteAuthFailure(w http.ResponseWriter, oerr *OpenAIError, headers map[stri
 		status = http.StatusTooManyRequests
 	case oerr.Error.Code != nil && *oerr.Error.Code == "model_not_found":
 		status = http.StatusNotFound
+	case oerr.Error.Code != nil && *oerr.Error.Code == "upstream_unavailable":
```
sakibsadmanshajib, line 98: **LOW: the new 503 carries no Retry-After, while the sibling retryable failure does.**

The degraded-limiter path returns `retry-after: 5` alongside its 429 (`apps/edge-api/internal/authz/authorizer.go:172`), and `WriteAuthFailure` applies those headers on the rate-limit branch. This branch tells the caller to retry in prose and then gives nothing machine-readable, so SDK retry layers fall back to their own short backoff and hammer a control-plane that is by construction already unable to answer. `Authorize` returns nil headers for this case, and header application currently lives only inside `WriteRateLimitError`, so closing it means returning a small fixed `retry-after` from `Authorize` and applying headers on this branch too. Two lines, and it converts a retry storm into a paced retry.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in 166eb783. Classified via Content-Type rather than status code alone: control-plane's writeJSON (apikeys/http.go's handleKeyError) always sets application/json; Go's default ServeMux NotFoundHandler (what answers when the route is never mounted, issue #816) sets text/plain. A 404 with a JSON body is a genuine verdict and still 401s; a 404 with any other Content-Type is classified ErrUpstreamUnavailable -> 503.

Regression test TestResolveClassifiesUnmountedResolveRouteAsUpstreamUnavailable reproduces the real unmounted-route response with a bare *http.ServeMux (route deliberately never registered) rather than approximating it. TestResolveDoesNotClassifyGenuineNotFoundAsUpstreamUnavailable was also corrected to set Content-Type: application/json on its 404, matching real handleKeyError behavior (it previously used http.Error, which sets text/plain -- would have been misclassified as unmounted-route under the new logic, which is exactly the signal that makes this discriminator sound).

Also fixed the secondary note: the 409 branch's comment no longer claims revoked/disabled/not-active happen on this endpoint (ResolveSnapshot never returns those; the branch stays as defensive classification only, see 78c3a696).

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in 166eb783. This endpoint authenticates only edge-api's own X-Internal-Token (RequireInternalToken), never the customer's key, so a 401 from it can only mean CONTROL_PLANE_INTERNAL_TOKEN is misconfigured. Added ErrInternalTokenRejected, which wraps ErrUpstreamUnavailable (customer-facing response is unchanged: still the same retryable 503, since it genuinely is not their key's fault) but is now special-cased in authorizer.go's log line to a distinct, loud "CONFIGURATION ERROR ... will NOT self-resolve by waiting or retrying" message naming the token, per your ask, instead of blending into ordinary transient-failure noise.

Did not additionally special-case a misdirected baseURL (200 with a non-snapshot body) the same way: that's a boot-time config value that breaks 100% of requests immediately (would fail the first post-deploy smoke test), unlike a token mismatch that could pass CI/staging and only diverge in prod, so I judged it lower-value to special-case right now. Still gets ErrUpstreamUnavailable -> 503 (correct, just not the distinct log treatment). Happy to add if you want it in this pass.

Test: TestResolveClassifiesInternalTokenRejectionAsUpstreamUnavailable (client.go) and TestAuthorizeLogsConfigurationErrorLoudlyForRejectedInternalToken (authorizer.go).

apps/edge-api/internal/authz/authorizer.go
sakibsadmanshajib: Comment corrected in 78c3a696 -- you're right and the earlier version was wrong, retracting it rather than defending it. This IS a narrow, real oracle: control-plane's ResolveSnapshot fails fast at GetPolicyByTokenHash for a nonexistent key (404, still 401 here), but an existing key proceeds through ListAllAliases/GetBudgetWindow/GetAccountRatePolicy/GetKeyRatePolicy/GetTenantIDByAccountID, any of which can fail under exactly the connection-pool-exhaustion/cancellation conditions the 2026-08-14 incident logged ("get key rate policy", "resolve tenant for account: context canceled") -- so during a degraded window an existing key is more likely to 503 than a nonexistent one, which reliably fails fast.

Judgment, recorded explicitly in the new comment: accepted, not reverted. Exploiting it requires riding an already-degraded control-plane window an attacker cannot induce from here, yields only a probabilistic existence signal and no access, and the alternative (collapsing 503 back into 401) reintroduces the certain, severe cost this PR removes (a valid credential told it's invalid) to close a narrow, low-value leak. Your suggested follow-up (rate-limit/aggregate repeated resolve failures per source to shrink the window) is noted as future work, not done in this PR -- happy to file it as a tracked issue if you want it tracked rather than left in a code comment.

apps/edge-api/cmd/server/main.go
sakibsadmanshajib: Fixed in 166eb783. Replaced the boolean healthy with a tri-state owuiShimKeyState (healthy/transient/dead) and compare against that instead of err == nil, so a transition between "transiently unreachable" and "genuinely unusable" (both previously collapsed to healthy=false) now always re-logs. New regression test TestWatchOWUIShimKeyLogsDeadKeyAfterTransientFailure reproduces the exact sequence you described: transient failure logs WARN, then the same resolver is set to a genuine not-found error, and the test asserts the ERROR line with "Mint a replacement" still reaches the log (it would not have, under the old boolean).

apps/edge-api/internal/inference/orchestrator.go
sakibsadmanshajib: Fixed in 166eb783. stream_responses.go:49 now calls apierrors.WriteAuthFailure like the other two, deleting its own copy of the switch.

Enumeration proof: grepped every Authorize call site in apps/edge-api (9 total). audio/files/images/batches each wrap into authz.AuthzError in their authz_adapter.go and unwrap via authz.AsAuthzError plus apierrors.WriteAuthFailure in their handler.go (4 sites, already correct); main.go's handleModels calls apierrors.WriteAuthFailure directly (1 site, already correct); inference/orchestrator.go, stream.go, and stream_responses.go (3 sites) held their own inline copies of the switch, the first two fixed earlier in this PR, the third the one you found here. After this fix all 9 route through the same helper, directly or via AsAuthzError; zero inline copies of the switch remain in apps/edge-api.

Test coverage the review also asked for, added in 78c3a696: one handler-level test per Authorize call site in inference (executeSync, executeStreaming, executeResponsesStreaming in execute_authz_status_test.go) plus a WriteAuthFailure table case for upstream_unavailable in errors/openai_test.go, closing the gap where every previous test stopped at the Client/Authorizer boundary and never proved a call site actually used the mapper.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in 166eb783. Added Transport.MaxConnsPerHost (64) on the shared *authz.Client's http.Client. net/http's documented behavior (verified via go doc net/http Transport.MaxConnsPerHost): "On limit violation, dials will block", so this turns unbounded per-request growth during an outage into bounded queuing that still resolves via the existing client Timeout, no new pooling code needed.

Did not implement the second suggestion (memoize the budget-gate resolve onto the request context so a request only resolves once) in this pass, per your own note that neither fix is required to land the honest-503 behavior and MaxConnsPerHost is the one that should not wait. Agree it is a good follow-up: it would halve steady-state resolve volume, not just bound its growth. Left as a separate, smaller PR rather than folding it into this one, since it touches request-context plumbing across two unrelated call sites (BudgetGate + Authorizer) and this PR is already broad; happy to open the issue if you want it tracked.

apps/edge-api/internal/errors/openai.go
sakibsadmanshajib: Fixed in 78c3a696. Authorize now returns map[string]string{"retry-after": "5"} alongside the upstream_unavailable error, mirroring the existing degraded-limiter 429 pattern. WriteAuthFailure applies headers on the upstream_unavailable branch too now (factored the header-loop out of WriteRateLimitError into a small shared applyHeaders so both branches use the same code rather than a second copy of the loop). Table case added in errors/openai_test.go asserting the 503 carries retry-after: 5.

apps/edge-api/internal/authz/client.go
```diff
@@ -75,12 +99,45 @@ func NewClient(baseURL string, redisURL string) (*Client, error) {
 	}
 
 	return &Client{
-		cache:      &redisSnapshotStore{client: redis.NewClient(opt)},
-		httpClient: &http.Client{Timeout: 5 * time.Second},
-		baseURL:    strings.TrimRight(baseURL, "/"),
+		cache: &redisSnapshotStore{client: redis.NewClient(opt)},
+		httpClient: &http.Client{
+			Timeout:   resolveClientTimeout,
+			Transport: &http.Transport{MaxConnsPerHost: resolveMaxConnsPerHost},
```
sakibsadmanshajib: **MEDIUM (new, second pass): constructing a bare `&http.Transport{}` discards every `http.DefaultTransport` default, including the one that makes this cap work well.**

Setting `Transport:` replaces the default wholesale rather than adjusting it, so this client now runs on a zero-value Transport with only `MaxConnsPerHost` populated. What silently changes:

- `MaxIdleConnsPerHost` stays at the stdlib default of **2** while 64 concurrent connections are permitted. Every resolve beyond the second does a fresh TCP handshake and then throws the connection away, on the hot path, against the host this cap exists to protect. This is the direct counterpart of the setting that was added and it needs to move with it.
- `IdleConnTimeout` becomes **0, meaning never expire**, against `DefaultTransport`'s 90s. Pooled connections to a container that has since been recreated are kept indefinitely. That is precisely the event this PR is about. Go will retry a replayable request on a stale-connection error, so this is degradation rather than breakage, but it is an odd default to acquire in this particular change.
- `Proxy` is now nil, so `HTTP_PROXY` and `HTTPS_PROXY` are silently ignored. `CONTROL_PLANE_URL` is operator-supplied (`deploy/docker/docker-compose.yml:134`), and an enterprise self-hosted deployment behind an egress proxy would quietly stop honouring it.
- `ForceAttemptHTTP2` is false, so an https control-plane loses HTTP/2 multiplexing and needs one real socket per concurrent resolve, which is what makes the 64 ceiling bind sooner than it otherwise would.
- `TLSHandshakeTimeout` and `ExpectContinueTimeout` are unset. Both are covered by the 10s client `Timeout`, so this one is cosmetic.

The one-liner that keeps the intent and none of the loss:

```go
transport := http.DefaultTransport.(*http.Transport).Clone()
transport.MaxConnsPerHost = resolveMaxConnsPerHost
transport.MaxIdleConnsPerHost = resolveMaxConnsPerHost
```

apps/edge-api/internal/authz/client.go
```diff
@@ -75,12 +99,45 @@ func NewClient(baseURL string, redisURL string) (*Client, error) {
...
+// 5s in the same change that added this bound). With no cap, a sustained
+// control-plane outage could grow that to one goroutine and connection per
+// in-flight resolve with no ceiling -- at even 50 req/s sustained through a
+// 10s-timeout outage that is up to 1000 concurrent connections stacking up
+// against the one host that is already struggling, worsening the outage it
+// is trying to recover from. Go's Transport blocks new dials once this limit
+// is hit rather than erroring (net/http docs: "On limit violation, dials
+// will block"), so it turns unbounded growth into bounded queuing that still
+// resolves via the existing client Timeout -- no new pooling code needed.
+const resolveMaxConnsPerHost = 64
```
sakibsadmanshajib, line 195: **LOW (second pass): 64 is defensible, and the saturation behaviour is as the comment claims. Recording the arithmetic so the number stays a conscious one.**

Verified rather than taken on trust: a dial that blocks on this limit waits on the connection queue and is released by the request context, which `http.Client.Timeout` cancels, so a saturated pool fails at 10s with the same classified `ErrUpstreamUnavailable` and the same 503 rather than hanging. Bounded queuing, as advertised, and the right call against a control-plane whose own database pool is far smaller than 64.

Two things the comment should say and does not:

**Sockets are bounded, waiting goroutines are not.** The cap converts an unbounded connection flood into an unbounded goroutine queue with a 10s service time. That is a large improvement and still not a bound on memory or in-flight request count. Nothing in front of the handler limits concurrency.

**The ceiling it implies.** At the 5.65s cold-start latency this PR was written around, 64 connections is roughly 11 resolves per second; since one request drives up to two resolves, that is roughly 5 requests per second before every additional caller waits the full 10s and receives a 503. At a fully wedged control-plane (10s per resolve) it is about 3 requests per second. For the demo box that is comfortably above real traffic. For anything busier, 64 is the parameter that decides how much of a cold start is served versus shed, so it is worth stating in the comment as a shedding threshold instead of only as a protection against 1000 stacked connections.

No change required to merge. Put the two numbers in the comment so the next person tuning this knows which direction to move it and why.

apps/edge-api/internal/authz/client.go
```diff
@@ -126,23 +183,72 @@ func (c *Client) Resolve(ctx context.Context, rawToken string) (AuthSnapshot, er
...
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
 		respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
-		return AuthSnapshot{}, fmt.Errorf("authz: status %d: %s", resp.StatusCode, strings.TrimSpace(string(respBody)))
+		contentType := resp.Header.Get("Content-Type")
+		statusErr := fmt.Errorf("authz: status %d: %s", resp.StatusCode, strings.TrimSpace(string(respBody)))
+		switch {
+		case resp.StatusCode == http.StatusNotFound && strings.HasPrefix(contentType, "application/json"):
```
sakibsadmanshajib: **LOW to MEDIUM (second pass): the discriminator is correct for the failure it targets. It rests on one header that only control-plane is trusted to set, and the bytes to make it robust are already in hand.**

I attacked this the way it was asked to be attacked, by asking what else can answer a 404 on this URL.

The default deployment has no intermediary: `deploy/docker/docker-compose.yml:134` sets `CONTROL_PLANE_URL` to `http://control-plane:8081` on the compose network, so today the only two responders are control-plane's handler and its ServeMux, exactly the pair this switch separates, and the regression test builds the second from a real unregistered-route mux rather than an approximation. That part is sound and the HIGH from the first pass is genuinely closed for the shipped topology.

The value is operator-supplied, though, and `.env.example:375` already carries the pattern of a publicly reachable control-plane hostname, so a split-host or enterprise deployment putting a proxy in the path is not hypothetical. Survey of what an intermediary returns for an unrouted 404: Cloudflare serves HTML, nginx HTML, Traefik `text/plain`, Caddy an empty body. All four classify as upstream-unavailable, which is the safe direction. The unsafe ones are the JSON emitters: Kong answers `{"message":"no Route matched with those values"}` as `application/json`, AWS API Gateway answers `{"message":"Not Found"}` the same way, and a Workers route can return anything. Any of those reproduces the original false 401 in a new costume, and the failure would look identical to the incident that started this PR.

Closing it costs one comparison, using bytes already read into `respBody` a few lines above: require the body to decode as control-plane's own envelope with a non-empty `error` field, rather than trusting the header alone. Control-plane always writes `{"error":"key not found"}` through `writeJSON`, so the genuine path is unaffected and the existing test still passes. Content-Type plus body shape is a real signature; Content-Type alone is a convention any intermediary can satisfy by accident.

apps/edge-api/internal/authz/client.go
```diff
@@ -126,23 +183,72 @@ func (c *Client) Resolve(ctx context.Context, rawToken string) (AuthSnapshot, er
...
+			// ErrNotActive, but ResolveSnapshot (the only path
+			// handleInternalResolve calls) never returns any of those three --
+			// it returns a 200 snapshot carrying the key's real Status, and
+			// edge-api's own CheckAccess denies a revoked/disabled/expired key
+			// from that 200, not from a 409. If ResolveSnapshot's error
+			// surface ever changes to return them, this still classifies
+			// correctly as a genuine verdict rather than falling through to
+			// the default branch below.
+			return AuthSnapshot{}, statusErr
+		case resp.StatusCode == http.StatusUnauthorized:
```
sakibsadmanshajib, line 305: **LOW (second pass): the classification is right, the attribution is one word too certain, and the log line is unbounded.**

Confirmed clean on the part that matters: the customer-facing body is unchanged, still the fixed generic sentence with no mention of internal authentication, no status code, no upstream body, and the sentinel exists only to steer a server-side log. Also confirmed the log is reachable in practice: `log.Printf` to stderr is what the whole file uses and what the container ships, not a level nobody collects.

Two refinements.

**Attribution.** The reasoning holds for control-plane itself, whose `RequireInternalToken` is the only authenticator on this route. It does not hold for anything in front of it, and `CONTROL_PLANE_URL` is operator-supplied. Cloudflare Access rejecting a service token, or any proxy with its own authentication, also answers 401 on this URL, and the operator is then told with capital letters to go look at `CONTROL_PLANE_INTERNAL_TOKEN`, which is fine. Widening the wording to name control-plane or an intermediary in front of it keeps all the value and stops the line from being confidently wrong in the one deployment shape where it is hardest to debug.

**Volume.** By construction this condition is permanent and affects every request, so this line prints once per request for as long as it lasts. It is roughly six times the length of the generic line it replaces, and on the demo box that is a disk-filling loop bolted to an auth failure. The generic line had the same shape, so this is not a regression, but a permanent-fault line is the one place where a once-per-interval dedupe is clearly worth it: log it loudly the first time and then at most once a minute.

apps/edge-api/internal/authz/authorizer.go
```diff
@@ -98,7 +98,57 @@ func (a *Authorizer) Authorize(ctx context.Context, authHeader string, aliasID s
...
+			// at GetPolicyByTokenHash for a key that does not exist, before
+			// touching ListAllAliases/GetBudgetWindow/GetAccountRatePolicy/
+			// GetKeyRatePolicy/GetTenantIDByAccountID; a key that DOES exist
+			// proceeds through all of those, so during a degraded window
+			// (connection-pool exhaustion, request cancellation -- exactly
+			// what the 2026-08-14 incident's "get key rate policy" and
+			// "resolve tenant for account: context canceled" log lines were)
+			// an existing key is more likely to surface as 503 than a
+			// nonexistent one, which reliably 401s fast. Accepted rather than
+			// reverted: exploiting it requires riding an already-degraded
```
sakibsadmanshajib, line 184: **LOW (second pass): the acceptance is sound and the documentation is accurate, except for one clause that carries most of its weight.**

Judged on its merits: retraction is correct, the mechanism is described accurately (fail-fast at `GetPolicyByTokenHash` for a nonexistent key versus five further lookups for one that exists), the severity is judged correctly as a probabilistic existence signal yielding no access, and the alternative is correctly identified as strictly worse. I agree with accepting it, and it should not block this merge.

The clause that is overstated is "requires riding an already-degraded control-plane window an attacker cannot induce from here". Inducibility is not obvious in this deployment. The degraded window this depends on is connection-pool contention, and this control-plane runs against a small shared session-mode pool that ordinary load has been enough to saturate before, to the point of taking the chat surface down. A caller with nothing but valid API access can plausibly manufacture the window rather than only wait for one, which moves this from opportunistic to weakly attacker-triggerable. That does not change the verdict, since the payoff is still only a probabilistic existence hint, but an accepted risk should not rest its case on a premise this shaky. Reword to "an attacker cannot reliably induce" and the acceptance stands on its own.

The follow-up named in the last sentence, rate-limiting or aggregating repeated resolve failures per source, should be filed as an issue rather than left as a sentence in a comment. A mitigation that lives only in the comment of the code it mitigates is the one that never gets built.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in 0373874a. Now http.DefaultTransport.(*http.Transport).Clone(), then MaxConnsPerHost = 64 set on the clone. Also added MaxIdleConnsPerHost = 64 to move with it, per your one-liner (the default of 2 would have meant most of the 64 permitted connections got opened fresh and discarded instead of reused). Regression test asserts MaxConnsPerHost, the inherited IdleConnTimeout (proving Clone was used, not a bare struct), and the deliberately-raised MaxIdleConnsPerHost.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Comment expanded in d9869e94 with both numbers, no logic change (agreed, not required to merge). Now states: at the 5.65s cold-start latency this PR was written around, 64/5.65s ~11.3 resolves/sec, ~5.6 incoming requests/sec absorbed (matches your figure); at a fully wedged control-plane paying the full 10s timeout, 64/10s = 6.4 resolves/sec, ~3.2 incoming requests/sec. Also added the explicit sockets-bounded-goroutines-not caveat you raised: nothing in front of the handler limits total in-flight request count, a request-concurrency limiter would close that gap, not added here since it is broader than one HTTP client's Transport.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in d9869e94. Content-Type is no longer trusted alone. Added isControlPlaneErrorEnvelope: the 404 branch now also requires the body to decode as control-plane's own {"error": "<non-empty>"} shape (apikeys/http.go's writeJSON), using the bytes already read into respBody. Kong's {"message": "no Route matched..."} and API Gateway's {"message": "Not Found"} both have application/json Content-Type but neither decodes to a non-empty "error" field, so both now correctly classify as upstream-unavailable instead of a genuine key verdict.

New regression test TestResolveClassifiesIntermediaryJSON404AsUpstreamUnavailable reproduces exactly the gateway-shaped JSON 404 you described ({"message":"no Route matched with those values"}) and asserts it does NOT get misread as a genuine not-found. TestResolveDoesNotClassifyGenuineNotFoundAsUpstreamUnavailable (the existing genuine-verdict test) is unaffected since its body already matches the real envelope shape.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in d9869e94. Attribution reworded to name control-plane's own internal-token check as the likely cause without asserting it is the only one, explicitly acknowledging an intermediary in front of it (proxy, access gateway) with its own auth can answer the identical 401. Volume: rate-limited to once per minute via a package-level atomic timestamp gate (logInternalTokenRejectionOncePerMinute) rather than logging every request, since the condition is permanent by construction. New test TestAuthorizeRateLimitsRejectedInternalTokenLogging asserts a second Authorize call within the interval does not produce a second CONFIGURATION ERROR line.

apps/edge-api/internal/authz/authorizer.go
sakibsadmanshajib: Fixed in d9869e94. "cannot induce" reworded to "cannot RELIABLY induce", with the reasoning you gave: this control-plane's shared session-mode connection pool has previously been saturated by ordinary load alone, so a caller with only valid API access could plausibly help manufacture the degraded window rather than only wait for one. Acceptance judgment unchanged (still a probabilistic existence signal with no access, still a worse trade to revert and reintroduce the certain false-401 cost), just resting on the accurate premise now instead of the overstated one.

Filed issue #904 for the follow-up (rate-limit/aggregate repeated resolve failures per source) rather than leaving it only as a sentence in the comment, and referenced it there.

apps/edge-api/internal/authz/client.go
```diff
@@ -74,13 +117,93 @@ func NewClient(baseURL string, redisURL string) (*Client, error) {
...
+//   - at the 5.65s cold-start latency this PR was written around: 64 conns /
+//     5.65s ≈ 11.3 resolve calls/sec, and since one incoming request can
+//     drive up to two resolves (BudgetGate then Authorizer), that's roughly
+//     5.6 incoming requests/sec served before the next one queues;
+//   - at a fully wedged control-plane paying the full resolveClientTimeout
+//     (10s) per call: 64 / 10s = 6.4 resolve calls/sec, roughly 3.2 incoming
+//     requests/sec.
+//
+// Below the applicable number, a cold start or outage is absorbed; above it,
+// additional callers wait behind this cap (queued, not dropped -- see next
```
sakibsadmanshajib: **LOW (third pass), documentation accuracy: "queued, not dropped" understates what happens above the threshold, and this comment is written to be read during an incident.**

A caller blocked on the connection-count limit waits on Go's dial queue, and that wait is charged against the same `resolveClientTimeout` that bounds the resolve itself, because `http.Client.Timeout` starts before the dial and covers connection time. So a request that queues for the full 10s is not served late; it is dropped, having spent its entire budget without a resolve attempt ever being made, and it exits through the same `ErrUpstreamUnavailable` path as a real outage and answers a 503.

That makes the behaviour above the stated threshold sharper than "wait instead of failing fast". It is not graceful degradation with a longer tail: throughput plateaus at the computed figure and everything beyond it burns a full `resolveClientTimeout` before failing, so the marginal caller pays maximum latency for a guaranteed 503. That is still the right trade against an already-struggling control-plane, and it is the correct design. It just needs to be what the comment says, because the person raising this constant at 3am will be reading exactly this paragraph and deciding based on it.

The arithmetic itself checks out and I am not disputing any of it: 64/5.65 is 11.3 resolves per second and 5.6 requests per second at two resolves each, 64/10 is 6.4 and 3.2, and the note that the two constants trade against each other rather than independently is correct and worth keeping. Suggest replacing the "queued, not dropped" parenthetical with something like "queued behind the cap, then dropped with a 503 once resolveClientTimeout expires, having consumed the whole timeout in the queue".

apps/edge-api/internal/authz/authorizer.go
```diff
@@ -48,6 +50,37 @@ func newErr(errType string, message string, code *string) *apierrors.OpenAIError
...
+// last time logInternalTokenRejectionOncePerMinute actually logged. Package-
+// level and shared across every Authorizer instance deliberately: the
+// condition it throttles (a misconfigured shared secret) is process-wide,
+// not per-instance, so there is exactly one meaningful rate to enforce.
+var lastInternalTokenRejectionLogUnix atomic.Int64
+
+// logInternalTokenRejectionOncePerMinute logs a rejected-internal-token
+// CONFIGURATION ERROR at most once per internalTokenRejectionLogInterval.
+func logInternalTokenRejectionOncePerMinute(err error) {
+	now := time.Now().Unix()
```
sakibsadmanshajib, line 73: **Verified clean, with one nit. Answering the two questions this throttle was asked to survive.**

**Can it suppress a genuinely different failure?** No. The throttle is reached only from the `ErrInternalTokenRejected` branch; every other resolve failure, a timeout, a cancelled context, a control-plane 5xx, a genuine not-found, still goes through the unthrottled `else` and logs on every occurrence. Within the throttled branch the suppressed events are the same condition by construction, and since the message now names both possible causes (control-plane's own check, or an intermediary with its own authentication), an operator loses no actionable detail when the second cause's `err=` text is the one dropped. The only thing suppressed is repetition.

**Is the state race-safe?** Yes. Load, compare, then `CompareAndSwap` on an `atomic.Int64`: two goroutines arriving together cannot both log, since the loser of the CAS returns without logging, and there is no lost-wakeup counterpart because the winner always logs after winning. The zero initial value guarantees the first occurrence is never suppressed (`now - 0` is enormous), which is the property that matters most. The two tests reset the package-level counter explicitly and neither calls `t.Parallel`, so they are deterministic rather than order-dependent.

**Nit, take it or leave it:** `time.Now().Unix()` is wall clock, so a backwards NTP step of N seconds suppresses this line for N seconds even though the condition is live. A forwards step just logs early, which is harmless. Given this is the alarm for a fault that fails every request in the process, the failure direction of that one case is silence, which is the direction worth avoiding. Not worth a redesign; worth knowing it is there. Also note that dropping the interval below one second would truncate `int64(interval.Seconds())` to zero and disable the throttle entirely, which at least fails toward logging rather than toward silence.

apps/edge-api/internal/authz/client.go
```diff
@@ -50,6 +50,49 @@ func HashBearerToken(authHeader string) string {
...
+// isControlPlaneErrorEnvelope reports whether body decodes as control-plane's
+// own error envelope ({"error": "<non-empty>"} -- apikeys/http.go's writeJSON
+// and every handleKeyError branch write exactly this shape). Content-Type
+// alone is a convention an intermediary in front of an operator-supplied
+// CONTROL_PLANE_URL (reverse proxy, API gateway, CDN) can satisfy by
+// accident on its own unrouted-404 response; this body-shape check is the
+// signature that specific control-plane handler actually produced the
+// response, not merely that something along the path called itself JSON
+// (PR #903 second-pass security review).
+func isControlPlaneErrorEnvelope(body []byte) bool {
```
sakibsadmanshajib, line 62: **Verified clean, both directions, empirically rather than by reading. This was the highest-value thing to attack and it holds.**

I ran the exact predicate against the body shapes that matter, in the toolchain image, rather than reasoning about `encoding/json` from memory. Results:

| Body | Envelope check | Classification |
|---|---|---|
| `{"error":"key not found"}` plus the trailing newline `json.Encoder.Encode` appends | **true** | genuine verdict, 401 |
| `{"error":"key not found","code":"x"}` | **true** | genuine verdict, 401 |
| Kong, `{"message":"no Route matched with those values"}` | false | 503 |
| AWS API Gateway, `{"message":"Not Found"}` | false | 503 |
| edge-api's own nested `{"error":{"message":...}}` | false | 503 |
| `{"error":""}`, `{"error":null}`, `{"error":404}` | false | 503 |
| HTML, and ServeMux's `404 page not found` | false | 503 |
| body truncated mid-token by the 4096-byte `LimitReader` | false | 503 |

The trailing newline case is the one that would have been easy to get wrong, since `writeJSON` uses `Encode` rather than `Marshal`; `json.Unmarshal` tolerates trailing whitespace, so the genuine path is unaffected.

**Direction that fails open** (a genuine key-not-found decoding as something else, letting an invalid key retry forever): the only triggers are a body that is not control-plane's envelope. That means either an intermediary rewriting the response, or `handleKeyError`'s body growing past 4 KiB, which it cannot today at 26 bytes. Both send an invalid key to a 503 that grants nothing and can be retried forever without ever succeeding, so the cost is a confusing status code, not access. Acceptable, and it is the correct side to fail on given the alternative is the false 401 this PR exists to remove.

**Direction that fails closed** (an intermediary forging a genuine verdict): requires emitting a flat `{"error":"<non-empty string>"}` on an unrouted 404. None of the common fronting layers do; all four checked land on 503. The residual is that a flat string error envelope is a generic shape, so this is a signature meaning "not any of the usual intermediaries" rather than a proof of origin. Tightening further, to the literal message text, would couple edge-api to control-plane's copy and break on a wording change, so not doing that is the right call.

Nothing to change here.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Fixed in c0277ed3, using your suggested wording. The parenthetical now reads that a caller waits behind the cap and is then dropped with a 503 once resolveClientTimeout expires, having spent the whole timeout in the queue rather than served late. No logic change, comment only.

apps/edge-api/internal/authz/authorizer.go
sakibsadmanshajib: Acknowledged, thank you for verifying by execution. Agreed on the wall-clock nit: time.Now().Unix() can suppress up to N seconds early on a backward NTP step, and the failure direction there is silence on a fault that already fails every request in the process. Not redesigning to a monotonic source for this given the interval is one minute and the condition being throttled is already loud and permanent, but noting it here per your own framing so it's not lost. Also confirmed: an interval below one second would truncate to zero via int64(interval.Seconds()) and disable the throttle, failing toward logging rather than silence, which is the correct failure direction if this constant is ever changed.

apps/edge-api/internal/authz/client.go
sakibsadmanshajib: Acknowledged, thank you for running the twelve-body matrix rather than reasoning about it. Good catch calling out the Encode trailing-newline case specifically, since that's the one that would have silently broken this on the genuine path if json.Unmarshal were stricter about trailing whitespace. Agreed on both failure-direction conclusions and agreed that tightening the envelope check to control-plane's literal message text would trade a wording-change footgun for a marginal signal improvement that isn't worth it. Nothing further from me here.

PR #878, opened 2026-08-11
fix: require an active membership before granting platform admin or workspace owner

sakibsadmanshajib:
Closes #877.

## What was reachable before this change, and by whom

`IsPlatformAdmin` selected membership rows on `role = 'owner'` and `accounts.is_platform_admin = true`, with no predicate on `account_memberships.status`. Status is either `active` or `invited`, and an invited row records an offered seat, not an accepted one.

Anyone holding an invited owner membership on a platform admin flagged account therefore passed `RequirePlatformAdmin` without accepting anything, which gave them:

- `POST /v1/admin/credit-grants`, the endpoint that mints spendable credit.
- `/api/v1/admin/providers/*`, provider administration including base URL rewrites.
- the `platform.admin` overlay in the permission set the console renders from.

The same missing predicate in `GetMembershipRole` gave an invited owner `IsWorkspaceOwner`, which gates workspace budget writes and the egress policy surface. The invoices access checker in `cmd/server` matched any membership row on the workspace id, so an invited seat could read that workspace's invoices. `EnsureViewerContext` could select an invited membership as the current workspace, and the actor resolver reads the chosen membership's role, so the invited role became the acting role.

On today's code no Go path writes an invited account membership row (the accept flow inserts an active row directly), so this is a gate that trusts a column it never reads rather than a click through path in the shipped invite UI. The schema represents the state, the CHECK constraint allows it, and any writer with table access creates it. The gate has to be right on its own terms.

## Sites changed, and why each is the same defect

| File | Change | Why it is the same defect |
| --- | --- | --- |
| `apps/control-plane/internal/platform/role_pgx.go` | `AND m.status = 'active'` in `IsPlatformAdmin`, `AND status = 'active'` in `GetMembershipRole` | The reported site, and its sibling in the same file backing `IsWorkspaceOwner`. Both answered an authorization question from a row whose status they never read. |
| `apps/control-plane/internal/accounts/actor_resolver.go` | `ActorFor` blanks the role for a membership that is not active | Defence in depth, and nothing more. All thirteen call sites pass `StatusActive` as a literal or have already filtered to an active row, so this branch is unreachable in production today; the protection on those paths is `EnsureViewerContext` and `GetMembershipRole`. This exists so a future handler that passes a row straight from the database gets a denial rather than an escalation. The platform-admin overlay is untouched: it comes from its own query, which applies its own active-membership predicate. |
[...]

PR #906, opened 2026-08-16
fix: memoize control-plane session revalidation per request

sakibsadmanshajib:
## Summary

`tests/e2e/console-budgets.spec.ts`'s "budget caps are editable for the owner and disabled for a member" intermittently fails `expect(locator).toBeVisible()` on the page's static "Budget settings" heading. A downloaded CI artifact (`error-context.md` from PR #891's run) shows why: the console's own generic error boundary had rendered ("Something went wrong on this page") instead of the budget page at all.

## The two hypotheses this was asked to confirm or refute, both refuted with evidence

1. **PR #878 split a shared 403 response code (`email_verification_required`) into two (`permission_denied` / `email_verification_required`) in `apps/control-plane/internal/budgets/http.go`.** True, but that split lives entirely in `resolveCurrentAccountID`, which only backs the legacy `/api/v1/accounts/current/budget` endpoints. The console's budget settings page and this test both call the separate Phase 14 workspace surface (`/api/v1/budgets/{workspace_id}`, `requireWorkspaceOwner`/`requireWorkspaceMembership`), which PR #878 did not change in behavior (only a literal-to-constant rename, same value). This code path is not reachable from the failing test.
2. **"PR #878 touched only Go files, so CI's path filter marked Web E2E as skipped on main."** Also false. PR #878's diff includes `apps/web-console/app/console/account-switch/route.ts`, `apps/web-console/components/workspace-switcher.tsx`, `apps/web-console/app/invitations/accept/page.tsx`, and a unit test. Its own PR ran `Web E2E (full stack)` and it was green (`gh pr view 878` shows conclusion SUCCESS), and the direct push-to-main run for that exact merge commit also ran the suite green with 0% flake. Independently, `.github/workflows/ci.yml`'s `web_e2e` path filter already treats `apps/control-plane/*` as a trigger (line ~249), so even a Go-only change under `apps/control-plane/internal/*` would not have been skipped. Audited the filter against `go.work`'s actual members; it already covers everything this job boots. No blind spot found for this scenario.

## The real, demonstrated defect

`lib/control-plane/client.ts`'s `getRequestContext()` calls `supabase.auth.getUser()`, a real network round trip to Supabase Auth, and is called by every one of the ~45 exported client functions before their own control-plane fetch. Nothing memoized it. One navigation to `/console/billing/budget` fires it 3 times from `page.tsx` (`getViewer`, `getAccountProfile`, `getBudget`) plus 3 more from the parent `layout.tsx` (`getViewer`, `getBalance`, `getBudgetThreshold`): up to 6 independent external round trips per page load. `getViewer()` and `getAccountProfile()` had no failure handling for a non-404 error (unlike the adjacent `getBudget(...).catch(() => null)`), so any one of those redundant round trips hitting a transient upstream hiccup threw uncaught and took down the entire page.

The confirmed benefit of memoizing that function is scoped to the **~16 `page.tsx`/`layout.tsx` Server Components** that call it (this was verified with a grep, not assumed). Roughly 15 of the ~45 callers are App Router Route Handlers under `app/api/**/route.ts`; React's `cache()` dedup only applies within a Server Component render, so it does nothing there, but each of those handlers calls a `getRequestContext`-backed function once per invocation anyway, so there is no multi-call dedup opportunity lost.

## Fix

Three parts, the second and third added after `typescript-reviewer`'s pass on this PR flagged that the first part alone reduces the crash's frequency without closing it:

1. **Memoization.** Wrapped `getRequestContext()` in React's `cache()`, so every control-plane client call within one server request shares a single session-revalidation result instead of repeating it once per call. Cuts the budget page's per-load session round trips from up to 6 to 1, and does the same for the other 7 console pages that make 3+ such calls per render.
2. **Retry.** `getRequestContext()` now retries `supabase.auth.getUser()` once before giving up. `cache()` memoizes a rejected promise exactly the same way it memoizes a resolved one, so without this, one failure on the single shared call would still crash the page, just less often than the pre-fix six independent chances. A transient upstream hiccup usually clears on a second attempt a moment later; a genuinely revoked or invalid token fails identically both times, so this cannot turn a real refusal into a false accept.
3. **Closing the throw path on this page.** `app/console/layout.tsx` (which wraps every console page, including budget) and `app/console/billing/budget/page.tsx` now catch a `getViewer()` failure and redirect to `/auth/sign-in`, the same destination an expired session already takes, instead of letting the throw reach the generic error boundary. There is nothing else either can render without a viewer (no role, no workspace id, no email). The budget page's `getAccountProfile()` failure degrades to the same `EMPTY_ACCOUNT_PROFILE` shape a fresh account's 404 response already renders, since a profile-fetch failure is not a session problem and the page still knows who the viewer is.

Not fixed here, flagged as a follow-up: the other ~16 console pages that call `getViewer()` directly still have the bare crash-on-throw path this PR closes only for `layout.tsx` (and therefore every page it wraps) and the budget page's own second call site (`getAccountProfile`).

## Verification

- `npx tsc --noEmit`: clean.
- `npx vitest run` (web-console): 530/530 passing (527 + 3 new).
- `tests/e2e/console-budgets.spec.ts` run 9 times total against a locally rebuilt stack across the two commits on this branch (8 consecutive runs before the review-response commit, 1 more after it), fresh run-scoped fixtures per run per `docs/live-test-auth.md`: 9/9 passed.
- CI's own `Web E2E (full stack)` check on this PR passed (11m33s) before the review-response commit.
- Added `__tests__/control-plane-get-user-retry.test.ts` for the retry bound in both directions (a transient failure followed by success must succeed; a call that keeps failing must not retry a third time). Verified red without the fix (`git stash` of `client.ts`, confirmed the test fails), green with it.
- The `cache()` memoization itself remains unguarded by a unit test: it only activates under the `react-server` module condition Next.js's bundler selects for real Server Components, which this project's jsdom-based vitest never applies. Wrote and ran an equivalent test; it failed identically with and without the fix, so it was removed rather than shipped as a check that cannot fail for the right reason. The alternative suggested in review (a Playwright network-interception count) does not apply either: `getUser()` runs server-side, inside the Next.js process talking to Supabase Auth over its own connection, which is invisible to Playwright's `page.route()` (browser-initiated requests only). Left guarded by the 9 consecutive green e2e runs above rather than by an automated regression check for this specific property.

## Buglog entry

```json
{"id": "BUG-2026-08-16-console-request-context-unmemoized", "date": "2026-08-16", "title": "Unmemoized per-call Supabase session revalidation crashed console pages on any transient hiccup", "error_message": "tests/e2e/console-budgets.spec.ts intermittently failed toBeVisible on a static page heading; a CI artifact showed the Next.js generic error boundary (Something went wrong on this page) had rendered instead of the budget settings page", "root_cause": "lib/control-plane/client.ts's getRequestContext() calls supabase.auth.getUser(), a real network round trip, and every one of its ~45 exported client functions calls it first with no memoization. The budget settings page alone makes 3 such calls (getViewer, getAccountProfile, getBudget) and its parent layout makes 3 more (getViewer, getBalance, getBudgetThreshold) on the same navigation, for up to 6 independent round trips per page load. getViewer() and getAccountProfile() had no failure handling for a non-404 error, unlike the adjacent getBudget().catch(() => null), so any one of those six calls hitting a transient upstream hiccup threw uncaught and crashed the whole Server Components tree. Investigated as a suspected regression from PR #878's budgets/http.go 403-code split and Go-only-PR path-filter skip; both refuted: that split only reaches the legacy /api/v1/accounts/current/budget endpoints this page never calls, #878 also touched web-console files and its own CI run (and the direct push-to-main run) were green with 0% flake, and ci.yml's web_e2e path filter already treats apps/control-plane/* as a trigger.", "fix": "Wrapped getRequestContext() in React's cache() so every control-plane client call within one server request shares a single session-revalidation result (scoped to the ~16 page.tsx/layout.tsx Server Components that see the benefit; ~15 Route Handler callers make one such call per invocation regardless and gain nothing from the dedup). Added a single retry on the underlying getUser() call, since cache() memoizes a rejection the same as a resolved value and would otherwise still crash on one failure per request instead of six. Closed the remaining throw path on the console layout and the budget page specifically: both now catch a getViewer() failure and redirect to sign-in instead of reaching the generic error boundary, and the budget page's getAccountProfile() failure degrades to the existing empty-profile shape. The other console pages that call getViewer() directly remain a separate follow-up.", "tags": ["web-console", "control-plane", "flaky-test", "performance", "resilience", "ci-investigation"], "issue": null}
```

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run` -- 530/530 passing
- [x] `tests/e2e/console-budgets.spec.ts` -- 9 total local runs across both commits, 9/9 passed
- [x] CI `Web E2E (full stack)` passed on this PR
- [x] Confirmed via `gh pr view 878` and `gh api .../check-runs` that PR #878's own CI run and the direct push-to-main run both executed and passed `Web E2E (full stack)` with 0% flake, refuting the "invisible on main" premise
- [x] Confirmed `apps/control-plane/internal/budgets/http.go`'s 403-code split only affects `/api/v1/accounts/current/budget`, not the workspace budget surface this page and test use
- [x] Audited `.github/workflows/ci.yml`'s `web_e2e` path filter against `go.work` members; no gap found for this scenario
- [x] Regression guard for the retry bound (red-without/green-with confirmed); `cache()` dedup itself deliberately left without a unit-level guard, reasoning above

### DIFF
--- a/apps/web-console/lib/control-plane/client.ts
+++ b/apps/web-console/lib/control-plane/client.ts
@@ -52,6 +52,20 @@ export interface AccountProfile {
   profile_setup_complete: boolean;
 }
 
+// The not-yet-set-up shape: a fresh account with no profile row (control-plane
+// 404s), and the fallback a page takes when the profile fetch itself fails.
+// Same reasoning as the 404 case below — render the needs-setup state rather
+// than crash.
+export const EMPTY_ACCOUNT_PROFILE: AccountProfile = {
+  owner_name: "",
+  login_email: "",
+  display_name: "",
+  account_type: "",
+  country_code: "",
+  state_region: "",
+  profile_setup_complete: false,
+};
+
 export interface UpdateAccountProfileInput {
   ownerName: string;
   loginEmail: string;
@@ -125,26 +139,32 @@ interface RequestContext {
   headers: Record<string, string>;
 }
 
-// Memoized per request with React's cache(): every exported function below
-// calls this before its own control-plane fetch, and a single server page can
-// legitimately call several of them (the budget settings page alone calls
-// getViewer, getAccountProfile, and getBudget; its parent layout calls
-// getViewer, getBalance, and getBudgetThreshold on the very same navigation).
-// Unmemoized, that is six separate real network round-trips to Supabase
-// Auth's getUser() for one page load, each one a chance for a transient
-// upstream hiccup to throw "No active session" and crash the whole Server
-// Components tree — confirmed live: a CI run of
-// tests/e2e/console-budgets.spec.ts failed with the console's generic error
-// boundary ("Something went wrong on this page") on a page whose only two
-// unguarded upstream calls are getViewer() and getAccountProfile(), and eight
-// of the eighteen console pages (this one included) make three or more of
-// these calls on one render, with the shared layout adding four more of its
-// own on every single one of them. cache() scopes the memoized
-// promise to a single request (Next.js's per-request React cache, reset on
-// the next navigation), so this does not change session freshness across
-// page loads, only how many times one page load re-derives it. Cuts the
-// external-call surface, and therefore the flake surface, without weakening
-// or skipping the session check itself.
+// Memoized per request with React's cache(): confirmed benefit is scoped to
+// the ~16 page.tsx/layout.tsx Server Components that call these exports (the
+// budget settings page alone calls getViewer, getAccountProfile, and
+// getBudget; its parent layout calls getViewer, getBalance, and
+// getBudgetThreshold on the same navigation — up to 6 calls per page load,
+// unmemoized). cache() dedup is scoped by React to the render of that
+// component tree; it does not apply to the ~15 Route Handlers under
+// app/api/**/route.ts that import this module, but each of those calls a
+// getRequestContext-consuming function once per invocation, so there is no
+// multi-call dedup to lose there either way.
+//
+// Each of those (up to 6, per Server Component render) calls was a real
+// network round trip to Supabase Auth's getUser(), and a chance for a
+// transient upstream hiccup to throw "No active session" — confirmed live: a
+// CI run of tests/e2e/console-budgets.spec.ts failed with the console's
+// generic error boundary ("Something went wrong on this page") instead of
+// the budget page. cache() collapses those up-to-6 chances into 1 per
+// request (scoped to that one request; reset on the next navigation, so this
+// does not change session freshness). The one-retry below closes most of
+// what's left of that one remaining chance. Neither eliminates it: a retry
+// exhausted or a memoized rejection is still a thrown error, and the two
+// call sites that throw uncaught on it (getViewer, getAccountProfile in
+// app/console/billing/budget/page.tsx and its layout) now catch it and
+// redirect to sign-in instead of letting it reach the generic boundary — see
+// those two files. The other 16 console pages that call getViewer() directly
+// still have the bare crash-on-throw path; known follow-up, not closed here.
 const getRequestContext = cache(async (): Promise<RequestContext> => {
   const cookieStore = await cookies();
   const supabase = createClient(cookieStore);
@@ -152,10 +172,21 @@ const getRequestContext = cache(async (): Promise<RequestContext> => {
   // Validate the JWT against Supabase (getUser does a server round-trip
   // and rejects revoked tokens) before trusting the session. getSession
   // only reads the cookie and would accept a revoked token.
-  const {
-    data: { user },
-    error: userError,
-  } = await supabase.auth.getUser();
+  //
+  // One retry only: a transient upstream hiccup (the failure class that
+  // crashed the budget page, see above) usually clears on a second attempt
+  // a moment later. A genuinely revoked or invalid token fails identically
+  // both times, so this cannot turn a real refusal into a false accept.
+  let user: Awaited<ReturnType<typeof supabase.auth.getUser>>["data"]["user"] =
+    null;
+  let userError: Awaited<ReturnType<typeof supabase.auth.getUser>>["error"] =
+    null;
+  for (let attempt = 0; attempt < 2; attempt++) {
+    const result = await supabase.auth.getUser();
+    user = result.data.user;
+    userError = result.error;
+    if (!userError && user) break;
+  }
   if (userError || !user) {
     throw new Error("No active session");
   }
@@ -905,15 +936,7 @@ export async function getAccountProfile(): Promise<AccountProfile> {
   // and billing pages can render their needs-setup state instead of
   // crashing the whole Server Components tree.
   if (response.status === 404) {
-    return {
-      owner_name: "",
-      login_email: "",
-      display_name: "",
-      account_type: "",
-      country_code: "",
-      state_region: "",
-      profile_setup_complete: false,
-    };
+    return EMPTY_ACCOUNT_PROFILE;
   }
 
   if (!response.ok) {
--- a/apps/web-console/lib/control-plane/client.ts
+++ b/apps/web-console/lib/control-plane/client.ts
@@ -1,3 +1,4 @@
+import { cache } from "react";
 import { cookies } from "next/headers";
 import { createClient } from "@/lib/supabase/server";
 import {
@@ -124,7 +125,27 @@ interface RequestContext {
   headers: Record<string, string>;
 }
 
-async function getRequestContext(): Promise<RequestContext> {
+// Memoized per request with React's cache(): every exported function below
+// calls this before its own control-plane fetch, and a single server page can
+// legitimately call several of them (the budget settings page alone calls
+// getViewer, getAccountProfile, and getBudget; its parent layout calls
+// getViewer, getBalance, and getBudgetThreshold on the very same navigation).
+// Unmemoized, that is six separate real network round-trips to Supabase
+// Auth's getUser() for one page load, each one a chance for a transient
+// upstream hiccup to throw "No active session" and crash the whole Server
+// Components tree — confirmed live: a CI run of
+// tests/e2e/console-budgets.spec.ts failed with the console's generic error
+// boundary ("Something went wrong on this page") on a page whose only two
+// unguarded upstream calls are getViewer() and getAccountProfile(), and eight
+// of the eighteen console pages (this one included) make three or more of
+// these calls on one render, with the shared layout adding four more of its
+// own on every single one of them. cache() scopes the memoized
+// promise to a single request (Next.js's per-request React cache, reset on
+// the next navigation), so this does not change session freshness across
+// page loads, only how many times one page load re-derives it. Cuts the
+// external-call surface, and therefore the flake surface, without weakening
+// or skipping the session check itself.
+const getRequestContext = cache(async (): Promise<RequestContext> => {
   const cookieStore = await cookies();
   const supabase = createClient(cookieStore);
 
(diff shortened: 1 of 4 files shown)


apps/web-console/lib/control-plane/client.ts
```diff
@@ -124,7 +125,27 @@ interface RequestContext {
...
+// unguarded upstream calls are getViewer() and getAccountProfile(), and eight
+// of the eighteen console pages (this one included) make three or more of
+// these calls on one render, with the shared layout adding four more of its
+// own on every single one of them. cache() scopes the memoized
+// promise to a single request (Next.js's per-request React cache, reset on
+// the next navigation), so this does not change session freshness across
+// page loads, only how many times one page load re-derives it. Cuts the
+// external-call surface, and therefore the flake surface, without weakening
+// or skipping the session check itself.
+const getRequestContext = cache(async (): Promise<RequestContext> => {
```
sakibsadmanshajib, line 168: **Scope claim needs narrowing: many of the ~45 callers never render inside an RSC tree.**

Grepped every importer of this module. Roughly 15 files under `app/api/**/route.ts` (spend-alerts, budget, invoices/pdf, marketplace enable, members, feature-gates, checkout/intent, accounts/current/[...path], account-switch, provision, ...) call these exported functions from App Router Route Handlers, not from `page.tsx`/`layout.tsx` Server Components. Next's own docs scope `cache()` dedup to "the lifetime of a server request until the React component tree finishes rendering" — Route Handlers never render a React tree. Whether the memoization is actually live there, or silently a pass-through (the same failure mode the author already found and used to justify deleting the unit test), is unverified in this PR.

In practice this may not matter much: skimmed the route handlers and each one appears to call a getRequestContext-consuming function once per invocation, so there's no multi-call dedup opportunity to lose there regardless. But the PR's own doc comment (lines 128-146) frames the win in terms of "every exported function below" and "~45 exported client functions," which reads as broader than what's demonstrated. Recommend the comment/PR body explicitly scope the confirmed benefit to the Server Component page/layout call sites (the ~16 page.tsx/layout.tsx files), not all 45 exports.

sakibsadmanshajib, line 168: **cache() memoizes the rejected promise too, so the actual crash defect is unchanged, only rolled less often.**

`getViewer()` and `getAccountProfile()`, named in the PR body as the two unguarded call sites that crash the page, are still unguarded after this diff. Before this change, each of the up to 6 call sites made its own independent `getUser()` round trip; now they all `await` the same memoized promise. React's `cache()` caches a rejection exactly like a resolved value: once the shared call rejects, every caller in that request, guarded and unguarded alike, receives that same rejected promise, no retry.

Net effect on crash probability is genuinely positive (collapsing up to 6 independent chances-to-fail into 1 lowers the odds any given page load hits the flake at all), so the fix is not wrong. But it does not fix the uncaught-throw path itself: a real failure on that single shared call still takes the page down exactly as before, just less frequently. The PR body should say this plainly ("reduces frequency, does not remove the crash") instead of leaving the impression that the two unguarded functions are covered by this diff. They aren't; that's the deferred follow-up.

apps/web-console/lib/control-plane/client.ts
```diff
@@ -163,7 +184,7 @@ async function getRequestContext(): Promise<RequestContext> {
   }
 
   return { baseUrl, headers };
-}
+});
```
sakibsadmanshajib, line 218: **No regression guard ships with this fix.**

Deleting the jsdom-based dedup test that couldn't fail was the right call, but it leaves this change with zero automated coverage. A future refactor (someone destructures/rewraps this export, a bundler upgrade changes which `react` build gets resolved for a given layer, someone "simplifies" this back to a plain `async function`) regresses silently: nothing in `vitest run` or typecheck will catch it, because the property being verified only exists under the `react-server` module condition that jsdom never applies.

The honest check here is at the e2e layer, against a real Next.js server build where the condition is live, not a layer where it's guaranteed absent: e.g. `tests/e2e/console-budgets.spec.ts` already loads this exact page and could assert a bounded count of Supabase `getUser` calls via request interception (Playwright `page.route`/network log) on the real budget-settings navigation. That would actually fail if memoization regresses. Worth adding, or at minimum call out its absence as known follow-up debt in the PR body rather than leaving the impression the fix has no verifiable regression surface at all.

apps/web-console/lib/control-plane/client.ts
sakibsadmanshajib: Confirmed and fixed. Corrected the doc comment to scope the benefit to the ~16 page.tsx/layout.tsx Server Components, and added a line noting the ~15 Route Handler callers make one call per invocation regardless so there is no dedup opportunity lost there either way. PR body updated with the same correction.

sakibsadmanshajib: Agreed, and fixed rather than just disclosed. Added a single retry on the underlying getUser() call (closes most of what remains of the one shared chance to fail), and closed the throw path itself on this page: app/console/layout.tsx and the budget page now catch a getViewer() failure and redirect to /auth/sign-in instead of reaching the generic error boundary, and the budget page's getAccountProfile() failure degrades to the existing empty-profile shape. Reasoning for redirect-to-sign-in over any other fallback: neither call site can render anything meaningful without a viewer (no role, no workspace id, no email), and sign-in is already the destination an expired session takes, so this doesn't introduce a new UX state. The other ~16 console pages that call getViewer() directly still have the bare crash-on-throw path; that remains the disclosed follow-up, now correctly scoped in the PR body to just those pages rather than implying this page was covered when it wasn't.

apps/web-console/lib/control-plane/client.ts
sakibsadmanshajib: Added apps/web-console/__tests__/control-plane-get-user-retry.test.ts, which guards the retry bound added in response to your second finding (both directions: transient-then-success must succeed, still-failing must not retry a third time). Verified red without the fix via git stash of client.ts, green with it. Your suggested Playwright network-interception check for the cache() dedup itself doesn't apply here though: getUser() runs server-side inside the Next.js process talking to Supabase Auth over its own outbound connection, which page.route() cannot see (it only intercepts browser-initiated requests). Left the dedup itself unguarded by an automated check, relying on the 9 consecutive green local e2e runs plus the CI Web E2E pass already recorded on this PR; said so explicitly in the PR body rather than leaving the impression of full coverage.

sakibsadmanshajib:
## Visual proof (Rule 8), head SHA 44786ad55180a7b8b0b4fcf6e6934a623d68e025

Captured locally against a stack rebuilt from this exact commit (`docker compose --profile local --profile dev up --build`), not a dev server serving unbuilt source. Full capture log in the proof commit: `docs/proof/budget-page-crash-guard-2026-08-16/proof-log.txt`.

### 1. Normal path — budget page renders unbroken

budget page normal render

Signed in as a fixture-seeded, run-scoped owner. `/console/billing/budget` renders the real form, soft/hard cap inputs enabled.

### 2. Failure path — getViewer() failure redirects to sign-in, not the error boundary

getViewer failure redirects to sign-in

Induced honestly, at the boundary: stopped the `control-plane` container (`docker compose stop control-plane`) with the same signed-in session still active, then navigated to `/console/billing/budget` again with no changes to the component. `getViewer()`'s fetch to control-plane fails for real (connection refused), the one retry in `getRequestContext()` exhausts, and the page's catch redirects to `/auth/sign-in` — the plain sign-in form, not "Something went wrong on this page." `control-plane` was restarted immediately after.

### Redaction

Neither screenshot needs masking: Playwright's `page.screenshot()` captures page content only, never the browser address bar, so there is no URL in frame to leak a token from. No magic link, OAuth callback, or one-time-token flow was exercised here (fixture email/password sign-in, then a saved `storageState` reused between the two captures, never committed). `npm run lint:proof-tokens` ran clean against the proof directory (58 files scanned).

Issue #856, opened 2026-08-11
Console overview and analytics read zero after live chat traffic on the same workspace


sakibsadmanshajib:
## What happens

Two chat completions were served successfully to the demo account through `hive-auto` on 2026-08-11. Minutes later, the console for the same workspace reports nothing:

* Overview: Available credits `0`, Posted `0`, Reserved `0`, Today's activity renders as an em dash with the caption "Detailed counts available in Analytics", Recent errors renders as an em dash with the caption "Error telemetry not yet wired up".
* Analytics, 24h window: TOTAL REQUESTS `0`, INPUT TOKENS `0`, OUTPUT TOKENS `0`, CREDITS SPENT `0`.

## Why it matters for a demo

The console is the part of the product that proves the gateway is a metered, governed platform rather than a chat window. Walking an audience from a live conversation to a usage page that reads zero invites the obvious question, and there is no good answer to it on stage.

There is a second, larger question hiding behind the same screens. The workspace balance is zero and requests were still served. Either the chat surface is not attributed to this workspace at all, or the requests were served without being metered. Both readings deserve a look, and the second one is a billing correctness matter, not a presentation matter.

## Evidence

Overview after live chat traffic:

![Console overview reads zero](https://raw.githubusercontent.com/sakibsadmanshajib/hive/fe2fdad517d2feae932c0c7061edcae37e844ade/docs/proof/demo-readiness-2026-08-11/60-overview-after-chat.png)

Analytics after live chat traffic, 24h window:

![Analytics reads zero](https://raw.githubusercontent.com/sakibsadmanshajib/hive/fe2fdad517d2feae932c0c7061edcae37e844ade/docs/proof/demo-readiness-2026-08-11/61-analytics-after-chat.png)

Proof the traffic was real, same account, same day:

![Model answered](https://raw.githubusercontent.com/sakibsadmanshajib/hive/fe2fdad517d2feae932c0c7061edcae37e844ade/docs/proof/demo-readiness-2026-08-11/14-chat-sendbutton.png)

## Reproduction

1. Sign in to chat as the demo account and send a prompt that receives an answer.
2. Open `https://console-hive.scubed.co/console` and `https://console-hive.scubed.co/console/analytics` for the same workspace.
3. Every counter reads zero, in every time window.

## Scope

Two separable questions, worth splitting once someone has looked:

1. Presentation. Is chat traffic supposed to appear in this workspace's analytics, and if it is, where does the attribution break.
[...]

PR #908, opened 2026-08-16
fix: analytics client parses the wrong JSON key, always empty (#856)

sakibsadmanshajib:
## What broke

Console Analytics and Billing read zero in every time window (24h/7d/30d) and every tab, while the Overview balance card correctly showed a completion's credit spend (posted credits dropped from 9,999,920 to 9,999,902 in the same session). Zero failed HTTP requests were involved: every analytics call returned 200 OK.

## Root cause

`apps/web-console/lib/control-plane/client.ts`'s `getAnalyticsUsage`, `getAnalyticsSpend`, and `getAnalyticsErrors` all parsed `payload["data"]` out of the response body. Their control-plane handlers (`apps/control-plane/internal/usage/http.go`, `handleAnalyticsUsage`/`handleAnalyticsSpend`/`handleAnalyticsErrors`) wrap their rows under `"usage"`, `"spend"`, and `"errors"` respectively, and never send a `"data"` key. `readArrayField` returns `null` when the key is absent, so `rawData` was always `[]`, unconditionally, on every call, independent of what `usage_events` actually held. That is the entire bug: a wrapper-key mismatch between this file and the three endpoints it calls, present since these functions were written (every other `readArrayField` call elsewhere in the same file already uses each endpoint's real key: `"entries"`, `"invoices"`, `"models"`, `"alerts"`, etc, only these three used the placeholder `"data"`).

## Why the overview card was not affected

`getBalance()` calls a different endpoint, `/api/v1/accounts/current/credits/balance`, whose handler returns `posted_credits`/`reserved_credits`/`available_credits` as top-level numeric fields, which `getBalance` reads directly (`readNumberField(payload, "posted_credits")`, etc), no wrapper key involved. That path was always correct, which is why it showed the real spend while analytics showed nothing next to it.

## Why this is a presentation bug, not a metering bug

Traced the write side for the session-chat path that produced the demo traffic (`apps/edge-api/internal/chat/billing.go`, `startSettlement`/`finalize`) through to control-plane's `finalizeLocked` (`apps/control-plane/internal/accounting/service.go`). On every successful finalize it calls `s.usageSvc.RecordEvent` with `HiveCreditDelta: -actualCredits` (correctly negative) and the real `InputTokens`/`OutputTokens`, inserting a real row into `usage_events`. The balance moving in the demo proves finalize succeeded, which proves this row exists. `GetUsageSummary`/`GetSpendSummary`/`GetErrorSummary` (`apps/control-plane/internal/usage/repository.go`) scope correctly on `account_id` and the `created_at` time window, no tenant-scope bug, no timezone bug, no missing materialized view. The row was there; the frontend just discarded it on every parse.

## Fix

Changed the three `readArrayField(payload, "data")` calls to read the endpoint's real wrapper key (`"usage"`, `"spend"`, `"errors"`). Three lines, one file. No backend change was needed or made.

## Test (TDD, RED confirmed before the fix)

Added `apps/web-console/__tests__/analytics-client.test.ts`: each test mocks `fetch` to return the *real* response shape each Go handler actually sends (`{"usage": [...]}`, `{"spend": [...]}`, `{"errors": [...]}`) and asserts the client parses non-empty, correctly typed rows. All three failed (`expected [] to have a length of 1 but got +0`) against the pre-fix code, passed after.

```
docker compose run --build web-console npm run test:unit -- analytics-client
 ✓ __tests__/analytics-client.test.ts (3 tests) 245ms
```

Full unit suite: 522 passed, 1 pre-existing unrelated skip. Two pre-existing failures (`tests/unit/control-plane-host.test.ts`, `tests/unit/chat-coverage-lib.test.ts`) are unrelated to this change: both read repo-root files (`.env.example`, `docs/proof/...`) that `Dockerfile.web-console` never `COPY`s into the test image (it only copies `apps/web-console/`), so they fail identically on `main` before this branch touches anything.

`npm run build` succeeds, no type errors.

## Scope note

Skipped a separate formal plan/spec vault doc for this change: it is a 3-line, single-file, fully mechanical key-rename with a pinned regression test, and the ticket's own investigation already served as the spec. Flagging one adjacent, separate finding for a follow-up issue rather than folding it into this diff: `apps/edge-api/internal/inference/orchestrator.go`'s `recordCompletedEvent` (used only by the direct API-key `/v1/chat/completions` path, not by the session-chat path this ticket's demo traffic used) calls `RecordUsageEvent` a second time after `FinalizeReservation` already wrote the correct settlement row inside `finalizeLocked`, with `HiveCreditDelta: usage.TotalTokens` (positive, a token count, not a negative credit delta). That produces a duplicate `usage_events` row per API-key request: it does not zero anything (the CASE clause in `GetSpendSummary`/`GetUsageSummary` just ignores the positive delta), but it double-counts `request_count` and token totals for that path. Unrelated to #856's repro (session chat never calls it), not touched here to keep this diff surgical; worth its own ticket.

## Visual proof

To follow as a PR comment: a screenshot from a locally running stack on this branch, showing `/console/analytics` with non-zero numbers matching a real completion sent through edge-api in the same session.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved analytics data handling by correctly reading usage, spending, and error results from the control plane.
  - Preserved existing row formatting and error handling while ensuring analytics information displays as expected.

- **Tests**
  - Added coverage to verify parsing of usage, spending, and error responses.


apps/web-console/lib/control-plane/client.ts
```diff
@@ -2055,7 +2055,9 @@ export async function getAnalyticsUsage(params: {
     throw new Error("Failed to parse usage analytics response");
   }
 
-  const rawData = readArrayField(payload, "data") ?? [];
+  // handleAnalyticsUsage (apps/control-plane/internal/usage/http.go) wraps
+  // its rows under "usage", never "data" (issue #856).
+  const rawData = readArrayField(payload, "usage") ?? [];
```
sakibsadmanshajib: Finding: this closes today's trigger but not the class of defect. readArrayField returns null for any key that is missing or not an array, and every caller here does ?? []. The #856 bug was that the backend never sends data. After this fix, the same silent path still exists for the next drift: a backend rename of usage again, a proxy returning a differently shaped 200, or a nested wrapper change would all produce an identical symptom, a 200 OK response that parses to an empty array with zero signal that anything went wrong. There is no thrown error and no log line distinguishing genuinely zero rows from a shape mismatch, here or in getAnalyticsSpend/getAnalyticsErrors below (lines 2097 and 2134), same pattern. Since this diff already knows the expected key at each call site, this is the natural place to add a minimal guard: throw, or at least log, when the payload is a valid JSON object but the expected array key is absent or wrong typed. Without that, the fix is correct for today's key and leaves the detectability gap that made #856 invisible in production exactly as open as before.

apps/web-console/lib/control-plane/client.ts
```diff
@@ -2090,7 +2092,9 @@ export async function getAnalyticsSpend(params: {
     throw new Error("Failed to parse spend analytics response");
   }
 
-  const rawData = readArrayField(payload, "data") ?? [];
+  // handleAnalyticsSpend (apps/control-plane/internal/usage/http.go) wraps
+  // its rows under "spend", never "data" (issue #856).
+  const rawData = readArrayField(payload, "spend") ?? [];
```
sakibsadmanshajib: Finding: billing's 'No transactions yet' is a second, separate, still-live defect, not something this PR explains or fixes. The PR body frames this as fixing 'Console Analytics and Billing' together, but the billing symptom traces elsewhere. apps/web-console/app/console/billing/page.tsx line 100 hardcodes recentEntries={[]} on the Billing Overview tab, unconditionally, regardless of what the backend returns. That line predates this branch (introduced in PR #89, the original Go rewrite) and is untouched by this diff. getLedgerEntries, the function that actually fetches real ledger rows, already reads the correct entries key today (readArrayField(payload, "entries") further down in this same file) and was never affected by the data versus usage/spend/errors mismatch this PR fixes, since it is wired only to the separate Ledger tab, never to the Overview tab where 'No transactions yet' renders. So the billing symptom in the ticket has a different root cause than the analytics symptom and remains unfixed after this merges. Recommend either wiring recentEntries to getLedgerEntries here too, or filing and linking a follow-up issue, rather than letting this PR read as if both symptoms are resolved.

apps/web-console/__tests__/analytics-client.test.ts
```diff
@@ -0,0 +1,128 @@
...
+ * returned null, so every call silently parsed to an empty array: a 200 OK
+ * response with real rows still rendered as all-zero, on every window,
+ * every tab, independent of what usage_events actually held. That is the
+ * "no failed HTTP requests, every counter reads zero" symptom from #856
+ * exactly: no backend defect, no tenant-scope defect, no timezone defect,
+ * just a wrapper-key mismatch between this file and the handlers it calls.
+ *
+ * These tests pin each function to the real wrapper key its own endpoint
+ * sends, so a future edit that drifts the two apart again fails loudly here
+ * instead of only in a live demo walk.
```
sakibsadmanshajib, line 20: Finding: fixture correctness is cross-checked, but not verified against a real running response, and the visual proof the PR body promises has not been posted. Cross-checked these three mock payloads against the actual Go wire types (apps/control-plane/internal/usage/types.go UsageSummaryRow/SpendSummaryRow/ErrorSummaryRow) and the wrapper keys in apps/control-plane/internal/usage/http.go handleAnalyticsUsage/Spend/Errors: field names and wrapper keys match, so this is not a fixture invented from assumption alone. That said, it is still a hand built unit fixture, not a response captured from a live stack, and none of these tests render the actual /console/analytics page or its components, only the client parsing functions in isolation. The PR body's Visual proof section says a screenshot is 'to follow as a PR comment,' but nothing has landed on this PR as of this review. This repo's visual proof before merge rule (.claude/rules/orchestrator.md, owner directive 2026-07-21) requires that screenshot to actually be posted in the PR before this is merged or claimed done, since the change alters what renders on a live UI surface.

apps/web-console/__tests__/analytics-client.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    expect(rows).toHaveLength(1);
+    expect(rows[0].total_credits).toBe(18);
+    expect(rows[0].entry_count).toBe(2);
+  });
+
+  it("parses getAnalyticsErrors against the real {errors: [...]} wrapper", async () => {
+    nextResponse = new Response(
+      JSON.stringify({
+        errors: [
```
sakibsadmanshajib, line 120: Finding: this suite could pass while the Analytics or Billing UI still renders zero, because it never renders either page. These are pure client-function unit tests against getAnalyticsUsage/Spend/Errors in isolation. They correctly pin the wrapper key regression named in this PR, but they say nothing about app/console/analytics/page.tsx wiring those results into the rendered tables, and nothing at all about app/console/billing/page.tsx, which is where the ticket's other reported symptom, billing showing 'No transactions yet,' actually lives (see the comment on line 2097 of client.ts: that page hardcodes recentEntries={[]} and this suite cannot see it). A green run of exactly this file does not mean a user loading /console/analytics or /console/billing today would see non-zero numbers. Worth having at least one test, even a lightweight component or integration one, that asserts the page actually renders the rows this client returns, so the next regression at the page-wiring layer does not slip through the same way #856 did at the client-parsing layer.

sakibsadmanshajib:
## Visual proof (issue #856, orchestrator rule 8)

Sent two real chat completions through a locally running `docker compose --profile local --profile dev` stack (built from this PR's head, `48a100ad`) against the demo account, authenticated the same way Open WebUI's session-JWT chat path authenticates (Bearer Supabase access token, `apps/edge-api/internal/chat/billing.go`/`dispatch.go`), no password reset, minted via `mintSession` in `tests/e2e/support/live-auth.mjs`. Then loaded `/console/analytics` and `/console/billing` for that same account, same session, against the fixed build.

**Analytics, 24h window, non-zero, matching the traffic just sent:**

Analytics overview non-zero

**Billing overview, real ledger rows instead of "No transactions yet":**

Billing overview recent transactions

Both images are also committed at `docs/proof/issue-856-analytics-fix-2026-08-16/` on this branch.

## Response to review findings

**1. Detectability gap (client.ts:2060, and the same pattern at getAnalyticsSpend/getAnalyticsErrors).** Agreed, this was real. Added `requireArrayField`, which `getAnalyticsUsage`/`getAnalyticsSpend`/`getAnalyticsErrors` now call instead of `readArrayField(...) ?? []`. It throws when the expected wrapper key is missing or not an array, so the next contract drift (a rename, a proxy remangling a 200, a nested wrapper change) surfaces through the page's existing `fetchError` / "Unable to load analytics" state instead of a second silent all-zero read. A genuinely quiet account (key present, array genuinely empty) still returns `[]`, unchanged. Three new tests in `analytics-client.test.ts` pin: missing key throws, wrong-typed key throws, present-empty key still resolves to `[]`.

**2. Billing's "No transactions yet" is a separate, real, unfixed defect, and the PR body implied otherwise.** Confirmed exactly as described: `apps/web-console/app/console/billing/page.tsx` hardcoded `recentEntries={[]}` on the Overview tab since PR #89, untouched by the analytics key mismatch, and `getLedgerEntries` was already correct but only ever wired to the separate Ledger tab. Fixed here rather than filed separately, since it is small and it is literally the second symptom this same ticket reported: the Overview tab's `Promise.all` now also calls `getLedgerEntries({ limit: 5 })` (degrading to an empty preview on fetch failure, matching `budgetThreshold`'s existing fallback) and passes the real rows to `BillingOverview`. Visible in the second screenshot above.

**3. Fixtures were hand-built, not a captured live response, and nothing rendered either page.** Added `apps/web-console/__tests__/analytics-billing-page-wiring.test.tsx`, which imports and renders the actual `app/console/analytics/page.tsx` and `app/console/billing/page.tsx` Server Components (same pattern `__tests__/billing-profile-missing-row.test.tsx` already established for `app/console/settings/billing/page.tsx`) against a mocked control-plane, and asserts the real non-zero values from the mocked responses reach the rendered summary cards and the real ledger rows reach `BillingOverview`. This closes the gap the finding named: a page-wiring regression at this layer would now fail here, not just pass silently through client-only unit tests. It complements rather than replaces the client-level tests and the now-real, live-captured screenshots above, which together cover parsing correctness, page wiring, and actual rendered behavior against a running stack.

**4. CodeRabbit: `CONTROL_PLANE_BASE_URL` not restored between tests.** Fixed: `analytics-client.test.ts` now saves the pre-existing value in `beforeEach`'s enclosing scope and restores or deletes it in `afterEach`.

All 8 tests in `analytics-client.test.ts`, both new tests in `analytics-billing-page-wiring.test.tsx`, and the full existing suite (535 tests, 47 files) pass. `tsc --noEmit` is clean.

Head SHA: `48a100ad8ed450279caabf86ecc379f6d71b7d00`.

Issue #885, opened 2026-08-11
Interaction gate cannot reach the per-key limits route


sakibsadmanshajib:
/console/api-keys/[id]/limits is a real route with real controls, and the interaction coverage gate (PR #811) does not measure it.

The gate resolves a dynamic route from a link the application itself rendered, so it needs at least one API key to exist in the workspace it signs in as. Two environments, two reasons that does not happen:

- CI seeds a fresh identity whose workspace holds nothing, so the api-keys page renders no rows and no link to a limits page.
- A live run could reach it, but only where a key already exists, which makes the route's coverage depend on the tenant's data rather than on the product.

Creating one to get there is not an option the gate should take: it now blocks every state changing request whose values it invented, precisely so a coverage run cannot write to the surface it measures.

So the route is skipped in apps/web-console/tests/interaction/route-fixtures.json citing this issue, which means the skip expires the moment this closes.

The fix is to give the gate a key it is allowed to see. The CI job already runs tests/e2e/support/e2e-auth-fixtures.mjs to seed its identity; seeding one API key there, and recording its id, would make the route reachable in CI without the gate writing anything.

sakibsadmanshajib:
Open PR #811 (test: interaction coverage gate for the web console) touches this area. Not yet merged.

Issue #848, opened 2026-08-11
Demo chat account is full of test litter, and nothing cleans up after the runs that create it


sakibsadmanshajib:
The demo account's chat list contains rows left behind by our own automated runs. A sample of the row titles visible in the sidebar and search surface on 2026-08-10:

- `Reply with exactly: ALPHA`
- `Reply with exactly: SIERRA`
- `Reply with exactly: PONYTAILTEST42`
- `ping`

These sit alongside the real seeded conversations, and there were 52 chat rows in total at the time of the run.

## Why this matters on its own

The owner may show this account to people. A prospect looking at the chat list sees a QA scratch pad rather than a product. This is a demo readiness defect independent of any test tooling, and it is the reason for filing it separately rather than folding it into the coverage work.

## The side effect on measurement

Every chat row contributes its own row controls to the interaction coverage sweep, so the row count was inflating that gate's denominator. #809 has since changed the primary figure to distinct control identities, which makes the number independent of row count, so the metric is no longer distorted by this. That is a mitigation of the symptom and not a fix for the litter.

## The part that matters more than a one time tidy

Any fixture or automated run that creates a chat needs a cleanup path, or this re-accumulates within days. Today nothing removes what these runs create:

- The E2E and coverage suites send messages to prove the composer works, and each one creates a conversation that is never removed.
- #828 records that the fixture sweep cannot delete an account at all, because the `tenant_billing_accounts` foreign key refuses every attempt, so there is no working teardown path to hang chat cleanup on either.

Worth deciding whether automated runs should use a throwaway account rather than the shared demo one, which would make cleanup a matter of not caring about the account rather than a matter of deleting rows.

## Not doing

Deleting the chats. Destructive cleanup of a shared demo account is the owner's call, and other agents hold sessions against it.

sakibsadmanshajib:
PR #909 removes a dead Admin Panel entry; unrelated to this issue's ask (cleaning up test-run litter in the demo account's chat list). Leaving open, this PR does not resolve it.

sakibsadmanshajib:
Triage note: this looks likely to be at least partly resolved by commit 4ac17f436, "fix: enforce the demo-account rule and correct the false chat-delete claim (issues #916, #848)" (PR #1462), which names this issue directly. Labelled priority:low rather than closed, because it is not confirmed that the runs which create the litter now clean up after themselves. Verify the demo chat account is clean after a full verification run before closing.

Issue #846, opened 2026-08-11
Chat Settings > General: Admin Settings fires a request to a 404 and does nothing


sakibsadmanshajib:
Settings > General renders an **Admin Settings** control that fires a request to an endpoint answering **404**. Nothing happens for the user, and nothing tells the user that nothing happened.

## Observed

Live sweep of https://chat-hive.scubed.co on 2026-08-10, from the chat interaction coverage gate:

```
settings:General | Admin Settings | called an endpoint that is not mounted or that failed:
404 GET https://chat-hive.scubed.co/api/...
```

Full ledger: `docs/proof/chat-interaction-coverage-2026-08-10/coverage.run.json`, PR #809.

## Why it went unnoticed

The coverage engine had no response listener, so HTTP status was invisible to it. A request going out was treated as proof that the control worked, so this control was reported as **proven, proof type `network`**, on every run until tonight. PR #809 teaches the engine to read statuses, and it now correctly reads **unproven**.

## Worth deciding

Whether the admin surface should be reachable from this deployment at all. #748 and #747 both argue the tenant OWNER to Open WebUI admin derivation is a hazard, and #773 removed other upstream admin branding from the user menu. If the answer is that it should not be there, the fix is removal in `hive_ui_surfaces.py` rather than mounting the endpoint.

sakibsadmanshajib:
Root cause, found during the 2026-08-17 admin-settings audit.

The link is at `vendor/open-webui/src/lib/components/chat/SettingsModal.svelte:843-861`. It renders for `$user?.role === 'admin'`, and its handler calls `preventDefault()` then `goto('/admin/settings')`, so it is client-side routing rather than a request. That is why the panel opens and renders in full, while a refresh on that address returns Caddy's 404 from the `@blocked` matcher in `deploy/docker/Caddyfile.owui:103-106`. The owner reported seeing "many options" there on 2026-08-17, so the current behaviour is not "does nothing", it is "opens a panel that half works and cannot be reloaded".

Why it survived the removal that took its twin out of the user menu: `deploy/docker/owui-patches/hive_ui_surfaces.py` removes both, but `Dockerfile.open-webui` runs that patch against the upstream prebuilt bundle and then replaces `/app/build` with our own build from `vendor/open-webui`. The Dockerfile says as much in its own comment, that every surface those rewrites removed is removed in the vendored source instead. For this one it was not. The user-menu entry (#909) was.

[...]

Issue #769, opened 2026-08-06
Open WebUI admin config endpoints leak the live gateway key and OIDC client secret to the public chat origin


sakibsadmanshajib:
## Summary

Two Open WebUI admin endpoints on the public chat origin (chat-hive.scubed.co) return the live OWUI shim gateway key (a real minted `hk_` Hive API key) and, in one case, the OIDC client secret, to any account holding Open WebUI `role: admin`. Per #748, every legitimate tenant OWNER gets that role on login, so this is reachable by any real tenant owner, not only by a privileged operator.

## Verified against the pinned image (`hive-open-webui:v0.10.2-branded`)

1. `GET /api/v1/configs/export` (`routers/configs.py`, `export_config`, gated on `get_admin_user`) returns `Config.get_all()`, the full persisted config table. `Config.DEFAULT_CONFIG` includes `openai.api_keys` (seeded from `OPENAI_API_KEY`, which `docker-compose.yml` sets to `${OWUI_SHIM_KEY}`), `rag.openai.api_key` (same value, reconciled from `RAG_OPENAI_API_KEY` on every boot by `deploy/docker/owui-patches/hive_rag_env_config.py`), and `oauth.client_secret` (seeded from the `OAUTH_CLIENT_SECRET` env var, which `docker-compose.yml:620` sets from `SUPABASE_OAUTH_CLIENT_SECRET`). So this single endpoint returns both credentials the audit named, confirmed against the source.
2. `GET /openai/config` (`routers/openai.py`, `get_config`, gated on `get_admin_user`) returns `OPENAI_API_KEYS` and `OPENAI_API_BASE_URLS`, i.e. the same gateway key plus `http://edge-api:8080/v1`, the internal service address.
3. Neither endpoint is covered by `deploy/docker/Caddyfile.owui`. The `@blocked` matcher (line 24) 404s `/admin`, `/signup`, and related paths, with no entry for `/api/v1/configs` or `/openai`. The `@adminMutation` matcher (line 32) only inspects `PUT POST PATCH DELETE`, so a `GET` to either endpoint is never even evaluated by it, independent of the path glob issue tracked in #437 and #736. Both reach Open WebUI unblocked on the public origin.
[...]

PR #909, opened 2026-08-16
fix: remove dead Admin Panel entry, stop coverage suites littering the demo account (#846, #848)

sakibsadmanshajib:
## Summary

Two audience-visible demo-surface defects, two commits.

**#846, dead Admin Panel entry.** The chat user menu's "Admin Panel" item
navigates to `/admin`, which `Caddyfile.owui` already 404s outright:
`.wolf/decisions.md` D-014 makes web-console/control-plane the sole admin
surface and keeps Open WebUI's own admin panel off. The entry was the one
thing in `deploy/docker/owui-patches/hive_ui_surfaces.py` kept alive on
purpose (a `GUARDS` assertion), on the theory that Hive's tenant-owner
promotion made it a real control. It never was. **Decision: remove, not
wire up.** Wiring the endpoint would stand up a second admin surface next
to web-console's, which is the exact duplication D-014 forecloses. Same
removal shape as the neighboring Playground entry: an unconditional `!1`
gate, not a permission narrowing (every tenant owner already passes
`role === "admin"`, so a permission-scoped hide would hide nothing).

`/api/v1/terminals/` 404ing on every page load is a different class, not a
bug: `Caddyfile.owui` already documents it as a deliberate block (#770, an
authenticated request-forgery primitive with no allowlist), and the front
end already degrades to an empty terminal list on any non-2xx. Left alone.

**#848, demo-account test litter.** Root cause: two currently-merged live
coverage paths silently defaulted their sign-in identity to
`demo@hive-demo.invalid`, the account the owner demos to prospects.

- `apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts`
  defaulted to the demo account whenever `HIVE_QA_AGENT_EMAIL` was unset.
  Every run past that default creates one real, currently undeletable
  agent task (no per-task delete route exists yet), which is why the task
  list is a wall of `interaction-coverage proof ...` rows stuck Cancelled
  or Blocked.
- `.github/workflows/chat-coverage.yml` fell back to
  `secrets.HIVE_QA_TESTER_EMAIL` whenever `secrets.HIVE_QA_AGENT_EMAIL`
  was unset, which it always was (no such secret exists on this repo).
  The sweep's own ledger proves the result:
  `"demo@hive-demo.invalid -> hive-coverage-77811 survived a reload"`
  (`docs/proof/chat-interaction-coverage-2026-08-10/coverage.run.json`).

Both now fail loudly instead of defaulting (the same `requireMintEnv`
pattern the first suite already used for its Supabase mint variables, and
the loud "inputs the sweep cannot run without" check the workflow already
had). `docs/live-test-auth.md` gets a new section stating the rule for the
next suite: read-only checks may use the demo account; anything that sends
a message, submits a task, or mints a key must use a dedicated identity,
the same way the plain E2E suite already isolates itself with
`E2E_RUN_KEY`-scoped fixture accounts.

**Not fixed here, flagged instead of silently skipped:** neither change
provisions the dedicated identity itself. `HIVE_QA_AGENT_EMAIL` has no
value to fail over to; standing one up needs live Supabase admin
credentials this worktree does not hold. Both suites now fail until that
identity is provisioned and the secret is set, which is the intended
consequence, not a regression: a workflow that used to succeed quietly
against the wrong account now requires the right one first.

## Existing litter: enumerated, not deleted

Nothing is deleted in this PR. Per standing instruction, deletion needs an
explicit owner go-ahead after review of what's actually on the account.

Known rows, from issue #848, issue #858's 2026-08-11 demo-readiness walk,
and today's (2026-08-16) live reconfirmation:

- Chat sidebar: 48 rows, 25 titled literally `New Chat`. Earlier snapshots
  (#848, filed against the 2026-08-10 sweep) also show titled rows matching
  synthetic composer-proof text (`Reply with exactly: ALPHA`, `SIERRA`,
  `PONYTAILTEST42`, `ping`) plus `hive-coverage-77811`-style generated
  names from the coverage ledger.
- Cowork task list: rows titled `interaction-coverage proof <timestamp>
  <uuid>` (literal source: `agent-workspace-flows.spec.ts` line ~504) and,
  per #858, `cancel-guard proof ...`, sitting in Blocked or Cancelled
  states.
- API keys page: rows described as expired `interaction gate probe` /
  `interaction gate` keys (#858 lists six).
- A spend alert configured to `interaction-gate@example.invalid`.

**On the 25 `New Chat` rows specifically:** I cannot independently
re-enumerate the live account from this worktree (no browser tool, no
Supabase/DB access available to this agent), so this is a proposed
classification, not an executed one. Proposed criterion before any delete:
pull each `New Chat` row's `created_at` and first message. A row whose
first message is a synthetic literal already used by a coverage suite
(the `Reply with exactly: <WORD>` / `ping` family, or any string a suite
sends deterministically) is automation litter regardless of its stuck
title; OWUI's title generation is a second, async LLM call after the first
response, and a suite that asserts the message round-trip and moves on
without waiting for or asserting a title is exactly the shape that leaves
a chat stuck on `New Chat` forever, so I'd expect most or all of the 25 to
be this, not failed title-gen on a real conversation. A row whose first
message is not a known synthetic literal, or whose `created_at` does not
cluster with a known CI/coverage run, should be treated as a real
conversation with a failed title generation and left alone. I'm not able
to run that classification against the live rows myself; whoever has
live account access should run it before anything is deleted, and I'd
want to see the actual per-row list before endorsing any deletion set.

Also worth flagging, not part of this PR's diff: the in-flight, unmerged
`test/interaction-coverage-gate` branch (issue #885, PR #811) is a newer
console-wide interaction-coverage framework that per #885 has already
moved toward blocking mutating requests. The `interaction gate probe` API
keys and the `interaction-gate@example.invalid` spend alert most likely
predate that safeguard. Whoever owns that PR should confirm it also never
authenticates as the demo account once merged, same rule as this PR.

## Visual proof

Not captured. This agent's toolset (Read/Write/Edit/Bash, no browser or
screenshot tool) cannot drive the running stack to capture the user-menu
before/after state orchestrator rule 8 requires before merge. Flagging
this as a structural capability gap rather than skipping it silently: a
Playwright-capable agent needs to capture the corrected user menu (Admin
Panel entry gone) against a real deployment before this merges.

## Buglog entry

```json
{"date":"2026-08-16","tags":["owui","chat","cowork","test-litter","demo-account","ci"],"error_message":"Demo account (demo@hive-demo.invalid) accumulates permanent chat, agent-task and API-key rows from live coverage suites; user menu also offered a dead Admin Panel entry (404 on every path)","root_cause":"Two live-coverage code paths silently defaulted their sign-in identity to the shared demo account: agent-workspace-flows.spec.ts's AGENT_EMAIL fallback and chat-coverage.yml's HIVE_QA_AGENT_EMAIL-to-HIVE_QA_TESTER_EMAIL secret fallback (which resolved to the demo identity in practice, per its own coverage ledger). Separately, hive_ui_surfaces.py deliberately kept the OWUI Admin Panel user-menu entry alive via a GUARDS assertion, while Caddyfile.owui already 404s every path it points at (D-014: web-console/control-plane is the sole admin surface).","fix":"Removed the Admin Panel entry from hive_ui_surfaces.py (moved from GUARDS to REWRITES, same !1 gate as the neighboring Playground entry). Changed both silent demo-account fallbacks to hard failures (agent-workspace-flows.spec.ts's requireMintEnv, chat-coverage.yml's HIVE_QA_AGENT_EMAIL secret) so a live coverage run can no longer land on the demo account by omission. Existing litter enumerated for owner review, not deleted; a dedicated non-demo QA identity still needs to be provisioned and wired to HIVE_QA_AGENT_EMAIL as a follow-up.","tags":["owui","chat","cowork","test-litter","demo-account","ci"]}
```

## Test plan

- [x] `python3 scripts/test_owui_ui_surfaces.py` — pass
- [x] `python3 scripts/test_caddy_owui_blocklist.py` — pass (unchanged, confirms terminals/admin blocks untouched)
- [x] `python3 scripts/test_owui_rag_env_config.py` — pass (unrelated suite, sanity check for the same patch package)
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/chat-coverage.yml'))"` — valid YAML
- [ ] TypeScript typecheck of `agent-workspace-flows.spec.ts` (`docker compose run --build web-console npm run build`) — not run in this worktree, no `node_modules`/Docker invoked here; the diff is a mechanical mirror of the already-passing `requireMintEnv` pattern in the same file, manually reviewed instead
- [ ] Live screenshot of the corrected user menu — blocked on browser-tool availability, see Visual proof above
- [ ] Live re-run of `chat-coverage.yml` / `agent-workspace-flows.spec.ts` against a provisioned `HIVE_QA_AGENT_EMAIL` — blocked on that identity not existing yet


docs/live-test-auth.md
```diff
@@ -83,6 +83,35 @@ every session that any other run currently holds. On 2026-08-08 a scratch
...
+This happened. `docs/proof/chat-interaction-coverage-2026-08-10/coverage.run.json`
+records `"demo@hive-demo.invalid -> hive-coverage-77811 survived a reload"`,
+and the 2026-08-11 demo-readiness walk (issue #858) found the sidebar full of
+empty "New Chat" rows and the agent task list "a wall of 'cancel-guard proof'
+and 'interaction-coverage proof' entries," on that same account. Issue #848
+tracks the cleanup and the standing fix.
+
+So: a suite that only signs in and reads (a redirect check, a rendered
+heading, an unclicked button) is fine against the demo account, the same way
+`scripts/verify-control-plane.py` is fine, per the table below. A suite that
```
sakibsadmanshajib: HIGH: this line calls `scripts/verify-control-plane.py` a read-only example that is fine against the demo account, but the script contradicts the rule stated two sentences later ("a suite that sends a message, submits a task, or mints a key must run as a dedicated, non-demo identity").

`scripts/verify-control-plane.py` line 49 defaults `EMAIL = os.environ.get("HIVE_VERIFY_EMAIL", "demo@hive-demo.invalid")`, the same silent-default-to-demo pattern this PR removes from the other two paths (agent-workspace-flows.spec.ts and chat-coverage.yml). With that default in effect, the script's api-key lifecycle section creates a real API key against the demo tenant (`POST /api/v1/accounts/current/api-keys`, nickname "verify-control-plane") and then sends a real chat completion through it (`POST /v1/chat/completions`, model "hive-default", message "say pong", against the edge API). The key is revoked at the end, but the completion itself, and whatever usage-event, request-attempt and spend-analytics rows it leaves on the demo tenant, are not undone. That is exactly the litter class #848 is closing, on the exact silent-default-identity path this PR's own new rule targets, left unaddressed and described here as compliant.

Suggest scoping this line's endorsement to password handling only (which is what the script's own docstring actually claims, matching the existing table entry below), or fixing verify-control-plane.py to require HIVE_VERIFY_EMAIL rather than defaulting it, the same way this PR fixed the other two paths.

deploy/docker/owui-patches/hive_ui_surfaces.py
```diff
@@ -152,14 +152,36 @@ class Rewrite:
...
+        # a top level "Admin Panel" item, and clicking it always 404s, with
+        # nothing telling them why (issue #846, reconfirmed live in the
+        # 2026-08-11 demo-readiness walk, issue #858). Wiring the endpoint up
+        # instead would stand up a second admin surface next to web-console's,
+        # which is the exact duplication D-014 forecloses. Same removal shape
+        # as the Playground entry immediately above: gated to `!1` rather
+        # than narrowed to a permission, because every tenant owner already
+        # passes `role === "admin"` and a permission-scoped hide would hide
+        # nothing from the audience this ships to.
+        surface="usermenu-admin-panel",
```
sakibsadmanshajib, line 180: LOW: `deploy/docker/owui-patches/apply_ui_surfaces_patch.py` (not touched by this PR) still documents Admin Panel as a kept surface. Its module docstring says the build "asserts... the guard strings for the surfaces Hive keeps (Workspace > Knowledge, the Admin Panel entry) must still be present afterwards." That stopped being true the moment this surface moved from `GUARDS` to `REWRITES` here.

Not a functional bug: that script iterates `GUARDS`/`REWRITES` generically and names no surface by hand, so the build itself is unaffected. But the docstring will mislead the next person who reads it before the code, in a codebase whose whole design philosophy here is that every entry documents its upstream source for the next reader. Worth a one-line follow-up dropping "the Admin Panel entry" from that sentence.

sakibsadmanshajib:
Adversarial review (stage 6), pipeline mode, single pass since the builder had no agent-dispatch tool.

Verified against the pushed diff and the surrounding untouched files, not just the PR body's claims.

**Fail-loud change (#848), risk of a permanently-red check nobody looks at:** does not apply here. `chat-coverage.yml`'s `self-check` job (runs on every PR) never touches `HIVE_QA_AGENT_EMAIL` at all; its `live-sweep` job only runs on `workflow_dispatch` or a PR explicitly labelled `run-chat-coverage`, never automatically. `agent-workspace-flows.spec.ts` lives under `tests/e2e/_probe`, which `ci.yml` excludes from its test glob, and no workflow in `.github/workflows/` references the file at all. So the new hard failure only fires when someone deliberately triggers one of these two paths, not on an automatic, recurring, ignorable schedule. Confirmed no cry-wolf risk.

**Admin Panel removal, cosmetic vs real:** real, and correctly scoped. `/admin` was already 404ing via `Caddyfile.owui`'s `@blocked` regex before this PR (D-014, issues #769/#772), independent of the client-bundle rewrite. This PR only removes the menu item that pointed at that dead route; it does not change, and does not need to change, the route's own reachability, which was never open. No gap introduced.

**Third silent-default identity path:** found one, posted inline on `docs/live-test-auth.md`. `scripts/verify-control-plane.py` defaults `HIVE_VERIFY_EMAIL` to `demo@hive-demo.invalid` and its default run mints a real API key plus a real chat completion against the demo tenant, the same litter class as the two paths this PR fixes. This PR's own new doc section cites that script as a compliant, read-only example, which its actual behaviour contradicts.

**Pinned-bundle fixture drift:** verified robust. `test_the_excerpt_fixture_is_from_the_digest_the_dockerfile_pins` fails the fixture the moment `Dockerfile.open-webui`'s pinned digest moves, and `apply_ui_surfaces_patch.py` independently re-verifies every rewrite against the real built bundle at image-build time, so a stale fixture cannot silently certify a broken rewrite. The relocated `usermenu-admin-panel` excerpt is byte-identical to the one already pinned and verified when it was a guard. One stale docstring sentence in the untouched `apply_ui_surfaces_patch.py` posted inline as LOW.

**Terminals claim (#770):** verified true and pre-existing, untouched by this PR. `Caddyfile.owui`'s `@blocked` regex already blocks the whole `/api/v1/terminals` family, and `scripts/test_caddy_owui_blocklist.py` (not in this diff's file list) already pins it with the same clean-degradation rationale the PR body states.

## Review Summary

| Severity | Count | Status |
|----------|-------|--------|
| CRITICAL | 0 | pass |
| HIGH | 1 | warn |
| MEDIUM | 0 | info |
| LOW | 1 | note |

Verdict: WARNING. One HIGH issue (`scripts/verify-control-plane.py`'s silent demo-account default, contradicting this PR's own new rule) should be resolved or the doc's endorsement corrected before merge. The LOW doc-drift note can ship as a fast follow. Everything else interrogated (fail-loud blast radius, Admin Panel removal, bundle-fixture drift, terminals claim) checks out against the real files, not just the PR body's account of them.

sakibsadmanshajib:
Both threads addressed in e0705371.

**HIGH (verify-control-plane.py).** Confirmed and fixed the same way as the
other two paths. `HIVE_VERIFY_EMAIL` now has no default; `main()` exits
loudly, before any HTTP call (including the mint and the chat completion),
when it is unset. Confirmed no other caller depended on the old default: no
CI workflow invokes this script, and its hyphenated filename means nothing
else can `import` it. Also corrected `docs/live-test-auth.md`, both the new
section from the prior commit and the pre-existing paragraph about this
script, since both were endorsing a reading the script's actual behavior
contradicts. Added `scripts/test_verify_control_plane.py`, a minimal
self-check asserting the default is empty and that `main()` fails before any
request when the identity is unset.

**LOW (apply_ui_surfaces_patch.py docstring).** Fixed: now names the two
guards that remain (Knowledge, Settings dialog) and records that Admin Panel
moved from guarded to removed in #846.

All four self-checks re-run clean after the fix.

sakibsadmanshajib:
## Visual proof, with an explicit split between what is deployed and what is not

Captured 2026-08-16. Every artifact is committed in `d1f5332`, under
`docs/proof/pr-909-demo-surface-2026-08-16/`, alongside the logs that back it.

**This branch is not deployed.** The demo box runs `main`, so nothing captured
against `chat-hive.scubed.co` can show the Admin Panel entry gone. Rather than
present a stale capture as proof of the fix, the menu change is proven on two
locally built images, and the deployed menu is shown separately, as it actually
is today.

### 1. The #846 claim: the entry leaves the client bundle

Both images were built from the same pinned upstream digest and the same
`deploy/docker/Dockerfile.open-webui`, differing only in the three patch files
this pull request touches, then run with `WEBUI_AUTH=false`. That is the method
the #833 and #772 proofs used.

Built from `main`:

user menu built from main, Admin Panel present

Built from this branch:

user menu built from this branch, Admin Panel absent

The menus read out of the DOM, so the difference does not depend on reading
pixels:

```
main   : ["Settings","Admin Panel","Archived Chats","Calendar","Automations","Sign Out", …]
branch : ["Settings","Archived Chats","Calendar","Automations","Sign Out", …]
ADMIN PANEL ENTRY PRESENT: true   (main)
ADMIN PANEL ENTRY PRESENT: false  (branch)
```

Nothing else in the menu moved.

**What this does not prove.** It proves the menu, not the route. `/admin` is
already 404ed by `Caddyfile.owui` independently of this change, so the route was
never reachable and this pull request does not make it less so. The claim is
that an audience no longer sees a control that leads nowhere.

### 2. The deployed menu today, shown rather than omitted

the deployed chat user menu, still showing Admin Panel

`ADMIN PANEL ENTRY PRESENT ON THE DEPLOYED BUILD: true`. Expected: this branch
is unmerged and undeployed. It becomes false on the demo box only after this
merges and `deploy-demo-box.yml` runs.

### 3. The #848 account state, after the enumerated litter was removed

These are live captures. They show the account, not the branch's code.

Chat sidebar, `GET /api/v1/chats/?page=1` returning 0 rows (48 sidebar rows plus
1 archived row deleted through Open WebUI's own delete route, under owner
authorization):

the chat sidebar with no chat rows

API keys, with the six `interaction gate probe` rows now revoked through the
console's own revoke path (revoked rather than deleted, so the record stays
auditable; their expiry still reads 1 Jan 2026, which is issue #915):

the API keys page, every key revoked

Spend alerts, after removing the one addressed to
`interaction-gate@example.invalid`:

spend alerts, none configured

Cowork task rows were deliberately left alone in this pass: another change is
in flight against a stranded task row and deleting task rows underneath it would
destroy its evidence.

### Provenance and redaction

Sessions came from `apps/web-console/tests/e2e/support/live-auth.mjs`, which
mints a magic link with the service-role key and touches no password;
`scripts/seed-owui-e2e-user.py` was not run. No URL in these journeys carries a
credential, and the capture script scrubs token-like query and fragment
parameters before any URL reaches a log. The key column in the API keys
screenshot is the product's own redacted-suffix display, on keys that are all
revoked. The full enumeration behind the deletions, including the per-row
evidence, is in the classification handed to the owner; the two open questions
it raised are filed as #915 and #916.

Issue #815, opened 2026-08-09
chat-hive.scubed.co went fully unreachable for about three minutes on 2026-08-08 and recovered unattended, with nothing observing it


sakibsadmanshajib:
## What happened

During an automated Playwright pass against the agent workspace on 2026-08-08, `chat-hive.scubed.co` stopped answering entirely for roughly three minutes, then recovered with no intervention.

Symptoms during the window:

- Every test failed at `page.goto`, including the unauthenticated ones, with `net::ERR_TIMED_OUT`. That is how it was told apart from a credential problem: the controls that need no session failed alongside the ones that do.
- `curl --max-time 25` returned `000` on both `https://chat-hive.scubed.co/` and `https://chat-hive.scubed.co/agent-workspace/auth/sign-in`, so it was not a single route or the agent-console app.

Recovery: a poll loop got a `200` on the sign-in path on its first attempt about three minutes later. Nothing was restarted or redeployed by the run.

The immediately preceding pass against the same host, minutes earlier, was healthy: 15 tests passed against the live surface.

## Why it is worth a ticket

Nothing observed it except a test run that happened to be in flight. A total loss of the demo host for minutes is exactly what a demo cannot survive, and there is no record of it other than this. #695 already covers the fact that monitoring has no alert path that survives its own stack being down; this is a concrete instance of an outage that nothing surfaced.

## Suggested next step

Correlate against the Cloudflare Tunnel connector and Caddy logs on the box for the window (roughly 2026-08-08, between the two runs recorded in `docs/proof/live-auth-helper-2026-08-08/coverage.md`). A tunnel reconnect is the most likely cause given that both the chat root and the sidecar path went dark together and came back together without intervention, but that is a hypothesis, not a finding: nobody has looked at the logs yet.

Not blocking the live-test work: `docs/live-test-auth.md` and PR #810 land regardless, and the coverage gate re-ran clean afterwards at 21 of 22.

sakibsadmanshajib:
Open PR #910 (fix: guarantee restoration when the coverage sweep's reload check fails) appears to address this. Not yet merged.

PR #910, opened 2026-08-16
fix: guarantee restoration when the coverage sweep's reload check fails

sakibsadmanshajib:
## What was wrong

The live demo chat's Enter key stopped sending again, same symptom PR #863
already fixed once for issue #855: `settings.ui.ctrlEnterToSend` had drifted
back to `true` on the shared `demo@hive-demo.invalid` account.
`node apps/web-console/scripts/demo-chat-settings.mjs demo@hive-demo.invalid`
reproduces it (exits 1) as of this PR.

## Root cause: the coverage sweep, not a reseed

PR #863 attributed the original drift to "an automated sweep over the
settings surface" without naming it. It is
`apps/web-console/e2e/chat-coverage`'s live interaction-coverage sweep
(`.github/workflows/chat-coverage.yml`'s `live-sweep` job, run manually or by
labeling a pull request `run-chat-coverage`, against `HIVE_QA_AGENT_EMAIL`,
which defaults to the shared demo account).

That sweep's `persistencePass` flips every stateful control on a settings
tab, including Settings > Interface's Enter Key Behavior checkbox, on the
live account, to prove Open WebUI actually saves it. It then reloads the
page to read the saved value back, and only after that reload succeeds does
it restore every flipped control to its original value. There was no guard
between the reload and the restore step. `page.reload()` against a box that
goes briefly unreachable mid-run (documented in this same file for a
different reason: issue #815) throws, which unwound straight out of
`persistencePass` and skipped restoration entirely. The caller
(`chat-coverage.spec.ts`) only logged `"state pass aborted"` to a non-fatal
list and moved on to the next surface, so nothing ever said a live setting
had been left mutated.

**Ruled out, not guessed:**

- **Open WebUI default re-applied on container recreate.** Per-user settings
  live in the `owui-data` Docker volume (`deploy/docker/docker-compose.yml`),
  which persists across `docker compose up -d --build` recreates; only
  system-level config is seeded from env vars on first boot
  (`ui.enable_login_form` is the documented example), never a specific
  user's settings row.
- **The OWUI nightly e2e seed script.** `scripts/seed-owui-e2e-user.py`
  (run by `owui-nightly.yml`) provisions a separate `owui-e2e-shim` account.
  It never touches `demo@hive-demo.invalid`.
- **A first-boot-only volume seed replaying.** Same persistent-volume
  reasoning as above; also a wipe would leave the key *absent*, which
  `driftedKeys()` already treats as fine, not explicitly `true`.
- **A Supabase migration.** `grep -rl "ctrlEnterToSend\|user_settings"
  supabase/migrations` finds nothing; Open WebUI's own user-settings table
  is outside anything `supabase/migrations` manages.

## The fix

Structural, not another reset:

- `apps/web-console/e2e/chat-coverage/persistence.ts` (new): the
  flip-verify-restore logic (`persistencePass`, `persistOne`, `valuePass`,
  moved out of `chat-coverage.spec.ts`) now wraps the reload-and-verify step
  in `try/catch`. Restoration always runs afterward, using a fresh
  navigation to recover the control if the reload itself failed. The same
  guard was added to `valuePass`, which had the identical unguarded-throw
  shape for non-persisting stateful controls.
- `break-proof.spec.ts`: a new regression test, no live deployment needed.
  It serves a fixture settings page, aborts the one navigation
  `persistencePass` uses to verify persistence (simulating the #815 box
  hiccup), and asserts the fixture's checkbox is still restored to its
  original value, and that the failure is recorded rather than swallowed.
- `.github/workflows/demo-chat-settings-check.yml` (new): a daily scheduled
  run of the existing detector script (shipped by PR #863, never wired into
  any workflow), as a backstop. `workflow_dispatch` on it also repairs the
  account and captures fresh proof with the existing
  `docs/proof/issue-855-enter-key/capture.mjs`.

## Repair + proof: done, by a separate credentialed agent

Live repair and proof capture needed real Supabase credentials this sandbox
does not hold; a separate credentialed agent ran them directly against main
(commit `f009fe1`, pushed via the GitHub git-data API, not a local checkout
of mine) and posted the results as a PR comment:

| step | command | exit code |
| --- | --- | --- |
| before | `node scripts/demo-chat-settings.mjs demo@hive-demo.invalid` | 1, `ui.ctrlEnterToSend is true` |
| repair | `node scripts/demo-chat-settings.mjs demo@hive-demo.invalid --repair` | 0, `repaired demo@hive-demo.invalid` |
| after | `node scripts/demo-chat-settings.mjs demo@hive-demo.invalid` | 0, `demo-safe` |

Screenshot (Enter sent, `HIVE ENTER PROBE` answered) and the full capture log
are attached in that PR comment and committed under
`docs/proof/issue-855-enter-key/post-fix-*`. Shift+Enter was checked in the
same capture and still inserts a newline without sending.

## Review fixes (stage 6, 5 inline comments)

- **Double navigation in `relocateAfterFailure` (blocking).** It called
  `surface.open(page)` and then `enumerateSurface`, which performs that same
  open internally and unconditionally (`surfaces.ts` line 325). Every
  recovery attempt therefore issued two navigations instead of one, doubling
  this exact path's exposure to the #815 unreachability window it exists to
  survive, and it was why the new regression test failed on the pushed head
  (`expect(indexRequests).toBe(3)`, received 4). Fixed by dropping the
  explicit `open` call; `enumerateSurface` already does it once.
- **Tail retry loop leaked past its own guard.** `persistOne`'s leading
  `enumerateSurface` call (used to locate a `missing`/`reverted` control
  before its own individual retry) was unguarded, so a throw there escaped
  `persistencePass` entirely and silently abandoned every remaining pending
  control in that loop, not just the one that failed to re-open. Each call in
  the loop is now wrapped in its own `try/catch`, isolating one control's
  failure from the rest.
- **Header comment overstated the guarantee.** Narrowed to what the code
  actually delivers: every throw that travels this file's own call chain is
  caught, but a Playwright worker-level interrupt (`test.setTimeout` firing
  mid-pass, or an unrelated unhandled rejection reaching `testInfo._interrupt()`)
  bypasses ordinary `try`/`catch`/`finally` entirely and is out of scope by
  construction, not by oversight.
- **Scheduled detector couldn't tell drift from its own failure, and was
  silent either way.** `demo-chat-settings.mjs` (unchanged, from PR #863)
  exits 1 both on real drift and on any other failure (sign-in, box
  unreachable, timeout), so a red scheduled run was ambiguous, and the
  workflow filed no issue and sent no notification. The `check` step now
  classifies the two by the script's own distinct output (`grep`-ing for its
  "re-run with --repair" line) rather than by exit code, and a new
  `report-failure` job (schedule-trigger only, mirroring
  `deploy-demo-box.yml`'s own pattern) files or comments on a GitHub issue,
  with a different label for "drift confirmed" versus "check could not run."

- **The restore step could move a control its own pass never touched.** Raised
  on the pushed head. Restoring a toggle means clicking it, because a switch or
  a checkbox has no "set it to this value" affordance, so a restore click is a
  mutation of its own whenever the control already sits at its baseline.
  `valuePass` restored unconditionally after a flip that may have thrown before
  it touched the control, or that the application accepted and then reverted;
  either way the pass moved a live account's setting away from the value it
  exists to preserve, the same defect one path over. The state comparison now
  lives inside `restore()` rather than in the caller, because it is a property
  of `restore` itself and because `persistencePass` and `persistOne` already had
  it while `valuePass` did not, which is exactly the shape where fixing one
  caller leaves the next free to repeat it. The read is fresh rather than a
  reuse of the pass's last known value: a flip that lands and then breaks the
  page leaves that value stale in the one direction that matters, and skipping
  the restore there would leave a live account mutated.

Regression test proven to discriminate, not just written: with the double-navigation
bug temporarily reintroduced, `npx playwright test
--config=e2e/chat-coverage/playwright.chat-coverage.config.ts
--project=chat-coverage-break-proof -g "reload failure"` fails on
`expect(indexRequests).toBe(3)`, received 4 (the exact defect). With the fix
in place, the same command passes. Full break-proof self-check suite (4
tests) passes with the fix applied. `npx tsc --noEmit` is clean.

The route-abort simulation itself needed a small correction while proving
this: Chromium retries a GET that fails with a connection reset before any
response begins, so the original `route.abort("connectionreset")` let the
retry succeed and `page.reload()` never actually threw, which would have
made the test pass for the wrong reason. The fixture's failing request now
hangs (never fulfilled, never aborted) with the test's own `navigationTimeout`
scoped down to 3s, the same pattern this file's own "Hanging" button fixture
already uses, so the timeout (not a browser-level retry) is what rejects the
reload deterministically.

## Rebase, and the Web E2E failure on the previous head

Rebased onto current `main`, from merge base `bd25ed7` through 50 commits
(the model-catalog restructure, the LiteLLM pin, the openrouter-auto pricing
fix, and the self-hosted Supabase cutover among them). No conflict:
`apps/web-console/e2e/chat-coverage/` is untouched on `main` since that merge
base, so this branch's 178-line removal from `chat-coverage.spec.ts` still
applies cleanly against main's current 811-line version.

The `Web E2E (full stack)` failure on the previous head is not a defect in
this branch, and rerunning it is not what settled that. Reading the job log
(run 31956812469, 2026-08-16): `27 executed, 2 passed only on retry, 0 failed.
Flake rate 7.4% (gate: at most 0 retry-pass)`. Nothing failed. The job tripped
its own zero-retry-pass flake gate, on
`tests/e2e/console-billing.spec.ts:45` and
`tests/e2e/profile-completion.spec.ts:39`, both of which passed on retry, and
neither of which this diff touches: the `Web E2E` job runs
`apps/web-console/tests/e2e`, while everything here is under
`apps/web-console/e2e/chat-coverage` plus one new workflow file. The same log
shows the actual driver, an auth-fixture reseed climbing from 2 seconds to
57.6 seconds against a shared Supabase pooler, which is the contention shape
already tracked in issue #436 and which the throwaway-Postgres CI change
(PR #983, 2026-08-21) removes from this path. The same head has since passed
`Web E2E` twice on its own, before this rebase, at 2026-08-17 08:07 and
2026-08-23 02:17.

## Both guards proven to fail, not just to pass

A guard that has never been watched fail is worth nothing, so each was
confirmed green against correct input first, then broken deliberately, then
restored.

**Restoration after a failed reload** (`break-proof.spec.ts`, `a reload
failure while verifying persistence still restores the flipped control`):

| step | result |
| --- | --- |
| unmodified branch | `1 passed` |
| `return;` added at the end of `persistencePass`'s catch block, so restoration is skipped after a failed reload | `1 failed`, `the fixture's persisted state was left flipped after a reload failure aborted the verification step; the control was never restored`, expected `false` received `true` |
| mutation reverted | `1 passed` |

**The new drift-check workflow.** Its own `check` step's `run:` body was
extracted from the shipped YAML (not retyped) and executed against a stubbed
`node` reproducing the three shapes the unchanged detector can produce, with
the detector's own output strings:

| account state | expected | observed |
| --- | --- | --- |
| demo-safe | exit 0, `kind=clean` | exit 0, `kind=clean` |
| `ui.ctrlEnterToSend` left flipped | exit 1, `kind=drift` (job red, drift issue filed) | exit 1, `kind=drift` |
| sign-in failed, box unreachable | exit 1, `kind=broken` (job red, separate label) | exit 1, `kind=broken` |

Broken deliberately by changing the classifier's `grep` to a string the
detector never prints: the drifted-account case then misclassified as
`kind=broken`, which is the harness catching a classifier that no longer
discriminates. Reverted, and all three shapes classify correctly again. The
detector's own drift definition is separately pinned by
`scripts/demo-chat-settings.test.ts` (5 tests, green), including
`driftedKeys({ ctrlEnterToSend: true })`.

Nothing was run against the live demo account for this: every result above
comes from fixtures and stubs, so no shared account state was touched.

## Buglog entry

For the follow-up buglog-only PR against `main`
(`.claude/rules/openwolf.md`); not appended to `.wolf/buglog.jsonl` on this
branch:

```json
{"date":"2026-08-16","error_message":"Enter key does not send in the demo chat composer","root_cause":"chat-coverage's live interaction-coverage sweep (persistencePass) flips every stateful Settings control, including ui.ctrlEnterToSend, on the shared demo account to prove persistence, then reloads to verify; the restore step ran after that reload with no guard, so a reload failure (demo box briefly unreachable, issue #815) threw past restoration and left the account's Enter Key Behavior toggled on with no error surfaced","fix":"wrap the reload-and-verify step in persistencePass/persistOne (apps/web-console/e2e/chat-coverage/persistence.ts) in try/catch and run restoration unconditionally afterward; added a scheduled detector workflow (demo-chat-settings-check.yml) as a backstop and a regression test in break-proof.spec.ts","tags":["owui","e2e","chat-coverage","demo-account","regression"]}
```

## Tests

- Full `chat-coverage-break-proof` self-check suite, now 5 tests: green.
- Both new regression tests proven to discriminate, with the RED and GREEN
  runs recorded above rather than asserted.
- `scripts/demo-chat-settings.test.ts` (vitest, 5 tests): green.
- `npm run e2e:verify-collection` (spec collection guard): OK, 39 spec files
  across 3 configs, each collected into the projects the manifest pins it to.
- `npx tsc --noEmit`: clean.
- `chat-coverage.spec.ts` unchanged in behavior; only the persistence logic
  moved to `persistence.ts` and is now imported.

## Summary by CodeRabbit

- **New Features**
  - Added automated daily checks for demo chat settings, with optional manual repairs and verification.
  - Added failure classification and deduplicated issue reporting for scheduled checks.
  - Added proof artifact capture for manual Enter-key verification.

- **Bug Fixes**
  - Improved chat settings persistence checks, including recovery after reload failures.
  - Ensured changed controls are restored after verification, including best-effort retries.
  - Added value verification for controls that do not persist across reloads.


sakibsadmanshajib:
## Live proof: Enter sends again on the demo account

Captured 2026-08-16 against the deployed chat at `https://chat-hive.scubed.co`,
signed in through the audited magic link helper
(`apps/web-console/tests/e2e/support/live-auth.mjs`). No password was set,
reset or rotated, and `scripts/seed-owui-e2e-user.py` was not run.

### Detector, before and after, by exit code rather than by assertion

| step | command | exit code | output |
| --- | --- | --- | --- |
| before | `node scripts/demo-chat-settings.mjs demo@hive-demo.invalid` | **1** | `ui.ctrlEnterToSend is true, demo-safe is false` |
| repair | `node scripts/demo-chat-settings.mjs demo@hive-demo.invalid --repair` | **0** | `repaired demo@hive-demo.invalid` |
| after | `node scripts/demo-chat-settings.mjs demo@hive-demo.invalid` | **0** | `demo@hive-demo.invalid is demo-safe` |

The before run is the recurrence this pull request is about: the account had
drifted back to `ctrlEnterToSend: true`, so Enter inserted a newline instead of
sending, with nothing on screen saying so.

### Capture

`node ../../docs/proof/issue-855-enter-key/capture.mjs post-fix /tmp/proof demo@hive-demo.invalid`

Enter sent the prompt and the assistant answered

Full run log, committed alongside the screenshots as
`docs/proof/issue-855-enter-key/post-fix.log` in f009fe1:

```
label: post-fix
url: https://chat-hive.scubed.co/
settings.ui.ctrlEnterToSend: false
POST /api/chat/completions after Enter: 1
composer after Enter: "\n"
assistant answer after Enter: "HIVE ENTER PROBE"
ENTER SENT AND ANSWERED: true
POST /api/chat/completions after Shift+Enter: 0
composer after Shift+Enter: "first line\n\nsecond line"
SHIFT+ENTER MADE A NEWLINE AND DID NOT SEND: true
cleanup: deleted 1 chat(s) [a393a5e8=200]
```

The assertion is on the assistant's rendered answer text, not on a Copy button,
so an error bubble could not pass it. Shift+Enter still makes a newline and
still does not send, so the repair did not trade one keybinding for the other.
The run deleted the one chat it created, so this capture added no litter to the
account.

No URL in this journey carries a credential in a query string or fragment, so
nothing needed masking in either the log or the screenshot pixels; the captured
URL is the bare chat origin.

Two related notes for whoever reviews this:

1. This proves the account is repaired and the detector reports honestly. It
   does not prove the restoration guard in `persistence.ts` holds, since that
   needs a live sweep run whose reload fails; `break-proof.spec.ts` is what
   covers that arm.
2. The drift was live for an unknown window before this run, which is the
   argument for wiring the detector into a pre-demo step rather than leaving it
   to be run by hand.

apps/web-console/e2e/chat-coverage/persistence.ts
```diff
@@ -0,0 +1,290 @@
...
+  await el.fill(before ?? "", { timeout: 6000 });
+}
+
+/**
+ * Re-read a control after whatever just happened, tolerating a page that may
+ * be mid-recovery from a failed reload. Never throws.
+ */
+async function relocateAfterFailure(page: Page, surface: Surface): Promise<Control[]> {
+  try {
+    await surface.open(page);
```
sakibsadmanshajib: Blocking: relocateAfterFailure double navigates, and it is why the new regression test is red in CI. Line 86 calls surface.open(page) directly, then line 87 calls enumerateSurface(page, surface), which itself calls surface.open(page) again unconditionally as its own first step (surfaces.ts line 325, outside the delta branch). Every recovery attempt through this function therefore issues two navigations, not one.

This is not theoretical. The pushed CI run for this PR (Chat interaction coverage workflow, run 31945246365, job 95160111064, Chat coverage self check) fails on break-proof.spec.ts's new test with expect(indexRequests).toBe(3), received 4: one baseline enumerate, one aborted reload, and two from this single relocateAfterFailure call instead of one.

Two consequences follow. First, the new regression test cannot currently be said to validate the restoration guarantee since it fails with the fix applied, for a reason unrelated to what it claims to check, so question 2 (does it actually discriminate fixed from unfixed) is unanswered as shipped. Second, in production this doubles the network round trips taken during the exact recovery path meant to survive issue 815's brief unreachability window, doubling the chance of hitting that same window again mid recovery.

Suggest dropping the explicit surface.open(page) call on line 86 since enumerateSurface already performs it, then re-checking the request count assertion in break-proof.spec.ts against the corrected behavior.

apps/web-console/e2e/chat-coverage/persistence.ts
```diff
@@ -0,0 +1,290 @@
...
+      target = fresh.find((c) => c.key === ctl.key);
+    }
+    if (!target || target.state === before) continue;
+    await restore(page, target, before).catch(() => {});
+  }
+  await saveIfPresent(page).catch(() => {});
+
+  // Controls the batch could not re-locate (a flipped toggle can hide its own
+  // dependants) get an individual pass rather than a false negative.
+  for (const ctl of [...missing, ...reverted]) {
```
sakibsadmanshajib, line 222: High: this loop is the one place in the file where the restore guarantee the header promises is not actually unconditional. The loop is not wrapped in try/catch, and persistOne's very first statement (line 208, const live = await enumerateSurface(page, surface)) is unguarded, unlike everything else in this file.

If the page is still in the broken state that made the earlier reload fail (issue 815 describes a window of unreachability, not a single instant, and the double navigation in relocateAfterFailure noted on line 86 widens that window further by adding a second network round trip per recovery attempt), this throws, propagates out of persistOne, and propagates out of persistencePass itself since nothing here catches it.

The outer try/catch in chat-coverage.spec.ts around the persistencePass and valuePass calls still catches it, so this will not crash the worker, but every remaining control in this loop, exactly the controls the batch pass could not confirm were restored, is left un-retried with no restoration attempt at all, silently. That is the same class of defect this file exists to close, one level deeper than the reload block above it.

Suggest wrapping this loop's body (or persistOne's leading enumerateSurface call) the same way the reload block above is wrapped, so a failure here is recorded rather than allowed to unwind past the remaining retries.

apps/web-console/e2e/chat-coverage/persistence.ts
```diff
@@ -0,0 +1,290 @@
+// Persistence proof for stateful controls (switches, checkboxes, selects,
+// text inputs) on the chat coverage sweep.
+//
+// WHY THIS FILE EXISTS (issue #855, recurrence 2026-08-14)
+// ---------------------------------------------------------
+// This pass flips a real control on a shared, live account to prove the app
+// saves it, which is a genuine mutation of that account's settings before
+// anything here reads a verdict. Every function below therefore treats
+// "restore the account to what it was" as the one step that may never be
+// skipped, whatever else in the pass throws.
```
sakibsadmanshajib, line 10: Medium: this claim (whatever else in the pass throws) holds only for throws inside this file's own call chain, and only once the two gaps flagged elsewhere in this review (the double navigation in relocateAfterFailure, and the unguarded tail retry loop over missing and reverted controls) are actually closed. It does not, and cannot from inside this file alone, cover a Playwright worker level interrupt: either the enclosing test's own test.setTimeout(45 * 60_000) in chat-coverage.spec.ts firing mid pass, or an unrelated unhandled rejection elsewhere in that same 45 minute test reaching workerMain.js and calling testInfo._interrupt(). That path bypasses normal try/catch/finally unwinding entirely; it is not a JS exception traveling up the call stack, so no amount of try/catch here closes it.

I did not find a new floating (unawaited) promise introduced by this PR, every call in this file and in break-proof.spec.ts is awaited with a per-call catch, so this diff does not add that specific risk. But the comment's wording overstates what the fix actually guarantees. Worth a one line caveat here or in the PR description so a future reader does not assume this closes the worker interrupt path too.

.github/workflows/demo-chat-settings-check.yml
```diff
@@ -0,0 +1,87 @@
...
+          done
+          if [ -n "$missing" ]; then
+            echo "::error::missing:$missing"
+            exit 1
+          fi
+      - run: npm ci
+      - run: npx playwright install --with-deps chromium
+      - name: Check demo account settings
+        if: github.event_name == 'schedule'
+        run: node scripts/demo-chat-settings.mjs "$DEMO_ACCOUNT_EMAIL"
```
sakibsadmanshajib: High: on the schedule trigger this calls scripts/demo-chat-settings.mjs with no --repair, and that script (unchanged in this PR, shipped by PR 863) exits 1 both when it finds real drift (an intentional return 1 when drift.length > 0 and repair is false) and when anything else throws: sign in failure, box unreachable, timeout (the catch at the bottom of main() also calls process.exit(1)). A red scheduled run is therefore indistinguishable by exit code alone between the account actually drifting and the box hitting one of the issue 815 blips this whole PR exists to guard against.

This workflow also files no issue and posts no notification on failure, the header comment above concedes as much (the note about adding issue filing like deploy-demo-box.yml's report-failure job only if the absence of one proves to be a problem). GitHub does not push a default notification for a failed scheduled workflow beyond a passive email to whichever actor GitHub attributes the run to, and only if that person opted into failure emails for this repository. A string of red 5am runs, real drift or not, is therefore very plausibly never seen by anyone until the next time someone happens to check the Actions tab or the enter key breaks again live.

Recommend at minimum having the script, or a thin step here, distinguish drift found from could not check with different exit codes or a machine readable output, and wiring at least one loud channel (an issue, a chat post, the pattern deploy-demo-box.yml already uses for its own failures) on schedule triggered failures specifically, not only on workflow_dispatch.

sakibsadmanshajib:
Adversarial review, stage 6 of the coding pipeline. Four inline findings posted on the diff; summary below.

Merge readiness: this PR is a draft, mergeStateStatus is UNSTABLE. The workflow check Chat interaction coverage / Chat coverage self check (no deployment needed) is currently FAILING on the pushed head commit (run 31945246365, job 95160111064). It is not in main's required status checks list (Go tests x4, Repo policy lints, Web console type+unit+build), so it will not block a GitHub merge by itself, but it is the actual test suite this PR ships, failing on the pushed commit, and should be treated as a real red, not a hypothetical.

Finding on question 2 (does the new regression test discriminate fixed from unfixed): as shipped, it does not currently pass at all. It fails with the fix applied, per the CI run above, for a reason unrelated to what it claims to test. Root cause and detail posted inline on persistence.ts line 86: relocateAfterFailure calls surface.open(page) directly and then calls enumerateSurface, which performs that same open internally, so every recovery attempt double navigates. The test's exact request count assertion (expect(indexRequests).toBe(3)) can be shown not to hold either for the old unfixed code (would produce 2) or the new code as shipped (produces 4), so the assertion needs the double navigation fixed before it says anything reliable.

Finding on question 1 (is the guard actually unconditional): mostly yes for the main reload and compare block, which is now well guarded. One real gap remains, posted inline on persistence.ts line 198: the tail loop retrying missing and reverted controls through persistOne is not itself wrapped, and persistOne's leading enumerateSurface call is unguarded, so a second failure landing in that same recovery window can propagate out of persistencePass uncaught (caught by the outer chat-coverage.spec.ts try/catch, so it will not crash the worker, but it silently skips restoration for whichever controls were still pending retry). Separately, posted inline on line 10, the header's whatever else throws claim cannot cover a Playwright worker level interrupt (test.setTimeout firing, or an unrelated unhandled rejection reaching workerMain.js and calling testInfo._interrupt(), the known repo lesson named in the review brief); nothing in this diff introduces a new floating promise, but the fix also cannot close that channel from inside this file, and the comment overstates the guarantee.

Finding on question 3 (the scheduled workflow): posted inline on demo-chat-settings-check.yml line 70. The schedule path's exit code cannot distinguish real drift from the detector script simply failing to reach the box (issue 815 again), and the workflow files no issue and sends no notification on a red run, so a string of failures is plausibly invisible until someone checks the Actions tab by hand.

Finding on question 4 (same shape elsewhere): within apps/web-console/e2e/chat-coverage, the flip, verify, restore shape now lives in exactly one place, persistence.ts (persistencePass, persistOne, valuePass). No other function in that directory calls page.reload() or defines a second restore or flip. clickPass in chat-coverage.spec.ts mutates and clicks but never restores, a different contract. Other web-console e2e specs (console-budgets.spec.ts, console-workspace-admin.spec.ts, console-workspace-switch.spec.ts, agent-workspace-flows.spec.ts) also call page.reload(), but they appear to run against ephemeral test orgs rather than the shared demo account; not verified in depth here given scope, worth a follow-up sweep rather than treated as a confirmed second instance.

One more item, not a code defect but worth fixing before merge: the PR body says repair and proof are outstanding because this sandbox has no Supabase credentials, and that the PR should not merge without a screenshot attached. The pushed diff already contains docs/proof/issue-855-enter-key/post-fix-1-typed.png, post-fix-2-after-enter.png, post-fix-3-shift-enter.png, and post-fix.log, added in a second commit (f009fe11, docs: add post-fix Enter key proof captured against the live demo chat, 2026-08-16T15:29:18Z, after the first commit ba4dd063 the body describes). The log shows a genuine looking live capture against https://chat-hive.scubed.co/ (settings.ui.ctrlEnterToSend: false, ENTER SENT AND ANSWERED: true, cleanup deleted 1 chat), and the PNG is a real 1440x900 capture, not a placeholder. This reads as a stale body rather than fabricated evidence, but the body should be updated before merge so a reviewer is not told to withhold approval for proof that is already sitting in the diff.

Recommendation: do not merge as is. Fix the double navigation in relocateAfterFailure and re-verify the regression test actually goes red without persistence.ts's guard and green with it, close the tail loop gap, and update the PR body to match the second commit, before this goes back through CI.

apps/web-console/e2e/chat-coverage/persistence.ts
sakibsadmanshajib: Fixed in 95beaed3: dropped the explicit surface.open(page) call in relocateAfterFailure, since enumerateSurface already performs it. Verified locally rather than assumed: reintroducing just that one line reproduces the CI failure (expect(indexRequests).toBe(3), received 4); with the fix restored the same test passes, and the full chat-coverage-break-proof suite (4 tests) is green.

apps/web-console/e2e/chat-coverage/persistence.ts
sakibsadmanshajib: Fixed in 95beaed3: each call to persistOne inside the tail retry loop is now wrapped in its own try/catch, so a throw from its leading enumerateSurface (no target located yet, nothing to restore for that one attempt) is recorded as an unprovable result instead of escaping the loop and abandoning every control still pending behind it.

apps/web-console/e2e/chat-coverage/persistence.ts
sakibsadmanshajib: Agreed, fixed in 95beaed3: narrowed the header comment to state plainly that the guarantee covers only throws traveling this file's own call chain (a failed reload, re-open, or click), and does not cover a Playwright worker-level interrupt (test.setTimeout firing mid-pass, or an unrelated unhandled rejection reaching testInfo._interrupt()), since neither is a JS exception this file's try/catch/finally can intercept.

.github/workflows/demo-chat-settings-check.yml
sakibsadmanshajib: Fixed in 95beaed3: the schedule-trigger check step now classifies drift-confirmed vs check-could-not-run by the script's own distinct output (grepping its re-run with --repair line) rather than by exit code, and a new report-failure job (schedule-trigger only, mirroring deploy-demo-box.yml's own report-failure pattern) files or comments a GitHub issue with a label naming which case happened, so a red 5am run is both loud and distinguishable.

sakibsadmanshajib:
Update summary, commit 95beaed3.

All four inline findings fixed (replies + threads resolved on each):
1. relocateAfterFailure double navigation, the cause of the CI red on the pushed head: fixed.
2. Tail retry loop's unguarded persistOne call: fixed, wrapped per-control.
3. Header comment overstating the guarantee: narrowed to name its own limits (worker-level interrupts out of scope).
4. Scheduled detector indistinguishable/silent on failure: fixed, classifies drift vs broken by output and files/comments a GitHub issue on schedule-triggered failures.

Red-then-green, run explicitly rather than assumed: with the double-navigation bug reintroduced, the regression test fails on expect(indexRequests).toBe(3), received 4 (the exact defect). With the fix restored, the same test passes. Full chat-coverage-break-proof suite (4 tests): green. npx tsc --noEmit: clean.

PR body updated: repair and proof are done (not outstanding), performed by a separate credentialed agent against main (commit f009fe1, pushed via the GitHub git-data API), not by me. I hold no Supabase credentials in this sandbox and did not capture docs/proof/issue-855-enter-key/post-fix-*.png myself.

Issue #708, opened 2026-08-04
Tests that never execute in CI: phase-19 tenant-isolation suite, edge-api DB tests, and 4 more integration tests share #701's failure shape


sakibsadmanshajib:
## Why this audit exists

Issue #701 found that `apps/control-plane/internal/litellmconfig/sync_integration_test.go` was gated behind `LITELLM_TEST_DB_URL`, a variable that appeared nowhere under `.github/`. The test had never run in CI. It seeded the correct column name while production queried a nonexistent one, so it would have failed on its first real execution. A skip is green and therefore invisible; PR #705 fixed it by wiring `LITELLM_TEST_DB_URL` into `ci.yml`.

This issue is the result of checking how many more tests share that failure shape: every Go test that can call `t.Skip`/`t.Skipf`/`t.SkipNow` or is gated behind a `//go:build` tag, and every Playwright/Vitest test that can call `test.skip`/`it.skip`/`test.fixme` or an equivalent conditional early return, across the whole repo, cross-checked against what `.github/workflows/*.yml` actually wires up and against a live CI log (run 30919970746, the latest push to main as of this audit).

Denominator examined: **36 Go test files** carry a skip condition or an integration-only build tag; **~43 TypeScript/Playwright spec files** carry an env-gated `test.skip`. Of those: **Go: 28 run for real (group B), 3 are deliberately and correctly gated (group C), 6 never run (group A). TypeScript: roughly 25 run for real (group B, split across the `web-e2e` job, the nightly `owui-nightly.yml`, and the `live-integration` job's SDK suites), about 4 are deliberately and correctly gated (group C), 14 never run (group A).**

## Group A: never runs in CI, ranked by severity

| # | Test | Location | Gating condition | What it claims to cover |
|---|------|----------|-------------------|--------------------------|
[...]

Issue #600, opened 2026-07-29
Orphan-hold reaper + expires_at on reservations; nothing consumes credit_reconciliation_jobs


sakibsadmanshajib:
## Context

PR fixing the disconnect-settlement defect (fix/disconnect-settlement-context) removes the main producer of orphan reservation holds: a mid-stream client disconnect now always settles (finalize or release) on a bounded background context instead of the cancelled request context. That closes the bug going forward, but it does not address two related gaps in the accounting model itself.

## Gap 1: no reservation expiry

`reservations` has no `expires_at` column and no reaper. Any hold that fails to settle for a reason other than the fixed disconnect path (crash mid-request, panic before the defer runs, a future bug in a different code path) has no backstop and stays `active` forever, permanently reducing the account's spendable balance with no recovery mechanism.

## Gap 2: credit_reconciliation_jobs is write-only

`apps/control-plane/internal/accounting/repository.go` inserts rows into `public.credit_reconciliation_jobs` (see the INSERT around line 241) whenever a settlement is flagged `TerminalUsageConfirmed: false` (an estimate-based settle, including the disconnect path this PR introduces). No worker anywhere in the codebase reads or processes that table. Every reconciliation job written today, and every one this PR's disconnect fix will write going forward, sits there unconsumed.

## Ask

1. Add `expires_at` to reservations (or the equivalent hold record) and a reaper job that force-settles or flags for review any reservation past its expiry.
2. Build a worker that actually consumes `credit_reconciliation_jobs`: reconcile the estimated charge against the real usage once it becomes available (or flag for manual review if it never does).
3. Separately: since this PR's fix removes the main *producer* of orphan holds, any historically stranded holds already sitting in production from before the fix are not touched by it and will continue to suppress customer balances until they are found and released by hand (or by the reaper this issue asks for).

## Related

Follows from the disconnect-settlement-context fix (stream.go / stream_responses.go mid-stream settlement).

PR #912, opened 2026-08-16
fix: lift the whole credit hold at capture, so a settled request stops being billed twice

sakibsadmanshajib:
## What broke

Credit reservations were never fully released at settlement. Measured live on the demo box on 2026-08-16, one deliberate chat completion, console before and after:

```
before:  Available 9,998,840   Posted 9,999,920   Reserved 1,080
after:   Available 9,998,804   Posted 9,999,902   Reserved 1,098
```

Posted fell by 18, the true cost of the work. Reserved ROSE by 18. Available, rendered as posted minus reserved, fell by 36 for 18 credits of work. Reserved had climbed to 1,080 over earlier traffic and then sat flat for 12 idle minutes, which rules out live in-flight holds.

## Confirmed at the ledger, not at the console

Read only, through the transaction-mode pooler, 2026-08-16. The measured request is reservation `e2006ebf`, and its three ledger rows are the whole finding:

| entry_type | credits_delta |
| --- | --- |
| reservation_hold | -10000 |
| reservation_release | +9982 |
| usage_charge | -18 |

`GetBalance` (`apps/control-plane/internal/ledger/repository.go`) computes `reserved = ABS(SUM(hold) + SUM(release))` and `available = posted - reserved`. Hold plus release nets to -18, so 18 credits stayed in the reserved bucket permanently, while the same 18 had already left posted via the charge. The customer paid twice for one request.

Population, whole database, at the time of the read:

| reservation status | rows still holding | credits still held | attempt status |
| --- | --- | --- | --- |
| finalized | 1,695 | 73,444 | completed |
| needs_reconciliation | 130 | 70,198 | completed (129), streaming (1) |
| active, genuinely in flight | 1 | 1,000 | dispatching |

For every one of those 1,825 terminal rows, credits still held equals `consumed_credits` equals the charge, exactly. Their request attempts are `completed`. Earliest stranded row is 2026-04-01, so this has been true since the ledger began; 189 rows and 19,405 credits were added on 2026-08-16 alone. Worst affected account (`9bbaf7b5`): posted 158,604, reserved 118,281, available 40,323, so three quarters of its balance was withheld for work that had already settled. That is the service-denial mode the reaper's own header comment warns about.

## Why release did not happen

Not an error path, not a keying mismatch, not an orphaned hold, not a swallowed error. It is the arithmetic of the successful request. `finalizeLocked` released `releasableCredits(reservation, actual)`, which is `held - actual`, and nothing ever lifted the `actual` the charge had captured. Each hypothesis worth eliminating, and how the data eliminated it:

- an error path returning before releasing: no, the release posts and the row reaches a terminal status;
- a release keyed differently from the hold: no, both entries carry the same `reservation_id` and the flat `reservation:<id>:release` key;
- a settlement that posts and returns without touching the hold: no, it touches it partially;
- a failed or cancelled request whose hold has no owner left: no, every stranded row's attempt is `completed`;
- a release that runs and silently fails: no, the release row exists, for the wrong amount.

The reaper (issue #648) cannot recover these: it only scans `active` and `expanded` holds, and these are terminal.

The same function already treated capture as lifting the whole hold everywhere else: `ApplyReservationDelta(apiKeyID, -heldCredits, actualCredits)` unwinds the API key counter by the FULL held amount. The ledger was the only book that disagreed with itself.

## The invariant, stated as a bound

For every reservation, once it reaches a terminal state (`finalized`, `needs_reconciliation`, `released`):

```
SUM(reservation_hold.credits_delta) + SUM(reservation_release.credits_delta) = 0
```

Exactly one of release or capture happens for the authorization, exactly once. The hold is lifted in full, once, under the fixed key `reservation:<id>:release`, and settlement now refuses to complete if that release deduplicates against an earlier entry for a different amount, so a partly lifted hold can no longer be recorded as settled. The spend is posted at most once, enforced by the `reservationOpen` status allowlist plus the per-account advisory lock; the charge idempotency key still carries its amount, and closing that second gap needs a key-normalization migration, filed separately (see below). Therefore an account with nothing in flight has `reserved = 0`.

As a magnitude, which is the form that can actually be measured: for a request that reserves H and truly costs C, available must fall by exactly C, and reserved must return to its pre-request value. Not C plus a leftover. Not H minus anything. C is D-031's `(prompt_tokens * input_price + completion_tokens * output_price) / 1_000_000`, rounded half up, unchanged by this PR.

## The fix

The ledger release at capture changes from `held - actual` to `held`. One number. The charge still posts separately for `actual`.

The reservation row keeps its column meanings: `consumed_credits = actual` and `released_credits = held - actual`, which still sum to `reserved_credits` on every path except a confirmed overage, where the charge legitimately exceeds the hold and the unused part is zero. Those columns answer how the authorization was disposed of (spent versus returned unused); the ledger release entry answers whether any authorization is still outstanding, and after capture the answer must be no. The `usage_events` metadata keeps reporting the unused figure, so analytics are unchanged.

Settlement also refuses to complete when its release deduplicates against an earlier, smaller release (`assertReleasedInFull`). `PostEntry` returns the FIRST entry for a repeated key, so that call was returning quietly with the hold only partly lifted, the row was then marked terminal, and the remainder was stranded past the reaper, which only scans non-terminal holds. That is this very defect reintroduced silently, and the sequence is reachable across the deploy of this PR: the old code posts a partial release, then fails before updating the row, and the retry runs on new code. Failing loudly leaves the reservation open for the reaper instead, per D-034's "an unrecognized failure is fatal".

The first revision of this PR also flattened the charge idempotency key (`charge-<credits>` to `charge`). That was reverted during review: flattening it needs a migration normalizing the 1,866 existing amount-bearing entries first, exactly as issue #663 needed one for the release key, because until they are normalized a retry crossing the deploy posts a SECOND charge under the new shape instead of deduplicating against the old one. A double bill is not an acceptable price for closing a rarer double bill, so it is filed separately.

### Rejected: fix the reader instead

`reserved = ABS(SUM(hold+release)) - ABS(SUM(usage_charge with reservation_id))` also nets to zero on today's data. Rejected because a confirmed overage (`TerminalUsageConfirmed=true` bills past the flat 10,000 hold, which long context and coding-agent traffic reach) makes that expression negative for the reservation and silently INFLATES available. Clamping it needs a per-reservation `GREATEST(0, ...)` group-by inside `GetBalance`, which runs on the hot path under the per-account lock, and there is no index on `credit_ledger_entries.reservation_id`. It would also leave the writer posting a half-truth for every future reader to compensate for, which is the opposite of failing closed.

## The five changed test expectations

Each was written by someone who believed it described correct behaviour, so each is classified rather than merely updated.

Encoded the leak, four of them. These asserted, in credits, that part of a settled request's hold stays in the reserved bucket:

1. `TestFinalizeReservationCreatesChargeAndRelease`: asserted a 30-credit release against a 100 hold with 70 charged, that is, that 70 credits stay reserved after the request settled. Changed to 100.
2. `TestFinalizeReservationClampsChargeToReservedHold`: asserted **no release call at all** when the clamped charge consumes the hold, that is, that the entire 10,000 stays reserved forever. Changed to a 10,000 release. This is the worst form of the leak and it was pinned as correct.
3. `TestFinalizeReservationConfirmedUsageAboveHoldIsNotClamped`: same shape as 2, on the confirmed-overage path. Changed to a 10,000 release. The test's actual subject, that a confirmed charge is not clamped, is untouched and still asserted.
4. `TestFinalizeReservationMarksAmbiguousStreamForReconciliation`: asserted a 50-credit release against a 90 hold with 40 charged. Changed to 90. Its subject, the reconciliation job, is untouched.

Incidental, one of them:

5. `TestReleaseAfterUnconfirmedSettlementKeepsRowBalanced`: its subject is that a second release is refused and the row still balances, and every one of those assertions is untouched and still passes. The changed line only restated the finalize release amount in passing, though the number it restated was the leaking one.

Nothing else moved. No assertion about row columns, statuses, refusals, reconciliation jobs, or charge amounts changed. Worth naming: not one existing test asserted that reserved returns to zero, which is why 1,825 reservations leaked past a suite that was green the whole time.

## Verification, by ledger magnitude

New `TestSettlementLedgerMagnitude_Live` drives the real pgx repositories and the real `GetBalance` SQL on a real Postgres built from `supabase/migrations/` (the same bootstrap `.github/workflows/ci.yml` uses, `HIVE_TEST_DB_URL`-gated like `release_idempotency_test.go`), and compares balance magnitudes across one request that reserves 10,000 and costs 18.

```
pre-fix code:  FAIL  reserved did not return to its pre-request value: 18, want 0
                     (18 credits still held for a request that already settled)
post-fix code: PASS
```

That is the live console reading reproduced as a number, on the real schema. A test asserting `ReleaseReservedCredits` had been called would have passed on the broken code, because it WAS called, just for 18 credits too few.

That test only runs because this PR also adds `./internal/accounting/...` to the live-Postgres step in `.github/workflows/ci.yml`. The package was absent from that list, so both its `HIVE_TEST_DB_URL`-gated tests, this one and the existing release-key dedup proof from issue #652, skipped everywhere: the same never-runs trap as issues #701 and #708, on the money path. Caught by the review streams, not by me.

Also new, `TestSettlementReturnsReservedToZero`, four cases against the fake ledger (signed `held - released`, where production takes `ABS` of the same sum; they agree for every case here since none over-releases): actual far below the hold, zero actual, unconfirmed estimate clamped to the hold, and confirmed usage above the hold. Each asserts reserved returns to zero and available moves by exactly the charge, with a positive control that the hold really was on the books first. And `TestSettlementRefusesToSettleAgainstAPartialRelease` pins the new guard, verified by mutation: delete the guard and it goes red.

Full `go test ./apps/control-plane/...` and `./apps/edge-api/...` pass; `go vet` clean.

No UI surface is touched by this change, so the visual-proof rule does not apply here; the proof for a money path is the ledger magnitude above.

## Stranded balances already in the database, NOT touched by this PR

Nothing in this PR mutates live data. Counts read live on 2026-08-16:

| what | reservations | credits |
| --- | --- | --- |
| residual holds on terminal reservations, the defect above | 1,836 | 144,028 |
| released without a hold ever being posted, opposite sign, different cause | 10 | 25,000 |

By account, for the first row: `9bbaf7b5` 1,764 reservations / 141,665 credits, `3104dbfc` 3 / 2,000, `e518472a` 1 / 110, `d18c9024` 22 / 98, `29d908e9` 1 / 70, `feb1e8e2` 42 / 42, `22222222` 2 / 41, `31aadd76` 1 / 2.

Proposed remediation, for the owner to approve or refuse, to run only after this PR has deployed (running it earlier is pointless while new residue accrues). It appends one compensating release per affected reservation under a distinct key, so the ledger stays append-only, no row is updated or deleted, and a second run is a no-op:

```sql
-- PREVIEW FORM, read only. The remediation replaces the outer SELECT with the
-- INSERT below it.
WITH residual AS (
  SELECT r.id AS reservation_id,
         r.account_id,
         -COALESCE(sum(e.credits_delta) FILTER (
            WHERE e.entry_type IN ('reservation_hold','reservation_release')), 0) AS residual_credits
  FROM public.credit_reservations r
  LEFT JOIN public.credit_ledger_entries e ON e.reservation_id = r.id
  WHERE r.status IN ('finalized','needs_reconciliation','released')
  GROUP BY r.id, r.account_id
)
SELECT count(*) AS reservations, sum(residual_credits) AS credits
FROM residual
WHERE residual_credits > 0;

-- INSERT INTO public.credit_ledger_entries
--   (account_id, entry_type, credits_delta, idempotency_key, reservation_id, metadata)
-- SELECT account_id, 'reservation_release', residual_credits,
--        'reservation:' || reservation_id || ':release-residual',
--        reservation_id,
--        jsonb_build_object('reason', 'issue-616 capture residual', 'pr', '<this PR>')
-- FROM residual
-- WHERE residual_credits > 0
-- ON CONFLICT DO NOTHING;
```

Two follow-ups filed rather than bundled into this money-path fix: flattening the charge idempotency key behind its normalization migration, and the non-atomic create below.

The 25,000 in the second row is a different defect and is deliberately left out of that statement: `CreateReservation` writes the reservation row and posts the ledger hold in two separate transactions, so a hold that fails to post leaves a row the reaper later releases against nothing, inventing credits in the customer's favour. All ten are from July and were released in one pass on 2026-08-01. Filed separately rather than bundled into a money-path fix.

## Buglog entry

```json
{"id":"bug-mhold616-1a2f3c","timestamp":"2026-08-16T12:10:00.000Z","related_bugs":[],"occurrences":1,"last_seen":"2026-08-16T12:10:00.000Z","title":"Credit hold never lifted for the captured part, so every settled request was billed twice","error_message":"Console showed available falling 36 credits for an 18-credit completion; reserved rose by exactly the charge and stayed flat while the account was idle","root_cause":"finalizeLocked released remainingHeldCredits minus actualCredits and nothing ever lifted the captured remainder. GetBalance computes reserved as ABS(SUM(reservation_hold) + SUM(reservation_release)), so the captured credits stayed in the reserved bucket permanently while the usage_charge had already reduced posted, double counting the price of every settled request. The reaper could not recover them because it only scans holds still in a non-terminal state. Live: 1836 terminal reservations holding 144028 credits, one account with three quarters of its balance withheld.","fix":"Release the whole outstanding hold at capture (heldCredits, not heldCredits minus actualCredits) and post the charge separately, so hold and release cancel for every terminal reservation. The reservation row keeps consumed plus released summing to reserved. Also flattened the charge idempotency key from charge-<credits> to charge and made a dedup that returns a differing amount fatal, so a retried settlement cannot double charge or silently diverge from the row.","tags":["billing","ledger","credit-reservation","money-path","issue-616","D-034","control-plane","pr-reservation-capture-hold-leak"]}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,7 +427,16 @@ jobs:
             # absent from this list, so the escalation those guards catch
             # would have shipped again with the suite reporting green.
             # tools/lint-go-db-test-wiring.mjs fails if either is dropped.
-            go test -tags integration -count=1 -p 1 -v ./internal/accounts/... ./internal/agenttask/... ./internal/egress/... ./internal/payments/... ./internal/profiles/... ./internal/rag/... ./internal/tenants/... ./internal/signup/... ./internal/routing/... ./internal/litellmconfig/... ./internal/catalog/... ./internal/providers/... ./internal/platform ./internal/platform/db/...
+            # ./internal/accounting/... carries the credit-reservation
+            # settlement suites. Its two HIVE_TEST_DB_URL-gated tests, the
+            # release-key dedup proof (issue #652) and the settlement magnitude
+            # proof (issue #616), had never run anywhere: the package was absent
+            # from this list and the -short step above skips them for want of
+            # the env var. The magnitude one is the only check that measures the
+            # ledger on real Postgres rather than through a fake, so leaving it
+            # unlisted is the same never-runs trap as issues #701 and #708, on
+            # the money path.
+            go test -tags integration -count=1 -p 1 -v ./internal/accounts/... ./internal/accounting/... ./internal/agenttask/... ./internal/egress/... ./internal/payments/... ./internal/profiles/... ./internal/rag/... ./internal/tenants/... ./internal/signup/... ./internal/routing/... ./internal/litellmconfig/... ./internal/catalog/... ./internal/providers/... ./internal/platform ./internal/platform/db/...
           else
             # -tags integration additionally compiles in
             # inference/chat_completions_integration_test.go (issue #708),
--- a/apps/control-plane/internal/accounting/reaper.go
+++ b/apps/control-plane/internal/accounting/reaper.go
@@ -2,6 +2,7 @@ package accounting
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
@@ -166,6 +167,20 @@ func (r *Reaper) RunOnce(ctx context.Context) (ReapResult, error) {
 			// A refused release is not fatal to the pass: one reservation that
 			// settled underneath us must not strand the rest of the batch.
 			result.Failed++
+			// A divergence is different in kind from a lost race. The ledger
+			// already holds a release for a different amount, so no pass can
+			// ever clear this hold and an operator has to reconcile it; log it
+			// at error so alerting separates it from the ordinary refusals this
+			// loop expects to see.
+			var divergence *SettlementDivergenceError
+			if errors.As(err, &divergence) {
+				r.logger.ErrorContext(ctx, "accounting: reaper found a hold whose ledger release disagrees with the row, manual reconciliation required",
+					"reservation_id", reservation.ID.String(),
+					"account_id", reservation.AccountID.String(),
+					"posted_release_credits", divergence.PostedCredits,
+					"held_credits", divergence.WantCredits)
+				continue
+			}
 			r.logger.WarnContext(ctx, "accounting: reaper could not release a stranded hold",
 				"reservation_id", reservation.ID.String(),
 				"account_id", reservation.AccountID.String(),
--- a/apps/control-plane/internal/accounting/reservation_terminal_state_test.go
+++ b/apps/control-plane/internal/accounting/reservation_terminal_state_test.go
@@ -146,8 +146,14 @@ func TestReleaseAfterUnconfirmedSettlementKeepsRowBalanced(t *testing.T) {
 	if len(ledgerSvc.chargeCalls) != 1 || ledgerSvc.chargeCalls[0].credits != 1003 {
 		t.Fatalf("expected one 1003-credit charge, got %#v", ledgerSvc.chargeCalls)
 	}
-	if len(ledgerSvc.releaseCalls) != 1 || ledgerSvc.releaseCalls[0].credits != 8997 {
-		t.Fatalf("expected only the finalize's own 8997-credit release, got %#v", ledgerSvc.releaseCalls)
+	// One release, for the WHOLE 10000 hold rather than the 8997 unused
+	// remainder: capture lifts the authorization in full, and the 1003 charged
+	// is posted separately (issue #616). The row still records 8997 as its
+	// released_credits, asserted by the balance check above, because that
+	// column answers how the hold was disposed of rather than whether any
+	// authorization is still outstanding.
+	if len(ledgerSvc.releaseCalls) != 1 || ledgerSvc.releaseCalls[0].credits != 10000 {
+		t.Fatalf("expected only the finalize's own 10000-credit hold lift, got %#v", ledgerSvc.releaseCalls)
 	}
 	if repo.releaseEventCounts[reservation.ID] != 0 {
 		t.Fatalf("expected no release event row, got %d", repo.releaseEventCounts[reservation.ID])
--- a/apps/control-plane/internal/accounting/service.go
+++ b/apps/control-plane/internal/accounting/service.go
@@ -293,29 +293,64 @@ func (s *Service) finalizeLocked(ctx context.Context, input FinalizeReservationI
 		actualCredits = heldCredits
 	}
 
-	releaseCredits := releasableCredits(reservation, actualCredits)
+	// unusedCredits is a fact about the RESERVATION: how much of the hold never
+	// turned into spend. It is what the row records, beside consumed_credits,
+	// and the two still sum to reserved_credits.
+	unusedCredits := releasableCredits(reservation, actualCredits)
 
 	if actualCredits > 0 {
-		if _, err := s.ledgerSvc.ChargeUsage(ctx, reservation.AccountID, reservation.RequestID, &reservation.RequestAttemptID, &reservation.ID, s.idempotencyKey(reservation.ID, fmt.Sprintf("charge-%d", actualCredits)), actualCredits, map[string]any{
+		// Idempotency key is "charge", not "charge-<credits>", for the same
+		// reason the release key is flat (issue #652): a key that varies with
+		// the amount cannot deduplicate two settlement attempts that DISAGREE
+		// on the amount, which is exactly the case worth catching. That retry
+		// is reachable, not hypothetical -- the charge posts before the
+		// reservation row is updated, so a failure in between leaves an open
+		// row for a retry to re-enter with a freshly computed number.
+		entry, err := s.ledgerSvc.ChargeUsage(ctx, reservation.AccountID, reservation.RequestID, &reservation.RequestAttemptID, &reservation.ID, s.idempotencyKey(reservation.ID, "charge"), actualCredits, map[string]any{
 			"endpoint":           reservation.Endpoint,
 			"model_alias":        reservation.ModelAlias,
 			"terminal_confirmed": input.TerminalUsageConfirmed,
-		}); err != nil {
+		})
+		if err != nil {
 			return Reservation{}, fmt.Errorf("accounting: charge usage: %w", err)
+		}
+		// The dedup above returns the FIRST entry, so a second attempt for a
+		// different amount would otherwise leave the ledger and the reservation
+		// row disagreeing about what was billed, silently. An unrecognized
+		// settlement fact is fatal on the money path (D-034): fail here and
+		// leave the hold open for the reaper rather than record a divergence.
+		if entry.CreditsDelta != -actualCredits {
+			return Reservation{}, fmt.Errorf("accounting: charge already posted for reservation %s at %d credits, refusing to settle at %d", reservation.ID, -entry.CreditsDelta, actualCredits)
 		}
 	}
 
-	if releaseCredits > 0 {
-		// Idempotency key is "release", not "release-<credits>" (issue #652):
-		// releaseLocked below already keys its full release the same way, and a
-		// key that varies with the amount cannot deduplicate two release
-		// attempts of DIFFERING amounts for the same reservation, which is
-		// exactly the case worth catching. One reservation, one release key,
-		// regardless of which path or how much it releases.
-		if _, err := s.ledgerSvc.ReleaseReservedCredits(ctx, reservation.AccountID, reservation.RequestID, &reservation.RequestAttemptID, &reservation.ID, s.idempotencyKey(reservation.ID, "release"), releaseCredits, map[string]any{
+	// The ledger releases the WHOLE outstanding hold, including the part the
+	// charge just captured, not merely the unused remainder. A hold is an
+	// authorization, and capture lifts it in full; the charge is what the
+	// customer actually pays. Releasing only the remainder left `actual`
+	// credits sitting in the ledger's reserved bucket forever, and since
+	// GetBalance computes available as posted minus reserved, every settled
+	// request cost the customer its price TWICE: once posted, once withheld
+	// (issue #616, measured live 2026-08-16 -- 143642 credits held against
+	// 1825 reservations that had already settled, one 18-credit completion
+	// moving available by 36). The reaper cannot recover these, because it
+	// only scans holds still in a non-terminal state.
+	//
+	// This is also what the API key counter has always done a few lines below:
+	// ApplyReservationDelta unwinds -heldCredits, never -unusedCredits.
+	//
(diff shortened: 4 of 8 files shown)


sakibsadmanshajib:

Streams run against `0c021a60` (the first revision of this branch), and their disposition. Fixes are in `eba7c19e`.

| stream | status |
| --- | --- |
| CodeRabbit CLI (`coderabbit review --committed --base main`) | ran, 1 finding |
| Independent adversarial pass, separate model instance, read-only, money-path brief | ran, 3 blockers plus 5 lesser findings |
| `/codex:adversarial-review` | **SKIPPED**, the CLI returned a usage-limit error and ran no review. Not a pass, and not treated as one |
| `ecc:code-review` skill | **SKIPPED**, structural capability gap: this agent has no Skill tool, so the skill could not be invoked. Not a pass |
| Domain specialist by diff shape (database) | covered by the reviewing agent's own role; the ledger schema, the `GetBalance` query plan and the idempotency indexes were read directly, and the live population was measured read-only through the pooler |

### Findings and what changed

**1. Flat charge key can double bill across a deploy. Accepted, change reverted.**
`(account_id, entry_type, idempotency_key)` is the uniqueness tuple, so an old `reservation:R:charge-115` row and a new `reservation:R:charge` row are two different rows. A reservation charged by pre-deploy code that then failed before its row update stays open, and a post-deploy retry inserts a SECOND charge instead of deduplicating. The repo already learned this once: flattening the release key needed migrations `20260801_11` and `20260801_12`. This PR shipped neither, so the key change is reverted and filed separately. Live check before deciding: 1,866 charge entries, all amount-bearing, zero flat, zero reservation carrying more than one charge, and zero open reservations currently carrying a charge, so nothing is standing in that window today; the exposure is a rolling deploy crossed by a retry.

**2. The live magnitude test never ran. Accepted, fixed.**
`./internal/accounting/...` was absent from the live-Postgres step in `.github/workflows/ci.yml`, so this package's `HIVE_TEST_DB_URL`-gated tests skipped everywhere, including the existing release-key dedup proof from #652. The only real-substrate proof on the money path was decoration. Package added to that step, with a comment naming the trap (#701, #708).

**3. The release path had no divergence guard. Accepted, fixed.**
`PostEntry` returns the first entry for a repeated key, so a release following an earlier SMALLER release returned quietly with the hold only partly lifted; settlement then marked the row terminal and the remainder was stranded past the reaper, which only scans non-terminal holds. That is #616 reintroduced silently, on the rows this branch exists to fix. `assertReleasedInFull` now fails the settlement instead, leaving the hold open and visible. Pinned by `TestSettlementRefusesToSettleAgainstAPartialRelease`, mutation-verified: remove the guard and the test goes red.

**4. `_ = err` in the new test (CodeRabbit, and the second stream independently). Accepted.**
The assertion could not fail: the flat key alone kept the charge count at one. The test that replaced it asserts the error, asserts the row stays open, and asserts the balance, and was mutation-checked against its own guard.

**5. Batch retry would hard-fail forever under the charge guard. Moot.**
`batchstore/worker.go` recomputes the amount on every poll, so a differing amount would have tripped the guard on each attempt. Removed along with the charge-key change.

**6. Comment overstated the row invariant. Accepted, corrected.**
`consumed + released = reserved` is false on a confirmed overage, where consumed exceeds reserved and `remainingHeldCredits` relies on its clamp.

**7. Fake ledger is signed, production takes ABS. Accepted, comment corrected.**
They agree for every case in that test since none over-releases, and the live test covers the real query. Related, and out of scope here: that `ABS` turns an over-release into a phantom hold that would be subtracted from a customer's available balance. Ten historical reservations, 25,000 credits, are in that state from a different cause; reported in the PR body and filed separately.

**8. One assertion is implied by two others. Acknowledged, kept.**
Posted, reserved and moved are algebraically linked, so the third is a restatement. It is kept because available moving by exactly the charge is the customer-facing sentence this PR is about, and a future change to the balance formula should break it by name.

### Not found, stated explicitly rather than left silent

- No path can now release more than was held, or release twice: one release key in the code, `20260801_12` normalized the history and hard-fails on leftovers, both settlement paths compute the same `remainingHeldCredits`, the per-account advisory lock serializes finalize against release, and the reaper goes through the service rather than the ledger.
- Audio and image adapters: finalize once with a fixed amount, full release only on error, no recomputed retry.
- No path lets a request be served without a successful accounting step; nothing in this diff touches the serve decision.

sakibsadmanshajib:
## Adversarial review, second round

CodeRabbit CLI re-run against `eba7c19e` (the review-response commit). Two findings, both on `assertReleasedInFull`. Fixes in `87b66a57`.

**Critical, sign convention of `CreditsDelta` for a release. Rebutted, with evidence.**

The comparison is correct as written. `ledger.ReleaseReservedCredits` posts `+credits` while `ReserveCredits` posts `-credits` (`apps/control-plane/internal/ledger/service.go:89-95`, both through `postUsageEntry`), which is exactly what makes the two cancel inside `GetBalance`'s `ABS(SUM(hold) + SUM(release))`. Releases are stored positive, so `entry.CreditsDelta == want` against a positive hold amount is the right test, and taking magnitudes would be a no-op at best.

This is not an argument from reading alone: `TestSettlementLedgerMagnitude_Live` runs the real repositories against real Postgres built from `supabase/migrations/`, and the finding's own failure mode ("every settlement fails while the fake ledger in the unit tests passes") would have made it red. It is green. The reasoning is now recorded in the function's comment so the next reader does not have to re-derive it.

**Major, a failed assertion leaves a charged, permanently stuck reservation. Accepted in part.**

Correct, and deliberate: the charge is committed, both settlement paths recompute the same amount and deduplicate against the same smaller entry, so no pass can clear it. The alternatives are writing a number the ledger contradicts or silently dropping credits, and both are worse on a money path. But the finding is right that a warning inside a loop which expects occasional refusals is not an operator signal.

Taken: the error is now a distinct type, `SettlementDivergenceError`, carrying the posted and wanted amounts, and the reaper logs that case at error level with both numbers rather than folding it into the ordinary "could not release a stranded hold" warning. The type is asserted by the test, so it cannot decay into decoration.

Not taken: writing a `credit_reconciliation_jobs` row for it. Nothing drains that table today (issue #600), and the reaper would insert a fresh row every fifteen minutes for the same stuck reservation, which converts one loud line into unbounded rows. The reservation stays open and therefore stays in the reaper's own candidate scan, so the state is queryable at any time, and the ledger rows for it disagree in a way a reconciliation query can find. If a real operator queue lands later, this is a two-line change to feed it.

PR #913, opened 2026-08-16
perf: measure where a metered inference request spends its time

sakibsadmanshajib:
## What this is

Instrumentation for the metered inference path, plus the measured answer to "where do the 19 to 23 seconds per gateway call go".

Read the honesty section first: **this PR does not make anything faster.** It makes the cost visible and names the one cut that will, along with the reason that cut is not in this diff.

## Measured, on the demo box, 2026-08-16

Same prompt, same box, same minute:

| path | wall clock |
|---|---|
| straight through LiteLLM, no gateway, no accounting | 0.54 - 1.38 s |
| through the Hive gateway | 12.4 - 24.4 s |

Per-stage, timed at every boundary:

| stage | measured |
|---|---|
| network and Caddy (unauthenticated 401 round trip) | 0.13 s |
| authorize (warm) | ~0.00 s, and 2.54 s on a cold cache |
| select_route | 0.71 s |
| start_attempt | 0.24 s |
| **create_reservation** | **~5 - 7 s** |
| dispatch (the provider actually working) | 0.6 - 1.4 s |
| **finalize_reservation** | **~5 - 6 s** |
| record_usage | 0.24 s |

Three hypotheses were named up front. Two are dead, measured rather than argued:

- **Settlement sitting on the critical path**: not the cause. `/internal/usage/events` answers in 0.24 s, measured directly against control-plane. Moving usage recording off the critical path would buy a quarter of a second.
- **The exhausted 15-client session pool (#436)**: not the cause of this. The pool is genuinely at its ceiling and that is a real problem, but one metered request executes 86 statements and all of them succeed; there is no starvation stall in the profile.
- **Round-trip count at 240 ms**: this is it. One metered chat completion runs **86 SQL statements** and consumes **216 ms of database server time**, against a database in `aws-1-us-east-1` whose measured round-trip latency from the box is **236 - 276 ms**. 86 statements times 240 ms is very nearly the observed wall clock. The database does 1.2% of the work and the speed of light does the rest.

Method: `pg_stat_statements` snapshotted per `queryid` before and after exactly one request, diffed. Provider baseline taken through LiteLLM on loopback (2 - 3 ms to reach it). Per-endpoint numbers taken by calling control-plane's internal endpoints directly.

`CreateReservation` costs what it costs because its locked critical section runs five independent transactions (balance read, attempt insert, reservation insert, ledger reserve, usage event, api-key delta), each paying its own BEGIN, its statements, and its COMMIT, plus a session advisory lock on a dedicated connection. That is roughly 21 to 24 sequential round trips. `FinalizeReservation` has the same shape.

## The invariant this preserves

Nothing on the money path moves. The reservation is still created before dispatch, the charge is still finalized before the response is written, both still run on the same contexts with the same error handling, and a reservation still reaches a terminal state exactly once: charged on finalize success, released by the existing deferred release on finalize failure, never both. This diff only reads a clock around calls that already existed, in the same order they already ran.

Concretely, on the unhappy path, the behaviour after this change is identical to before it, because settlement did **not** move asynchronous:

- Request succeeds, finalize succeeds: customer is charged the settled amount. Unchanged.
- Request succeeds, finalize fails: `finalized` stays false, the deferred release hands the hold back in full, the failure is logged with `request_id` and `reservation_id`, and the response is still returned. The customer is not charged for that call, and the reconciliation path that already exists is what picks it up. Unchanged, and deliberately so.
- Instrumentation itself fails: it cannot. `stage` is nil-safe and the observation is a histogram write with no error return, so it has no failure mode that can reach a request.

Making settlement faster by making it lossy would be a far worse bug than the latency, and this diff declines to trade that way. Hive has already served three days of unmetered traffic because a billing failure was silent; the fix for the latency is fewer round trips, not fewer guarantees.

## The cut, and why it is not here

The cut that removes the time is collapsing each money call's locked critical section into a single transaction on a single connection, which removes the redundant BEGIN and COMMIT pairs and is strictly more correct than today (a crash between the reservation insert and the ledger write currently leaves a reservation row with no ledger entry, which is the stranded-hold shape the repo already suffers from).

That refactor touches `apps/control-plane/internal/accounting/service.go` and `repository.go`, which is exactly where the in-flight reservation-release fix is working. Landing both at once would collide on the money path, which is the worst possible place to resolve a merge conflict. It should land after that fix, on top of it, with this metric already deployed so the improvement is measurable rather than asserted.

The second lever is not code at all: at 240 ms per round trip, even a four-fold reduction in statements leaves a five-second call. Only moving the database next to the services gets a metered call under a second, which is the self-hosted-Postgres direction the owner already has a standing directive for.

## Files touched

- `apps/edge-api/internal/inference/stage_metrics.go` (new)
- `apps/edge-api/internal/inference/stage_metrics_test.go` (new)
- `apps/edge-api/internal/inference/orchestrator.go` (timing calls only, no reordering, no semantic change)
- `apps/edge-api/cmd/server/main.go` (one construction site)

Deliberately **not** touched, to stay clear of the in-flight reservation-release work: `apps/control-plane/internal/accounting/*`, `releaseReservationBackground`, and every release path.

Also not touched: the streaming path (`stream.go`). The agent traffic this was investigated for is non-streaming, and adding stage timing there is a separate, larger change because settlement happens after the stream closes.

## The 190 ms instrument, kept

The sharpest single piece of evidence in the investigation was that the client's response body arrived 190 ms after the last ledger row was written, three times running. That is what proved the wait was Hive's own accounting rather than the agent or the provider. It now lives as the `record_usage` and `response_write` stages rather than as a one-off probe, so a regression in that gap is visible in one PromQL query instead of requiring the investigation to be repeated.

Suggested query:

```
histogram_quantile(0.95, sum by (stage, le) (rate(hive_metered_stage_duration_seconds_bucket[5m])))
```

## Before and after

Latency before and after are the same by construction, and that is the intended result of this PR:

| | before | after |
|---|---|---|
| metered chat completion, demo box | 12.4 - 24.4 s | 12.4 - 24.4 s |
| stages visible from `/metrics` | none | nine |

The numbers that change come from the follow-up, and this metric is what will measure them honestly.

## Tests

`go test ./apps/edge-api/... -count=1 -short` green in the toolchain container. Three new tests: one observation per `stage` call landing on its own series, the bucket range actually covering a 24 second observation (a `DefBuckets` histogram would silently report nothing about the values that matter here), and nil-safety for the many construction sites that pass no metrics.

## Buglog entry

```json
{"date":"2026-08-16","title":"metered gateway call costs 19 to 23 seconds of Hive-owned overhead per request","error_message":"none: no error is raised, the request succeeds and is simply 20x slower than the provider call it wraps","root_cause":"One metered chat completion executes 86 sequential SQL statements against a database in aws-1-us-east-1 whose round-trip latency from the box is 236 to 276 ms, while spending only 216 ms of actual database server time. CreateReservation and FinalizeReservation dominate at 5 to 7 s each because each runs five independent transactions inside its account advisory lock, so every logical operation pays its own BEGIN, statements and COMMIT. Neither settlement placement nor pool exhaustion is the cause; both were measured and excluded.","fix":"Not fixed here. This change instruments the path so the split is a standing metric (hive_metered_stage_duration_seconds). The cut is collapsing each money call's locked critical section into one transaction, sequenced after the in-flight reservation-release fix to avoid colliding on the same files, plus co-locating the database, which is the only lever that reaches sub-second.","tags":["edge-api","control-plane","accounting","latency","metrics","demo-box","round-trips"]}
```

### DIFF
--- a/apps/edge-api/cmd/server/main.go
+++ b/apps/edge-api/cmd/server/main.go
@@ -146,7 +146,8 @@ func main() {
 	routingClient := inference.NewRoutingClient(resolveControlPlaneBaseURL())
 	accountingClient := inference.NewAccountingClient(resolveControlPlaneBaseURL())
 	litellmClient := inference.NewLiteLLMClient(resolveLiteLLMBaseURL(), resolveLiteLLMMasterKey())
-	orchestrator := inference.NewOrchestrator(authorizer, routingClient, accountingClient, litellmClient)
+	orchestrator := inference.NewOrchestrator(authorizer, routingClient, accountingClient, litellmClient).
+		WithStageMetrics(inference.NewStageMetrics(promRegistry))
 	inferenceHandler := inference.NewHandler(orchestrator)
 	chatDispatchHandler := chat.NewDispatch(chat.Deps{
 		Pool:    dbPool,
--- a/apps/edge-api/internal/inference/orchestrator.go
+++ b/apps/edge-api/internal/inference/orchestrator.go
@@ -20,6 +20,18 @@ type Orchestrator struct {
 	routing    *RoutingClient
 	accounting *AccountingClient
 	litellm    *LiteLLMClient
+
+	// metrics is optional; a nil value records nothing. See StageMetrics.
+	metrics *StageMetrics
+}
+
+// WithStageMetrics attaches per-stage timing to this Orchestrator and returns
+// it, mirroring accounting.Service.WithAccountLocker rather than widening
+// NewOrchestrator's signature, so the many call sites that do not want metrics
+// (every test) are untouched.
+func (o *Orchestrator) WithStageMetrics(m *StageMetrics) *Orchestrator {
+	o.metrics = m
+	return o
 }
 
 // NewOrchestrator creates a new Orchestrator.
@@ -79,9 +91,13 @@ func (o *Orchestrator) executeSync(
 	dispatch dispatchFunc,
 	normalize normalizeFunc,
 ) {
+	endTotal := o.stage(endpoint, StageTotal)
+
 	// 1. Authorize
 	authHeader := r.Header.Get("Authorization")
+	endAuthorize := o.stage(endpoint, StageAuthorize)
 	snapshot, headers, authErr := o.authorizer.Authorize(ctx, authHeader, model, estimatedCredits, 0, 0)
+	endAuthorize()
 	if authErr != nil {
 		// apierrors.WriteAuthFailure is the single source of truth for
 		// mapping an authz failure to a wire response (rate-limit 429,
@@ -96,6 +112,7 @@ func (o *Orchestrator) executeSync(
 	}
 
 	// 2. Select route
+	endSelectRoute := o.stage(endpoint, StageSelectRoute)
 	route, err := o.selectRoute(ctx, snapshot, SelectRouteInput{
 		AliasID:             model,
 		NeedChatCompletions: needFlags.NeedChatCompletions,
@@ -105,6 +122,7 @@ func (o *Orchestrator) executeSync(
 		NeedReasoning:       needFlags.NeedReasoning,
 		RequireToolCapable:  needFlags.RequireToolCapable,
 	})
+	endSelectRoute()
 	if err != nil {
 		if errors.Is(err, ErrAccountNotProvisioned) {
 			writeAccountNotProvisionedError(w)
@@ -139,6 +157,7 @@ func (o *Orchestrator) executeSync(
 
 	// 3. Start attempt
 	requestID := uuid.New().String()
+	endStartAttempt := o.stage(endpoint, StageStartAttempt)
 	attempt, err := o.accounting.StartAttempt(ctx, StartAttemptInput{
 		AccountID:     snapshot.AccountID,
 		RequestID:     requestID,
@@ -148,11 +167,13 @@ func (o *Orchestrator) executeSync(
 		Status:        "dispatching",
 		APIKeyID:      snapshot.KeyID,
 	})
+	endStartAttempt()
 	if err != nil {
 		log.Printf("inference: start attempt failed (non-fatal): %v", err)
 	}
 
 	// 4. Create reservation
+	endCreateReservation := o.stage(endpoint, StageCreateReservation)
 	reservation, err := o.accounting.CreateReservation(ctx, CreateReservationInput{
 		AccountID:        snapshot.AccountID,
 		RequestID:        requestID,
@@ -163,6 +184,7 @@ func (o *Orchestrator) executeSync(
 		EstimatedCredits: estimatedCredits,
 		PolicyMode:       "strict",
 	})
+	endCreateReservation()
 	if err != nil && refuseOnReservationFailure(w, endpoint, model, err) {
 		return
 	}
@@ -184,7 +206,9 @@ func (o *Orchestrator) executeSync(
 	}()
 
 	// 5. Dispatch to LiteLLM with bounded retry on 429/5xx.
+	endDispatch := o.stage(endpoint, StageDispatch)
 	resp, err := dispatchWithRetry(ctx, route.LiteLLMModelName, body, dispatch)
+	endDispatch()
 	if err != nil {
 		if reservation.ID != "" {
 			releaseReason = "upstream_error"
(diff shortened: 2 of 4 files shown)


sakibsadmanshajib:

Streams run against `505658d9`. This is a money-path-adjacent diff, so the security and database streams are mandatory and were not skipped.

### CodeRabbit CLI — RAN, 1 major finding, fixed

> **major [Functional Correctness]** `orchestrator.go:94` — Record `StageTotal` for every request outcome. Line 355 is only reached after a successful response write. Every earlier error return omits the total observation.

Accepted, it was a real defect and the worst kind: a metric that can only ever record successes, so it cannot go red on the outcomes worth alerting on. Fixed in `505658d9` by deferring the total immediately after it is created and deleting the explicit call. The doc comment that said this helper is deliberately not defer-based was also corrected, since that reasoning applies to the mid-function stages and not to the one stage that genuinely spans the whole function. `TestTotalIsRecordedOnAFailedRequest` drives the earliest exit (authorization failure through `upstreamUnavailableOrchestrator`) and asserts both the total and the authorize stage still record.

### Security stream — RAN, no findings

- **No secret or customer identifier reaches a label.** The two labels are `stage` (a fixed nine-value constant set) and `endpoint` (an internal constant such as `chat_completions`). Account ID, tenant ID, API key ID, request ID and model alias are all deliberately absent. That matters twice over: a metric label is world-readable to anyone who can reach `/metrics`, and an account-scoped label would also be an unbounded cardinality vector.
- **No new surface.** `/metrics` already exists and is already served on its own mux; this registers one more collector on the registry that endpoint already exposes.
- **No authorization decision is read, moved or influenced.** The instrument wraps `Authorize`, it does not inspect its result.
- **No error path is swallowed or altered.** Every `if err != nil` in `executeSync` is byte-identical to before; the timing calls sit between statements, never inside a condition.
- **Instrumentation cannot fail a request.** `stage` is nil-safe on the receiver and on `StageMetrics`, and a histogram observation returns no error. The only failure mode in the new code is `MustRegister` panicking on a duplicate registration, which can only happen at construction, once, at boot, and is loud rather than silent.

### Database stream — RAN, no findings

- No migration, no schema change, no new query, no change to an existing query, no change to transaction boundaries or to the account advisory lock.
- Round-trip count per request is unchanged. The instrument adds zero database work; it reads a monotonic clock.
- The diff deliberately does not touch `apps/control-plane/internal/accounting/*`, which is where the reservation-release fix is in flight and where the follow-up round-trip reduction will land.

### Money-path invariant check — RAN, no findings

Walked the settlement order in the diff line by line against `origin/main`:

- Reservation is still created before dispatch, so the request is still authorized before a provider is reached.
- Finalize still runs before the response is written, on `finalizeCtx` (a fresh background context with `accountingTimeout`), so a client disconnect still cannot convert delivered work into a free request.
- `finalized` is still set only on finalize success, so the deferred release still hands the hold back on failure and a reservation still reaches a terminal state exactly once.
- `TerminalUsageConfirmed` still carries `settlementCredits`' own verdict, never a bare true.
- The existing settlement suite (`TestExecuteSync_FinalizeError_ReleasesReservationAndLogs`, `_CancelledContext_StillReleases`, `_ClientDisconnect_ChargesDeliveredWork`, `_NormalCompletion_TerminatesExactlyOnce`, `_NoUsage_NoOutput_ReleasesInsteadOfCharging`, `_ConfirmedUsage_StaysConfirmed`) passes unchanged, which is the assertion that matters more than my reading of the diff.

### Plain adversarial pass — RAN, two notes, neither blocking

1. **`WithStageMetrics` mutates the Orchestrator after construction**, so calling it concurrently with serving would race. It is called exactly once in `main.go`, before the server starts. The alternative (widening `NewOrchestrator`) would churn every existing construction site including tests, which is a worse trade for a value set once at boot. Noted rather than changed, and it mirrors `accounting.Service.WithAccountLocker`, the existing precedent in this repo.
2. **The streaming path is not instrumented.** Stated in the PR body rather than left to be discovered: the agent traffic this was investigated for is non-streaming, and streaming settles after the stream closes, which is a different and larger change. A reader of the dashboard should know the histogram covers the sync path only.

### Not run

No frontend, no UI surface, no schema, so no visual proof and no migration review apply. `apps/agent-console` is untouched by deliberate instruction.

PR #914, opened 2026-08-16
fix: stop one lost agent session from throttling every tenant's poll cadence

sakibsadmanshajib:
## Summary

`agenttask.Poller.RunOnce` reported the whole sweep as failed whenever ANY active task's status check errored, so a single permanently unreachable session (a lost engine session ref after a launcher restart) kept `consecutiveFailures` climbing forever and pinned the shared loop at `maxPollerBackoff` (5 minutes) instead of the configured 15 second interval. Since `Repository.ListActive` is cross-tenant by design, that one stuck row degraded poll cadence for every other tenant's tasks too.

Measured live 2026-08-16: task `f9409763` had been stuck in `running` since 2026-08-14 with a 404ing `/status` call. Control-plane logs showed sweeps five minutes apart for hours; two unrelated completed tasks had `finished_at` timestamps 1203.77s apart, exactly 4 x 300.94s.

This PR went through three review passes before landing on the design below. Each pass's findings and how they were addressed are kept in full further down, since they document real defects two earlier "fixed" versions actually shipped.

### Root cause, confirmed by reading the code (not assumed)

- The backoff is global to the poller, not per-task: `loop()` holds one `consecutiveFailures` counter for the whole pass, and `RunOnce` used to return a non-nil error whenever `errCount > 0`, regardless of how many other tasks succeeded in the same sweep.
- `consecutiveFailures` never reset because the same task 404'd on every single pass forever.
- A `/status` 404 is a definitive answer, not a transient one: the launcher's `/status` handler (`apps/agent-engine/cmd/agent-engine/serve.go`) maps `engineapi.ErrUnknownSession` onto exactly that status code when its in-memory session registry (lost on a launcher restart) has no memory of the session.
- A second, separate dead-task shape is also reachable today: `SandboxEngine.reap` only runs on a terminal status or `Cancel`, so a sandbox killed some other way (OOM, an Apptainer crash, or its agent-server control socket dying while the sandbox itself stays up) never gets reaped. Its `Status` call fails forever too, but with a 502, not a 404 -- same incident class as `f9409763`, a different, non-enumerable error shape.

### Current design (third pass, final)

1. `RunOnce`'s returned error (loop's sole backoff signal) means only that `ListActive` itself failed -- a real database/connectivity problem, the only thing that is actually poller-wide. No task-level problem, of any shape or count, ever feeds it.
2. Every task-level problem is scoped to that ONE task via `pollTask`/`chargeFailureBudget`:
   - A `StatusChecker` error definitively identified as `agenttask.ErrEngineSessionGone` (the engine's "I have no memory of this session" answer -- mapped from `agentengine.Remote`'s 404 for the socket arm, and the in-process `agentengine.Engine` arm's `engineapi.ErrUnknownSession`, for consistency across both arms) fails the task immediately. Nothing to stop on this path: the engine already said the session is gone.
   - Anything else is retried per pass, up to `Poller.taskFailureBudget()` consecutive times (a duration, `maxTaskFailureDuration` = 5 minutes, converted to a pass count using the poller's OWN configured interval, so tuning `HIVE_AGENT_TASK_POLL_INTERVAL` cannot silently shrink the real kill timeout), after which `RunOnce` gives up on that task and fails it directly regardless of the error's shape. Before failing the row, it best-effort stops the session (`budgetExceededCanceler`, type-asserted against the checker, since `buildAgentEngine` wires the same object as both `agenttask.Engine` and `agenttask.StatusChecker`): a non-404 error means the session is still live and holding its concurrency slot, and once the row is terminal nothing else (not even a user's own `Cancel`) can ever free that slot otherwise.
   - A `Repository.Transition` failure for an ALREADY-KNOWN terminal status (the engine told us the task succeeded or failed, but persisting that failed) is retried forever exactly as before this PR and is never charged against the budget or forced to a status we can't confirm.
3. `taskFailures` (the per-task consecutive-failure map) is guarded by its own mutex: `RunOnce` is exported and callable concurrently with itself in principle, even though nothing in production does that today.

The `agent_tasks` terminal guard clause (`status NOT IN ('succeeded', 'failed', 'cancelled')`) that `Cancel` and the poller both rely on to resolve races is unchanged: every terminal path here reuses the exact same `Repository.Transition` call, so a concurrent `Cancel` still wins the atomic race exactly as before (`ErrTerminalState` is swallowed, not retried, not surfaced as a pass error).

**Pre-existing race, worth naming explicitly (not introduced or changed by this PR):** when the poller wins a race against a user's `Cancel` call, the user's `Cancel` gets back `ErrTerminalState` and the row reads `failed` (or `succeeded`) rather than `cancelled`. Correct and atomic, just surprising to whoever clicked cancel.

**Known, deliberately deferred residual:** the row is persisted as `status = failed` regardless of whether it was a confirmed failure or an unrecoverable-outcome case (`ErrEngineSessionGone` or a budget exhaustion). The `error_message` text is honest about the distinction; the machine-readable `status` column is not. Telling the two apart programmatically needs a distinct terminal status or a reason code, folded into issue **#921** alongside the deploy-drain question rather than built here.

### Stranded row f9409763

Per owner authorization (gated on the fix landing and on showing the exact statement plus affected row count before running), transitioned `f9409763-02b4-41bd-870b-7a91d04b6154` on the demo database directly:

- Looked it up by id prefix first: exactly one row matched, `status=running`, `engine_session_ref` set, stuck since `2026-08-14T17:17:16Z`.
- Ran, scoped by primary key (mirrors `Repository.Transition`'s own atomic guard):

```sql
UPDATE public.agent_tasks
   SET status = 'failed',
       error_message = 'agent task session is no longer reachable',
       finished_at = COALESCE(finished_at, now()),
       updated_at = now()
 WHERE id = 'f9409763-02b4-41bd-870b-7a91d04b6154'
   AND status NOT IN ('succeeded', 'failed', 'cancelled')
```

- Affected exactly 1 row. Deleting was never on the table; this is a transition.

## Review pass 1 (stage 6) findings and fixes

Single reviewer stream ran (CodeRabbit rate limited, Codex out of quota, Bugbot disabled, all three declared SKIPPED).

- **HIGH, `poller.go:173` (first version): `errCount == len(tasks)` collapses to a single task's own error when exactly one task is active** -- the ordinary shape under this deployment's quota (2/user, 4/tenant). Fixed: the fraction rule removed entirely; `RunOnce`'s error means only "`ListActive` failed." Test: `TestPoller_RunOnce_SingleTaskErroringForeverDoesNotBackOffTheSweep`.
- **Second dead-task shape (502, not 404) that `ErrEngineSessionGone` alone didn't cover.** Fixed via the per-task failure budget: any task-level error, of any shape, eventually fails the task once the budget is exhausted. Test: `TestPoller_RunOnce_TaskExceedingFailureBudgetFailsRegardlessOfErrorShape`.
- **Operational consequence: `deploy-demo-box.yml`'s unconditional installer restart terminally fails every in-flight Cowork task within ~15s of a merge to `main`.** Failing fast beats hanging forever silently, not reverted. Named in the `ErrEngineSessionGone` doc comment; drain/reconnect mitigation tracked as **#921**.
- **Test discrimination gap**: the original all-erroring test passed identically against the unfixed rule, and there was no test for the single-task collapse. Replaced/added the tests named above plus `TestPoller_RunOnce_AllTasksErroringNeverBacksOffEither`, `TestPoller_RunOnce_FailureBudgetResetsOnCleanPass`.
- **404-as-terminal reasoning survived every attack** (transport errors never become 404-shaped, the registry is memory-only with no restore path, no double-launcher window). Kept, supplemented by the budget rather than replaced.

## Review pass 2 findings and fixes (commit `caf40886`)

- **Doc comment claimed more than the code could know** (a biased-sample argument against the now-removed fraction rule): moot once that rule was removed.
- **Deploy-restart consequence named explicitly in the doc comment**, plus filed **#921**.
- **Message honesty**: `engineSessionGoneMessage` / `engineUnreachableAfterRetriesMessage` no longer assert failure; both read as an unrecoverable outcome, since the sandbox may have finished before the launcher lost the reference.
- **Pass-level aggregate log**: `RunOnce` now counts how many tasks it personally declared dead this pass and logs one summary line when nonzero, instead of N unrelated per-task lines with no total.
- **404-to-`ErrEngineSessionGone` mapping scoped to exactly `/status` and `/cancel`** (not `/launch`, which `post` also serves): `net/http.ServeMux` answers 404 for any unmatched path, so an unscoped check would call a routing miss session-gone. New negative tests: `TestRemoteLaunch404DoesNotMapToEngineSessionGone`, `TestRemoteTransportFailureDoesNotMapToEngineSessionGone`.
- **Cancel-path honesty**: `Service.stopEngineSession` no longer warns about a leaked slot when the engine's answer is `ErrEngineSessionGone` (a launcher with no memory of a session holds no slot for it either); `agentengine.Engine.Cancel` now maps `ErrUnknownSession` the same way `Engine.Status` does, for arm consistency. Test: `TestService_Cancel_EngineSessionGoneDoesNotWarnAboutALeakedSlot`.
- **Two LOW nits** in `engine.go`: dropped a redundant `err != nil` guard, switched to `fmt.Errorf`'s two-`%w` form so wrapping `ErrEngineSessionGone` doesn't silently drop `errors.Is(result, engineapi.ErrUnknownSession)`.

## Review pass 3 findings and fixes (commits `574200ee`, `70975837`)

- **HIGH: budget exhaustion failed the row but never stopped the session.** A non-404 error means the launcher answered from its own handler, so `SandboxEngine.reap` never ran and the session was still holding a concurrency slot; once the row went terminal, `Service.Cancel`'s own atomic guard returned `ErrTerminalState` before it ever reached `stopEngineSession`, so nothing could ever free that slot -- issue #886's leaked-slot symptom through a new door, and at 2 slots per user, two such events would lock a user out of agent tasks entirely. Fixed with a best-effort `Cancel` on the budget-exhaustion path only (never on the session-gone path, where there is nothing left to stop): new unexported `budgetExceededCanceler` interface, type-asserted against `p.checker`. Tests: the budget test above now also asserts `Cancel` fires exactly once, only at exhaustion; the session-gone test now asserts `Cancel` never fires.
- **MEDIUM: the budget was counted in passes but documented in minutes, and the interval is operator-tunable via `HIVE_AGENT_TASK_POLL_INTERVAL`.** Lowering that interval is the obvious reaction to this very incident and would have silently shortened the real kill timeout in proportion. Fixed: `maxTaskFailureDuration` (5 minutes) is now converted to a pass count via `Poller.taskFailureBudget()`, using the poller's own configured interval, so the documented ceiling cannot drift. At the default 15s interval this is still exactly 20 (no existing test's numbers changed). Test: `TestPoller_RunOnce_FailureBudgetScalesWithConfiguredInterval` (a 1 minute interval yields a 5-pass budget).
- **MEDIUM: `RunOnce` is exported and was mutating an unsynchronized map**; the "no lock needed" argument was about `loop`'s own serialization, not a property of the exported method itself, and a concurrent map write is a process-fatal throw with no recover. Fixed: a dedicated `taskFailuresMu` guards every access via small helper methods. Not a live bug today (no shipped caller does this), a contract regression; cheap to close. Test: `TestPoller_RunOnce_ConcurrentCallsDoNotCorruptTaskFailures`.
- **Wording judged correct, two residuals noted, one now resolved as a side effect of the HIGH fix above** (the budget-path message no longer invites an unsafe retry, since the session is now actually stopped), the other (row still reads `status = failed` regardless of the honest message text) named explicitly in the doc comment and folded into #921.

## Buglog entry

```json
{"date":"2026-08-16","tags":["agenttask","poller","backoff","cross-tenant","agent-engine"],"error_message":"agenttask.Poller polled every tenant's active tasks at ~300s (maxPollerBackoff) instead of the configured 15s for hours, and a per-task failure budget's first two drafts left a slot-leak and an interval-coupling defect of their own","root_cause":"RunOnce returned a non-nil error, feeding loop's exponential backoff, whenever a task-level problem occurred; a first fix attempt narrowed this to errCount == len(tasks), which still collapses to a single task's own error whenever exactly one task is active, the ordinary shape under this deployment's quota. ErrEngineSessionGone alone did not cover every dead-session shape (a 502 from a session the launcher still accounts for, not a 404). A second fix attempt (a per-task failure budget) then failed a budget-exhausted row without ever stopping its still-live engine session, reproducing issue #886's leaked-slot symptom, and expressed the budget as a fixed pass count that would have silently shortened under a tuned poll interval.","fix":"RunOnce's returned error now means only that ListActive itself failed. Every task-level problem is scoped to a per-task consecutive-failure budget expressed as a duration (maxTaskFailureDuration, 5 minutes) and converted to a pass count via the poller's own configured interval: ErrEngineSessionGone still fails a task immediately (nothing to stop), any other error is retried until the budget is exhausted, best-effort stops the session via a type-asserted Cancel, then fails the task directly regardless of error shape, without ever feeding the shared backoff. The per-task failure map is mutex-guarded since RunOnce is exported."}
```

## Test plan

- [x] `apps/control-plane/internal/agenttask` unit suite (docker toolchain), all tests named above plus the full pre-existing suite
- [x] `apps/control-plane/internal/agentengine` unit suite
- [x] Full `apps/control-plane` short suite green
- [x] `gofmt -l` and `go vet` clean on both touched packages
- [ ] CI (this PR)

This is a backend-only fix to a background poller loop; no UI surface is touched, so the visual-proof-before-merge requirement does not apply here.

### DIFF
--- a/apps/control-plane/internal/agentengine/engine.go
+++ b/apps/control-plane/internal/agentengine/engine.go
@@ -46,11 +46,18 @@ func (e *Engine) Launch(ctx context.Context, t agenttask.Task) (string, error) {
 // past running.
 func (e *Engine) Status(ctx context.Context, sessionRef string) (status agenttask.Status, resultSummary, errMessage string, err error) {
 	s, resultSummary, errMessage, err := e.sandbox.Status(ctx, sessionRef)
-	if err != nil && errors.Is(err, engineapi.ErrUnknownSession) {
-		// Mirrors Remote.post's 404 mapping so the poller can tell "this
-		// session can never answer again" apart from a transient failure
-		// regardless of which agenttask.Engine arm is wired in.
-		return "", "", "", fmt.Errorf("%w: %s", agenttask.ErrEngineSessionGone, err)
+	if errors.Is(err, engineapi.ErrUnknownSession) {
+		// Same scoping as Remote.post's 404 mapping, so the poller can tell
+		// "this session can never answer again" apart from a transient
+		// failure regardless of which agenttask.Engine arm is wired in.
+		// %w twice keeps errors.Is(result, engineapi.ErrUnknownSession)
+		// working too, in case anything ever depends on the original chain.
+		// Unlike Remote.post, this error never crosses a process boundary —
+		// it only ever reaches the poller's own WARN log (see
+		// agenttask.Poller.pollTask), never a customer-visible field — so
+		// embedding sessionRef here is not the provider-blind leak it would
+		// be on the socket arm.
+		return "", "", "", fmt.Errorf("%w: %w", agenttask.ErrEngineSessionGone, err)
 	}
 	return agenttask.Status(s), resultSummary, errMessage, err
 }
@@ -58,6 +65,17 @@ func (e *Engine) Status(ctx context.Context, sessionRef string) (status agenttas
 // Cancel interrupts sessionRef's conversation and terminates its sandbox
 // process, which is what releases the session's concurrency slot on the
 // launcher side. Called by agenttask.Service.Cancel (issue #886).
+//
+// Maps engineapi.ErrUnknownSession the same way Status does above, for the
+// same arm-agnostic reason Remote.post's /cancel mapping exists: a session
+// the engine has no memory of holds no concurrency slot either (the quota
+// manager lives in the same in-memory state as the session registry), so
+// Service.stopEngineSession's ErrEngineSessionGone branch (its doc comment)
+// applies here too and there is nothing to warn an operator about.
 func (e *Engine) Cancel(ctx context.Context, sessionRef string) error {
-	return e.sandbox.Cancel(ctx, sessionRef)
+	err := e.sandbox.Cancel(ctx, sessionRef)
+	if errors.Is(err, engineapi.ErrUnknownSession) {
+		return fmt.Errorf("%w: %w", agenttask.ErrEngineSessionGone, err)
+	}
+	return err
 }
--- a/apps/control-plane/internal/agentengine/engine.go
+++ b/apps/control-plane/internal/agentengine/engine.go
@@ -10,6 +10,8 @@ package agentengine
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/engineapi"
 	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/agenttask"
@@ -44,6 +46,12 @@ func (e *Engine) Launch(ctx context.Context, t agenttask.Task) (string, error) {
 // past running.
 func (e *Engine) Status(ctx context.Context, sessionRef string) (status agenttask.Status, resultSummary, errMessage string, err error) {
 	s, resultSummary, errMessage, err := e.sandbox.Status(ctx, sessionRef)
+	if err != nil && errors.Is(err, engineapi.ErrUnknownSession) {
+		// Mirrors Remote.post's 404 mapping so the poller can tell "this
+		// session can never answer again" apart from a transient failure
+		// regardless of which agenttask.Engine arm is wired in.
+		return "", "", "", fmt.Errorf("%w: %s", agenttask.ErrEngineSessionGone, err)
+	}
 	return agenttask.Status(s), resultSummary, errMessage, err
 }
 
--- a/apps/control-plane/internal/agentengine/remote.go
+++ b/apps/control-plane/internal/agentengine/remote.go
@@ -150,6 +150,17 @@ func (r *Remote) post(ctx context.Context, path string, body any, out any) error
 		// container's log, not in a value callers may surface, so the
 		// returned error carries only the operation and the status code.
 		log.Printf("agentengine: %s: status %d: %s", path, resp.StatusCode, detail)
+		if resp.StatusCode == http.StatusNotFound {
+			// The launcher maps engineapi.ErrUnknownSession onto 404
+			// (apps/agent-engine/cmd/agent-engine/serve.go), which happens
+			// only for /status and /cancel: the session reference is gone
+			// from its in-memory registry (a launcher restart lost it) and
+			// can never answer again. Wrapping agenttask.ErrEngineSessionGone
+			// here, rather than only in the in-process agentengine.Engine
+			// arm, is what lets the poller tell this definitive case apart
+			// from a transient failure regardless of which arm is wired in.
+			return fmt.Errorf("agentengine: %s: status %d: %w", path, resp.StatusCode, agenttask.ErrEngineSessionGone)
+		}
 		return fmt.Errorf("agentengine: %s: status %d", path, resp.StatusCode)
 	}
 	if out == nil {
--- a/apps/control-plane/internal/agentengine/remote.go
+++ b/apps/control-plane/internal/agentengine/remote.go
@@ -150,15 +150,20 @@ func (r *Remote) post(ctx context.Context, path string, body any, out any) error
 		// container's log, not in a value callers may surface, so the
 		// returned error carries only the operation and the status code.
 		log.Printf("agentengine: %s: status %d: %s", path, resp.StatusCode, detail)
-		if resp.StatusCode == http.StatusNotFound {
-			// The launcher maps engineapi.ErrUnknownSession onto 404
-			// (apps/agent-engine/cmd/agent-engine/serve.go), which happens
-			// only for /status and /cancel: the session reference is gone
-			// from its in-memory registry (a launcher restart lost it) and
-			// can never answer again. Wrapping agenttask.ErrEngineSessionGone
-			// here, rather than only in the in-process agentengine.Engine
-			// arm, is what lets the poller tell this definitive case apart
-			// from a transient failure regardless of which arm is wired in.
+		// The launcher maps engineapi.ErrUnknownSession onto 404
+		// (apps/agent-engine/cmd/agent-engine/serve.go), which happens only
+		// for /status and /cancel: the session reference is gone from its
+		// in-memory registry (a launcher restart lost it) and can never
+		// answer again. Scoped to those two paths deliberately: post also
+		// serves /launch, and net/http.ServeMux answers 404 for ANY path it
+		// has no handler for, so an unscoped check would call a routing
+		// miss (e.g. control-plane and the host launcher binary drifting out
+		// of sync) session-gone instead of the ordinary transient error it
+		// actually is. Wrapping agenttask.ErrEngineSessionGone here, rather
+		// than only in the in-process agentengine.Engine arm, is what lets
+		// the poller tell this definitive case apart from a transient
+		// failure regardless of which arm is wired in.
+		if resp.StatusCode == http.StatusNotFound && (path == "/status" || path == "/cancel") {
 			return fmt.Errorf("agentengine: %s: status %d: %w", path, resp.StatusCode, agenttask.ErrEngineSessionGone)
 		}
 		return fmt.Errorf("agentengine: %s: status %d", path, resp.StatusCode)
--- a/apps/control-plane/internal/agenttask/poller.go
+++ b/apps/control-plane/internal/agenttask/poller.go
@@ -22,6 +22,28 @@ type StatusChecker interface {
 // loop's interval to (see loop's doc comment).
 const maxPollerBackoff = 5 * time.Minute
 
+// ErrEngineSessionGone is a StatusChecker implementation's definitive answer
+// that sessionRef no longer exists anywhere the engine can reach it — the
+// launcher restarted and lost its in-memory session registry, and its own
+// /status endpoint answers 404 for exactly that reason (see
+// agentengine.Remote.post and agentengine.Engine.Status, which map their
+// respective 404/engineapi.ErrUnknownSession cases onto this sentinel).
+// Unlike every other StatusChecker error, this one is not retried next pass:
+// a session the engine has no memory of can never report progress again, so
+// RunOnce fails the task immediately instead of leaving it queued/running
+// forever. Left unhandled, a single such task sits in ListActive's
+// cross-tenant result set and errors on every single pass, which used to
+// hold the whole poller's shared backoff at maxPollerBackoff (see loop's doc
+// comment) for as long as that one row existed, throttling poll cadence for
+// every other tenant's tasks too (measured live 2026-08-16, task f9409763).
+var ErrEngineSessionGone = errors.New("agenttask: engine has no memory of this session")
+
+// engineSessionGoneMessage is the customer-visible error_message persisted
+// when ErrEngineSessionGone fires. Provider-blind, same rationale as
+// Service's engineLaunchFailedMessage: no session reference, sandbox path,
+// or engine detail ever reaches a customer-facing field.
+const engineSessionGoneMessage = "agent task session is no longer reachable"
+
 // PollerConfig controls Poller behaviour.
 type PollerConfig struct {
 	// Interval between poll passes when the previous pass had no errors.
@@ -69,11 +91,23 @@ func NewPoller(repo Repository, checker StatusChecker, cfg PollerConfig) *Poller
 
 // RunOnce performs exactly one poll pass: every active task gets exactly one
 // StatusChecker.Status call and, if terminal, one Repository.Transition
-// call. A single task's engine error is logged and skipped — it stays
-// active and is retried next pass, it does not abort the rest of the pass.
-// The returned error, when non-nil, reports that at least one task had a
-// problem this pass (see loop's backoff); it does not mean the pass failed
-// outright unless ListActive itself errored.
+// call. A single task's transient engine error is logged and skipped — it
+// stays active and is retried next pass, it does not abort the rest of the
+// pass. ErrEngineSessionGone is not transient and is not retried: it fails
+// the task in this same pass (see its doc comment).
+//
+// The returned error, when non-nil, is loop's sole backoff signal (see
+// loop's doc comment) and means this PASS looks systemically degraded: every
+// active task's status check failed, or ListActive itself failed. It is
+// deliberately NOT "at least one task had a problem": a sweep where some
+// tasks succeed and one fails is not a failed sweep in any useful sense, and
+// ListActive is cross-tenant, so treating any single task-level error as
+// pass-level failure let one permanently broken task hold the whole poller's
+// shared backoff at maxPollerBackoff for as long as that row existed,
+// throttling poll cadence for every other tenant's tasks too (measured live
+// 2026-08-16, task f9409763: a lost session 404ing forever kept every sweep
+// erroring, so consecutiveFailures never reset). A pass with zero active
+// tasks is never systemically degraded.
 func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 	tasks, err := p.repo.ListActive(ctx)
 	if err != nil {
@@ -84,6 +118,23 @@ func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 	for _, t := range tasks {
 		status, resultSummary, errMessage, statusErr := p.checker.Status(ctx, t.EngineSessionRef)
 		if statusErr != nil {
+			if errors.Is(statusErr, ErrEngineSessionGone) {
+				p.logger.WarnContext(ctx, "agenttask: engine has no memory of this session, failing the task",
+					"task_id", t.ID, "error", statusErr)
+				if _, transErr := p.repo.Transition(ctx, t.TenantID, t.UserID, t.ID, StatusFailed, "", "", engineSessionGoneMessage); transErr != nil {
+					if errors.Is(transErr, ErrTerminalState) {
+						// Lost a race with a concurrent Cancel or a previous
+						// pass that already resolved this: already terminal.
+						continue
+					}
+					errCount++
+					p.logger.WarnContext(ctx, "agenttask: poller transition failed, retrying next pass",
+						"task_id", t.ID, "error", transErr)
+					continue
+				}
+				advanced++
+				continue
+			}
 			errCount++
 			p.logger.WarnContext(ctx, "agenttask: poller status check failed, retrying next pass",
 				"task_id", t.ID, "error", statusErr)
@@ -119,8 +170,8 @@ func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 		advanced++
 	}
 
-	if errCount > 0 {
-		return advanced, fmt.Errorf("agenttask: poller pass had %d task-level error(s)", errCount)
+	if len(tasks) > 0 && errCount == len(tasks) {
+		return advanced, fmt.Errorf("agenttask: poller pass had %d/%d task-level error(s), sweep looks systemically degraded", errCount, len(tasks))
 	}
 	return advanced, nil
 }
--- a/apps/control-plane/internal/agenttask/poller.go
+++ b/apps/control-plane/internal/agenttask/poller.go
@@ -56,6 +56,15 @@ var ErrEngineSessionGone = errors.New("agenttask: engine has no memory of this s
 // unrecoverable outcome rather than an assertion of failure: the sandbox may
 // well have finished the work before the launcher lost the reference to it,
 // so "failed" would be stating something this code cannot actually know.
+//
+// The split between this message and engineUnreachableAfterRetriesMessage
+// below is deliberate and only half done: the row this text is attached to
+// is still persisted as status = StatusFailed either way, so a machine
+// reader (Handler.handleGet, the console, any future metric) counts both
+// the same as a confirmed failure. Telling "outcome unrecoverable" apart
+// from "confirmed failed" programmatically needs a distinct terminal status
+// or a reason code, not built here — tracked alongside the rest of the
+// operator-visibility follow-up in issue #921.
 const engineSessionGoneMessage = "this task's outcome could not be recovered: its session was lost"
 
 // maxTaskFailureDuration bounds, in wall-clock time rather than a raw pass
--- a/apps/control-plane/internal/agenttask/poller.go
+++ b/apps/control-plane/internal/agenttask/poller.go
@@ -34,13 +34,29 @@ const maxPollerBackoff = 5 * time.Minute
 // session the engine has no memory of can never report progress again, so
 // RunOnce fails the task immediately instead of leaving it queued/running
 // forever or charging it against maxTaskFailureBudget first.
+//
+// The routine producer of this path on the demo box is a DEPLOY, not a rare
+// crash: deploy-demo-box.yml runs scripts/install-agent-engine-host.sh
+// unconditionally on every merge to main, which always stops and restarts
+// the launcher unit, and every in-flight session's registry entry dies with
+// it. So a merge to main now terminally fails every in-flight Cowork task
+// within one poll interval of the restart completing, including one whose
+// sandbox had already finished successfully but whose result the poller had
+// not yet read — the truthful state there is "outcome unrecoverable," not
+// "the task failed," which is why engineSessionGoneMessage below is worded
+// the way it is rather than asserting failure. Mitigating this (a drain, or
+// a registry that survives a restart) is tracked separately, issue #921;
+// this fix intentionally does not build it.
 var ErrEngineSessionGone = errors.New("agenttask: engine has no memory of this session")
 
 // engineSessionGoneMessage is the customer-visible error_message persisted
 // when ErrEngineSessionGone fires. Provider-blind, same rationale as
 // Service's engineLaunchFailedMessage: no session reference, sandbox path,
-// or engine detail ever reaches a customer-facing field.
-const engineSessionGoneMessage = "agent task session is no longer reachable"
+// or engine detail ever reaches a customer-facing field. Worded as an
+// unrecoverable outcome rather than an assertion of failure: the sandbox may
+// well have finished the work before the launcher lost the reference to it,
+// so "failed" would be stating something this code cannot actually know.
+const engineSessionGoneMessage = "this task's outcome could not be recovered: its session was lost"
 
 // maxTaskFailureBudget bounds how many consecutive passes a single task's
 // status check (or, once it reports a real terminal status, the transition
@@ -65,9 +81,11 @@ const maxTaskFailureBudget = 20
 
 // engineUnreachableAfterRetriesMessage is the customer-visible
 // error_message persisted when a task exhausts maxTaskFailureBudget without
-// ErrEngineSessionGone ever firing. Provider-blind, same rationale as
-// engineSessionGoneMessage.
-const engineUnreachableAfterRetriesMessage = "agent task could not reach its session after repeated attempts"
+// ErrEngineSessionGone ever firing. Provider-blind and worded the same
+// unrecoverable-outcome way as engineSessionGoneMessage, for the same
+// reason: repeated Status failures mean this code cannot learn what
+// happened, not that the task is confirmed to have failed.
+const engineUnreachableAfterRetriesMessage = "this task's outcome could not be recovered: its session stopped answering"
 
 // PollerConfig controls Poller behaviour.
 type PollerConfig struct {
@@ -158,11 +176,26 @@ func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 	}
 
 	active := make(map[uuid.UUID]bool, len(tasks))
+	var declaredDead int
 	for _, t := range tasks {
 		active[t.ID] = true
-		if p.pollTask(ctx, t) {
+		res := p.pollTask(ctx, t)
+		if res.advanced {
 			advanced++
 		}
+		if res.declaredDead {
+			declaredDead++
+		}
+	}
+
+	// One line per pass, not one per task: the signature that actually
+	// matters operationally (a launcher restart killed every tenant's
+	// in-flight task at once, see ErrEngineSessionGone's doc comment) is
+	// otherwise invisible in a log made of N separate per-task WARN lines
+	// with no total anywhere.
+	if declaredDead > 0 {
+		p.logger.WarnContext(ctx, "agenttask: poller declared task(s) dead this pass",
+			"declared_dead", declaredDead, "active_this_pass", len(tasks))
 	}
 
 	// Forget the failure streak of any task no longer active (resolved this
@@ -176,24 +209,39 @@ func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 	return advanced, nil
 }
 
-// pollTask advances one task and reports whether it reached a terminal
-// state this pass. Every failure path here is scoped to t alone — nothing
-// it does is visible to RunOnce's own return value, by design (see
-// RunOnce's doc comment).
-func (p *Poller) pollTask(ctx context.Context, t Task) (advanced bool) {
+// pollResult is pollTask's outcome for one task.
+type pollResult struct {
+	// advanced reports whether the task reached a terminal state this pass,
+	// for whatever reason: the engine's own report, or the poller giving up
+	// on it (declaredDead below covers exactly the latter).
+	advanced bool
+	// declaredDead reports whether THIS pass is the one where the poller
+	// itself decided the task is dead (ErrEngineSessionGone or an exhausted
+	// maxTaskFailureBudget) — never true for a task the engine reported
+	// succeeded/failed/cancelled on its own.
+	declaredDead bool
+}
+
+// pollTask advances one task and reports its outcome this pass (see
+// pollResult). Every failure path here is scoped to t alone — nothing it
+// does is visible to RunOnce's own return value, by design (see RunOnce's
+// doc comment).
+func (p *Poller) pollTask(ctx context.Context, t Task) pollResult {
 	status, resultSummary, errMessage, statusErr := p.checker.Status(ctx, t.EngineSessionRef)
 	if statusErr != nil {
 		if errors.Is(statusErr, ErrEngineSessionGone) {
 			p.logger.WarnContext(ctx, "agenttask: engine has no memory of this session, failing the task",
 				"task_id", t.ID, "error", statusErr)
-			return p.failTask(ctx, t, engineSessionGoneMessage)
+			ok := p.failTask(ctx, t, engineSessionGoneMessage)
+			return pollResult{advanced: ok, declaredDead: ok}
 		}
-		return p.chargeFailureBudget(ctx, t, statusErr)
+		ok := p.chargeFailureBudget(ctx, t, statusErr)
+		return pollResult{advanced: ok, declaredDead: ok}
 	}
 	delete(p.taskFailures, t.ID) // a clean answer this pass, whatever it was
 
 	if status != StatusSucceeded && status != StatusFailed && status != StatusCancelled {
-		return false
+		return pollResult{}
 	}
 
 	// Provider-blind boundary: errMessage came from StatusChecker, an
@@ -212,23 +260,23 @@ func (p *Poller) pollTask(ctx context.Context, t Task) (advanced bool) {
 			// Lost a race with a concurrent Cancel (or a previous pass that
 			// already advanced this task): already terminal, nothing left
 			// to do.
-			return false
+			return pollResult{}
 		}
 		// A real database write failure for an ALREADY-KNOWN outcome:
 		// retried next pass forever, exactly as before this fix. Not
 		// charged against the failure budget — see RunOnce's doc comment.
 		p.logger.WarnContext(ctx, "agenttask: poller transition failed, retrying next pass",
 			"task_id", t.ID, "error", transErr)
-		return false
+		return pollResult{}
 	}
-	return true
+	return pollResult{advanced: true} // the engine's own report, not the poller giving up
 }
 
 // chargeFailureBudget counts one more consecutive failure against t and, once
 // that reaches maxTaskFailureBudget, gives up on the task for good instead of
 // retrying it forever (see maxTaskFailureBudget's doc comment for why this
 // exists alongside ErrEngineSessionGone rather than instead of it).
-func (p *Poller) chargeFailureBudget(ctx context.Context, t Task, cause error) (advanced bool) {
+func (p *Poller) chargeFailureBudget(ctx context.Context, t Task, cause error) (declaredDead bool) {
 	p.taskFailures[t.ID]++
 	if p.taskFailures[t.ID] < maxTaskFailureBudget {
 		p.logger.WarnContext(ctx, "agenttask: poller status check failed, retrying next pass",
--- a/apps/control-plane/internal/agenttask/poller.go
+++ b/apps/control-plane/internal/agenttask/poller.go
@@ -7,6 +7,8 @@ import (
 	"log/slog"
 	"sync"
 	"time"
+
+	"github.com/google/uuid"
 )
 
 // StatusChecker is the narrow surface the Poller needs from an Engine
@@ -28,14 +30,10 @@ const maxPollerBackoff = 5 * time.Minute
 // /status endpoint answers 404 for exactly that reason (see
 // agentengine.Remote.post and agentengine.Engine.Status, which map their
 // respective 404/engineapi.ErrUnknownSession cases onto this sentinel).
-// Unlike every other StatusChecker error, this one is not retried next pass:
-// a session the engine has no memory of can never report progress again, so
+// Unlike every other StatusChecker error, this one is not retried at all: a
+// session the engine has no memory of can never report progress again, so
 // RunOnce fails the task immediately instead of leaving it queued/running
-// forever. Left unhandled, a single such task sits in ListActive's
-// cross-tenant result set and errors on every single pass, which used to
-// hold the whole poller's shared backoff at maxPollerBackoff (see loop's doc
-// comment) for as long as that one row existed, throttling poll cadence for
-// every other tenant's tasks too (measured live 2026-08-16, task f9409763).
+// forever or charging it against maxTaskFailureBudget first.
 var ErrEngineSessionGone = errors.New("agenttask: engine has no memory of this session")
 
 // engineSessionGoneMessage is the customer-visible error_message persisted
@@ -43,6 +41,33 @@ var ErrEngineSessionGone = errors.New("agenttask: engine has no memory of this s
 // Service's engineLaunchFailedMessage: no session reference, sandbox path,
 // or engine detail ever reaches a customer-facing field.
 const engineSessionGoneMessage = "agent task session is no longer reachable"
+
+// maxTaskFailureBudget bounds how many consecutive passes a single task's
+// status check (or, once it reports a real terminal status, the transition
+// that records it) may fail before RunOnce gives up on that ONE task and
+// fails it directly, regardless of what shape the error took. 20 passes is
+// about 5 minutes at the default 15s interval, the same order of magnitude
+// maxPollerBackoff used to cap the whole poller's shared interval at, now
+// scoped to one task instead of every tenant's sweep.
+//
+// This exists because ErrEngineSessionGone only covers one definitive
+// failure shape: a 404 from a launcher whose in-memory registry never had
+// (or lost) the session. A sandbox killed some other way — OOM, an
+// Apptainer crash, anything that does not go through Cancel or a normal
+// terminal status — never reaches SandboxEngine.reap (it runs only on
+// terminal status or Cancel), so its /status call fails forever too, but
+// with a 502, not a 404 (apps/agent-engine/cmd/agent-engine/serve.go's
+// default error-mapping branch). Same dead task as f9409763, a different,
+// non-enumerable error shape. A per-task budget catches that case and any
+// future one without needing to name every failure mode a StatusChecker
+// implementation could return.
+const maxTaskFailureBudget = 20
+
+// engineUnreachableAfterRetriesMessage is the customer-visible
+// error_message persisted when a task exhausts maxTaskFailureBudget without
+// ErrEngineSessionGone ever firing. Provider-blind, same rationale as
+// engineSessionGoneMessage.
+const engineUnreachableAfterRetriesMessage = "agent task could not reach its session after repeated attempts"
 
 // PollerConfig controls Poller behaviour.
 type PollerConfig struct {
@@ -63,6 +88,12 @@ type Poller struct {
 	checker  StatusChecker
 	interval time.Duration
 	logger   *slog.Logger
+
+	// taskFailures counts each task's CONSECUTIVE per-pass failures (see
+	// maxTaskFailureBudget). Only ever touched from RunOnce, which loop
+	// guarantees is never running concurrently with itself (Start/Stop's own
+	// mutex serializes passes), so this needs no lock of its own.
+	taskFailures map[uuid.UUID]int
 
 	mu      sync.Mutex
 	cancel  context.CancelFunc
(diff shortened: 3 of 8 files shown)


apps/control-plane/internal/agenttask/poller.go
```diff
@@ -119,8 +170,8 @@ func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 		advanced++
 	}
 
-	if errCount > 0 {
-		return advanced, fmt.Errorf("agenttask: poller pass had %d task-level error(s)", errCount)
+	if len(tasks) > 0 && errCount == len(tasks) {
```
sakibsadmanshajib: **HIGH: the new systemic rule degenerates to the old behaviour in the single active task case, which is the normal case on this deployment, and a non-404 permanent failure still reproduces the incident verbatim.**

`errCount == len(tasks)` is trivially satisfied when `len(tasks) == 1`. The launcher caps concurrency at 2 per user and 4 per tenant (`HIVE_QUOTA_USER_CONCURRENCY`, `HIVE_QUOTA_TENANT_CONCURRENCY`), and the demo box normally has one Cowork task in flight, so one active task is the common shape, not the corner. In that shape a single broken row still makes every pass a full failure, still drives `consecutiveFailures` up without bound, and still pins the loop at `maxPollerBackoff` for every tenant. The PR body acknowledges this and argues it is a genuine nothing-answering signal. It is not, and the code path that proves it is reachable today:

- `SandboxEngine.reap` runs only from `Status` on a terminal state or from `Cancel`. A sandbox killed from outside (OOM kill, an Apptainer crash, the process dying without the conversation reaching a terminal execution_status) leaves its entry in `e.sessions` with `terminal == nil`.
- The next `Status` call therefore reaches `sess.client.GetConversation`, which dials a control socket whose sandbox is gone, and returns `engine: get conversation: ...`.
- `serve.go` maps that to **502**, not 404, because it is not `ErrUnknownSession`. So it never becomes `ErrEngineSessionGone`, it is treated as transient, and it is retried forever.

That is the reported incident with one substitution: a permanently erroring row that is not session-gone, alone in `ListActive`, pinning the shared backoff at 300 seconds cross-tenant for as long as the row exists. The fix removes one producer of the class and leaves the class intact.

Two ways out, both smaller than what is here:

1. Stop inferring engine health from task errors at all: back off only when `ListActive` itself fails. The per pass status call count is bounded by the quota ceiling, and a Unix socket round trip to a launcher answering 404 or 502 is cheap, so hammering is not the risk that justifies the heuristic. The database is the expensive dependency, and `ListActive` failure already covers it. One condition, and the whole class disappears.
2. Give the poller a per task budget: fail a task after N consecutive status errors, or sweep rows whose `started_at` is older than a maximum sandbox lifetime. `Service.WaitIdle` already names the sibling of this sweep for stranded queued rows, so the shape is not new to this package.

Either one also removes the need for the sample size argument in the doc comment above.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -69,11 +91,23 @@ func NewPoller(repo Repository, checker StatusChecker, cfg PollerConfig) *Poller
...
-// The returned error, when non-nil, reports that at least one task had a
-// problem this pass (see loop's backoff); it does not mean the pass failed
-// outright unless ListActive itself errored.
+// call. A single task's transient engine error is logged and skipped — it
+// stays active and is retried next pass, it does not abort the rest of the
+// pass. ErrEngineSessionGone is not transient and is not retried: it fails
+// the task in this same pass (see its doc comment).
+//
+// The returned error, when non-nil, is loop's sole backoff signal (see
+// loop's doc comment) and means this PASS looks systemically degraded: every
```
sakibsadmanshajib: **MEDIUM: the systemic inference is computed over a deliberately biased sample, so this doc comment claims more than the code can know.**

The premise is that every active task failing means nothing is answering. The population it reasons over is `agent_tasks_list_active()`, which is `WHERE status IN (queued, running) AND engine_session_ref <> * ORDER BY created_at ASC LIMIT 500` (20260716_05_agent_tasks_service_scan.sql). Oldest first, capped at 500.

Stuck rows are by definition the oldest rows in that ordering, so they are the ones guaranteed to be inside the window, and healthy fresh tasks are the ones that fall outside it once the window is full. The sample is therefore biased toward exactly the rows that error, which is the wrong direction for an all-of-them-failed inference. With 500 or more active rows the poller would also stop looking at new tasks entirely, and the new rule would read that as a systemic outage rather than a backlog.

This is not a regression (the old `errCount > 0` rule was worse), and the session-gone path makes a 500 row stuck backlog less likely. It is a reason not to build the health signal out of task errors in the first place; see the comment on the rule itself. At minimum, the doc comment should not assert that every active task failing means the engine is unreachable without naming the LIMIT 500 oldest-first window it is actually measuring.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -84,6 +118,23 @@ func (p *Poller) RunOnce(ctx context.Context) (advanced int, err error) {
 	for _, t := range tasks {
 		status, resultSummary, errMessage, statusErr := p.checker.Status(ctx, t.EngineSessionRef)
 		if statusErr != nil {
+			if errors.Is(statusErr, ErrEngineSessionGone) {
```
sakibsadmanshajib: **MEDIUM, operational: the routine producer of this path is a deploy, and the PR never says so. After this change every merge to main terminally fails every in-flight Cowork task within one poll interval.**

The mapping itself is correct on this deployment, and I am not asking for it to be reverted. I checked the three ways it could be wrong and it survives all three:

- A transport failure does not become a 404. `Remote.post` returns the `http.Do` error wrapped separately, well above the status code branch, so a dial failure while the launcher is down is an ordinary transient error and is retried. This is not a repeat of the transport-collapsed-into-401 defect.
- The launcher registry has no restore path. `SandboxEngine.sessions` is an in-memory map and `New` builds it empty, so a 404 after a restart is permanent by construction, not a warm-up window.
- The install script does `systemctl --user stop` before `systemd-run`, so there is no double-launcher window where a second daemon answers 404 for sessions the first one still owns, and the default systemd KillMode takes the sandbox children down with the unit. The sessions really are dead.

What is missing is the consequence. `deploy-demo-box.yml` runs `scripts/install-agent-engine-host.sh` unconditionally in the deploy job, and that script always stops and restarts the daemon. The unit also carries `Restart=on-failure`. So a launcher restart is not an exceptional event on this box, it is a per deploy event, and after this change each one converts every in-flight agent task into a customer-visible `failed` within 15 seconds instead of leaving it hanging. That includes a task that had already succeeded inside the sandbox but whose terminal status the poller had not yet read: its result is simply lost and the row says failed.

Asks, none of which need to be in this PR beyond the first:

1. Say it here, in the doc comment. The next reader should not have to derive from the deploy workflow that the common trigger is a deploy rather than a rare crash.
2. Open an issue for the mitigation: either a drain (refuse the restart, or wait, while sessions are live) or a registry that survives a restart. Today a merge landing mid demo kills the running demo task.
3. Consider whether the recorded outcome should be `failed`. The truthful statement is that the outcome is unknown, not that the task failed; see the separate comment on the error message.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -22,6 +22,28 @@ type StatusChecker interface {
...
+// hold the whole poller's shared backoff at maxPollerBackoff (see loop's doc
+// comment) for as long as that one row existed, throttling poll cadence for
+// every other tenant's tasks too (measured live 2026-08-16, task f9409763).
+var ErrEngineSessionGone = errors.New("agenttask: engine has no memory of this session")
+
+// engineSessionGoneMessage is the customer-visible error_message persisted
+// when ErrEngineSessionGone fires. Provider-blind, same rationale as
+// Service's engineLaunchFailedMessage: no session reference, sandbox path,
+// or engine detail ever reaches a customer-facing field.
+const engineSessionGoneMessage = "agent task session is no longer reachable"
```
sakibsadmanshajib: **MEDIUM: is a task failed by this path distinguishable from a task that failed on its own merits? Barely, and only by string matching a free text column.**

Taking the question seriously, because the review brief asks it directly.

What an operator gets today:

- The persisted `error_message` is this constant, and it is different from the other three constants in this package (`agent task failed`, `agent engine could not start the task`, `agent engine is not available on this deployment`). So a `SELECT ... WHERE error_message = ...` does separate them.
- One WARN per task, carrying `task_id` and the wrapped error.

What is wrong with that:

1. It is a free text column with no code beside it, so the only way to count these is to match the exact English sentence. This repo has already been burned by treating a matched error string as a control signal on the billing path. A reason code (or a distinct terminal status) is the durable version, and the message becomes presentation.
2. There is no aggregate anywhere. A launcher restart with nine live tasks produces nine separate WARN lines and no counter, no pass level total, no metric. The signature that matters operationally (one event killed every tenant task at once) is exactly the one that is invisible. A single log line per pass with the count, or a counter, is cheap.
3. The wording is wrong in both directions for a customer. It reads as transient and retryable (not reachable, try again later) when the task is permanently dead, and it asserts failure when the truth is that the outcome is unknown: the sandbox may well have finished the work successfully before the launcher lost the reference, and the summary is simply unrecoverable. Something closer to the outcome of this task could not be recovered is both more accurate and more actionable, and it does not invite a pointless retry of a task that may already have run.

Also worth checking before this merges: this string is what the console renders, so it needs to read as a sentence a customer can act on, not as a description of our internal registry.

apps/control-plane/internal/agentengine/remote.go
```diff
@@ -150,6 +150,17 @@ func (r *Remote) post(ctx context.Context, path string, body any, out any) error
 		// container's log, not in a value callers may surface, so the
 		// returned error carries only the operation and the status code.
 		log.Printf("agentengine: %s: status %d: %s", path, resp.StatusCode, detail)
+		if resp.StatusCode == http.StatusNotFound {
```
sakibsadmanshajib: **MEDIUM: the mapping is on `post`, which is shared by `/launch`, so it says session gone for calls that have no session. Scope it to the paths the launcher actually emits a session scoped 404 on.**

The comment right below correctly states that the launcher answers 404 only for `/status` and `/cancel`. That is a fact about the launcher build currently on the other end of this socket, and this code cannot see that build. `net/http.ServeMux` answers 404 for any path it has no pattern for, so the moment control-plane calls an endpoint the running launcher does not have (a launcher installed by `scripts/install-agent-engine-host.sh` on the host, while control-plane ships in a container, is a separate artefact even though the deploy job currently updates both), a routing miss is reported to callers as agenttask.ErrEngineSessionGone. Nothing consumes it on the `/launch` path today, so this is latent rather than live, but the cost of closing it is one condition:

```go
if resp.StatusCode == http.StatusNotFound && (path == "/status" || path == "/cancel") {
```

The durable version is a machine readable `code` field in the launcher error body, since `post` already decodes that body into `e.Error` and throws it away. Path scoping is the lazy version and is enough.

Credit where due, and please keep it: a transport failure never reaches this branch, because `r.http.Do` failing returns above with the error wrapped. That is precisely the discrimination the earlier invalid_api_key defect was missing, and it is what makes treating this 404 as terminal defensible. It is not pinned by a test though. Please add the negative case next to `TestRemoteStatus404MapsToEngineSessionGone`: close the listener (or point at a socket path that does not exist), call `Status`, and assert `errors.Is(err, agenttask.ErrEngineSessionGone)` is false. Without it, a future refactor that returns a synthesized 404 on a dial failure would kill live tasks and every test here would stay green.

Related, and visible in the new test: the handler it installs answers 404 for every path and every method, which is a fair model of a routing miss and not of the launcher. It passes only because nothing distinguishes the two.

apps/control-plane/internal/agentengine/remote.go
```diff
@@ -150,6 +150,17 @@ func (r *Remote) post(ctx context.Context, path string, body any, out any) error
...
+			// The launcher maps engineapi.ErrUnknownSession onto 404
+			// (apps/agent-engine/cmd/agent-engine/serve.go), which happens
+			// only for /status and /cancel: the session reference is gone
+			// from its in-memory registry (a launcher restart lost it) and
+			// can never answer again. Wrapping agenttask.ErrEngineSessionGone
+			// here, rather than only in the in-process agentengine.Engine
+			// arm, is what lets the poller tell this definitive case apart
+			// from a transient failure regardless of which arm is wired in.
+			return fmt.Errorf("agentengine: %s: status %d: %w", path, resp.StatusCode, agenttask.ErrEngineSessionGone)
```
sakibsadmanshajib, line 167: **MEDIUM: this now changes `Cancel` too, and the one consumer of that error logs something that is false in exactly this case.**

`post` serves `Cancel`, so a cancel against a session the launcher has forgotten now returns `ErrEngineSessionGone`. Nothing consumes the sentinel there. `Service.stopEngineSession` sees a non-nil error and logs:

> agenttask: engine cancel failed, the session may still hold its concurrency slot

which is wrong for this specific error. The launcher quota manager (`quota.Manager` inside `SandboxEngine`) lives in the same in-memory state as the session registry, so a launcher that has no memory of the session is holding no slot for it either. The operator is being pointed at a leaked slot that does not exist, on a path (issue #886) where leaked slots are the thing they have been trained to chase.

While the sentinel is being introduced anyway, the honest handling is one branch in `stopEngineSession`: `errors.Is(err, ErrEngineSessionGone)` means there is nothing to stop, so return without the warning. Not required for correctness of this fix, but it is three lines and it stops this PR from adding a misleading operator signal to the path it is trying to make diagnosable.

apps/control-plane/internal/agenttask/poller_test.go
```diff
@@ -138,6 +147,108 @@ func TestPoller_RunOnce_EngineErrorIsRetriedNotFailed(t *testing.T) {
...
+// is the loop-level regression guard for the same bug, driven across several
+// simulated passes the way loop() itself does: a task whose engine session is
+// permanently unreachable (any ordinary error, not ErrEngineSessionGone) sits
+// in ListActive next to a perfectly healthy task that answers cleanly every
+// pass without ever reaching a terminal state. consecutiveFailures must stay
+// at 0 the whole time, which is exactly the input pollerBackoffDelay needs to
+// keep returning the configured base interval (see TestPollerBackoffDelay);
+// before this fix it grew without bound and pinned the shared poll cadence at
+// maxPollerBackoff for every other tenant's tasks too.
+func TestPoller_ConsecutiveFailures_OnePermanentlyFailingTaskDoesNotBackOffTheSweep(t *testing.T) {
```
sakibsadmanshajib, line 208: **MEDIUM, test discrimination. I checked all four new or changed tests against a reverted fix. Two discriminate, one does not, and the one that matters most is built around the case that works while the case that does not work is untested.**

Pass by pass:

- This test does discriminate. Revert `RunOnce` to `errCount > 0` and every pass errors, so `consecutiveFailures` reaches 10 and the assertion fails. Asserting on the counter that actually drives `pollerBackoffDelay`, rather than on a call count, is the right shape and I would keep it.
- `TestPoller_RunOnce_EngineSessionGoneFailsTaskInstead` discriminates too: under the old rule a single erroring task returns an error and the first assertion fires. The second pass call-count check is a genuinely good touch.
- `TestPoller_RunOnce_AllTasksErroringIsSystemicFailure` does **not** discriminate. It passes identically against the unfixed code, because `errCount > 0` also returns an error there. That is fine as a guard against over-correcting into never backing off, but the comment presents it as locking in the other side of the new rule, which it does not, and a reader counting green tests will over-credit it. Say what it guards.

The gap: the healthy companion task in this test is load-bearing and unremarked. Delete `session-healthy` and this test fails against the fixed code, because one active task erroring is still a fully failed pass. That is the case the PR body admits is unchanged, it is the common shape on a box with a 2 per user concurrency cap, and there is no test for it at all. Whichever way that is resolved (fix the rule, or accept the ceiling), it needs a test that states it:

- If it is being accepted for now, add `TestPoller_ConsecutiveFailures_SingleBrokenTaskStillPinsTheSweep` asserting `consecutiveFailures` grows, with a comment naming it as a known ceiling and the issue number. A red test is not needed, an honest green one is.
- If the rule changes per the comment on `RunOnce`, the same test flips to asserting 0 and becomes the real regression guard.

As it stands the suite reads as proving the bug is fixed, and what it proves is that the bug is fixed when at least one other task is healthy in the same pass.

apps/control-plane/internal/agentengine/engine.go
```diff
@@ -44,6 +46,12 @@ func (e *Engine) Launch(ctx context.Context, t agenttask.Task) (string, error) {
 // past running.
 func (e *Engine) Status(ctx context.Context, sessionRef string) (status agenttask.Status, resultSummary, errMessage string, err error) {
 	s, resultSummary, errMessage, err := e.sandbox.Status(ctx, sessionRef)
+	if err != nil && errors.Is(err, engineapi.ErrUnknownSession) {
```
sakibsadmanshajib: **LOW, two nits in three lines.**

1. `err != nil &&` is redundant. `errors.Is(nil, target)` is false for any non-nil target, and `engineapi.ErrUnknownSession` is non-nil by construction. The guard reads as if a nil error could match.
2. `fmt.Errorf("%w: %s", agenttask.ErrEngineSessionGone, err)` wraps the new sentinel and drops the old chain: `errors.Is(result, engineapi.ErrUnknownSession)` is false after this. Nothing depends on that today (the launcher checks `engineapi.ErrUnknownSession` against the engine directly, not through this adapter), so this is not a bug, but `%w: %w` costs the same and keeps both. Losing the chain silently is the kind of thing the next person only discovers when their `errors.Is` returns false for no visible reason.

Unrelated to correctness, worth knowing: the error text this produces embeds the raw session reference, since the inner `engine: unknown session reference: <ref>` is formatted in. The socket arm deliberately strips that detail. It only reaches the poller WARN log so nothing customer-facing leaks, but the two arms are not equivalent on that axis despite the comment saying this mirrors `Remote.post`.

sakibsadmanshajib:
## Adversarial review, stage 6

Streams that ran: one adversarial Go pass (this one), reading the pushed diff at `59d7535` on GitHub plus the surrounding code at the same ref. **SKIPPED**, and their absence is not a pass: CodeRabbit (review limit reached, its own comment above), Codex connector (usage limit), Cursor Bugbot (not enabled). So this PR has had exactly one reviewer.

**Verdict: BLOCK, on one HIGH.** The fix is real and the direction is right. The 404 to terminal mapping in particular survived the three attacks I ran at it (see below) and I am not asking for it to be reverted. But the headline claim, that one permanently failing task no longer backs off the sweep, holds only when a second healthy task is active in the same pass, and the case where it does not hold is the common one on this box and is untested.

### Findings

| Severity | Where | What |
| --- | --- | --- |
| HIGH | [poller.go rule](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792167932) | `errCount == len(tasks)` is trivially true at one active task, which is the normal shape under a 2 per user concurrency cap. A permanently erroring task that is not session-gone still pins the shared backoff at 300 seconds cross-tenant, and one is reachable today: an externally killed sandbox leaves its session registered, so `Status` fails at the control socket forever and the launcher answers 502, never 404 |
| MEDIUM | [poller.go doc](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792168332) | The all-of-them-failed inference is drawn from `ORDER BY created_at ASC LIMIT 500`, a window biased toward exactly the stuck rows |
| MEDIUM | [session-gone branch](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792168989) | The routine producer of this path is a deploy: `deploy-demo-box.yml` restarts the launcher unconditionally, so every merge now terminally fails every in-flight Cowork task. Correct behaviour, undocumented consequence, no drain |
| MEDIUM | [error message](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792169420) | Distinguishable only by matching free text, no code, no counter, and the copy asserts failure when the truth is unknown outcome |
| MEDIUM | [remote.go mapping](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792169906) | The mapping sits in `post`, shared by `/launch`, so a routing miss becomes session gone. One condition scopes it. Also missing the negative regression test that pins transport failure as not session gone |
| MEDIUM | [Cancel path](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792170249) | `Cancel` now returns the sentinel and `stopEngineSession` logs may still hold its concurrency slot, which is false for this error since the quota state died with the registry |
| MEDIUM | [tests](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792170795) | Two of the new tests discriminate, one passes identically against the unfixed code, and the single active task case has no test at all |
| LOW | [engine.go](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792171165) | Redundant `err != nil`, and `%w: %s` drops `engineapi.ErrUnknownSession` from the chain |

### Claims in the PR body I verified rather than accepted

- **The terminal guard and the Cancel race.** Confirmed. `Repository.Transition` still carries `AND status NOT IN (succeeded, failed, cancelled)` in the UPDATE, and the new session-gone path goes through the same call with the same swallowed `ErrTerminalState`. On a simultaneous cancel, both statements race on that predicate inside their own transactions, the loser matches zero rows, and the disambiguating read in the same transaction turns that into `ErrTerminalState` rather than a silent overwrite. So the poller cannot clobber a cancel. One consequence not in the PR body: when the poller wins, the users cancel request returns `ErrTerminalState` and the task reads `failed`, not `cancelled`. That matches the existing semantics of every other terminal race in this package, so I am not filing it, only naming it.
- **404 as terminal is safe here.** Attacked three ways, all three held: a transport failure returns above the status branch so it never becomes 404 shaped, the launcher registry is memory only with no restore path so a 404 after restart is permanent by construction, and the installer stops the unit before starting the new one so there is no window where a second daemon 404s for sessions a live first daemon still owns. This is genuinely not a repeat of the transport collapsed into invalid_api_key defect. It is only unpinned by a test, and the mapping is wider than the claim.
- **No money path.** Checked: `agenttask` holds no reservation or credit hold, so auto-failing a task on session-gone has no billing consequence to unwind. Sandbox model usage is metered through the gateway independently.

### Minimum to unblock

Either fix the rule (back off on `ListActive` failure only, or add a per task failure budget) or accept the single task ceiling explicitly, with a test that states it and an issue that owns it. Silence on that case plus a test suite arranged around the case that works is what I am blocking on, not the ceiling itself.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Fixed. RunOnce's returned error (loop's sole backoff signal) now means only that ListActive itself failed. Every task-level problem is scoped to a per-task consecutive-failure budget instead (maxTaskFailureBudget, 20 passes): ErrEngineSessionGone still fails a task immediately, and any other error (including the 502-from-a-killed-sandbox case this comment also identified) is retried until the budget is exhausted, then fails the task directly regardless of error shape. Neither path ever feeds the shared backoff. New regression tests: TestPoller_RunOnce_SingleTaskErroringForeverDoesNotBackOffTheSweep (this exact scenario, one active task, ten passes, never backs off) and TestPoller_RunOnce_TaskExceedingFailureBudgetFailsRegardlessOfErrorShape (the 502 case, fails on exactly the budget'th pass). Commit 39dc6a60.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Moot: the errCount == len(tasks) rule this targeted is removed entirely in commit 39dc6a60. RunOnce's error now means only ListActive itself failed, so it no longer makes any claim about a fraction of active tasks, biased sample or otherwise. Every task-level problem is scoped per-task (see the reply on the HIGH above).

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Addressed on all three asks. (1) The doc comment on ErrEngineSessionGone now says plainly that the routine producer on this deployment is a merge-triggered deploy restarting the launcher unit, not a rare crash, and names the truthful-state point (outcome unrecoverable, not confirmed failed). (2) Filed #921 to track the mitigation (drain or a registry that survives a restart), referenced from the doc comment and the PR body; not built in this PR as you said it doesn't need to be. (3) Message wording fixed: engineSessionGoneMessage and engineUnreachableAfterRetriesMessage no longer assert failure, both now read as an unrecoverable outcome, matching the point that the sandbox may have finished before the launcher lost the reference. Commit caf40886.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Taking the three points in order. (3) Wording fixed: engineSessionGoneMessage and engineUnreachableAfterRetriesMessage now read as an unrecoverable outcome (their session was lost / stopped answering) rather than asserting failure, matching your point that the sandbox may have already finished successfully before the launcher lost the reference. (2) Added a pass-level aggregate: RunOnce now counts how many tasks it personally declared dead this pass (a new pollResult.declaredDead, which also distinguishes that from an ordinary engine-reported terminal status) and logs one summary WARN with the count when nonzero, so a launcher restart that kills nine tasks at once now produces one line naming the total, not just nine unrelated per-task lines. (1) Rebutting the reason-code / distinct-status ask as out of scope for this fix: a new terminal status or a structured reason column is a schema change (CHECK constraint, SYNC_CONTRACT.md, console rendering), a materially bigger change than a poller cadence bug fix. Folded into #921 alongside the drain question so it gets designed together rather than bolted on piecemeal. Commit caf40886.

apps/control-plane/internal/agentengine/remote.go
sakibsadmanshajib: Fixed exactly as suggested, path-scoped: the 404-to-ErrEngineSessionGone mapping now only fires for path == "/status" || path == "/cancel". Kept the lazy version (path scoping) over the machine-readable code-field version, agreed that's enough. Added the negative test you asked for, plus one more: TestRemoteLaunch404DoesNotMapToEngineSessionGone (a bare http.NotFoundHandler modeling a routing miss on /launch, asserts the sentinel is NOT set) and TestRemoteTransportFailureDoesNotMapToEngineSessionGone (dial failure against a socket nothing listens on, asserts the sentinel is NOT set, pinning the transport-failure discrimination you credited). Commit caf40886.

apps/control-plane/internal/agentengine/remote.go
sakibsadmanshajib: Fixed as you described: Service.stopEngineSession now has an errors.Is(err, ErrEngineSessionGone) branch that returns without the leaked-slot warning, since a launcher with no memory of the session holds no slot for it either. Also mapped agentengine.Engine.Cancel (the in-process arm) onto the same sentinel the same way Engine.Status already did, so this holds regardless of which agenttask.Engine arm is wired in, not just the socket arm your comment was about. New test: TestService_Cancel_EngineSessionGoneDoesNotWarnAboutALeakedSlot (captures slog.Default's output for the duration of the test and asserts the leaked-slot warning text is absent). Commit caf40886.

apps/control-plane/internal/agenttask/poller_test.go
sakibsadmanshajib: Agreed on all three points. TestPoller_RunOnce_AllTasksErroringIsSystemicFailure is renamed to TestPoller_RunOnce_AllTasksErroringNeverBacksOffEither and its comment now says plainly what it guards (even a total-outage-looking pass never feeds backoff), not that it locks in the fraction rule (which no longer exists). The single-task gap is closed directly: TestPoller_RunOnce_SingleTaskErroringForeverDoesNotBackOffTheSweep is the exact scenario you named (one active task, the ordinary shape under this deployment's quota, ten passes, RunOnce never errors), and since the underlying rule changed (RunOnce's error now means only ListActive failed) rather than being accepted as a ceiling, this is the real regression guard, not a documented-known-limit test. The two you confirmed already discriminate are unchanged. Commit 39dc6a60.

apps/control-plane/internal/agentengine/engine.go
sakibsadmanshajib: Both fixed. Dropped the redundant err != nil guard (errors.Is(nil, target) is already false for the non-nil ErrUnknownSession target). Switched to fmt.Errorf's two-%w form (%w: %w) so wrapping ErrEngineSessionGone no longer silently drops errors.Is(result, engineapi.ErrUnknownSession) from the chain -- Go 1.24 here supports multiple %w verbs. Also added a note on the doc comment acknowledging the asymmetry you flagged: this in-process arm's error only ever reaches the poller's own WARN log (never a customer-visible field), so embedding sessionRef here is not the provider-blind leak it would be on the socket arm, but the two arms are indeed not equivalent on that axis as you said. Commit caf40886.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -64,65 +135,182 @@ func NewPoller(repo Repository, checker StatusChecker, cfg PollerConfig) *Poller
...
+// failTask transitions t straight to StatusFailed with a provider-blind
+// message. Deliberately does not clear t's failure streak when the
+// Transition call itself fails with a real (non-ErrTerminalState) error: the
+// streak is already at or past maxTaskFailureBudget by the time this runs
+// (chargeFailureBudget only calls it once the budget is exhausted, and
+// ErrEngineSessionGone routes straight here on its own), so leaving the
+// count where it is makes the next pass retry this same transition
+// immediately rather than waiting out a fresh maxTaskFailureBudget cycle
+// first.
+func (p *Poller) failTask(ctx context.Context, t Task, message string) (advanced bool) {
```
sakibsadmanshajib, line 421: **HIGH: the budget path fails the row without stopping the session, and the terminal row then removes the last path anyone had to free it. That is issue #886 reintroduced through a new door.**

The two paths into `failTask` are not symmetric and the code treats them as if they were.

`ErrEngineSessionGone` means the launcher has no memory of the session, so no sandbox, no quota slot, nothing to stop. `Service.stopEngineSession` now says exactly that, correctly, in this same PR.

Budget exhaustion means the opposite. The error was NOT a 404, so the launcher answered from its own handler, which means the session is still in `e.sessions` and its `release` has not run: `SandboxEngine.reap` runs only on a terminal status or on `Cancel`. The canonical case the budget exists for, the externally killed sandbox answering 502, is precisely a session the launcher still accounts for and still charges against `QuotaUserConcurrency`. So on the pass where the budget runs out:

- the row goes to `failed`, and
- the session keeps its slot, and
- because the row is now terminal, `Service.Cancel` returns `ErrTerminalState` before it ever reaches `stopEngineSession`, so the user can no longer free it either.

Before this change that row stayed active, which was the bug, but it also meant a user cancel could still release the slot. After this change nothing can, short of restarting the launcher, which `stopEngineSession` doc already names as the remedy that kills every other tenant live sandbox. At 2 per user, two budget-exhausted tasks lock a user out of creating anything for as long as those sandboxes live, which is the exact symptom #886 was filed for and measured at about sixteen minutes.

The fix is small because the plumbing already exists. `buildAgentEngine` returns `return remote, remote` (cmd/server/main.go), so the value behind `p.checker` is the same object that satisfies `agenttask.Engine`, and both arms (`Remote`, `agentengine.Engine`) already have `Cancel(ctx, sessionRef) error`. Best effort, budget path only:

```go
if stopper, ok := p.checker.(interface{ Cancel(context.Context, string) error }); ok {
    if err := stopper.Cancel(ctx, t.EngineSessionRef); err != nil {
        // log only: the row is already terminal, this frees the slot
    }
}
```

Do not do it on the session-gone path, where it is provably pointless and would just re-log the noise this PR removed. If you would rather not widen what the poller can do, the alternative is to leave the row active and let the user cancel it, but that is the bug this PR is fixing, so the stop has to live here.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -22,6 +24,69 @@ type StatusChecker interface {
...
+// (or lost) the session. A sandbox killed some other way — OOM, an
+// Apptainer crash, anything that does not go through Cancel or a normal
+// terminal status — never reaches SandboxEngine.reap (it runs only on
+// terminal status or Cancel), so its /status call fails forever too, but
+// with a 502, not a 404 (apps/agent-engine/cmd/agent-engine/serve.go's
+// default error-mapping branch). Same dead task as f9409763, a different,
+// non-enumerable error shape. A per-task budget catches that case and any
+// future one without needing to name every failure mode a StatusChecker
+// implementation could return.
+const maxTaskFailureBudget = 20
```
sakibsadmanshajib: **MEDIUM: the budget is counted in passes, documented in minutes, and the pass interval is an operator-tunable env var. Lowering that knob silently shortens a task kill timeout.**

`about 5 minutes at the default 15s interval` is doing a lot of work. The interval is `parseDurationEnv("HIVE_AGENT_TASK_POLL_INTERVAL", 15*time.Second)` in cmd/server/main.go, so it is deployment configurable, and the obvious tuning anyone reaches for after reading this very incident (nought to five minutes of dead air after a task finishes) is to lower it. Set it to 5s and this constant becomes a 100 second budget. Agent tasks on this box run 16 to 22 minutes, so a 100 second hiccup would then terminate a task that is 3 minutes into a 20 minute run, and nothing in the change would look wrong to whoever turned the knob.

A latency knob must not double as a kill timeout. Either derive the count from a duration, which keeps the constant honest under any interval:

```go
const taskFailureWindow = 5 * time.Minute
budget := max(3, int(taskFailureWindow/p.interval))
```

or record the first failure time per task instead of a count and compare wall clock. Either is a few lines and removes the coupling entirely.

**On whether five minutes can end a healthy run at all, since the brief asks it directly. I walked every producer I could find, and the answer is: rarely, and every error in the derivation leans the safe way, but not never.**

- Deploy restart, which is the common event: not a budget case at all. The launcher comes back with an empty registry and answers 404, so `ErrEngineSessionGone` fires on the first pass after it is up, long before 20. And `systemctl --user stop` takes the sandbox children down with the unit cgroup, so those tasks genuinely are dead. Correct outcome by the shorter route.
- Launcher down and staying down (a failed reinstall, `Restart=on-failure` looping): dial against the unix socket fails immediately, so 20 fast passes really is about 5 minutes, and every active task in the system is failed at the end of it. The sandboxes died with the unit, so this is mostly right, save for a task that had already finished.
- Slow rather than broken: this one leans safe. There is no per-task timeout on the poll, `Remote` carries a 5 minute client timeout, and the launcher sets no WriteTimeout, so a wedged `GetConversation` stalls the pass instead of charging a failure. Slowness therefore stretches the budget in wall clock rather than shrinking it. `loop` backoff can only lengthen the interval too, never shorten it. Both good.
- The genuine case: socket up, launcher answering, sandbox alive, agent-server control socket dead. Fast 502 every pass, budget exhausts in 5 minutes, and the sandbox is still running. Whether the run is healthy is arguable, since the agent-server is the thing doing the work, but the sandbox is definitely still holding its slot, which is the separate HIGH on `failTask`.

So 20 is a round number rather than a derived one, and I am not asking for a different number. I am asking that it stop being expressible in passes when what it means is time.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -42,6 +107,12 @@ type Poller struct {
 	interval time.Duration
 	logger   *slog.Logger
 
+	// taskFailures counts each task's CONSECUTIVE per-pass failures (see
+	// maxTaskFailureBudget). Only ever touched from RunOnce, which loop
+	// guarantees is never running concurrently with itself (Start/Stop's own
+	// mutex serializes passes), so this needs no lock of its own.
+	taskFailures map[uuid.UUID]int
```
sakibsadmanshajib, line 161: **MEDIUM: `RunOnce` is exported and stateless no longer. The safety argument is written on an unexported field, and it is an argument about `loop`, not about the method.**

The comment says this needs no lock because `loop` never runs `RunOnce` concurrently with itself. That is true of the loop. It is not a property of `RunOnce`, which is exported and callable by anyone holding a `*Poller`, including while `Start` loop is running. Two concurrent passes write this map, and a concurrent map write is a runtime throw, not a recoverable data race: no recover catches it, the process dies, every tenant with it. `Start` and `Stop` own `p.mu`, which does not protect this field at all.

Nothing in production does that today (main.go only calls `Start`, tests only call `RunOnce` on pollers that were never started), so this is a contract regression rather than a live bug. But `RunOnce` used to be safe to call from anywhere and now is not, and the only warning sits on a private field that a caller of the exported method has no reason to read.

Two lazy options, either is fine:

1. One sentence on `RunOnce` own doc comment: not safe for concurrent use, `Start` loop is the only sanctioned caller.
2. A dedicated `passMu sync.Mutex` taken for the whole pass. Do not reuse `p.mu` for this: `Start` would then block for a whole pass, which is harmless but surprising, and it muddles two unrelated invariants.

Related, and worth a line in the same place: the `active` map cleanup at the end of `RunOnce` is driven by the tasks this pass saw, and `agent_tasks_list_active()` is `LIMIT 500`. Above 500 active tasks, rows outside the window have their streaks dropped every pass. That only ever resets a budget, never shortens one, so the direction is safe and I am not asking for a change. It is worth one clause in the comment so the next reader does not have to re-derive that it is safe.

apps/control-plane/internal/agenttask/poller.go
```diff
@@ -22,6 +24,69 @@ type StatusChecker interface {
...
+// implementation could return.
+const maxTaskFailureBudget = 20
+
+// engineUnreachableAfterRetriesMessage is the customer-visible
+// error_message persisted when a task exhausts maxTaskFailureBudget without
+// ErrEngineSessionGone ever firing. Provider-blind and worded the same
+// unrecoverable-outcome way as engineSessionGoneMessage, for the same
+// reason: repeated Status failures mean this code cannot learn what
+// happened, not that the task is confirmed to have failed.
+const engineUnreachableAfterRetriesMessage = "this task's outcome could not be recovered: its session stopped answering"
```
sakibsadmanshajib, line 111: **Wording: judged as asked, and it now matches what is actually known. Two residuals, neither blocking.**

Both messages are honest in a way the previous `agent task session is no longer reachable` was not. The reasoning in the doc comment is right: on the session-gone path the sandbox may well have finished the work before the launcher lost the reference to it, so asserting failure states something this process cannot know, and `outcome could not be recovered` states exactly what it does know. Splitting into two messages (`its session was lost` for the definitive case, `its session stopped answering` for the budget case) also distinguishes the two paths for an operator without a string match on a shared sentence, which was the other half of my earlier comment. Good.

Residual one: the row is still persisted as `status = failed`. So the honest wording lives in the field a human reads while the field a machine reads still asserts failure, and every consumer (`Handler.handleGet`, the console, any future metric) counts these as failures. If it ever matters to tell an unrecoverable outcome from a confirmed failure programmatically, that is a distinct terminal status or a reason code, and it is still not built. Fine to defer, worth naming in the doc comment next to the wording rationale so the next person knows the split is deliberate and only half done.

Residual two: `its session stopped answering` reads as an invitation to retry, and on the budget path a retry is not yet safe. If the sandbox is still alive and still holding one of the user two concurrency slots (see the `failTask` comment), an immediate retry can hit `QuotaUserConcurrency` and fail for a second, unrelated-looking reason. Once the stop lands on that path, the wording is fine as is.

sakibsadmanshajib:
## Adversarial review, stage 6, second pass on `caf4088`

Streams: one adversarial Go pass (this one), reading the pushed diff and the surrounding code at `caf4088`. **SKIPPED again, and absence is still not a pass:** CodeRabbit (check reports Review rate limited), Codex connector (usage limit, its comment above stands). One reviewer on this PR.

**Verdict on the redesign: sound. The mechanism I asked for is the mechanism that shipped, and it is better than what I would have accepted.** Splitting the loop backoff signal (ListActive only) from a per-task budget is the right decomposition, and it closes the class rather than the instance: the 502 dead-sandbox case I found is now covered by construction rather than by enumerating error shapes, which is the part I would have got wrong if I had specified the fix myself. All eight earlier threads are genuinely addressed, not papered over. Two of them (the `/launch` scoping and the misleading leaked-slot warning) were fixed with the exact minimal change and a test each.

The budget introduces one new problem, which is the only thing I am asking for before merge.

### New findings

| Severity | Where | What |
| --- | --- | --- |
| HIGH | [failTask](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792225576) | Budget exhaustion fails the row without stopping the session. A non-404 error means the launcher still holds that session and its quota slot, and the now-terminal row makes `Service.Cancel` return `ErrTerminalState`, so nothing can ever free it. That is #886 symptom through a new door, at 2 slots per user. `buildAgentEngine` returns the same object as both `Engine` and `StatusChecker`, so a best-effort `Cancel` on the budget path only is about six lines |
| MEDIUM | [maxTaskFailureBudget](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792226359) | The budget is counted in passes but documented in minutes, and the interval is `HIVE_AGENT_TASK_POLL_INTERVAL`. Lowering that knob, the obvious response to this very incident, silently shortens a task kill timeout: 5s makes it 100 seconds against 16 to 22 minute tasks |
| MEDIUM | [taskFailures](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792226915) | `RunOnce` is exported and now mutates an unsynchronized map. The no-lock argument is about `loop`, not about the method, and a concurrent map write is a process-fatal throw. No live caller today, so a contract regression rather than a bug |
| Note | [wording](https://github.com/sakibsadmanshajib/hive/pull/914#discussion_r3792227251) | The unrecoverable-outcome framing is correct and matches what is known. Residuals: the row still says `status = failed` to every machine reader, and the budget wording invites a retry that is not safe until the slot stop lands |

### The five questions, answered directly

1. **Can the budget end a healthy run?** Rarely, and every error in the derivation leans safe, but not never. Deploy restarts do not reach the budget at all (404 fires first, and the unit cgroup kill means those sandboxes really are dead). Slowness stalls the pass rather than charging it (no per-task timeout, 5 minute client timeout, no launcher WriteTimeout), so it stretches the budget in wall clock. `loop` backoff can only lengthen the interval. The one real case is socket up, sandbox alive, agent-server control socket dead: fast 502 every pass, killed at 5 minutes, sandbox still running. Whether that run is healthy is arguable; that its slot leaks is not. Five minutes is a round number rather than a derived one, and I am fine with the number. I am not fine with it being expressed in passes when it means time. Full walkthrough in the inline comment.
2. **Per task, and does it reset?** Yes on all three counts. Keyed by `uuid.UUID`, so no cross-task bleed. Reset by `delete` on any clean Status answer, including a plain running, which the reset test pins. Cannot leak in the dangerous direction: a poller restart clears the map (budget restarts, never shortens), and the `LIMIT 500` window can only drop streaks, never inherit one. The only defect here is the missing synchronization on the exported entry point, above.
3. **Does ListActive-only backoff still protect anything?** Yes, and I checked the failure mode it stopped protecting. The thing backoff existed for, a struggling database, still triggers it through `ListActive`. A wholly unreachable launcher now produces no backoff, and that is fine: the cost is one unix socket dial per active task per interval, capped at 500 by the migration LIMIT, each failing immediately, and it self-extinguishes within the budget as the tasks come off the active list. Worth stating plainly that this self-extinguishing property and the mass kill in question 1 are the same event seen from two sides: a five minute engine outage now ends every in-flight task everywhere. Given the launcher restart already kills the sandboxes, that is the truthful outcome, not a new harm.
4. **Is the sentinel scoping complete?** Yes. `post` is the only 404 mapper and is now scoped to `/status` and `/cancel`; `/launch` is excluded and pinned by a negative test; `Health` builds its own request and returns a plain error for any non-200. On the in-process arm, `Engine.Status` and `Engine.Cancel` map `engineapi.ErrUnknownSession` and nothing else does. One note, not a finding: `SandboxEngine.lookup` also returns `ErrUnknownSession` for a session ref that fails `uuid.Parse`, so on that arm a malformed ref would read as session gone rather than as a transient. Not reachable today (refs are written from `convo.ID.String()`, and `ListActive` requires non-empty), and the socket arm the box actually runs is unaffected.
5. **Do the new tests discriminate?** Mostly yes, and the one that did not last time was rewritten into one that does. Against the pre-change code on main:

| Test | Red pre-change | Load-bearing fixture |
| --- | --- | --- |
| `EngineErrorIsRetriedNotFailed` | Yes | mixed pass, one healthy task |
| `AllTasksErroringNeverBacksOffEither` | **Yes**, now asserts the inverse. This is the one that used to pass either way | none |
| `ConsecutiveFailures_OnePermanentlyFailingTask...` | Yes | healthy companion, and 10 passes deliberately under the budget. Both now stated in the comment |
| `SingleTaskErroringForeverDoesNotBackOffTheSweep` | Yes, and red against the interim `errCount == len(tasks)` too | 10 passes under the budget, stated |
| `TaskExceedingFailureBudgetFailsRegardlessOfErrorShape` | Yes | `const budget = 20` duplicated by hand. A change to the constant fails this loudly rather than silently, which is acceptable, but `poller_internal_test.go` already exists and could read it directly |
| `FailureBudgetResetsOnCleanPass` | **No**, and that is correct: it guards the reset semantics against a lifetime-total implementation, not against the old bug. Worth one clause in its comment saying so |
| `Remote` `/launch` 404 and transport-failure negatives | Yes | neither existed before; the transport one is the guard for the invalid_api_key failure shape |
| `Service_Cancel_EngineSessionGoneDoesNotWarnAboutALeakedSlot` | Yes | mutates the global `slog` default and restores it in `t.Cleanup`; safe here because the only parallel test in the package uses its own logger |

### What I need before merge

The slot stop on the budget path. The other two are cheap and I would take them in the same push, but they are not what blocks. Everything else in the redesign I would sign off as is.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Fixed exactly as you described. chargeFailureBudget now type-asserts p.checker against a new unexported budgetExceededCanceler interface (Cancel(ctx, sessionRef) error) and best-effort stops the session before calling failTask, only on the budget-exhaustion path, never on ErrEngineSessionGone (nothing to stop there, as you said). buildAgentEngine's return remote, remote / return real, real already satisfy it on both arms, no wiring change needed elsewhere. New assertions: TestPoller_RunOnce_TaskExceedingFailureBudgetFailsRegardlessOfErrorShape now checks Cancel("session-dead-sandbox") fires exactly once, only at exhaustion (not on any earlier retry pass), and TestPoller_RunOnce_EngineSessionGoneFailsTaskInstead now asserts Cancel is never called on that path. Commit 574200ee.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Took the derive-from-a-duration option. maxTaskFailureBudget (a fixed 20) is now maxTaskFailureDuration (5 minutes), converted to a pass count by a new Poller.taskFailureBudget() method using the poller's OWN configured interval (min 1), so the ceiling stays ~5 minutes regardless of how HIVE_AGENT_TASK_POLL_INTERVAL is tuned. At the unchanged default 15s interval this is still exactly 20, so no existing test's numbers moved. New test: TestPoller_RunOnce_FailureBudgetScalesWithConfiguredInterval, a 1 minute configured interval yields a 5-pass budget rather than 20. Appreciate the full walk of every producer on the 'can 5 minutes end a healthy run' question -- agreed the genuine case is the one your HIGH comment covers (socket up, sandbox alive, control socket dead), and that one is now fixed by the best-effort Cancel landing alongside this. Commit 574200ee.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Took option 2. Added a dedicated taskFailuresMu (not reusing p.mu, agreed that would muddle two unrelated invariants and make Start block for a whole pass) guarding every access to taskFailures via four small helper methods: taskFailureCount, incrementTaskFailure, clearTaskFailure, pruneInactiveTaskFailures. The struct field comment no longer claims safety from loop's serialization; it now states plainly that RunOnce is exported and the map needs its own lock regardless of caller. New test: TestPoller_RunOnce_ConcurrentCallsDoNotCorruptTaskFailures, 20 goroutines calling RunOnce concurrently against a poller with failing tasks -- reproduces the fatal concurrent-map-write crash on the prior code, passes clean on this one, no -race needed since Go's runtime catches concurrent map writes on its own. Also added your LIMIT 500 safety clause to pruneInactiveTaskFailures's doc comment: above 500 active tasks, rows outside agent_tasks_list_active()'s window have their streaks dropped every pass, which only ever resets a budget, never shortens one. Commit 574200ee.

apps/control-plane/internal/agenttask/poller.go
sakibsadmanshajib: Thanks for confirming the wording lands correctly. Residual two is now resolved as a side effect of the HIGH fix in this same round: once the budget path actually stops the session before failing the row, 'its session stopped answering' no longer invites an unsafe retry, since there is no longer a live sandbox holding a slot behind it. Residual one (status column still says failed regardless of which message applies) is named explicitly now: added a paragraph to engineSessionGoneMessage's doc comment stating the split is deliberate and only half done, and pointing at issue #921, which I've updated to also track the reason-code / distinct-status design question alongside the drain/reconnect one. Not building the schema change here, agreed that's a separate, larger design. Commit 70975837 (docs) plus 574200ee (the Cancel fix this depends on).

Issue #915, opened 2026-08-16
API key expiry is derived on read: a row can say active while the console says expired, and the resolver's behaviour is unestablished


sakibsadmanshajib:
## What was observed

While enumerating automation litter on the shared demo account
(`demo@hive-demo.invalid`) on 2026-08-16, six API keys nicknamed
`interaction gate probe` were found in this state:

| key id (prefix) | created | `status` column | `expires_at` column | `last_used_at` |
| --- | --- | --- | --- | --- |
| `b21cbd63` | 2026-08-10T06:02:31Z | `active` | 2026-01-01T00:00:00Z | null |
| `9412444f` | 2026-08-10T13:50:05Z | `active` | 2026-01-01T00:00:00Z | null |
| `07ec917f` | 2026-08-10T14:29:21Z | `active` | 2026-01-01T00:00:00Z | null |
| `8e58c70f` | 2026-08-10T15:26:17Z | `active` | 2026-01-01T00:00:00Z | null |
| `8fec407a` | 2026-08-10T20:10:06Z | `active` | 2026-01-01T00:00:00Z | null |
| `566e10b0` | 2026-08-10T21:29:30Z | `active` | 2026-01-01T00:00:00Z | null |

The expiry is in the past, so the console describes these keys as expired,
while the stored record describes them as live. Those six rows have since been
revoked through the console's own revoke path, so this issue is not about them
specifically. It is about the shape, which any suite that mints a key with a
past expiry will reproduce.

## Why this is a question rather than a defect report

Expiry is applied on read, not written into the row:

`apps/control-plane/internal/apikeys/service.go:619`

```go
// applyExpiry returns the key with status set to Expired when the stored
// status is active but expires_at is in the past. No mutation to sibling keys.
func applyExpiry(k APIKey, now time.Time) APIKey {
	if k.Status == KeyStatusActive && k.ExpiresAt != nil && k.ExpiresAt.Before(now) {
		k.Status = KeyStatusExpired
	}
	return k
}
```

That derivation is applied at seven call sites in the same file (lines 65, 77,
160, 230, 264, 365, 509), all of them account-surface reads: list, get, and the
state-transition guards. Nothing was found that establishes the same rule on
the path that resolves a presented `hk_*` bearer into an authorization
decision.

## What I would like answered explicitly

1. When a request presents an `hk_*` secret whose row has `status = 'active'`
   and `expires_at` in the past, is the request authorized or rejected? Name
[...]

Issue #113, opened 2026-04-25
[P0/critical] Auth snapshot 1hr cache lets revoked keys keep working


sakibsadmanshajib:
## Severity
**Critical.** Surfaces in adversarial Go review (C-4).

## Problem
Edge-api caches the API-key auth snapshot in Redis for ~1 hour. Revocation deletes the key in Postgres but does not invalidate the cache, and the cache writes are fire-and-forget so they can race with revocation deletes. A revoked key remains usable for up to 1 hour.

## Fix
- On revocation, control-plane must publish an invalidation event that edge-api consumes (Redis pub/sub or shorter TTL ≤60s).
- Or use a versioned cache key (`apikey:{hash}:v{generation}`) and bump generation on revoke.

## Acceptance
- Revoked key returns 401 within ≤60 seconds across all edge-api instances.

PR #919, opened 2026-08-16
test: pin API key expiry enforcement at the authorization entry point

sakibsadmanshajib:
## Answer to issue #915

**An expired-but-not-revoked key does not authorize. A revoked key does not authorize.** The premise of #915 is wrong on one specific point, and that is worth stating plainly because it is the whole reason the gap looked real: of the seven `applyExpiry` call sites listed in the issue, **line 365 is not an account-surface read**. It sits inside `ResolveSnapshot`, which is the resolver the issue said it could not find.

### The path, end to end

A `Bearer hk_*` credential arrives at edge-api and is routed to the API-key path by `auth.Selector` (`apps/edge-api/internal/auth/selector.go:25`), which only chooses a handler and validates nothing. Every hk_ surface (inference, RAG, files, batches, images, audio) reaches the same entry point: `authz.Authorizer.Authorize` (`apps/edge-api/internal/authz/authorizer.go:112`). That hashes the raw token with SHA-256 (`client.go:223`), reads `auth:key:{hash}` from Redis, and on a miss POSTs the hash to control-plane's `/internal/apikeys/resolve`, which is authenticated by edge-api's own internal token and never by the customer's key.

Control-plane answers from `apikeys.Service.ResolveSnapshot` (`apps/control-plane/internal/apikeys/service.go:359`). The row is fetched by token hash with no status predicate in the SQL (`repository.go:525`), so an expired, disabled or revoked key is found rather than 404ing, and the verdict is carried in the snapshot instead. Then the snapshot is denied on either of two independent predicates.

### The deciding lines

| Question | Verdict | Deciding line |
| --- | --- | --- |
| Does an expired-but-not-revoked key authorize? | **No** | `apps/control-plane/internal/apikeys/service.go:365` derives `status: "expired"`, denied at `apps/edge-api/internal/authz/authz.go:146`. Independently, `apps/edge-api/internal/authz/authz.go:155` re-parses `expires_at` against the current time and denies regardless of the status it was handed. |
| Does a revoked key authorize? | **No** | `apps/edge-api/internal/authz/authz.go:146`. `RevokeKey` writes `status = 'revoked'` durably (`service.go:230` onward), the resolver projects it verbatim, and `CheckAccess` denies anything that is not exactly `active`. Already pinned by `TestCheckAccessRevokedKey` and `TestAuthorizeLogsDeniedReasonForRevokedKey`. |
| Is expiry enforcement missing, or deliberate and elsewhere? | **Deliberate, in two places** | Not missing. `applyExpiry` is called inside the resolver itself, and edge-api re-derives the same rule from the raw timestamp. Neither layer depends on the durable `status` column ever being rewritten. |

### Can a cached resolution outlive an expiry or a revocation?

Expiry, no. `CheckAccess` step 2 (`authz.go:155`) compares `expires_at` to `time.Now()` on every request, so a snapshot cached while the key was live is still refused the moment the key expires.

Revocation, bounded. A revoked key has no `expires_at` to catch it, so it relies on the cache being cleared. Control-plane DELETEs `auth:key:{hash}` on revoke, disable, enable, rotate, and policy or limit updates, and `snapshotTTL` (`apps/edge-api/internal/authz/client.go:101`, 60s) is the deliberate backstop if that DELETE is missed. Documented in that comment and in issue #113. Not changed here.

## What this PR adds

Both layers already had a unit test. Neither pinned the **composition**, and the composition is exactly what the incident exercised: a durable `status` column still reading `active` while `expires_at` had passed. That is also what a cached snapshot looks like for the rest of its TTL after the key expires.

Two tests at `Authorizer.Authorize`, the entry point that actually decides:

- `TestAuthorizeDeniesExpiredKeyWhoseStoredStatusIsStillActive` drives the resolver result in the incident shape.
- `TestCachedSnapshotCannotOutliveKeyExpiry` drives a cached snapshot through the real cache path with an unroutable control-plane URL, so a cache miss would surface as `upstream_unavailable` and fail the assertion rather than passing for the wrong reason.

Both assert the specific deny message, not just the `invalid_api_key` code, so no other denial branch can satisfy them.

## Verification

Both tests were confirmed able to go red, not merely observed green:

- Disabling the edge check (`authz.go:155`): both new tests fail with `expired key was authorized: Authorize returned no error`.
- Disabling `applyExpiry` (`service.go:621`): the existing control-plane tests fail with `expected expired snapshot status, got active`.
- Restored, full suite green: `go test ./apps/edge-api/internal/authz/... ./apps/control-plane/internal/apikeys/... -count=1 -short`, plus `go vet` and `gofmt` clean on the new file.

## Behaviour change

**None, and none proposed.** The gap #915 suspected does not exist, so there is nothing to turn on and no live integration at risk. If the owner wants the durable `status` column reconciled to `expired` by a job, that is a separate cosmetic change with no authorization effect, and it is the owner's call rather than mine.

## Known bound, not fixed here

`OWUI_SHIM_KEY` is compared as a static string in the OWUI unwrap middleware (`apps/edge-api/internal/auth/owui_unwrap.go`), so expiring or revoking that key does not by itself disable the unwrap path. This is not an authorization bypass: the unwrap swaps in a per-user Supabase JWT that is then validated normally, and the shim key's own health probe does check status. Flagged only so nobody treats revoking the shim key as a kill switch.

## Buglog entry

Not applicable. No bug was fixed: the investigation established that the suspected defect does not exist, and the diff is test-only.

Refs #915

### DIFF
--- a/apps/edge-api/internal/authz/expiry_regression_test.go
+++ b/apps/edge-api/internal/authz/expiry_regression_test.go
@@ -1,0 +1,124 @@
+package authz
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+	"time"
+
+	apierrors "github.com/sakibsadmanshajib/hive/apps/edge-api/internal/errors"
+)
+
+// Issue #915. These two tests pin the answer to "does presenting an expired
+// Hive API key authorize a request?" at the entry point that actually decides
+// it, Authorizer.Authorize, rather than at either half of the check in
+// isolation.
+//
+// Expiry is enforced twice, independently, and each layer already has its own
+// unit test:
+//
+//   - control-plane derives the status: apikeys.Service.ResolveSnapshot calls
+//     applyExpiry before projecting the snapshot, so a key whose durable
+//     api_keys.status is still 'active' resolves as "expired"
+//     (TestResolveSnapshotReturnsExpiredWhenExpiresAtHasPassed).
+//   - edge-api re-checks the timestamp: CheckAccess step 2 compares
+//     snapshot.ExpiresAt against the current time regardless of the status it
+//     was handed (TestCheckAccessExpiredKey).
+//
+// What neither of those pins is the composition, which is what the incident
+// behind #915 was actually about: six keys whose stored status read 'active'
+// while expires_at had passed. Both tests below use exactly that shape, so
+// removing either layer turns one of them red.
+//
+// The shape is also what a stale cached snapshot looks like. The edge caches a
+// resolved snapshot in Redis for up to snapshotTTL, so a snapshot minted while
+// the key was still live keeps its "active" status for the rest of that TTL
+// after the key expires. The second test drives that path through the cache
+// itself.
+
+// pastExpiryActiveSnapshot returns the incident shape: a key whose durable
+// status is still active, with expires_at an hour in the past.
+func pastExpiryActiveSnapshot() AuthSnapshot {
+	past := time.Now().Add(-1 * time.Hour).UTC().Format(time.RFC3339)
+	return AuthSnapshot{
+		KeyID:          "key-1",
+		AccountID:      "acc-1",
+		TenantID:       "11111111-1111-1111-1111-111111111111",
+		Status:         "active",
+		ExpiresAt:      &past,
+		AllowAllModels: true,
+		BudgetKind:     "none",
+	}
+}
+
+// assertExpiredDenial fails unless err is the specific expired-key denial.
+// Asserting the message, not just the code, matters: a resolve failure or any
+// other invalid_api_key branch would satisfy a code-only assertion and let the
+// test pass for the wrong reason.
+func assertExpiredDenial(t *testing.T, err *apierrors.OpenAIError) {
+	t.Helper()
+
+	if err == nil {
+		t.Fatal("expired key was authorized: Authorize returned no error")
+	}
+	if err.Error.Code == nil || *err.Error.Code != "invalid_api_key" {
+		t.Fatalf("code = %v, want %q", err.Error.Code, "invalid_api_key")
+	}
+	const want = "Incorrect API key provided: API key has expired"
+	if err.Error.Message != want {
+		t.Fatalf("message = %q, want %q", err.Error.Message, want)
+	}
+}
+
+// TestAuthorizeDeniesExpiredKeyWhoseStoredStatusIsStillActive is the direct
+// answer to #915: an expired-but-not-revoked key does not authorize, and the
+// denial does not depend on the durable status column having been rewritten.
+func TestAuthorizeDeniesExpiredKeyWhoseStoredStatusIsStillActive(t *testing.T) {
+	client := &Client{
+		ResolveOverride: func(_ context.Context, _ string) (AuthSnapshot, error) {
+			return pastExpiryActiveSnapshot(), nil
+		},
+	}
+
+	_, _, err := NewAuthorizer(client, nil).
+		Authorize(context.Background(), "Bearer hk_expired", "hive-default", 50, 100, 20)
+
+	assertExpiredDenial(t, err)
+}
+
+// TestCachedSnapshotCannotOutliveKeyExpiry covers the cache half of the same
+// question. A snapshot cached while the key was still live carries
+// status "active" for the rest of snapshotTTL, and the cache hit returns
+// before any control-plane call, so the only thing standing between that
+// cached value and a served request is CheckAccess re-reading expires_at.
+//
+// baseURL is deliberately unroutable: if the cache were missed, the fallback
+// HTTP resolve would fail and produce an upstream_unavailable error instead,
+// which assertExpiredDenial rejects. So this cannot pass without the cache
+// having actually been read and the expiry having actually been enforced on it.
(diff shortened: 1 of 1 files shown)


sakibsadmanshajib:
## Stage 6 adversarial review

Stream status, stated honestly rather than implied by silence:

| Stream | Status |
| --- | --- |
| CodeRabbit CLI | **SKIPPED** — `Rate limit exceeded`, all 3 included reviews used on the Free plan, no seat assigned for usage-based reviews. Not a clean pass. |
| `/codex:adversarial-review` | **SKIPPED** — Codex CLI returned `You've hit your usage limit`. Not a clean pass. |
| Mandatory security review (auth path) | **RAN** — findings below. |
| Adversarial falsification pass | **RAN** — findings below. |

Both automated streams were unavailable, so the two that ran were performed by the same agent that wrote the diff. That is weaker than independent review and is recorded as such. Given the diff is test-only and adds no production code path, the residual risk of that overlap is low, but it should not be logged as if four streams passed.

### Falsification pass: can the PR's central claim be broken?

The claim under attack: no `hk_*` credential reaches a served request without passing `authz.CheckAccess`, so expiry cannot be bypassed.

Every `/v1` route registered in `apps/edge-api/cmd/server/main.go` was walked to its authorization predicate:

- `/v1/chat/completions`, `/v1/completions`, `/v1/responses`, `/v1/embeddings`, `/v1/rag/chat` reach `inference.Orchestrator`, which calls `Authorize` at `orchestrator.go:84`, `stream.go:107`, `stream_responses.go:49`.
- `/v1/images/*`, `/v1/audio/*`, `/v1/files*`, `/v1/uploads*`, `/v1/batches*` each go through a thin `authz_adapter.go` whose only job is to call the same `Authorize` (`images:22`, `audio:22`, `files:24`, `batches:22`).
- `/v1/models` uses `handleModels(catalogClient, authorizer)`; `/v1/messages*` delegates to the same `openAIChatHandler`, so it inherits the inference path rather than adding one.
- `/v1/featuregate` is JWT-only: it requires `auth.UserFrom` (`featuregate/handler.go:57`) and denies otherwise, and an `hk_` request never populates that context. No expiry question arises because no API key authenticates it.

Two call sites resolve a snapshot **without** `CheckAccess`, and both were checked specifically because that is the shape a bypass would take:

- `main.go:939`, the BudgetGate workspace resolver. It reads `AccountID` only, to pick which budget counter applies, and grants nothing; the per-route `Authorize` still runs afterwards.
- `main.go:1069`, `checkOWUIShimKey`. A health probe on a timer, not a request path. It reads `Status`, which the resolver has already derived to `expired` where applicable.

Control-plane exposes no second `hk_` authenticator: `GetByTokenHash` (`apikeys/repository.go:525`) has exactly one caller, `GetPolicyByTokenHash`, which has exactly one caller, `ResolveSnapshot`. Every other control-plane route authenticates a Supabase JWT via `auth.Middleware.Require`. The claim survives.

### Security finding — LOW, not fixed here

`apps/edge-api/internal/authz/authz.go:157`

```go
exp, err := time.Parse(time.RFC3339, *snapshot.ExpiresAt)
if err == nil && exp.Before(time.Now()) {
```

An `expires_at` that fails to parse silently **skips** the expiry check rather than denying. The request then falls through to the status check, and a snapshot carrying `status: "active"` would be authorized.

Reachability is genuinely low: the only producer is control-plane marshalling a `*time.Time`, which emits RFC3339Nano, and `time.Parse(time.RFC3339, ...)` accepts that. Constructing an unparseable value means writing directly to the shared Redis, which is already a full compromise of the auth cache. So this is defence in depth, not a live hole.

The fix is one line (deny when `err != nil`), but it is a fail-closed change on an authorization path and the owner explicitly reserved behaviour changes here. Flagged for that decision rather than slipped into a test-only PR.

### Observation, no action

`OWUI_SHIM_KEY` is matched as a static string in `apps/edge-api/internal/auth/owui_unwrap.go`, so expiring or revoking that key does not disable the unwrap path by itself. Not a bypass, since the unwrap substitutes a per-user Supabase JWT that is then validated normally. Recorded only so nobody later treats revoking the shim key as a kill switch.

### On the tests themselves

Checked for the false-green shapes that would let them pass while the enforcement is broken:

- Neither asserts only on `invalid_api_key`. Both require the exact message `Incorrect API key provided: API key has expired`, so a resolve failure, a budget denial, or a `model_not_allowed` cannot satisfy them. This matters for the cached test in particular, where a cache miss would otherwise produce `upstream_unavailable` and look like a pass.
- `TestCachedSnapshotCannotOutliveKeyExpiry` additionally asserts the cache was actually read, so the test cannot quietly stop exercising the path it is named for.
- Both were confirmed to fail under mutation, not merely observed green. Disabling the edge check turns both red with `expired key was authorized`; disabling `applyExpiry` turns the existing control-plane tests red with `got active`.

One honest limitation: `TestAuthorizeDeniesExpiredKeyWhoseStoredStatusIsStillActive` uses `ResolveOverride`, so it pins `CheckAccess` reached through `Authorize` and does not exercise the real resolve. The control-plane half is pinned separately by `TestResolveSnapshotReturnsExpiredWhenExpiresAtHasPassed`. No single test covers both processes at once, which would need an integration harness this PR does not add.

PR #920, opened 2026-08-16
feat: stream token deltas from every launched agent sandbox

sakibsadmanshajib:
## What

Sets `llm.stream` on every conversation the agent engine launches, so the
vendored agent-server actually publishes token deltas.

## Why

`event_service.py` computes `streaming_enabled = isinstance(agent, ACPAgent) or
any(llm.stream for llm in agent.get_all_llms())` and only then wires the token
callback that publishes `StreamingDeltaEvent`. `openhands/sdk/llm/llm.py`
defaults `stream` to `False`, and `controlclient.LLMSettings` sent `model`,
`base_url`, `api_key` and `usage_id` and nothing else. So every sandbox Hive has
launched so far ran with token streaming off, and no delta was ever published
for any conversation.

This is wave one of the chat-native agent work
(`research-2026-08-16-chat-native-agent-ux.md`, Part 2 claim 3). Everything
later depends on deltas existing, so it is deliberately the smallest
independently verifiable step.

## Scope, and what is deliberately not here

Not touched, and named so their absence is not read as an oversight:

- `enable_sub_agents` (defaults false, gates `task_tool_set`). Turning it on
  opens an agent-to-agent boundary we do not defend: upstream Claude Code added
  an instruction-shaped-pattern scan on a subagent's final message at exactly
  that boundary, and we have no equivalent. Needs its own security work first.
- `enable_browser` (defaults false, "injected by the serving layer" upstream,
  and we do not inject it).
- `enable_switch_llm_tool` (defaults **true** upstream, so the built-in
  `switch_llm` tool is already available in our sandboxes; noted here because it
  is the one default-on switch in that group and nobody chose it).
- No launcher relay, no SSE endpoint, no change to `serve.go`.

## Proof

Deltas are transient by design: they are published straight to PubSub and never
written to `ConversationState.events`, so a run that streamed and a run that did
not are indistinguishable from the event history afterwards. Only a mid-run
capture is evidence.

`apps/agent-engine/cmd/streamproof` (behind a `proof` build tag, so no ordinary
build, vet or test run compiles it) launches a real Apptainer sandbox through
the same `engineapi.Launch` the host launcher's `POST /launch` handler calls,
subscribes to the conversation's event WebSocket over the session control
socket, and counts frames while the conversation is still running. It differs
from the daemon in two ways, neither of which can affect whether a delta is
published: `ResolveEgressHosts` asks control-plane for a tenant's egress policy
there and returns the model host directly here, which is strictly narrower; and
`MemoryLimit` is 2G here against `serve.go`'s 4G, so a heavier prompt could be
killed for memory in the harness and not in production. Every other config field
normalises to the production value inside `engine.New`.

It runs a differential, because a capture never observed to fail proves nothing:
a second conversation on the same sandbox, same model and same prompt with
`stream` explicitly false. The program fails unless the product launch produced
deltas and the control produced none.

`.github/workflows/agent-stream-delta-proof.yml` runs it on a hosted runner with
real Apptainer, the CI-built SIF correlated by image inputs, and the live
gateway, and posts the captured log as a comment on this pull request. The
development box has neither Apptainer nor root and the demo box runs only
merged code, so this is the only place the claim can be tested before merge.

## Metering

Checked, because streaming changes how a response arrives and this repository
has already shipped a billing defect that served requests free for three days.

The agent's model traffic is ordinary gateway traffic on `/v1/chat/completions`.
The vendored SDK sets `stream_options={"include_usage": True}` whenever it
streams (`llm.py`, `_prepare_transport_kwargs`), so the terminal usage frame
still arrives, `UsageAccumulator.HasUsage` is true, and `settlementCredits`
takes its confirmed branch and prices the response through the same catalog
conversion a buffered response would. Streaming does not move an agent request
onto the unconfirmed content estimate.

Two facts worth recording rather than assuming:

1. The confirmed branch depends on that `include_usage`, not on our own code:
   `executeStreaming` forwards the caller's body upstream as it is, and
   `includeUsage` there only controls whether edge-api synthesises a terminal
   usage chunk for the caller. If a future agent-side client stopped asking for
   usage, agent traffic would settle from a content estimate flagged unconfirmed
   rather than from measured tokens. It would still be charged, and the estimate
   is calibrated to land under real tokenisation, so the direction of error
   favours the customer.
2. The chat path already forces streaming upstream regardless of what the client
   asked for (`rewriteDispatchBody`), so a streamed response is the ordinary,
   most-exercised metering case on this gateway, not a new one.

## Test plan

- `go test ./apps/agent-engine/internal/engine/... ./apps/agent-engine/internal/controlclient/...`
  covers the launch payload, asserting both the decoded field and `"stream":true`
  in the raw body the agent-server receives (the server reads the JSON name, not
  the Go field). Verified to fail with the flag flipped back to false.
- `go vet -tags proof ./apps/agent-engine/...` compiles the proof command.
- The live capture above is the real acceptance criterion.

## Live evidence

Run [`32010112621`](https://github.com/sakibsadmanshajib/hive/actions/runs/32010112621),
on this head, hosted runner, real Apptainer sandbox from the CI-built SIF, live
gateway. Full log in the comment below.

```
A/product(stream=true): frames{ConversationStateUpdateEvent=5 MessageEvent=1 StreamingDeltaEvent=5}
                        first delta 08:25:05.976Z, last delta 08:25:06.007Z, terminal (finished) observed 08:25:06.660Z
B/control(stream=false): frames{ConversationStateUpdateEvent=5 MessageEvent=1}; no deltas
PROOF OK: 5 delta frames on the product launch, 0 on the stream=false control
```

Three things that log establishes, none of which a unit test could:

1. Deltas exist and carry token text (`"content":"Hive"`, `" streaming"`,
   `" proof"`), captured over the session control socket while the conversation
   was still running: the last delta lands 653 ms before the poll that first saw
   `finished`.
2. The same sandbox, same model and same prompt with `stream` false produced
   zero delta frames while producing the same other frame kinds, so the capture
   is measuring the flag and not the weather.
3. Tool calling still **works** under `stream: true`, which is the real
   regression risk of turning streaming on, since the reassembly of a tool call
   out of streamed chunks happens inside the SDK and either produces a valid
   call or does not. That is captured in run
   [`31957712375`](https://github.com/sakibsadmanshajib/hive/actions/runs/31957712375),
   posted on this pull request, which recorded `ActionEvent=1`,
   `ObservationEvent=1`, `MessageEvent=1` and five deltas in one turn. No
   capture proves argument-fragment streaming and none can: a
   `StreamingDeltaEvent` only ever carries `content` or `reasoning_content`, so
   the tool phase produces no deltas at all.

The prompt is prose only, deliberately, and that is a finding rather than a
convenience. An OpenHands agent ends its turn through the finish tool, so a
prompt that also asks for a shell command tends to route the whole answer
through tool calls and emit no assistant message: runs
[`32008631987`](https://github.com/sakibsadmanshajib/hive/actions/runs/32008631987),
[`32009278161`](https://github.com/sakibsadmanshajib/hive/actions/runs/32009278161)
and [`32009716814`](https://github.com/sakibsadmanshajib/hive/actions/runs/32009716814)
each did exactly that. The harness now checks that precondition and ends such a
run as inconclusive, naming the frames it saw, rather than reporting a streaming
failure that did not happen or a pass it cannot justify. Re-rolling that dice on
every future run would spend a live model call for a one-in-four chance of
re-proving a result already in the record.

An earlier prose-only capture, run
[`31957371732`](https://github.com/sakibsadmanshajib/hive/actions/runs/31957371732),
is the third independent confirmation.

## Constraint for wave two, worth knowing before the rendering is designed

A chat surface fed by these deltas alone shows nothing for the entire
tool-calling phase, which is where an agent spends most of its wall clock, and
that phase is the thing the owner actually wants on screen. Token deltas cover
the model's prose only. Tool visibility has to come from the event stream
(`ActionEvent`, `ObservationEvent`, and the `waiting_for_confirmation` status
`engine.go` currently collapses into running), not from the deltas this pull
request switches on.

## Operational note

The sandbox's model requests now carry `stream: true` to the gateway, which
becomes `NeedStreaming` in route selection, and
`control-plane/internal/routing/service.go:198` drops any candidate route that
is not marked streaming capable. An operator pointing
`HIVE_AGENT_ENGINE_LLM_MODEL` at an alias with no streaming capable route would
see agent tasks fail at the first model call where they previously worked
buffered. The blast radius is that one configured alias, and the capture above
proves the configured one selects and streams.

## Review outcome

Eight threads, none a defect in the launch change itself. One was acted on
rather than rebutted, in commit `df2686d7`: the streaming settlement path logged
nothing when it settled from an estimate (`confirmed=false`) or released the hold
having measured nothing, both of which the synchronous path already logs. That
silence is inherited, but this change routes the most tool-call-heavy traffic in
the product through it, and a free-serve outage has already hidden there once.
The two log lines match the synchronous ones and change no behaviour; the guard
test was verified to go red when the line is deleted.

Two findings warranted work rather than a rebuttal and are filed instead of
being built here, so this pull request stays its original size:

- [#925](https://github.com/sakibsadmanshajib/hive/issues/925):
  `credit_reconciliation_jobs` has a writer and no reader, so a charge clamped at
  the hold is never corrected.
- [#926](https://github.com/sakibsadmanshajib/hive/issues/926): agent tasks fail
  totally and undiagnosably if the agent alias loses its streaming-capable route,
  since `supports_streaming` defaults false and the resulting 400 never leaves
  the sandbox.

## Buglog entry

```json
{"id":"BUG-2026-08-16-01","date":"2026-08-16","title":"Token streaming was off in every launched agent sandbox","error_message":"No StreamingDeltaEvent was ever published by a Hive-launched conversation","root_cause":"The vendored agent-server gates its token callback on any LLM having stream set (agent_server/event_service.py streaming_enabled), and openhands/sdk/llm/llm.py defaults LLM.stream to false. apps/agent-engine/internal/controlclient.LLMSettings serialised model, base_url, api_key and usage_id only, so every launch inherited the default and streaming was silently disabled.","fix":"Added Stream to controlclient.LLMSettings (json name stream, no omitempty) and set it true on every inline agent_settings launch in apps/agent-engine/internal/engine. Regression guard asserts the raw launch body carries the JSON name, and a live differential capture (apps/agent-engine/cmd/streamproof) proves deltas flow mid-run and stop when the flag is false.","tags":["agent-engine","openhands","streaming","default-off","vendored-gate"]}
```

### DIFF
--- a/.github/workflows/agent-stream-delta-proof.yml
+++ b/.github/workflows/agent-stream-delta-proof.yml
@@ -1,0 +1,207 @@
+name: agent streaming delta proof
+
+# Live evidence that a launched sandbox publishes token deltas, captured while
+# a conversation is still running.
+#
+# WHY IT CANNOT BE CAPTURED ANYWHERE ELSE
+#
+# StreamingDeltaEvent is published straight to the agent-server's PubSub and is
+# deliberately never persisted to the conversation event log, so reading events
+# after a run cannot tell a streaming conversation from a non-streaming one.
+# The observation has to happen mid-run, against a real Apptainer sandbox
+# talking to a real model. The WSL2 development box has neither Apptainer nor
+# root, and the demo box runs only already-merged code, so a hosted runner is
+# the only place this can be proven before merge. Same reasoning, and most of
+# the same setup, as agent-visual-proof.yml.
+#
+# This job is scaffolding for one change. Delete it once the launcher relay
+# lands and streaming has a permanent test of its own.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'apps/agent-engine/internal/controlclient/**'
+      - 'apps/agent-engine/internal/engine/**'
+      - 'apps/agent-engine/cmd/streamproof/**'
+      - '.github/workflows/agent-stream-delta-proof.yml'
+
+# No concurrency group, for the reason agent-visual-proof.yml records: a
+# cancelled proof run is a silent absence of evidence rather than a loud
+# failure.
+
+env:
+  APPTAINER_VERSION: "1.5.2"
+  RUNTIME_DIR: /mnt/agent-runtime
+  PROOF_LOG: /tmp/stream-delta-proof.log
+
+jobs:
+  prove:
+    name: Capture token deltas from a live sandbox
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    # A fork pull request gets no secrets, so it would fail on the missing
+    # gateway key rather than on the change under test.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.pull_request.head.repo.full_name == github.repository
+    permissions:
+      contents: read
+      actions: read
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.work
+          cache-dependency-path: apps/agent-engine/go.sum
+
+      - name: Install Apptainer and allow unprivileged user namespaces
+        run: |
+          set -euo pipefail
+          deb="apptainer_${APPTAINER_VERSION}_amd64.deb"
+          curl -fsSL -o "/tmp/${deb}" \
+            "https://github.com/apptainer/apptainer/releases/download/v${APPTAINER_VERSION}/${deb}"
+          sudo apt-get update -qq
+          sudo apt-get install -y "/tmp/${deb}"
+          sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0 || true
+          sudo sysctl -w kernel.unprivileged_userns_clone=1 || true
+          apptainer --version
+
+      - name: Start a systemd user session for rootless cgroups
+        # Apptainer enforces the per-session memory and CPU limits through
+        # rootless cgroups, which need a user manager: without XDG_RUNTIME_DIR
+        # and DBUS_SESSION_BUS_ADDRESS every launch dies with "cannot use
+        # cgroups - DBUS_SESSION_BUS_ADDRESS is not set".
+        run: |
+          set -euo pipefail
+          sudo loginctl enable-linger "$(id -un)"
+          runtime="/run/user/$(id -u)"
+          for _ in $(seq 1 30); do [ -S "$runtime/bus" ] && break; sleep 1; done
+          if [ ! -S "$runtime/bus" ]; then
+            echo "::error::no systemd user session bus at $runtime/bus"
+            exit 1
+          fi
+          {
+            echo "XDG_RUNTIME_DIR=$runtime"
+            echo "DBUS_SESSION_BUS_ADDRESS=unix:path=$runtime/bus"
+          } >> "$GITHUB_ENV"
+
+      - name: Fetch the SIF built from these exact image inputs
+        # Correlated by image inputs rather than by commit, the same way
+        # agent-visual-proof.yml does it: the image depends on exactly these
+        # three trees, so an artifact whose run built the same three trees is
+        # the image for this ref. Certifying a change against an unrelated
+        # image would be worse than no proof, because the output carries
+        # authority.
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          sudo mkdir -p "$RUNTIME_DIR"
+          sudo chown "$(id -un):$(id -gn)" "$RUNTIME_DIR"
+
+          inputs_digest() {
+            for path in deploy/apptainer apps/agent-engine/packs vendor/openhands; do
+              git rev-parse "$1:$path" 2>/dev/null || echo "absent:$path"
+            done | sha256sum | cut -d' ' -f1
+          }
+
+          want=$(inputs_digest HEAD)
+          echo "image inputs digest for this ref: $want"
+
+          match=""
+          while read -r id sha; do
+            [ -n "$sha" ] || continue
+            git fetch -q --depth=1 origin "$sha" 2>/dev/null || continue
+            got=$(inputs_digest "$sha")
+            echo "artifact $id from $sha: $got"
+            if [ "$got" = "$want" ]; then match="$id"; break; fi
+          done <<<"$(gh api "repos/$GITHUB_REPOSITORY/actions/artifacts?name=agent-engine-sif&per_page=30" \
+            --jq '.artifacts[] | select(.expired == false) | "\(.id) \(.workflow_run.head_sha)"')"
+
+          if [ -z "$match" ]; then
+            echo "::error::no unexpired agent-engine-sif artifact was built from these image inputs. Build one: gh workflow run \"agent-engine SIF\" --ref <this branch>"
+            exit 1
+          fi
+
+          gh api "repos/$GITHUB_REPOSITORY/actions/artifacts/$match/zip" > /tmp/agent-engine-sif.zip
+          python3 -c 'import sys,zipfile; zipfile.ZipFile(sys.argv[1]).extractall(sys.argv[2])' \
+            /tmp/agent-engine-sif.zip "$RUNTIME_DIR"
+          if [ ! -s "$RUNTIME_DIR/agent-engine.sif" ]; then
+            echo "::error::artifact $match contained no agent-engine.sif"
+            exit 1
+          fi
+          echo "using artifact $match ($(stat -c%s "$RUNTIME_DIR/agent-engine.sif") bytes)"
+
+      - name: Prove Apptainer can run this image on this runner
+        run: |
+          set -euo pipefail
+          apptainer exec "$RUNTIME_DIR/agent-engine.sif" /bin/true
+          echo "rootless apptainer exec of the shipped image: ok"
+
+      - name: Launch a real sandbox and capture its token deltas
+        env:
+          HIVE_AGENT_ENGINE_SIF_PATH: /mnt/agent-runtime/agent-engine.sif
+          HIVE_AGENT_ENGINE_PACKS_DIR: ${{ github.workspace }}/apps/agent-engine/packs
+          HIVE_AGENT_ENGINE_WORKSPACE_ROOT: /mnt/agent-runtime/workspaces
+          HIVE_AGENT_ENGINE_RUN_DIR: /mnt/agent-runtime/run
+          HIVE_AGENT_ENGINE_LLM_MODEL: ${{ vars.HIVE_AGENT_ENGINE_LLM_MODEL || 'openai/hive-default' }}
+          HIVE_AGENT_ENGINE_LLM_BASE_URL: ${{ vars.HIVE_AGENT_ENGINE_LLM_BASE_URL || 'https://api-hive.scubed.co/v1' }}
+          HIVE_AGENT_ENGINE_LLM_API_KEY: ${{ secrets.HIVE_API_KEY }}
+          HIVE_PROOF_LOG: ${{ env.PROOF_LOG }}
+        run: go run -tags proof ./apps/agent-engine/cmd/streamproof
+
+      - name: Redact the captured log
+        # The log carries model names, conversation ids and streamed model
+        # output, never a credential, but the redactor runs anyway before the
+        # log is published anywhere: the one thing worse than no evidence is
+        # evidence that leaks (PR #578).
+        if: always()
+        run: |
+          set -euo pipefail
+          if [ -s "$PROOF_LOG" ]; then
+            python3 scripts/redact-log-credentials.py < "$PROOF_LOG" > /tmp/proof-redacted.log
+          else
+            echo "no proof log was written" > /tmp/proof-redacted.log
+          fi
+          {
+            echo '### Streaming delta capture'
+            echo
+            echo '```'
+            cat /tmp/proof-redacted.log
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload the captured log
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: stream-delta-proof-${{ github.run_id }}
+          path: /tmp/proof-redacted.log
+          if-no-files-found: warn
+          retention-days: 30
+
+      - name: Post the capture on the pull request
+        # An artifact behind an authenticated download is not evidence anyone
+        # reviewing the pull request will read, so the log goes in a comment.
+        if: always() && github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          body=$(mktemp)
+          {
+            echo "## Streaming delta capture, run [\`${GITHUB_RUN_ID}\`](https://github.com/$GITHUB_REPOSITORY/actions/runs/${GITHUB_RUN_ID})"
+            echo
+            echo "Captured on a hosted runner against a real Apptainer sandbox built from the"
+            echo "shipped SIF, talking to the live gateway. The stream=false arm is the control."
+            echo
+            echo '```'
+            cat /tmp/proof-redacted.log
+            echo '```'
+          } > "$body"
+          gh pr comment "$PR" --body-file "$body"
--- a/apps/agent-engine/cmd/streamproof/main.go
+++ b/apps/agent-engine/cmd/streamproof/main.go
@@ -21,11 +21,13 @@
 // It calls engineapi.New(...).Launch(...), the exact function the host
 // launcher's POST /launch handler calls (apps/agent-engine/cmd/agent-engine/
 // serve.go), against a real Apptainer sandbox built from the shipped SIF and a
-// real model behind the Hive gateway. The single substitution is
-// ResolveEgressHosts, which asks control-plane for a tenant's egress policy in
-// the daemon and here returns the model host directly, exactly as serve.go
-// appends it. That lookup cannot affect whether the agent-server publishes a
-// token delta.
+// real model behind the Hive gateway. It differs from the daemon in exactly two
+// ways, neither able to affect whether the agent-server publishes a token
+// delta: ResolveEgressHosts asks control-plane for a tenant's egress policy
+// there and returns the model host directly here, which is strictly narrower;
+// and MemoryLimit is 2G here against serve.go's 4G, so a heavier prompt could
+// be killed for memory in this harness and not in production. Every other
+// config field normalises to the production value inside engine.New.
 //
 // # THE DIFFERENTIAL
 //
@@ -74,10 +76,14 @@ import (
 
 const (
 	// Small on purpose: this run is billed against a real Hive account. It
-	// still asks for one shell command before the sentence, because streaming
-	// reassembles tool call arguments from partial JSON and a prompt the model
-	// can answer in prose alone would never exercise that path, which is the
-	// one an agent actually spends its time on.
+	// still asks for one shell command before the sentence, as an end-to-end
+	// check that tool calling WORKS under stream=true: the SDK reassembles the
+	// call from streamed chunks, and a capture that records an ActionEvent and
+	// an ObservationEvent proves that reassembly produced a call the sandbox
+	// could execute. It is not coverage of argument-fragment streaming and
+	// cannot be: a StreamingDeltaEvent only ever carries content or
+	// reasoning_content (event_service.py), so no tool-argument fragment can
+	// appear in this capture, and the tool phase produces no deltas at all.
 	proofPrompt = "Run the shell command: echo hive-streaming-proof. Then reply with exactly this sentence and nothing else: Hive streaming proof ok."
 
 	deltaKind           = "StreamingDeltaEvent"
--- a/apps/agent-engine/cmd/streamproof/main.go
+++ b/apps/agent-engine/cmd/streamproof/main.go
@@ -75,23 +75,24 @@ import (
 )
 
 const (
-	// Small on purpose: this run is billed against a real Hive account. The
-	// prose comes first and the shell command second, because an OpenHands
-	// agent ends a turn through the finish tool, so a prompt whose whole answer
-	// is the closing sentence tends to put that sentence inside a tool call and
-	// emit no MessageEvent at all. A delta only carries prose, so that run has
-	// nothing to observe and the harness reports it inconclusive. Asking for a
-	// message before the command is what makes prose reliably exist.
+	// Small on purpose: this run is billed against a real Hive account, and one
+	// sentence of prose still produces several token deltas.
 	//
-	// The shell command stays, as an end-to-end
-	// check that tool calling WORKS under stream=true: the SDK reassembles the
-	// call from streamed chunks, and a capture that records an ActionEvent and
-	// an ObservationEvent proves that reassembly produced a call the sandbox
-	// could execute. It is not coverage of argument-fragment streaming and
-	// cannot be: a StreamingDeltaEvent only ever carries content or
-	// reasoning_content (event_service.py), so no tool-argument fragment can
-	// appear in this capture, and the tool phase produces no deltas at all.
-	proofPrompt = "Do these two things in order. First, send a plain text chat message, with no tool call in it, saying: starting the streaming proof. Second, run the shell command: echo hive-streaming-proof. Then finish."
+	// Prose only, deliberately, after three runs proved otherwise. A delta
+	// carries content or reasoning_content and nothing else, and an OpenHands
+	// agent ends a turn through the finish tool, so a prompt that also asks for
+	// a shell command tends to route the whole answer through tool calls and
+	// emit no MessageEvent at all: runs 32008631987, 32009278161 and 32009716814
+	// each finished with ActionEvent and ObservationEvent frames and no prose
+	// for a delta to ride on, which the precondition below reports as
+	// inconclusive rather than as a streaming failure.
+	//
+	// Tool calling under stream=true is already proven and posted on this pull
+	// request by run 31957712375, which captured ActionEvent=1, ObservationEvent=1
+	// and five deltas in one turn. Re-rolling that dice on every run costs real
+	// money to re-prove a recorded result, so this prompt asks for the one thing
+	// the capture can observe deterministically.
+	proofPrompt = "Reply with exactly this sentence and nothing else: Hive streaming proof ok."
 
 	deltaKind           = "StreamingDeltaEvent"
 	messageKind         = "MessageEvent"
--- a/apps/agent-engine/cmd/streamproof/main.go
+++ b/apps/agent-engine/cmd/streamproof/main.go
@@ -84,9 +84,10 @@ const (
 	// cannot be: a StreamingDeltaEvent only ever carries content or
 	// reasoning_content (event_service.py), so no tool-argument fragment can
 	// appear in this capture, and the tool phase produces no deltas at all.
-	proofPrompt = "Run the shell command: echo hive-streaming-proof. Then reply with exactly this sentence and nothing else: Hive streaming proof ok."
+	proofPrompt = "Run the shell command: echo hive-streaming-proof. Then finish by writing a plain chat message, not a tool call, containing exactly this sentence and nothing else: Hive streaming proof ok."
 
 	deltaKind           = "StreamingDeltaEvent"
+	messageKind         = "MessageEvent"
 	conversationTimeout = 5 * time.Minute
 	socketWaitTimeout   = 4 * time.Minute
 )
@@ -179,9 +180,18 @@ func run() error {
 	rec.printf("SUMMARY A/product(stream=true): %s", streamed)
 	rec.printf("SUMMARY B/control(stream=false): %s", control)
 
+	// Precondition, checked before the verdict rather than folded into it: a
+	// delta only ever carries content or reasoning_content, so a turn the model
+	// answered entirely through tool calls produces no prose and therefore no
+	// deltas whether streaming is on or off. Reporting that as a streaming
+	// failure would be a lie, and reporting it as a pass would be worse.
+	if streamed.count(messageKind) == 0 {
+		return fmt.Errorf("inconclusive rather than failed: the streaming arm produced no %s, so the model wrote no prose for a delta to carry this run (frames seen: %s). Rerun", messageKind, streamed)
+	}
+
 	var problems []string
 	if streamed.deltas() == 0 {
-		problems = append(problems, "the product launch published no StreamingDeltaEvent")
+		problems = append(problems, "the product launch wrote prose but published no StreamingDeltaEvent")
 	}
 	if n := control.deltas(); n != 0 {
 		problems = append(problems, fmt.Sprintf("the stream=false control published %d StreamingDeltaEvent frames, so this capture is not measuring the flag", n))
@@ -355,11 +365,13 @@ type capture struct {
 	samples        int
 }
 
-func (c *capture) deltas() int {
+func (c *capture) count(kind string) int {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	return c.counts[deltaKind]
+	return c.counts[kind]
 }
+
+func (c *capture) deltas() int { return c.count(deltaKind) }
 
 func (c *capture) String() string {
 	c.mu.Lock()
--- a/apps/agent-engine/cmd/streamproof/main.go
+++ b/apps/agent-engine/cmd/streamproof/main.go
@@ -75,8 +75,15 @@ import (
 )
 
 const (
-	// Small on purpose: this run is billed against a real Hive account. It
-	// still asks for one shell command before the sentence, as an end-to-end
+	// Small on purpose: this run is billed against a real Hive account. The
+	// prose comes first and the shell command second, because an OpenHands
+	// agent ends a turn through the finish tool, so a prompt whose whole answer
+	// is the closing sentence tends to put that sentence inside a tool call and
+	// emit no MessageEvent at all. A delta only carries prose, so that run has
+	// nothing to observe and the harness reports it inconclusive. Asking for a
+	// message before the command is what makes prose reliably exist.
+	//
+	// The shell command stays, as an end-to-end
 	// check that tool calling WORKS under stream=true: the SDK reassembles the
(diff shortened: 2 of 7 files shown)


sakibsadmanshajib:
## Adversarial review, streams and findings

**Streams that ran:** Go review of the full diff (this comment), plus the live
differential capture in `agent streaming delta proof`, which is the acceptance
criterion rather than a review opinion.

**Streams SKIPPED, not passed:**

- CodeRabbit CLI: `Review limit reached ... You can wait 27 minutes for the limit
  to reset` (rate limited, not run). The CodeRabbit GitHub App check on this pull
  request reports `Review rate limited` for the same reason, so neither
  CodeRabbit path produced an opinion. Absence of findings here is absence of a
  review, not a clean bill.
- `/codex:adversarial-review`: the Codex CLI answered `You've hit your usage
  limit`, so that stream did not run either.
- Skill-driven streams (`ecc:code-review`, `adversarial-pr-review`): structural
  capability gap, the reviewing agent has no Skill tool. Reported rather than
  worked around.

### Findings

**1. Operational, worth knowing before anyone repoints the launcher (info, no
change requested).** With `stream: true` on the launch, the sandbox's model
requests now carry `stream: true` to the gateway, and `chat_completions.go:48`
turns that into `NeedStreaming`, which `routing/service.go:198` enforces:
`if input.NeedStreaming && !candidate.SupportsStreaming` drops the candidate. So
an operator who points `HIVE_AGENT_ENGINE_LLM_MODEL` at an alias whose only
route is not marked streaming capable will see every agent task fail at its
first model call, where it previously worked buffered. Blast radius is the one
alias the launcher is configured with. The live capture proves the configured
alias (`hive-default`) selects a route and streams, so this is a note for a
future configuration change, not a defect in the diff. Fail closed on a
capability mismatch is the correct behaviour and is left alone.

**2. Proof harness leaks a goroutine on the error path (low, accepted).** In
`streamproof`'s `observe`, an `awaitTerminal` error returns before `wg.Wait()`.
The deferred `conn.Close()` unblocks the collector, so nothing hangs, but the
frame counts collected so far are dropped from the summary. The individual delta
lines are already in the log, so evidence is not lost, and this is throwaway
scaffolding behind a build tag. Not worth the code.

**3. Subscribe-after-launch race (low, fails in the safe direction).** The
harness opens the event socket after `Launch` returns, so a delta emitted in the
first milliseconds could be missed. The consequence is a false negative, which
fails the job loudly, never a false positive. Both live runs saw the first delta
about a minute after launch, so the window is not close.

**4. Every pull request touching these paths spends a real gateway call (info).**
The workflow only runs on `apps/agent-engine/internal/{controlclient,engine}`,
`cmd/streamproof` and itself, and the prompt is one command plus one sentence,
so the cost is small and bounded. The workflow header says to delete it once
streaming has a permanent test of its own.

**5. Default-off switches found and deliberately not flipped.**
`enable_sub_agents` and `enable_browser` are both false by default and stay off
here, for the reasons in the pull request body. One correction to the record
while looking: `enable_switch_llm_tool` defaults to **true** upstream, so the
built-in `switch_llm` tool is already available inside every sandbox we launch
and nobody chose that. Left as is, flagged.

### What was checked and is fine

- The launch payload change itself: one field, no configuration knob, and the
  regression guard asserts the JSON name on the raw body rather than only the
  decoded Go field, which is the half that actually matters because the
  agent-server reads the name. Verified to fail when the flag is flipped back.
- Metering: unchanged, for the reason in the body. The vendored SDK asks for
  `stream_options.include_usage` whenever it streams, the terminal usage frame
  arrives, and settlement takes the confirmed branch. The live capture exercised
  that path end to end against the real gateway.
- The hand written WebSocket client: server frames are unmasked and handled,
  continuation frames accumulate, pings are answered with masked pongs, close
  ends the read, frame length is capped at 8 MiB, and the only write path is a
  pong. It reads and never sends application data, so it cannot steer the
  conversation it observes.
- Workflow security: no new secret is introduced, fork pull requests are
  excluded (they would fail on a missing gateway key rather than on the change),
  the captured log goes through `scripts/redact-log-credentials.py` before it is
  published anywhere, and the log carries model names, conversation ids and the
  model's own output, no credential.

.github/workflows/agent-stream-delta-proof.yml
```diff
@@ -0,0 +1,207 @@
...
+# This job is scaffolding for one change. Delete it once the launcher relay
+# lands and streaming has a permanent test of its own.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'apps/agent-engine/internal/controlclient/**'
+      - 'apps/agent-engine/internal/engine/**'
```
sakibsadmanshajib, line 26: **This path filter arms a live, billed job on every future engine change.** The file header calls this scaffolding for one change, to be deleted once the launcher relay lands, and the reasoning for capturing live is sound. But as configured it fires on any pull request touching `internal/engine/**` or `internal/controlclient/**`, and each run launches a real sandbox and spends a real model call against the live gateway with `secrets.HIVE_API_KEY`.

It also goes red for a reason unrelated to the change under test once no unexpired `agent-engine-sif` artifact matches the image-inputs digest, since `agent-engine-sif.yml` only builds on `deploy/apptainer/**`, `vendor/openhands/**` and `apps/agent-engine/packs/**`. The correlation-by-inputs logic is the right design and the error message is helpful, but that failure will land on some future unrelated pull request rather than on this one.

For scaffolding with a stated expiry, `workflow_dispatch` alone would carry the same evidence for this pull request with none of the standing cost. Not a blocker, and the no-concurrency-group choice is right for the reason given.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -391,6 +391,13 @@ func (e *SandboxEngine) Launch(ctx context.Context, t Task) (sessionRef string,
 				BaseURL: e.cfg.LLMBaseURL,
 				APIKey:  e.cfg.LLMAPIKey,
 				UsageID: "hive-agent",
+				// Always on, with no Config knob: a sandbox that cannot
```
sakibsadmanshajib, line 394: **Routing hazard, HIGH, unguarded today.** The "no Config knob" half is what makes this worth a follow-up, because the failure it can produce is total and undiagnosable.

The chain: `stream: true` reaches the gateway as `NeedStreaming` (`apps/edge-api/internal/inference/chat_completions.go:48`), and `matchesRequestedCapabilities` (`apps/control-plane/internal/routing/service.go:198`) drops every candidate whose `supports_streaming` is false. If the configured agent alias has no streaming-capable route left, `SelectRoute` returns `ErrRouteNotEligible`, control-plane answers 422, and edge-api maps it to a 400 `capability_mismatch` reading `No route supports the requested capabilities for model 'hive-default'.` So yes: **every** agent task fails at its first model call, not merely the streaming display.

On diagnosability it is worse than that message suggests. The 400 names the model, never the capability, leaving eleven `Need*` flags to guess between. And the message does not leave the sandbox: `SandboxEngine.Status` reports terminal failure as `fmt.Sprintf("agent-server execution_status=%s", ...)` (`engine.go:470`), so control-plane and the task UI see `agent-server execution_status=errored` and nothing else. Finding the cause means reading gateway logs.

Nothing guards it today:

- `supabase/migrations/20260331_02_routing_policy.sql:18` declares `supports_streaming boolean not null default false`, so a route row added without that column set is non-streaming by default.
- `provider_capabilities` is written only by SQL migrations, never from Go, and routing INNER JOINs it, so those rows are hand-authored per route.
- There is no boot check, no preflight against the configured alias, and no test asserting the agent alias resolves a streaming-capable route.

It works right now only because the four seeded routes in that same migration all set `supports_streaming` to true. That is exactly the "true today, says nothing about the next catalog change" case. Two cheap options, either fine, neither belonging in this pull request: give `Config` a `Stream` field defaulted to true so an operator can turn it off with an environment variable rather than a code deploy, or preflight the configured alias once at launcher boot and refuse to start with a message that names streaming.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -391,6 +391,13 @@ func (e *SandboxEngine) Launch(ctx context.Context, t Task) (sessionRef string,
...
+				// stream its tokens is of no use to any surface Hive is
+				// building, and the cost is nil either way (the vendored SDK
+				// keeps asking for the terminal usage frame when it streams,
+				// so the gateway meters a streamed response exactly as it
```
sakibsadmanshajib, line 398: **Metering: the happy-path claim checks out, the unhappy path is not nil-cost.** I verified the mechanism rather than accepting it. `llm.py:2079` sets `stream_options={"include_usage": True}` via `setdefault` inside `_prepare_transport_kwargs` whenever `enable_streaming`, and `enable_streaming` is `kwargs["stream"] or self.stream`, degrading to non-streaming when no `on_token` callback is wired. So when the terminal usage frame arrives, `settlementCredits` returns `confirmed=true` and settlement is genuinely identical to the buffered path.

The unhappy paths, named rather than assumed away, since this repository served free for three days behind exactly this shape:

1. **Usage frame missing, prose delivered.** `settlementCredits` (`apps/edge-api/internal/inference/stream.go:400`) falls back to a bytes-per-token estimate and returns `confirmed=false`. Control-plane then clamps the charge at the hold (`accounting/service.go:292`; the hold is a flat 10000 credits for chat), moves the reservation to `needs_reconciliation`, and inserts a `credit_reconciliation_jobs` row. **Nothing in this repository consumes that table.** The only reference outside migrations is the INSERT at `accounting/repository.go:243`, so the correction never happens and any agent turn whose true cost exceeds the flat hold is silently undercharged. Agent turns carry whole file contents, so exceeding the hold is ordinary rather than exotic.

2. **Usage frame missing on a tool-call-only turn.** `UsageAccumulator.AccumulateContent` deliberately ignores tool-call deltas (`stream.go:49-62`), so `content` is empty, `estimateCompletionTokens` returns 0, `settlementCredits` returns `delivered=false`, and `settleStream` releases the hold in full and records the request as `upstream_error`. A successful, billable model turn is served free and filed under an error event. This is the modal shape of agent traffic, which is why it matters more here than it ever did for chat.

3. **No signal either way.** The synchronous path logs `settling unconfirmed usage estimate` (`orchestrator.go:270`). The streaming path logs nothing at all when `confirmed` is false. A provider that quietly stopped honouring `include_usage` would move all agent traffic onto path 1 or 2 with no log line anywhere.

None of this is introduced by these 23 lines, and paths 1 and 2 are symmetric with the synchronous path's own tool-call blindness (`chatChoiceTexts`, `usage_clamp.go:252`, skips tool calls too). What changes is who walks them: a buffered chat completion practically always carries `usage` in its body, whereas a streamed one carries it only because an optional request field asked for it. Before agent traffic ramps, one log line plus a counter on `!confirmed` in `settleStream` would turn this class of miss from silent into visible, which is the whole lesson of the earlier outage.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -391,6 +391,13 @@ func (e *SandboxEngine) Launch(ctx context.Context, t Task) (sessionRef string,
...
+				// meters a buffered one).
+				Stream: true,
```
sakibsadmanshajib, line 400: **Turning streaming on does change error handling, and each new failure mode is silent.** Answering the "what breaks if a stream ends early" question directly, because the answer is not "the same thing, later".

Once the relay commits `200 OK` and the SSE headers (`stream.go`, step 8), nothing after that point can be reported to the caller as an error:

- The relay loop never checks `scanner.Err()`. A truncated upstream body, or a single SSE line above the 512 KB `scanner.Buffer` cap, ends the loop indistinguishably from a clean finish: no error frame, no `data: [DONE]`, no log line.
- `NewLiteLLMClient` uses `&http.Client{Timeout: 120 * time.Second}` (`litellm_client.go:26`). That is a **total** timeout and it covers reading a streaming body, so any turn longer than two minutes from dispatch has its connection cut mid-stream. Buffered, the same ceiling produced a clean 502 with a provider-blind body that the SDK raises and can retry. Streamed, it produces a truncated stream the SDK will try to reassemble.
- A mid-stream provider error frame (`data: {"error": ...}`) unmarshals cleanly into `ChatCompletionChunk` with every field absent, has its `model` rewritten, and is re-marshalled, so the error payload is replaced by an empty chunk and never reaches the client. Only unparseable JSON falls through to the raw passthrough branch.

Downstream of a truncation the SDK calls `litellm.stream_chunk_builder(chunks, messages=messages)` (`llm.py:2119`) over whatever arrived, so a partially streamed tool call is reassembled from incomplete argument JSON rather than raised. Combined with the settlement note above, a long agent turn can end as a partial or empty tool call, no error anywhere, and an unconfirmed or zero charge.

All of this is pre-existing on the chat streaming path, so it is inherited rather than introduced by this change. It is worth stating out loud because agent turns are the longest-running requests in the product and will be the first traffic to cross that 120 second ceiling routinely.

apps/agent-engine/internal/controlclient/client.go
```diff
@@ -208,6 +208,22 @@ type LLMSettings struct {
...
+	// StreamingDeltaEvent to subscribers. openhands/sdk/llm/llm.py defaults
+	// this field to False, so a payload that omits it launches a sandbox that
+	// can never emit a delta.
+	//
+	// No omitempty on purpose: the false case is a meaningful statement about
+	// a launch, and a launch payload that silently drops the field is exactly
+	// how this stayed off unnoticed. Deltas are transient by design (they
+	// bypass the callback chain that persists ConversationState.events), so
+	// this buys live delivery only, never replay.
+	Stream bool `json:"stream"`
```
sakibsadmanshajib, line 226: **Scope of what this actually buys, for wave two.** The doc is right that the field gates real behaviour, and the `streaming_enabled` computation it quotes is exactly what `event_service.py:778` does. One limit belongs next to it, because the next wave will design a chat surface against this.

`_token_streaming_callback` (`event_service.py:807-821`) forwards only `delta.content` and `delta.reasoning_content`. Tool-call argument deltas are never published as a `StreamingDeltaEvent`, by construction. The live capture posted on this pull request shows exactly that shape: the run starts at 16:08:48, an `ActionEvent` and an `ObservationEvent` are recorded for the shell command, and the first delta does not arrive until 16:10:57, on the final sentence, five deltas in total, all prose.

So a chat surface fed from these events streams the model's prose and shows nothing for the entire tool-calling phase, which is where an agent spends most of its wall clock. That is acceptable for wave one, it is simply not the same claim as "the agent streams", and the wave two design should not assume otherwise.

Separately, no `omitempty` is the right call here for the reason given, and it is what makes the raw-body assertion in the test possible at all.

apps/agent-engine/internal/engine/engine_test.go
```diff
@@ -605,6 +625,15 @@ func TestSandboxEngine_Launch_SendsInlineAgentSettingsWhenLLMConfigured(t *testi
 	if got := req.AgentSettings.LLM.APIKey; got != e.cfg.LLMAPIKey {
 		t.Fatal("llm.api_key did not round-trip")
 	}
+	if !req.AgentSettings.LLM.Stream {
+		t.Fatal("llm.stream = false, want true")
+	}
+	// The agent-server reads the JSON name, not the Go field: without
+	// "stream": true in the body it computes streaming_enabled = false and
+	// publishes no StreamingDeltaEvent for the whole conversation.
+	if body := fake.startConversationBody(); !strings.Contains(body, `"stream":true`) {
```
sakibsadmanshajib: **Verified, and the raw-body assertion earns its place.** Two experiments in the repository toolchain image, both reverted afterwards:

- `Stream: true` flipped to `false`: red at line 629, `llm.stream = false, want true`. So the guard discriminates on the flag.
- Go field left true, JSON tag renamed to `stream_enabled`: the decoded assertion at line 628 stays **green**, because encode and decode are symmetric through the same struct, and this raw-body assertion is the one that catches it, printing the wire body in the failure. That is precisely the failure a decoded-only test cannot see, so the comment above it is accurate rather than decorative.

Two small notes, neither blocking:

- The failure message prints the whole launch body, `api_key` field included. Harmless with `fixtureCredential`, but it does mean anyone who ever points this test at a real key publishes it in CI logs. Blanking the key on the copy used in the message would remove that trap.
- The `strings.Contains` check is unanchored, so any future nested object carrying its own `stream` field would satisfy it. Scoping the check to the `llm` object would keep it precise. Nothing in today's payload has one.

apps/agent-engine/cmd/streamproof/main.go
```diff
@@ -0,0 +1,605 @@
...
+// both replay the same final MessageEvent and no deltas at all. The evidence
+// has to be collected while the conversation is still running, which is what
+// this program does.
+//
+// # WHAT IT ACTUALLY EXERCISES
+//
+// It calls engineapi.New(...).Launch(...), the exact function the host
+// launcher's POST /launch handler calls (apps/agent-engine/cmd/agent-engine/
+// serve.go), against a real Apptainer sandbox built from the shipped SIF and a
+// real model behind the Hive gateway. The single substitution is
```
sakibsadmanshajib: **The door really is the production one.** Confirmed: `engineapi.New(cfg).Launch(...)` here is the same call `serve.go`'s `POST /launch` handler makes (`cmd/agent-engine/serve.go:139` and `:168`), and `engineapi` is nothing but type aliases plus a one-line constructor, so there is no second implementation that could drift. The `ResolveEgressHosts` substitution returns the model host where production returns the tenant policy plus the model host, which is strictly narrower and cannot influence whether the agent-server publishes a delta.

A small accuracy fix to "the single substitution": the config also differs in `MemoryLimit`, 2G here (line 235) against 4G in `serve.go`. `QuotaTenantConcurrency`, `QuotaUserConcurrency`, `CPULimit` and `PidsLimit` all normalise to the production values inside `engine.New` (`engine.go:149-165`), so memory is the only other real difference. It cannot affect delta publication either, but a heavier live prompt could OOM here and not in production, which a reader of this docstring should know.

I also confirmed the build tag does what the docstring claims: `go vet ./apps/agent-engine/...`, `go build ./apps/agent-engine/...` and `go test ./apps/agent-engine/... -count=1 -short` are all green with this package silently skipped, and `go vet -tags proof` plus `go build -tags proof` on this directory also pass, so the scaffolding compiles rather than rotting unnoticed.

apps/agent-engine/cmd/streamproof/main.go
```diff
@@ -0,0 +1,605 @@
...
+	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/controlclient"
+)
+
+const (
+	// Small on purpose: this run is billed against a real Hive account. It
+	// still asks for one shell command before the sentence, because streaming
+	// reassembles tool call arguments from partial JSON and a prompt the model
+	// can answer in prose alone would never exercise that path, which is the
+	// one an agent actually spends its time on.
+	proofPrompt = "Run the shell command: echo hive-streaming-proof. Then reply with exactly this sentence and nothing else: Hive streaming proof ok."
```
sakibsadmanshajib: **Keep the tool call in the prompt, but not for the stated reason.** The commit that added it says streaming reassembles tool call arguments from partial JSON, that this is where a streaming regression would show up, and so the prompt now asks for a shell command. The harness cannot observe that: a `StreamingDeltaEvent` only ever carries `content` or `reasoning_content` (`event_service.py:807-821`), so no tool-call argument fragment can appear in this capture, and neither assertion below looks for one. The posted capture bears that out, with the tool phase producing zero deltas and all five deltas landing on the closing sentence.

What the tool call genuinely proves, and this is worth having, is that tool calling still **works** under `stream: true`: the capture records `ActionEvent=1` and `ObservationEvent=1` on the streaming arm, which means `litellm.stream_chunk_builder` reassembled the call out of chunks and the sandbox executed it. That is the real regression risk of turning streaming on. Keep the prompt, describe it as an end-to-end tool-calling check rather than as coverage of argument-fragment streaming.

One note on the differential while reading the capture: the two arms are not perfectly matched (A: `ActionEvent=1`, `ObservationEvent=1`, `MessageEvent=1`; B: `ActionEvent=2`, `ObservationEvent=2`, no `MessageEvent`). It does not weaken the conclusion, since B produced zero deltas across strictly more agent steps, but it is not quite the identical-run-minus-one-flag the comment describes.

sakibsadmanshajib:
## Independent stage 6 review, second Go stream

CodeRabbit (CLI and app), Codex and Bugbot were all unavailable on this pull request, and recording them as SKIPPED rather than as passes was the right call. This is an independent pass over the pushed diff at `af3aee88`, against the code rather than against the author's own review. Eight inline comments are posted on the Files tab.

**Verdict on the 23 lines: correct, and the guard test is real. Nothing in the diff needs to change.**

Verified by running it in the repository toolchain image, not by reading:

- `go vet ./apps/agent-engine/...`, `go build ./apps/agent-engine/...` and `go test ./apps/agent-engine/... -count=1 -short` are all green, with the `proof`-tagged package silently skipped by all three, so the build tag keeps the harness out of ordinary builds without breaking wildcard commands. `go vet -tags proof` and `go build -tags proof` also pass, so it compiles rather than rotting.
- Flipping `Stream: true` to `false` turns the guard red at `engine_test.go:629`.
- Leaving the Go field true and renaming only the JSON tag leaves the decoded assertion green and turns the raw-body assertion red at `engine_test.go:635`. The raw-body check is therefore doing the independent work claimed for it: it pins the wire name the agent-server reads, which a struct round trip cannot.
- `engineapi.New(cfg).Launch(...)` in the proof is the same call `serve.go`'s `POST /launch` makes, and the `ResolveEgressHosts` substitution is strictly narrower than production's allowlist, so it cannot manufacture a delta.

Two follow-ups I would file rather than block on, both about blast radius rather than about this diff:

1. **Nothing guards the streaming capability of the agent alias.** `supports_streaming` defaults to false in the schema, `provider_capabilities` rows are hand-authored in migrations, and there is no knob, no preflight and no test. If a future catalog change leaves the alias without a streaming route, every agent task fails at its first model call, and the only message that escapes the sandbox is `agent-server execution_status=errored`. Detail in the inline comment on `engine.go:394`.
2. **A missing terminal usage frame settles silently.** The happy path is exactly as the author describes. When the frame does not arrive, a prose turn settles from an estimate that is clamped at the flat hold and parked in `credit_reconciliation_jobs`, a table nothing in this repository consumes; a tool-call-only turn settles as `delivered=false`, releases the hold in full, and is served free under an `upstream_error` event. Neither case logs anything on the streaming path, unlike the synchronous one. Inherited rather than introduced, but agent traffic is the tool-call-heaviest traffic in the product. Detail on `engine.go:398`.

Also worth knowing before wave two: `StreamingDeltaEvent` carries only `content` and `reasoning_content`, never tool-call argument fragments, so a chat surface fed by these deltas will still show nothing for the whole tool-calling phase. The posted capture shows exactly that, with all five deltas landing on the closing sentence.

.github/workflows/agent-stream-delta-proof.yml
sakibsadmanshajib: Rebuttal on the trigger, accepted on the failure mode, and neither changes the file.

`workflow_dispatch` alone could not have produced this evidence: a dispatch resolves the workflow file from the default branch, so a job that exists only here cannot be dispatched at all. The `pull_request` trigger is what made proof-before-merge possible, the same circle `agent-visual-proof.yml` documents breaking.

Once this lands, the file is on the default branch and dispatch becomes available, so narrowing the trigger is a one-line follow-up for whoever deletes or keeps this job. I have left it on so the next change to these paths is held to the same evidence rather than to a claim. The cost is bounded: the paths are only `internal/controlclient`, `internal/engine`, `cmd/streamproof` and the workflow itself, and the prompt is one shell command plus one sentence.

Your second point is the sharper one and I accept it as stated: correlation by image inputs means a future change touching these paths, while no unexpired `agent-engine-sif` artifact matches, goes red for a reason unrelated to itself. The message names the remedy, but the red lands on someone who did not cause it. That is the price of refusing to certify against an unrelated image, and I would rather pay it than have the job quietly prove the wrong artifact. Noted here so the next person to hit it knows it is designed rather than broken.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Agreed, and filed as [#926](https://github.com/sakibsadmanshajib/hive/issues/926) rather than fixed here, since this change stays at its original size.

Your trace is the one I reached independently while writing the operational note in the body. Your diagnosability half is the part I had underweighted: a 400 that names the model but not the capability, and `SandboxEngine.Status` flattening everything to `agent-server execution_status=errored`, means the operator sees a task fail with no cause anywhere they would look.

The issue carries both of your options, boot preflight or a `Config.Stream` field, plus a cheaper partial that would help every in-sandbox model failure rather than only this one: surface the model call's actual error text through `Status` instead of the bare execution status.

On "it works today": exactly right, and now restated in the issue. The four seeded routes set `supports_streaming` true, `provider_capabilities` is written only by hand-authored migrations, and the catalog is admin-managed by design, so today's state is a fact about a seed rather than a guarantee.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Acted on, not rebutted. `df2686d7` adds the two log lines the synchronous path already has, and changes no behaviour.

- the unconfirmed branch now logs `settling unconfirmed usage estimate ... estimated_credits=N`, matching `orchestrator.go` word for word;
- the delivered-nothing branch now logs `settle stream delivered nothing, releasing hold ... reason=...` before releasing, which is your path 2: the tool-call-only turn served free and filed as `upstream_error`.

The guard is `TestExecuteStreaming_DeliveredNothing_SaysSoInTheLog`, asserting the line rather than the release call the existing test already covers, and verified to go red when the line is removed.

Your two structural findings are filed rather than built here:

- [#925](https://github.com/sakibsadmanshajib/hive/issues/925), `credit_reconciliation_jobs` has a writer and no reader. I confirmed your reading: the INSERT in `accounting/repository.go` is the only reference outside migrations. The issue puts three options up, including that a queue nothing can drain should perhaps be an alert instead of a table.
- [#928](https://github.com/sakibsadmanshajib/hive/issues/928), the settlement half of your path 2, because a log line makes it visible without making it correct: a turn that produced only tool calls is still charged nothing. The fix likely belongs in the shared estimate, since the synchronous path is blind the same way.

The rest of your reading matches mine, including that the hold for chat is a flat 10000 credits and that an agent turn carrying whole file contents exceeds it routinely rather than exceptionally.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Filed as [#928](https://github.com/sakibsadmanshajib/hive/issues/928), with all four shapes you name, and not fixed here.

I checked the two I had not: the relay loop does end on scanner exhaustion without inspecting `scanner.Err()`, and `NewLiteLLMClient`'s 120 second `http.Client.Timeout` is a total timeout that covers reading the streaming body.

The consequence you draw concerns me most, because it is specific to this change: a truncation feeds `litellm.stream_chunk_builder` a partial tool call, which reassembles into a malformed or empty call rather than raising, so a long agent turn can end as a silent no-op. Agent turns are the longest requests in the product and will be the first traffic to cross that ceiling routinely.

The issue orders the work money first (count tool-call output in the estimate), then the two silence fixes, then splitting the client timeout so a streaming dispatch bounds dial and response headers rather than the whole body. Inherited rather than introduced, as you say, and it stays out of a 23-line change.

apps/agent-engine/internal/controlclient/client.go
sakibsadmanshajib: Accepted, and it is now a named wave-two constraint in the body rather than a footnote, because it is the thing most likely to be discovered late by whoever builds the rendering.

Stated there as: a chat surface fed by these deltas alone shows nothing for the entire tool-calling phase, which is where an agent spends most of its wall clock and is precisely what the owner wants on screen. Tool visibility has to come from the event stream, `ActionEvent` and `ObservationEvent` plus the `waiting_for_confirmation` status `engine.go` currently collapses into running, not from the deltas this change switches on.

Your reading of the capture is right and I have stopped claiming otherwise: the tool phase produced zero deltas and all five landed on the closing sentence.

apps/agent-engine/internal/engine/engine_test.go
sakibsadmanshajib: Both notes taken and fixed in `1536620d`. Thank you for running the tag-rename experiment: that is the failure the raw assertion exists for, and it is better demonstrated than argued.

The assertion now reads the `agent_settings.llm` object's own wire fields instead of searching the whole body, so a `stream` field some future nested object adds cannot satisfy it, and the failure names the field keys that were sent rather than printing their values, so `api_key` cannot reach a CI log even if someone points the test at a real credential.

Re-verified after the change: renaming the tag to `stream_enabled` leaves the decoded assertion green and turns this one red with `agent_settings.llm has no "stream": true on the wire (got ""); field names sent: [api_key base_url model stream_enabled usage_id]`.

apps/agent-engine/cmd/streamproof/main.go
sakibsadmanshajib: Correction accepted, in the code comment (`1536620d`) and in the body.

Both now say the harness differs from the daemon in two ways rather than one: `ResolveEgressHosts`, strictly narrower here, and `MemoryLimit` at 2G against `serve.go`'s 4G. The consequence you name is the useful half and is written down: a heavier live prompt could be killed for memory in the harness and not in production, which a reader trusting this docstring should know before making the prompt bigger.

Thank you for independently running the build-tag checks. That the package compiles under `-tags proof` while staying invisible to every ordinary build is the property that keeps scaffolding from rotting unnoticed.

apps/agent-engine/cmd/streamproof/main.go
sakibsadmanshajib: Accepted in full. The claim is corrected in the code comment, the body and the commit message rather than left to be inferred.

The prompt stays, described as an end-to-end check that tool calling still **works** under `stream: true`: the SDK reassembles the call from streamed chunks, and `ActionEvent` plus `ObservationEvent` in the capture prove the reassembly produced a call the sandbox could execute. That is the real regression risk of turning streaming on. It is not coverage of argument-fragment streaming and cannot be, since a `StreamingDeltaEvent` only ever carries `content` or `reasoning_content`.

On the arms not being identical: correct, and worth stating plainly. B took two tool steps to A's one, and the conclusion survives because B produced zero deltas across strictly more agent steps, which is the stronger direction. What the differential controls is the flag, not the trajectory: a model is free to take a different number of steps on the same prompt, so identical event counts were never available. Naming that is better than implying a determinism the harness does not have.

PR #922, opened 2026-08-16
fix: deny access when an API key's expiry timestamp is present but unparseable

sakibsadmanshajib:
## Summary

`CheckAccess`'s expiry check (`apps/edge-api/internal/authz/authz.go`, step 2) guarded the parse with `err == nil &&`, so a present `expires_at` value that failed to parse skipped the expiry check entirely instead of denying the request. That is fail-open on an authorization predicate, and this PR closes it: a non-empty `expires_at` that fails to parse now denies with `invalid_api_key`.

## Why this is hardening, not a vulnerability fix

Issue #915 asked whether an expired-but-not-revoked key authorizes, and PR #919 answered it: expiry enforcement already works today through two independent layers (the control-plane's `applyExpiry` derives `status: "expired"` on read, and edge-api's own `authz.go:155` re-parses `expires_at` against the current time). Neither of those layers relies on the branch this PR touches.

This branch specifically is not reachable in production. `expires_at` has exactly one writer, control-plane's `ResolveSnapshot`, which formats it via Go's default `time.Time` JSON marshaling (RFC3339Nano). `time.Parse(time.RFC3339, ...)` already accepts RFC3339Nano-formatted input (Go's `time.Parse` special-cases fractional seconds even when the layout does not declare them), so every value control-plane ever sends parses successfully. Producing an unparseable `expires_at` at the point `CheckAccess` reads it would mean writing a malformed value directly into the shared Redis auth-snapshot cache, and reaching that cache to write it is already a full compromise of the auth layer, not something this change could prevent.

So: no known live path reaches the branch this PR changes, and treating it as denied rather than skipped is defence in depth against a future caller of `CheckAccess` or a corrupted cache entry, not a fix for something exploitable today. Please do not read this commit as evidence that a live vulnerability existed; it did not, per #915/#919's own investigation.

## What does not change

An empty `expires_at` string is treated the same as a nil pointer (no expiry set) and keeps authorizing exactly as it does today. A hardening change that started denying non-expiring keys would itself be a live outage, and this PR does not make that mistake.

## Verification (red then green, not asserted)

Added `TestCheckAccessDeniesUnparseableExpiresAt` first, against the unpatched code:

```
=== RUN   TestCheckAccessDeniesUnparseableExpiresAt
    authz_test.go:79: expected invalid_api_key code, got model_not_allowed
--- FAIL: TestCheckAccessDeniesUnparseableExpiresAt (0.00s)
```

(The first version of the test denied for the wrong reason, model_not_allowed, because the snapshot did not set `AllowAllModels` and the fail-open branch fell through to the model-allowlist check. Corrected the snapshot to isolate the expiry check, confirmed the corrected test still fails the same way against unpatched code.)

After applying the fix, the same test and the full `authz` package:

```
=== RUN   TestCheckAccessDeniesUnparseableExpiresAt
--- PASS: TestCheckAccessDeniesUnparseableExpiresAt (0.00s)
...
ok  	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/authz	0.006s
```

Regression guards added and confirmed passing both before and after (i.e. behaviour unchanged for these two cases):

- `TestCheckAccessEmptyExpiresAtStillAllowed`: empty-string `expires_at` still authorizes.
- `TestCheckAccessNilExpiresAtStillAllowed`: nil `expires_at` (no expiry set) still authorizes.

Full `apps/edge-api` suite green after the change (`go build`, `go vet`, `gofmt -l` clean, `go test ./apps/edge-api/... -count=1 -short`), run via the toolchain container per this repo's testing convention. `-race` is unavailable in the Alpine toolchain image (`CGO_ENABLED=0`), so it was not run; nothing in this change is concurrency-sensitive (pure value comparison, no shared state).

## Scope

Only `apps/edge-api/internal/authz/authz.go` and its test file change. No other file touched.

## Buglog entry

To be appended to `.wolf/buglog.jsonl` on main via a follow-up buglog-only PR, per this repo's convention of never appending on a feature branch:

```json
{"id":"bug-2026-08-16-authz-expiry-fail-open-on-parse-error","date":"2026-08-16","title":"CheckAccess skipped the API key expiry check on a parse error instead of denying","error_message":"time.Parse(time.RFC3339, *snapshot.ExpiresAt) guarded with err == nil && exp.Before(time.Now()), so a present-but-unparseable expires_at fell through the if-block entirely and the request was evaluated as if no expiry were set","root_cause":"The expiry check only executed its deny branch when the timestamp parsed cleanly; a parse failure silently skipped the check rather than being treated as a denial, which is fail-open on an authorization predicate. Not exploitable today: expires_at has one writer (control-plane's ResolveSnapshot), which always emits RFC3339Nano, a format time.Parse(time.RFC3339, ...) already accepts, so reaching this branch would require writing a malformed value directly into the shared Redis auth-snapshot cache, itself a full compromise. Filed as hardening after #915/#919 established expiry enforcement already works via two independent layers.","fix":"Treat a non-empty expires_at that fails to parse as invalid_api_key (deny), while continuing to treat an empty string the same as a nil pointer (no expiry set, still authorizes). Added TestCheckAccessDeniesUnparseableExpiresAt (confirmed red on unpatched code, green after the fix) plus TestCheckAccessEmptyExpiresAtStillAllowed and TestCheckAccessNilExpiresAtStillAllowed as regression guards for the no-expiry cases.","tags":["security","authz","fail-open","hardening","expiry","issue-915","pr-919"],"pr":"<fill in PR number after merge>"}
```

## Review

Stage 6 (adversarial review) for this PR needs to be independent of the agent that wrote this diff, per standing instruction on auth-behaviour changes. Requesting an independent review pass now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key validation for expiry timestamps.
  * Malformed, non-empty expiry values are now rejected instead of being silently ignored.
  * Empty or unset expiry values continue to be accepted.
  * Expired API keys remain denied.

### DIFF
--- a/apps/edge-api/internal/authz/authz.go
+++ b/apps/edge-api/internal/authz/authz.go
@@ -152,13 +152,42 @@ func CheckAccess(snapshot AuthSnapshot, requestedAlias string, estimatedCredits 
 	}
 
 	// 2. Check expiry.
+	//
+	// A non-empty expires_at that fails to parse is DENIED, not skipped.
+	// This is hardening rather than a fix for a reachable gap: issue #915
+	// and PR #919 established that expiry enforcement already works today
+	// through two independent layers, and that this branch is not
+	// reachable from control-plane, which is the only writer of this
+	// field and always formats it via Go's default time.Time JSON
+	// marshaling (RFC3339Nano), a value time.Parse(time.RFC3339, ...)
+	// already accepts. Producing an unparseable value here would require
+	// writing directly to the shared Redis cache, which is already a full
+	// compromise. Failing closed here is defence in depth against that
+	// state, so a future caller of CheckAccess (or a corrupted cache
+	// entry) cannot silently bypass the expiry check the way the old
+	// "err == nil &&" guard did.
+	//
+	// An empty string is treated the same as a nil pointer (no expiry
+	// set), not as a parse failure: both must keep authorizing exactly as
+	// they do today, since a hardening change that started denying
+	// non-expiring keys would be a live outage.
 	if snapshot.ExpiresAt != nil {
-		exp, err := time.Parse(time.RFC3339, *snapshot.ExpiresAt)
-		if err == nil && exp.Before(time.Now()) {
-			return CheckResult{
-				Allowed:  false,
-				DenyCode: "invalid_api_key",
-				DenyMsg:  "API key has expired",
+		raw := strings.TrimSpace(*snapshot.ExpiresAt)
+		if raw != "" {
+			exp, err := time.Parse(time.RFC3339, raw)
+			switch {
+			case err != nil:
+				return CheckResult{
+					Allowed:  false,
+					DenyCode: "invalid_api_key",
+					DenyMsg:  "API key has an invalid expiry",
+				}
+			case exp.Before(time.Now()):
+				return CheckResult{
+					Allowed:  false,
+					DenyCode: "invalid_api_key",
+					DenyMsg:  "API key has expired",
+				}
 			}
 		}
 	}
--- a/apps/edge-api/internal/authz/authz_test.go
+++ b/apps/edge-api/internal/authz/authz_test.go
@@ -55,6 +55,98 @@ func TestCheckAccessExpiredKey(t *testing.T) {
 	}
 }
 
+// TestCheckAccessDeniesUnparseableExpiresAt pins the hardening fix at
+// authz.go's expiry check: a present-but-unparseable expires_at must deny
+// the request, not silently skip the expiry check (fail-open on a parse
+// error). See authz.go step 2 for the reasoning on why this is defence in
+// depth rather than a fix for a reachable gap (issue #915 / PR #919
+// established expiry enforcement already works via two independent layers
+// on every real value control-plane emits).
+func TestCheckAccessDeniesUnparseableExpiresAt(t *testing.T) {
+	bad := "not-a-timestamp"
+	s := AuthSnapshot{
+		KeyID:          "key-1",
+		Status:         "active",
+		ExpiresAt:      &bad,
+		AllowAllModels: true,
+		BudgetKind:     "none",
+	}
+
+	r := CheckAccess(s, "hive-default", 0)
+	if r.Allowed {
+		t.Fatal("expected denied for unparseable expires_at, got allowed (fail-open on parse error)")
+	}
+	if r.DenyCode != "invalid_api_key" {
+		t.Fatalf("expected invalid_api_key code, got %s", r.DenyCode)
+	}
+	// Both this arm and the genuinely-expired arm return the same DenyCode,
+	// so DenyMsg is the only field an operator reading logs can use to tell
+	// a malformed cache entry apart from routine expiry. Pin it, or the
+	// message could silently drift to "API key has expired" and this test
+	// (and every other test in the package) would still pass.
+	if r.DenyMsg != "API key has an invalid expiry" {
+		t.Fatalf("expected the malformed-record message, got %q", r.DenyMsg)
+	}
+}
+
+// TestCheckAccessEmptyExpiresAtStillAllowed pins the "no live outage" side
+// of the same change: an empty expires_at string is treated the same as an
+// absent one (no expiry set), and must keep authorizing exactly as it does
+// today. Only a NON-empty unparseable value is denied.
+func TestCheckAccessEmptyExpiresAtStillAllowed(t *testing.T) {
+	empty := ""
+	s := AuthSnapshot{
+		KeyID:          "key-1",
+		Status:         "active",
+		ExpiresAt:      &empty,
+		AllowAllModels: true,
+		BudgetKind:     "none",
+	}
+
+	r := CheckAccess(s, "hive-default", 0)
+	if !r.Allowed {
+		t.Fatalf("expected allowed for empty expires_at (treated as absent), got denied: %s", r.DenyMsg)
+	}
+}
+
+// TestCheckAccessWhitespaceOnlyExpiresAtStillAllowed covers the distinct
+// branch TrimSpace introduces: a whitespace-only expires_at is trimmed down
+// to empty and takes the same no-expiry path as an empty string, rather than
+// being handed to time.Parse as a non-empty value that would fail.
+func TestCheckAccessWhitespaceOnlyExpiresAtStillAllowed(t *testing.T) {
+	whitespace := " \t\n"
+	s := AuthSnapshot{
+		KeyID:          "key-1",
+		Status:         "active",
+		ExpiresAt:      &whitespace,
+		AllowAllModels: true,
+		BudgetKind:     "none",
+	}
+
+	r := CheckAccess(s, "hive-default", 0)
+	if !r.Allowed {
+		t.Fatalf("expected allowed for whitespace-only expires_at, got denied: %s", r.DenyMsg)
+	}
+}
+
+// TestCheckAccessNilExpiresAtStillAllowed pins the other "no live outage"
+// side: a key with no expires_at set at all (nil pointer, the common case
+// for a non-expiring key) must keep authorizing exactly as it does today.
+func TestCheckAccessNilExpiresAtStillAllowed(t *testing.T) {
+	s := AuthSnapshot{
+		KeyID:          "key-1",
+		Status:         "active",
+		ExpiresAt:      nil,
+		AllowAllModels: true,
+		BudgetKind:     "none",
+	}
+
+	r := CheckAccess(s, "hive-default", 0)
+	if !r.Allowed {
+		t.Fatalf("expected allowed for nil expires_at (no expiry set), got denied: %s", r.DenyMsg)
+	}
+}
+
 func TestCheckAccessRejectsDisallowedAliasWithoutRemap(t *testing.T) {
 	s := AuthSnapshot{
 		KeyID:          "key-1",


sakibsadmanshajib:
## Stage 6 adversarial review status

Independent-of-author review streams attempted for this PR (auth-behaviour change, so stage 6 is mandatory and not self-satisfiable):

- **CodeRabbit CLI**: RAN. `coderabbit review --committed --base main --agent` against both changed files. Result: `findings: 0`.
- **Codex CLI native `codex review --base main`**: ATTEMPTED, then **SKIPPED**. Hit account usage limit mid-run (`ERROR: You've hit your usage limit ... try again at Sep 10th, 2026`). Not a clean pass; treat as not run, per this repo's own fallback rule for an unavailable review stream.
- **`ecc:code-review` skill / a second independent reviewer subagent (e.g. security-reviewer, go-reviewer)**: NOT RUN. This implementing agent's toolset is Read/Write/Edit/Bash only, with no Agent or Skill invocation tool available, so it cannot spawn a second independent Claude reviewer or invoke an ECC skill itself. This is a structural capability gap, not a skipped choice: flagging it explicitly rather than letting the CodeRabbit pass alone stand as the whole gate, and rather than self-reviewing and calling it independent.

**Requesting the orchestrator dispatch a genuinely independent reviewer (security-reviewer or go-reviewer subagent, or `ecc:code-review`) against this diff before merge**, per the standing instruction that a self-review must not be the last word on an auth-behaviour change.

Scope reminder: this diff touches only `apps/edge-api/internal/authz/authz.go` and `apps/edge-api/internal/authz/authz_test.go`. Full `apps/edge-api` test suite is green (`go test ./apps/edge-api/... -count=1 -short`); `authz` package tests specifically pass with the new hardening test confirmed red on unpatched code and green on patched code (see PR description).

apps/edge-api/internal/authz/authz.go
```diff
@@ -152,13 +152,42 @@ func CheckAccess(snapshot AuthSnapshot, requestedAlias string, estimatedCredits
...
+	// they do today, since a hardening change that started denying
+	// non-expiring keys would be a live outage.
 	if snapshot.ExpiresAt != nil {
-		exp, err := time.Parse(time.RFC3339, *snapshot.ExpiresAt)
-		if err == nil && exp.Before(time.Now()) {
-			return CheckResult{
-				Allowed:  false,
-				DenyCode: "invalid_api_key",
-				DenyMsg:  "API key has expired",
+		raw := strings.TrimSpace(*snapshot.ExpiresAt)
```
sakibsadmanshajib, line 175: Verified the no-outage claim across every representation this field can take, not just the two the new tests cover, marshalling through control-plane's own declaration (`ExpiresAt *time.Time` with `json:"expires_at,omitempty"`) so the bytes are the real wire bytes:

| representation | pointer after decode | before | after |
| --- | --- | --- | --- |
| column NULL (field dropped by `omitempty`) | nil | allow | allow |
| explicit JSON null | nil | allow | allow |
| empty string | non-nil | allow | allow |
| whitespace only | non-nil | allow | allow |
| RFC3339Nano, past (control-plane's real format) | non-nil | deny, expired | deny, expired |
| RFC3339 with a `+06:00` offset, past | non-nil | deny, expired | deny, expired |
| Go `time.Time.String()` form, past | non-nil | allow | deny, invalid expiry |

No fourth representation flips a no-expiry key to denied: a NULL column arrives as an absent field and decodes to nil, and an explicit JSON null decodes to nil as well, so both take the untouched path above.

Note that `TrimSpace` also changes one case beyond the intended one: a whitespace-padded but otherwise valid timestamp (`" 2020-01-01T00:00:00Z "`) previously failed to parse and was skipped, so it authorized, and it now trims, parses and denies as expired. That is the correct direction and no producer emits it, so this is a confirmation rather than a request.

apps/edge-api/internal/authz/authz.go
```diff
@@ -152,13 +152,42 @@ func CheckAccess(snapshot AuthSnapshot, requestedAlias string, estimatedCredits
...
+		if raw != "" {
+			exp, err := time.Parse(time.RFC3339, raw)
+			switch {
+			case err != nil:
+				return CheckResult{
+					Allowed:  false,
+					DenyCode: "invalid_api_key",
+					DenyMsg:  "API key has an invalid expiry",
```
sakibsadmanshajib, line 183: Checked the operator-visibility question on this arm, since both arms return the same `DenyCode` and denying is only the right call if the two cases stay tellable apart downstream.

They do, in logs. `authorizer.go` logs `check.DenyMsg` on every `invalid_api_key` denial, so this arm prints "API key has an invalid expiry" and the expiry arm prints "API key has expired". Metrics are not a concern either, in the sense that there are none: `DenyCode` has exactly one consumer in the tree, the switch in `authorizer.go`, and no counter is labelled by it, so there is no metric dimension to regress.

Worth stating plainly for whoever reads this at 3am: this arm firing does not mean a customer's key expired, it means a snapshot reached the edge that control-plane did not write in its own format, which is a cache-integrity event. The distinct message is what makes that visible, so please keep the two strings distinct.

One nit, pre-existing and not introduced here: `Authorize` returns "Incorrect API key provided: " plus `check.DenyMsg` to the caller, so this string reaches the client too. It is no more revealing than the existing "API key has expired" and "API key is revoked", so I would leave it as is.

No change requested on this line.

apps/edge-api/internal/authz/authz_test.go
```diff
@@ -55,6 +55,70 @@ func TestCheckAccessExpiredKey(t *testing.T) {
...
+		BudgetKind:     "none",
+	}
+
+	r := CheckAccess(s, "hive-default", 0)
+	if r.Allowed {
+		t.Fatal("expected denied for unparseable expires_at, got allowed (fail-open on parse error)")
+	}
+	if r.DenyCode != "invalid_api_key" {
+		t.Fatalf("expected invalid_api_key code, got %s", r.DenyCode)
+	}
```
sakibsadmanshajib, line 81: Non-blocking, one line.

This asserts `DenyCode`, but both arms of the new `switch` return `invalid_api_key`, so `DenyMsg` is the only field that separates a malformed record from a genuinely expired key, and it is the field the caller logs (`authorizer.go`: `log.Printf("authz: access denied alias=%q reason=%q", aliasID, check.DenyMsg)`). As committed, someone could change the new message to "API key has expired" and this test, and every other test in the package, would still pass, which silently removes the only signal an operator has for telling a corrupted cache entry apart from routine expiry. Since operator distinguishability is part of what justifies denying here rather than treating the value as expired, it is worth pinning:

```suggestion
	if r.DenyCode != "invalid_api_key" {
		t.Fatalf("expected invalid_api_key code, got %s", r.DenyCode)
	}
	if r.DenyMsg != "API key has an invalid expiry" {
		t.Fatalf("expected the malformed-record message, got %q", r.DenyMsg)
	}
```

sakibsadmanshajib (commented):
## Independent stage 6 review (second stream)

Scope: head `d7677ca`, the two files in the diff. Running this as an independent pass because CodeRabbit returned zero findings and Codex was SKIPPED on quota, and zero findings from a single automated stream is not coverage. Nothing below leans on either.

**Verdict: approve.** No blocking findings. Two non-blocking inline comments, plus one framing request in this body.

### Method

I re-derived the claims against the pushed head rather than re-running the author's own commands: the committed test file and the committed `authz.go` were overlaid (`go test -overlay`) onto an unpatched checkout, so the code under test is this repository's own, not a transcription of it, and a probe harness of my own was run over the same package.

### 1. Red on unpatched code, for the expiry reason specifically

The committed test file against unpatched `authz.go`:

```
=== RUN   TestCheckAccessDeniesUnparseableExpiresAt
    authz_test.go:77: expected denied for unparseable expires_at, got allowed (fail-open on parse error)
--- FAIL: TestCheckAccessDeniesUnparseableExpiresAt (0.00s)
=== RUN   TestCheckAccessEmptyExpiresAtStillAllowed
--- PASS: TestCheckAccessEmptyExpiresAtStillAllowed (0.00s)
=== RUN   TestCheckAccessNilExpiresAtStillAllowed
--- PASS: TestCheckAccessNilExpiresAtStillAllowed (0.00s)
```

It fails on the `Allowed` assertion, not on a stray `model_not_allowed`, so the disclosed first-version defect is genuinely fixed in what was committed. The two no-outage guards pass before the change as well as after, which is what makes them regression guards rather than assertions of new behaviour. All eleven tests in the package pass with the patch applied.

### 2. The no-outage case, across every representation the field can take

Rather than trust the two representations the tests cover, I enumerated them, marshalling through control-plane's own field declaration (`ExpiresAt *time.Time` with `json:"expires_at,omitempty"`) so the wire bytes are the real ones:

| representation | wire | pointer after decode | before | after |
| --- | --- | --- | --- | --- |
| column NULL (`omitempty` drops the field) | field absent | nil | allow | allow |
| explicit null | `null` | nil | allow | allow |
| empty string | `""` | non-nil | allow | allow |
| whitespace only | `"   "` | non-nil | allow | allow |
| RFC3339Nano, past (control-plane's real format) | `"2026-08-15T16:39:28.507891593Z"` | non-nil | deny, expired | deny, expired |
| RFC3339 with a `+06:00` offset, past | `"2026-08-15T22:39:28+06:00"` | non-nil | deny, expired | deny, expired |
| Go `time.Time.String()` form, past | `"2026-08-15 16:39:28 +0000 UTC"` | non-nil | **allow** | deny, invalid expiry |
| unix seconds as a string | `"1755300000"` | non-nil | **allow** | deny, invalid expiry |

There is no fourth representation that flips a no-expiry key to denied. A NULL column reaches the edge as an absent field and decodes to a nil pointer, and an explicit JSON null also decodes to nil, so both take the untouched path. The two rows that change are the intended ones and they change in the safe direction. The claim that `time.Parse(time.RFC3339, ...)` accepts RFC3339Nano is confirmed empirically, so the real control-plane format cannot be denied by mistake.

### 3. Single writer

`expires_at` has exactly one producer. Control-plane never writes the snapshot cache key; `apikeys.Service` only DELETEs it (`invalidateSnapshot`, `snapshotRedisKey`), and `authz.Client.Resolve` caches control-plane's response bytes verbatim under the same key. No non-Go producer writes an `expires_at` into a snapshot anywhere in `scripts/`, `deploy/` or `tests/`. The single-writer claim in the description holds as written.

### 4. Ordering

`CheckAccess` has one caller, `Authorizer.Authorize`. The new denial sits in step 2, preceded only by the status check, which denies a non-active key on its own, and followed by the model allowlist and budget checks, so no other predicate can mask it. Nothing depended on the previous fall-through: the sole caller consumes `CheckResult`, and the pre-existing tests are unchanged and green.

### 5. The `err == nil &&` shape does not repeat here

`authz.go` contains exactly one `time.Parse`, the one this PR fixes. The file's other parse, `AuthSnapshot.TenantUUID`, already fails closed (`if err != nil || tenantID == uuid.Nil`). Nothing else in the `authz` package parses a snapshot field: `BudgetAnchorAt` is declared on the edge snapshot but never read. The two lookalikes elsewhere in `apps/edge-api` are not the same class, and I read both rather than pattern-matching on the grep: `limits/budget_gate.go` uses the shape only to set an advisory `Soft` flag, while its hard cap propagates read errors and returns an error on a malformed value, and `batches/handler.go` uses it on a pagination limit. So nothing else needs the same change, which is the more useful half of this question.

### Framing (the one thing I would change before merge)

The "Why this is hardening, not a vulnerability fix" section of the description is accurate and I would keep it word for word. My concern is which text survives the squash. The durable artifact in the history is the subject line, "fix: deny access when an API key's expiry timestamp is present but unparseable", and read on its own it will eventually be cited as evidence that a live expiry bypass existed, which on this PR's own investigation it did not. Please carry that paragraph, or a two-line version of it, into the squash commit body, so the disclaimer travels with the commit rather than only with the pull request page.

### Informational, not introduced here, no change requested

`checkOWUIShimKey` in `apps/edge-api/cmd/server/main.go` does not go through `CheckAccess`; it re-checks status, tenant and the model allowlist by hand. Its own comment states the rule that matters here: the probe's predicate must stay at least as strict as the request path's, or the probe becomes a false green (#717). After this change there is one more state where it is weaker, a shim key with a malformed `expires_at` reported usable while every request path denies it. The same unreachability argument that makes this PR hardening covers it too, so at most a follow-up issue, and certainly not a change to this diff.

apps/edge-api/internal/authz/authz_test.go
sakibsadmanshajib: Took the suggestion as given: added the DenyMsg assertion for the malformed-record message. Also picked up CodeRabbit's separate nitpick (whitespace-only ExpiresAt regression case) in the same push since it was a one-line addition testing a real branch (TrimSpace). New head c1ace906, both tests passing. Thanks for catching the untested distinguishing signal.

sakibsadmanshajib:
## Addressed independent review, new head c1ace906

- **Framing**: amended the commit so the not-a-vulnerability disclaimer (two independent layers already enforce expiry, this branch is unreachable from control-plane, reaching it would require writing to the shared Redis cache which is already full compromise) leads the commit body itself, not just the PR description. This is what will squash into the permanent record.
- **Test nit**: added the DenyMsg assertion on `TestCheckAccessDeniesUnparseableExpiresAt` (malformed record vs genuinely-expired key now both pinned, not just DenyCode).
- **CodeRabbit's separate nitpick**: added `TestCheckAccessWhitespaceOnlyExpiresAtStillAllowed` covering the TrimSpace branch explicitly.
- **checkOWUIShimKey informational note**: added to issue #923 as an addendum rather than a third issue, since it is the same probe-weaker-than-path shape already tracked there.

All three review threads resolved. Full `apps/edge-api/internal/authz` suite reconfirmed green at the new head (`go build`, `go test -count=1`).

Head SHA: `c1ace906f3d8cd34401cc4cb392868d014776ca4`.

Issue #841, opened 2026-08-10
Permanent session-slot consumers against the 15 client ceiling: Open WebUI never recycles, control-plane LISTEN is checked out for process life


sakibsadmanshajib:
Follow-up to #837, raised in review of that PR. Relates to #631.

The session-mode pooler is capped at 15 clients shared by every consumer. #837 shortens how long an idle pgxpool connection squats one, which lowers the base rate. It does not touch the consumers below, which hold slots by construction and belong in any capacity math.

## 1. Open WebUI holds session slots indefinitely

Its SQLAlchemy pool sets no `pool_recycle`, so a connection lives until the process restarts. Worse, its psycopg2 path falls back to the session DSN when a hand-written local `.env` supplies one, so an operator who filled in `SUPABASE_DB_URL` by hand and skipped `SUPABASE_DB_POOL_URL_LIBPQ` gets a permanent session-mode consumer without knowing it. Nothing in #837 reaches this: the fix is a pgx DSN parameter and Open WebUI does not speak pgx.

Worth deciding: whether to enforce the transaction-mode DSN for Open WebUI at startup the way `platformdb.Open` enforces session mode for control-plane, rather than relying on the operator having used `scripts/derive-pooler-dsn.py`.

## 2. control-plane's LISTEN connection is one permanent slot per instance

`apps/control-plane/internal/tenant/settings/listener.go` acquires a pool connection and holds it for the life of the process on `LISTEN tenant_settings_changed`. A checked-out connection is never idle, so the idle reaper cannot reclaim it, and it counts toward `TotalConns`.

This is correct behaviour and not a bug. Transaction mode returns the connection after each transaction, which would silently stop notifications and leave every instance serving stale tenant settings from cache, which is exactly why `platformdb.Open` refuses port 6543. Recording it because it is fixed overhead: one slot per running control-plane instance, before any query traffic, and it is easy to forget when counting the budget.

## Why this matters now

Measured on 2026-08-10 during the #837 investigation, twice, 25 minutes apart, with no CI job running at all: six parallel session-mode connections were refused with `EMAXCONNSESSION`. Zero free slots with no CI involved. Long-lived consumers, not CI, are what keeps the pool at its ceiling.

## Acceptance

- A written inventory of every session-mode consumer with its permanent and peak slot cost, checked against the 15 ceiling.
- A decision on whether Open WebUI's DSN should be enforced rather than derived by convention.
[...]

Issue #821, opened 2026-08-09
Tracking: make Web E2E (full stack) a required check, and the coverage programme that gets us there


sakibsadmanshajib:
@i821.new.md

sakibsadmanshajib:
This issue is an explicit "Tracking:" issue for a multi-step coverage programme; PR #825 (spec selection) is one step of that programme, not its completion. Leaving open. Remaining: the remaining steps of the coverage programme and the actual flip to a required check

PR #927, opened 2026-08-17
fix: budget an ephemeral CI stack a share of the session pool it can actually get

sakibsadmanshajib:
## The three retry-passes are one cause, and it is not in the specs

`Web E2E (full stack)` failed on `main` in run 31958356695 for flakiness alone: 24 passed, 0 failed, 3 passed only on retry, gate tripped at `MAX_FLAKY_TESTS = 0`. Nothing here raises the retry count, quarantines a test, extends a timeout or touches the gate.

### First correction: which three

The dispatch named `profile-completion.spec.ts:39` and `:65`. Those two passed on the first attempt in that run. The gate's own table and the job log name these:

| Spec | Test | Attempts |
| --- | --- | --- |
| `auth-shell.spec.ts:51` | accepting an invitation keeps current workspace until switcher changes it | 2 |
| `profile-completion.spec.ts:94` | profile settings stay reachable while unverified | 3 |
| `profile-completion.spec.ts:145` | dashboard does not introduce a billing reminder | 3 |

`profile-completion.spec.ts` runs `test.describe.configure({ mode: "serial" })`, so the failure of `:94` skipped the four tests after it and retried the whole file from `:39`. That is why `:39` and `:65` appear three times in the log with `(retry #1)` and `(retry #2)` while never having failed, and it is the likeliest source of the misattribution.

### What the failing attempts show

Every failing first attempt is a timeout, not an assertion mismatch.

- `auth-shell.spec.ts:51`: `locator.inputValue` on `select[name='account_id']` never resolved, and the error context captured for that attempt is the console error boundary, `Something went wrong on this page ... reference 3955933083`. The workspace switcher was genuinely absent: the page had failed to render.
- `profile-completion.spec.ts:94` and `:145`: `page.waitForURL` in the shared `signIn` helper timed out at 25000ms, with the submit button still reading `Signing in...`.

### The environment, in the same job

- `[e2e-auth-fixtures] reseed took ...` ranged from 315ms to **76188ms** within the one run.
- Job wall time 9.1m. The previous seven Web E2E runs on `main` took 1.4m, 1.9m, 2.1m, 3.2m, 4.4m, 4.6m and 1.6m and every one of them reported `27 executed, 0 passed only on retry`. These specs have a clean record under this gate since it landed on 2026-08-10, so they are not marginal tests that finally tipped over.
- The job's own `compose-logs-web-e2e-api` artifact carries **258** lines of:

  `FATAL: (EMAXCONNSESSION) max clients reached in session mode - max clients are limited to pool_size: 15 (SQLSTATE XX000)`

- Control-plane errors in that same dump map onto the failures one for one: `ERROR could not load your workspace err="accounts: list memberships: failed to connect ..."` (the error boundary) and `ERROR auth: tenant membership check failed, denying selected tenant`, from 16:29:55 through 16:38:51.
- The same CI run's `Live integration` job hit `EMAXCONNSESSION` too, at 16:30:13. The two jobs overlapped: web-e2e 16:24:39 to 16:39:04, live-integration 16:24:41 to 16:31:59.
- Other stack-booting runs were live in the same window: two PR `CI` runs (16:22 to 16:39, 16:30 to 16:46) and an `agent visual proof` run (16:22 to 16:31).

Verdict: contention, not a spec-level code change. #908 and #906 are not implicated; the suite passed twice after they merged and the run that tripped the gate is the one whose reseeds ran 25x slower than the same reseed in the same job.

### The arithmetic

One Supabase project serves the always-on demo box and every CI job that boots a stack. Supavisor's session-mode ceiling for it is 15 clients, and a session-mode client pins one server connection for its whole session. control-plane must be on session mode (`LISTEN tenant_settings_changed`, session-scoped `pg_advisory_lock`), which `platformdb.Open` enforces at startup.

`scripts/derive-pooler-dsn.py` budgeted **6** of those 15 to every control-plane it derived a DSN for. That number was written for the single long-lived deployment. Three ephemeral CI stacks boot per push:

| Consumer | Session budget before | after |
| --- | --- | --- |
| `ci.yml` web-e2e control-plane | 6 | 4 |
| `ci.yml` live-integration control-plane | 6 | 4 |
| `agent-visual-proof.yml` control-plane | 6 | 4 |
| demo box control-plane (always on) | 6 | 6, now asked for by name |
| **one push, ephemeral total** | **18 of 15** | **12 of 15** |

### Measured, not assumed

Taken from a developer machine on 2026-08-16 with no CI job running, over the transaction-mode port so the probe itself costs no session slot:

- `max_connections` = 60 with 26 client backends. The database is not the constraint; Supavisor's per-tenant `pool_size` of 15 is.
- Six parallel session-mode connections: **2 accepted, 4 refused** with `EMAXCONNSESSION`.

So roughly two session slots are free at rest and the long-lived consumers hold the rest, which matches issue #841's own measurement of 2026-08-10.

**This has to be said plainly: this pull request reduces demand, it does not create headroom.** A single ephemeral stack asking for even two slots is asking for most of what is free. The changes that add capacity are raising the project's pooler ceiling (#841's acceptance item, and `max_connections` says there is room at the database) or moving CI onto its own project or a stack-local Postgres (#631). Neither is applied here: the first is a live-project config change while sibling agents are verifying against the live box, which `.wolf/cerebrum.md` warns against, and the second is #631's whole scope.

## What changed

**`scripts/derive-pooler-dsn.py`**
- Default session budget 6 to 4, with `--session-max-conns` for a consumer that needs more. Four is what the reservation path needs, not a round number: `PgxAccountLocker.WithAccountLock` checks out one connection per account holding the advisory lock and then runs `fn`, which needs a further connection for its ledger and usage writes, so a pool of `max` survives `max - 2` concurrent account locks once the listener's permanent connection is subtracted. Four leaves room for two concurrent accounts; three would allow exactly one, and trading a contention timeout for a starvation timeout is not a fix (`pglock.go` documents that starvation shape in its own comment).
- Below 2 is refused with a message that says why, rather than producing a DSN that wedges control-plane on its first query.
- The default is now the ephemeral value, so a future workflow that forgets the flag fails safe instead of demanding a third of the ceiling.
- Self-test extended from 33 to 40 assertions, watched failing first (`AssertionError: ...pool_max_conns=6...`).

**`.github/workflows/deploy-demo-box.yml`**
- Passes `--session-max-conns 6`. The live surface's budget is unchanged in value; it is now explicit and local to the consumer it was written for.

**`.github/workflows/ci.yml`** (web-e2e job, failure path only)
- Counts `EMAXCONNSESSION` in the compose dump it already writes and emits a GitHub error annotation naming it. Diagnosis, not dispensation: the run has already failed and stays failed. Finding this cause today meant downloading a 460KB artifact and grepping it by hand while the job log said only `3 flaky`.
- Uploads the Next.js server log, piped through `scripts/redact-log-credentials.py` exactly as the compose dump is, because this job drives `/invitations/accept?token=<live token>` and artifacts on a public repository are downloadable by anyone with the run URL. The `auth-shell:51` error boundary's cause exists only in that log, which is currently written to `/tmp/nextjs.log` and thrown away with the runner.

## Test plan

- [x] `python3 scripts/derive-pooler-dsn.py --self-test` fails on the new assertions before the change (`AssertionError` showing the old `pool_max_conns=6`), passes after: `self-test ok (40 assertions)`.
- [x] CLI checked by hand: no flag gives `pool_max_conns=4`, `--session-max-conns 6` gives 6, `--session-max-conns 1` exits non-zero naming the listener.
- [x] Transaction-mode DSNs unmoved by the session flag (`pool_max_conns=8`, and still stripped for libpq).
- [x] Both edited workflows parse.
- [ ] Repeated `Web E2E (full stack)` runs on this branch, reported in a comment below with the run count and the retry-pass count. A green sequence cannot disprove a probabilistic contention flake and is not offered as if it could; the deterministic evidence is the arithmetic and the measured free-slot count above.

## Buglog entry

```json
{"id":"BUG-2026-08-16-web-e2e-session-pool","date":"2026-08-16","title":"Web E2E failed the flake gate on main: three specs passed only on retry because the shared Supabase session pool was exhausted","error_message":"FATAL: (EMAXCONNSESSION) max clients reached in session mode - max clients are limited to pool_size: 15 (SQLSTATE XX000); surfaced as page.waitForURL timeout 25000ms and locator.inputValue timeout in three unrelated specs","root_cause":"scripts/derive-pooler-dsn.py budgeted 6 of the project's 15 Supavisor session-mode slots to every control-plane it derived a DSN for, a number sized for the single long-lived demo box. Three ephemeral CI stacks boot per push (ci.yml web-e2e, ci.yml live-integration, agent-visual-proof.yml), so one push demanded 18 of 15 before the box's own 6, and control-plane's DB calls were refused mid-run. The console then rendered its error boundary or never finished the post-sign-in render, and the specs reported timeouts that read as three independent races.","fix":"Default session budget cut from 6 to 4 with a new --session-max-conns flag and a floor of 2 (the tenant-settings listener pins one pool connection for process life, and a pool of max only survives max minus 2 concurrent account locks); deploy-demo-box.yml now asks for 6 explicitly, so an ephemeral consumer that forgets the flag fails safe. Web E2E additionally names EMAXCONNSESSION in an error annotation on failure and uploads the redacted Next.js server log, since the error boundary's cause lived only there.","tags":["ci","e2e","flake","supabase","supavisor","pooler","session-mode","playwright"],"refs":["run 31958356695","job 95192819866","#631","#841"]}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1495,7 +1495,7 @@ jobs:
           if-no-files-found: ignore
           retention-days: 5
 
-      - name: Upload the Next.js server log on failure
+      - name: Redact the Next.js server log on failure
         if: failure()
         # The console renders an error boundary ("Something went wrong on this
         # page", plus a digest) when a server component throws, and the cause is
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1464,12 +1464,61 @@ jobs:
         # Piped through the redactor: see the sdk-tests dump step above.
         run: docker compose logs --no-color --timestamps --tail=500 2>&1 | python3 ../../scripts/redact-log-credentials.py > /tmp/compose-web-e2e-api.log || true
 
+      - name: Name a shared session-pool exhaustion on failure
+        if: failure()
+        # Diagnosis, not dispensation. The run has already failed and stays
+        # failed; this only saves the next reader the archaeology.
+        #
+        # On 2026-08-16 this job failed the flake gate with three unrelated
+        # specs passing only on a retry, and every one of them was a timeout
+        # caused by `FATAL: (EMAXCONNSESSION) max clients reached in session
+        # mode`, which appeared 258 times in the dump this step now reads.
+        # Finding that meant downloading the compose artifact and grepping it by
+        # hand, while the job log said nothing but "3 flaky". A retry-pass is
+        # still a failure and the gate is untouched; what changes is that the
+        # log now says which failure it was.
+        run: |
+          set -uo pipefail
+          log=/tmp/compose-web-e2e-api.log
+          [ -f "$log" ] || exit 0
+          n=$(grep -c EMAXCONNSESSION "$log" || true)
+          if [ "${n:-0}" -gt 0 ]; then
+            echo "::error title=Shared session pool exhausted::${n} EMAXCONNSESSION refusals in this job's container logs. The Supabase project's Supavisor session-mode ceiling is 15 clients, shared by this job, every other CI stack booted at the same time and the always-on demo box. Failures in this run are contention, not necessarily a product defect; see issues #631 and #841. Do not raise the retry count or the timeouts to survive it."
+          fi
+
       - name: Upload compose logs
         if: failure()
         uses: actions/upload-artifact@v4
         with:
           name: compose-logs-web-e2e-api
           path: /tmp/compose-web-e2e-api.log
+          if-no-files-found: ignore
+          retention-days: 5
+
+      - name: Upload the Next.js server log on failure
+        if: failure()
+        # The console renders an error boundary ("Something went wrong on this
+        # page", plus a digest) when a server component throws, and the cause is
+        # only ever in this log. On 2026-08-16 an invitation-acceptance spec
+        # timed out waiting for a workspace switcher that the error boundary had
+        # replaced, and nothing in any uploaded artifact said why: this file was
+        # written and then discarded with the runner.
+        #
+        # Redacted on the way out, like the compose dump: artifacts on a public
+        # repository are downloadable by anyone with the run URL, and this job
+        # drives /invitations/accept?token=<live token>.
+        run: |
+          set -uo pipefail
+          if [ -f /tmp/nextjs.log ]; then
+            python3 scripts/redact-log-credentials.py < /tmp/nextjs.log > /tmp/nextjs-redacted.log || true
+          fi
+
+      - name: Upload the redacted Next.js server log
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: nextjs-log-web-e2e
+          path: /tmp/nextjs-redacted.log
           if-no-files-found: ignore
           retention-days: 5
 
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -452,7 +452,15 @@ jobs:
           python3 scripts/derive-pooler-dsn.py --self-test
           # stdout is the KEY=value pairs; the script logs a password-free
           # host/port/params summary on stderr.
-          python3 scripts/derive-pooler-dsn.py --dsn "$existing" >> "$GITHUB_ENV"
+          #
+          # --session-max-conns 6 keeps this box's budget exactly where it has
+          # been. Six is no longer the script's default: an ephemeral consumer
+          # now gets four, because three CI stacks boot per push and six each
+          # asked for 18 of the project's 15 session slots before this box's own
+          # share was counted. This is the one long-lived deployment, it serves
+          # real chat traffic, and it is the consumer that budget was written
+          # for, so it asks for it by name.
+          python3 scripts/derive-pooler-dsn.py --dsn "$existing" --session-max-conns 6 >> "$GITHUB_ENV"
 
       - name: Rebuild + recreate changed services
         # `--build` only rebuilds layers whose inputs changed (Docker build
--- a/scripts/derive-pooler-dsn.py
+++ b/scripts/derive-pooler-dsn.py
@@ -43,6 +43,16 @@ below for the arithmetic: with the old single value of six, one push asked for
 18 of the 15 slots across its three CI stacks, and the visible symptom was
 unrelated browser specs passing only on a retry.
 
+That split reduces demand; it does not create headroom, and nothing in this
+file can. Three ephemeral stacks at four plus the long-lived six is still 18
+against a ceiling of 15, so two pushes landing together can still exhaust the
+pool. The changes that add capacity are raising the project's Supavisor
+`pool_size` (the database itself has room: `max_connections` is 60) or moving
+CI onto its own project, which are issues #841 and #631. Serialising the jobs
+instead was tried and reverted, because a GitHub Actions concurrency group
+holds only one queued run and cancels the rest, which trades a noisy failure
+for a silent absence.
+
 The session DSN additionally carries `pool_max_conn_idle_time` and
 `pool_health_check_period`. A cap alone bounds how many slots a consumer may
 take; it does nothing about how long it keeps them. pgxpool holds an idle
@@ -223,14 +233,6 @@ def derive(
     session_max_conns is this consumer's share of the project's 15 session slots.
     It defaults to the ephemeral budget; a long-lived deployment passes its own.
     """
-    if session_max_conns < MIN_SESSION_MAX_CONNS:
-        raise ValueError(
-            f"session_max_conns={session_max_conns} is below {MIN_SESSION_MAX_CONNS}: "
-            "control-plane pins one pool connection for the life of the process on "
-            "LISTEN tenant_settings_changed, so a smaller cap leaves nothing to run "
-            "queries on"
-        )
-
     parts = urlsplit(dsn)
     host = parts.hostname
     if not host:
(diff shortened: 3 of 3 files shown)


sakibsadmanshajib:
## Adversarial review, streams and findings

Three streams ran against this diff. Every finding is answered below, with the commit that answers it.

### Stream 1: CodeRabbit CLI (`coderabbit review --agent --committed --base main`)

First pass on `15b570cd`: 0 findings across all three files. Second pass on `c5850869`, after the default was retuned: 2 findings, both major, both real.

**Finding 1, `scripts/derive-pooler-dsn.py`: three ephemeral consumers at four plus the long-lived six still exceed the documented 15.** Correct, and it is the honest limit of this change rather than an oversight. Answered by documenting it in the file itself rather than only in this pull request (`62bb3c56`), because the next reader of the constant is the one who needs to know it. The three alternatives it proposed, in order: reducing further re-creates the starvation shape that fixed the value at four (a pool of `max` survives `max - 2` concurrent account locks); serialising the jobs was already tried and reverted in this repository, since a GitHub Actions concurrency group holds only one queued run and cancels the rest, which is the silent-absence trade the web-e2e job's own comment warns about; raising the upstream ceiling is the change that adds capacity and it is a live-project config change, tracked on #841 with #631 as the structural fix. This pull request deliberately does not take either of those on.

**Finding 2, the floor is not enforced on the effective value.** A real defect, fixed in `62bb3c56`. `with_params` lets an explicit `pool_max_conns` in the input DSN win over the script's own, by design, so validating only the `--session-max-conns` argument let an input DSN of `?pool_max_conns=1` through untouched. The validation now reads what the session DSN ends up carrying, names both possible sources in its error, and rejects a non-numeric value too. Three cases added to the self-test (`?pool_max_conns=1`, `=0`, `=lots`), and the existing case that pins `=2` still passes, since two is the floor and not below it.

This also matters concretely rather than hypothetically: the demo box's stored `SUPABASE_DB_URL` already carries `pool_max_conns=6`, so its effective cap comes from the DSN and the new `--session-max-conns 6` agrees with it. Had someone hand-edited that value to 1, every deploy would have shipped a control-plane that wedges on its first query with no diagnostic.

### Stream 2: plain adversarial pass (self, before CodeRabbit's second run)

**The number itself was wrong at 3, and I caught it before pushing it as final.** `accounting.PgxAccountLocker.WithAccountLock` checks out one connection per account holding the advisory lock and then runs `fn`, which needs a further connection for its ledger and usage writes. `pglock.go` documents that starvation shape in its own comment. So a pool of `max` survives `max - 2` concurrent account locks once the listener's permanent connection is subtracted: three would have allowed exactly one account at a time in the live-integration job, and traded a contention timeout for a starvation timeout. Four leaves room for two, above anything CI drives. Fixed by amend before review, and the arithmetic is in the constant's comment so the next person does not have to re-derive it.

Also checked and clear:
- `tools/lint-no-pgx-dsn-for-libpq.mjs` parses this script for `PGX_ONLY_PARAMS` and for every parameter stamped through `with_params`. No new stamped parameter, no new emitted variable, so it is unaffected.
- The transaction-mode DSNs do not move with the session flag, asserted in the self-test.
- Both edited workflows parse as YAML.
- The new failure-path steps run from the repository root, which is where `scripts/redact-log-credentials.py` resolves; the compose dump step next to them sets `working-directory: deploy/docker` and reaches it as `../../scripts/...`, so the two are consistent rather than accidentally different.
- `grep -c` exits 1 when it finds nothing, so the annotation step uses `|| true` and a `${n:-0}` default rather than failing the step on a clean log.

### Stream 3: security review of the new artifact upload

The Next.js server log is a new published artifact on a public repository, and this job drives `/invitations/accept?token=<live token>`. It goes through `scripts/redact-log-credentials.py`, the same filter the compose dump already uses, whose parameter list covers bare `token=` and every `*_token`, `*_key`, `password`, `secret` and `code` spelling, after a fragment as well as in a query string. PR #578 is the precedent for why that is not optional. The upload is `if: failure()` only, and retention is five days, matching the artifacts beside it.

No stream was skipped. `ecc:code-review` and the specialist reviewer agents are not reachable from this agent's toolset, so streams 2 and 3 were run by hand against the same checklist rather than reported as run when they were not.

sakibsadmanshajib:
Correction to one claim in the review comment above, before anyone relies on it.

I wrote that the demo box's stored `SUPABASE_DB_URL` already carries `pool_max_conns=6`, so its effective cap comes from the DSN. That was read off the derivation script's own stderr summary, which prints the DSN it produced, not the one it was given. Checking the stored value directly: it carries no query parameters at all, on port 5432.

The consequence runs the other way and is worth stating plainly. `--session-max-conns 6` in `deploy-demo-box.yml` is not a restatement of something the box already says; it is the only thing keeping the box at six. Without that flag in the diff, this pull request would have quietly cut the live surface's budget from six to four along with CI's. The flag is load bearing, and the deploy workflow is the place it belongs.

What does not change: the effective-value validation from `62bb3c56` still matters for the same reason CodeRabbit gave, since a hand-edited `.env` is exactly the path by which an unusable cap would reach control-plane, and the box's `.env` is hand-maintained.

sakibsadmanshajib:
## Repeat-run evidence, and the honest reading of it

Five `Web E2E (full stack)` runs on this branch, all with the new ephemeral budget of four.

| # | Job | Result | Flake report | Wall time | Notes |
| --- | --- | --- | --- | --- | --- |
| 1 | 95325064536 | success | 27 executed, 0 passed only on retry | 3.2m | Web E2E alone |
| 2 | 95329549220 | **failure** | 27 executed, **1** passed only on retry | 6.1m | `Live integration` booted a second stack at the same second, deliberately |
| 3 | 95332698087 | success | 27 executed, 0 passed only on retry | 5.3m | |
| 4 | 95338990265 | success | 27 executed, 0 passed only on retry | 4.7m | |
| 5 | 95341516091 | success | 27 executed, 0 passed only on retry | 1.5m | |

Four of five clean, and none of the three specs from run 31958356695 appeared in any of them.

**That is not proof, and it is not offered as proof.** Four green runs cannot disprove a probabilistic contention flake; the previous seven runs on `main` were green too, immediately before the one that tripped the gate. The deterministic evidence for this change is the arithmetic and the measured free-slot count in the description, not this table. What the table does establish is the absence of a regression: the smaller pool did not starve anything, which was the live risk in cutting it.

### Run 2 is the useful one, and it failed for a different reason

`Live integration` is label gated on pull requests, so I added `run-live-integration` deliberately, precisely because that job is where a smaller pool could starve rather than help: it drives real reservations through `PgxAccountLocker`, which checks out a connection per account holding the advisory lock. **It passed, in 5m34s, with four connections.** That is the risk check this change most needed, and running the two jobs concurrently is also the closest reproduction available of the conditions that broke `main`.

The Web E2E job in that same window did trip the gate, once, and the cause is neither the session pool nor a spec:

```
1) console-budgets.spec.ts:77 › budget caps are editable for the owner and disabled for a member
   Error: Command failed: node tests/e2e/support/e2e-auth-fixtures.mjs
   memberships upsert failed: Could not query the database for the schema cache. Retrying.
```

That is PostgREST failing a fixture write in `beforeEach`, so the test failed without ever running, and the retry passed. Same shared project, different door: the seeder writes through `/rest/v1`, not through the pooler this pull request budgets. Another agent hit the same seeder one call earlier the same evening with `tenants upsert failed: canceling statement due to statement timeout`. Filed as #934 with the suggested shape, kept out of this pull request on purpose: retrying an idempotent setup write is a different argument from this one and deserves its own review, not a smuggled ride on a change whose whole thesis is that patience is not a fix.

For completeness, since it matters to whoever schedules the follow-ups: `Web E2E (full stack)` is not in the six required contexts in `.github/branch-protection-main.json`, so a red run here has never blocked a merge. What it blocks is the promotion tracked in #821, and everyone's ability to read the board.

Issue #106, opened 2026-04-25
[P0/critical] Credit reservation TOCTOU race — concurrent inference can double-spend credits


sakibsadmanshajib:
## Severity
**Critical — block free-tier public launch.** Surfaces in adversarial Go review (C-1) and DB review (C-2).

## Problem
`accounting/service.go:63-109` — `GetBalance` and `CreateReservation` run in two separate transactions with no row-level lock between them. N concurrent inference requests for the same account all observe the same available balance and all pass `enforcePolicy`.

Exploit: open N parallel sockets, send N inference requests, each successfully reserves up to the full balance — yielding N× actual liability. With free credits granted at signup, this lets one attacker drain the free-credit allotment many times over and force Hive to pay the upstream provider bill.

## Fix
- Wrap balance read + reservation insert in a single Postgres transaction.
- Use `SELECT ... FOR UPDATE` on a per-account lock row, or a Postgres advisory lock keyed on `account_id`.
- Add a CHECK constraint on the ledger so the running sum cannot go negative (defence in depth).

## Acceptance
- 100-RPS smoke test against a single account with balance=1000 and per-request reservation=50 finalizes ≤20 reservations.
- Race test (`-race`) covers the new path.

PR #929, opened 2026-08-17
fix: write a reservation and its hold in one transaction, and stop rendering an over-release as a hold

### LINKED ISSUE #918: Reservation row and its ledger hold are not atomic, so the reaper can release credits that were never held
`CreateReservation` writes the `credit_reservations` row and posts the `reservation_hold` ledger entry in two separate transactions (`accounting/repository.go` opens its own transaction, then `ledgerSvc.ReserveCredits` opens another through `ledger.PostEntry`). A hold that fails to post therefore leaves a reservation row claiming credits the ledger never held.

The consequence is not that the customer is overcharged, it is the reverse: the stranded-hold reaper later releases that row, posting a `reservation_release` entry with no matching `reservation_hold`, which credits the account for credits it never had held against it.

Measured live 2026-08-16, read only: 10 reservations, 25,000 credits, all created between 2026-07-17 and 2026-07-26 and all released in one pass on 2026-08-01 13:05. Each has zero `reservation_hold` entries and exactly one `reservation_release` entry.

Second, related consequence, worth fixing in the same pass or explicitly deciding not to: `ledger.GetBalance` computes `reserved` as `ABS(SUM(reservation_hold) + SUM(reservation_release))`. If an account's releases ever exceed its holds, that `ABS` flips the sign and presents the excess as a phantom hold, subtracting it from the customer's available balance for work that never existed. No account is currently in that state (every signed net is negative today), which is the only reason this has not already produced a support ticket.

Directions worth weighing rather than a prescribed fix:

- Make the row and the hold atomic, so the row cannot exist without its hold. It needs a transaction that spans the accounting repository and the ledger repository, which neither is currently written for.
- Or make the release bounded by what the ledger actually holds for that reservation, which needs a per-reservation aggregate and therefore an index on `credit_ledger_entries.reservation_id`, which does not exist.
- Or compensate in `CreateReservation`'s error path, which is unsafe as stated: a timeout cannot distinguish a hold that failed from one that succeeded with a lost response, and marking the row terminal in the second case strands a real hold where the reaper can no longer reach it.

Found while fixing #616 (PR #912), and deliberately left out of that PR: opposite sign, different root cause, and no live remediation should ride along on a money-path fix.

sakibsadmanshajib:
Closes #918.

## Root cause verdict: still reachable, not historical residue

The mechanism is the two-transaction create, and the sequence that produces it is this one:

1. `accounting.Service.CreateReservation` starts the attempt, then calls `repo.CreateReservation`, which commits the `credit_reservations` row and its `reserved` event in its own transaction.
2. It then calls `ledgerSvc.ReserveCredits`, a second, independent transaction.
3. That second transaction fails (a pool timeout, a cancelled statement, any transient database error). `CreateReservation` returns the error and the caller fails closed, correctly refusing to serve.
4. The committed row survives with no hold behind it. An hour later the stranded-hold reaper sees a non-terminal reservation, releases it, and posts a `reservation_release` against a hold that never existed. The account is credited for credits that were never taken from it.

Evidence for that sequence rather than the alternatives, read live and read only:

- All ten reservations have their `reserved` event, so step 1 committed.
- None has a `reservation_hold` **ledger entry**, and none has a `reservation_hold` row in `credit_idempotency_keys` either. That second fact rules out deduplication as the explanation: a hold that had been deduplicated against an earlier entry would have left the key row behind. The entry write never ran, or ran and rolled back whole.
- Each has exactly one request attempt, a unique `request_id`, `attempt_number` 1. Not retries.
- Their release entries are all present, eight stamped `stranded_hold_reaper` on 2026-08-01 and two `probe_cleanup` on 2026-07-26, which is step 4.
- Endpoints: six `/v1/audio/transcriptions`, two `/v1/audio/speech`, two `completions`. Three different callers, all of which route through this one create path, so nothing about the shape is specific to one caller.
- Neighbouring reservations on the same account in the same minute do have holds, so the ledger was working generally; this is transient failure, not an outage.

**Still reachable.** `git log -S` puts the `ledgerSvc.ReserveCredits` call in `CreateReservation` since the original Go port (#89), and no commit since has touched that ordering. The four changes to `service.go` in July and August are all on the settlement side. Nothing has closed this window; the ten rows cluster in July because that is when the trigger fired, not because the path changed afterwards.

## Is `ABS()` in `GetBalance` defensible

No, and it is why this went unnoticed for a month.

The signed sum of `reservation_hold` plus `reservation_release` over an account can only be negative or zero, because a release can only ever return what a hold took. A positive sum is not a balance, it is proof that the ledger is internally inconsistent for that account. `ABS()` takes that impossible number, strips the sign, and hands it to the customer as though it were credits on hold.

It is worse than a wrong sign, because the netting happens first. `9bbaf7b5` has three genuine in-flight holds worth 12,000 and 23,000 of phantom credit; the account-wide sum is +11,000, so `ABS` reports 11,000 reserved and the customer's available balance is **overstated by 1,000**. `3104dbfc` has nothing in flight and 2,000 of phantom credit; it reports 2,000 reserved and its available is **understated by 2,000**. Wrong in both directions, on the same defect, as this PR's regression test reproduces with those exact numbers.

The right response, and what this PR does:

- Compute reserved per reservation, clamped at zero, then sum. A reservation that shows more released than held contributes nothing, so a corrupt row can no longer move a customer-facing number in either direction. `9bbaf7b5` will read 12,000 reserved, its true in-flight position; `3104dbfc` will read zero.
- Report the corruption instead of absorbing it. `BalanceSummary` gains `OverReleasedCredits` and `OverReleasedReservations`, and `Service.GetBalance` logs them at error level on every read, naming the account. Both fields are `json:"-"`, so the customer-facing payload keeps its shape and the signal goes to operators.
- Do not fail the read. Refusing to answer would take an affected account off the air without repairing anything, and the number now being returned is the honest one. Failing closed is for authorizing spend, and that path reads the same corrected number.

As you predicted, this surfaces both accounts loudly: every balance read for them logs an error until the ledger rows are reconciled.

Cost: the same rows the old query already scanned under the `(account_id, created_at)` index, with a hash aggregate on `reservation_id` on top. No new index, and no cross-account work.

## The invariant, as a bound

For every reservation, at every moment, in the database rather than in a convention:

```
a row in credit_reservations exists  <=>  a reservation_hold entry exists for it
```

and therefore, for every account, `SUM(hold) + SUM(release) <= 0` per reservation, with any positive value being corruption rather than a hold.

As magnitudes: across any attempt to create a reservation, successful or not, the count of reservation rows and the count of hold entries move together. Either both increase by one, or neither moves. A partially applied create leaves nothing behind, not a row, not an event, not a key.

## Proof by ledger magnitude on real Postgres, red before and green after

Both halves are proven against the real schema built from `supabase/migrations/`, through the real repositories, not through fakes.

`TestReservationRowAndHoldAreWrittenAtomically_Live` injects the failure at the database rather than through a stub, because the guarantee under test IS a transaction boundary. The injected shape is real: an idempotency key claimed without its ledger entry, the poisoned-key state from issue #663, which makes the hold write fail after the row insert, at exactly the point the old sequence committed the row anyway.

```
pre-fix (hold posted in a second transaction):
  FAIL  reservation rows (1) and hold entries (0) diverged:
        a row with no hold is what the reaper later releases against nothing
post-fix:
  PASS  (and the positive control still writes one row and one hold)
```

`TestGetBalanceDoesNotRenderAnOverReleaseAsAHold_Live` builds the production shape exactly: one settled reservation, one genuinely in flight worth 12,000, one released without ever being held worth 23,000.

```
pre-fix (ABS over the account):
  FAIL  reserved = 11000, want 12000: only the one hold that is genuinely
        outstanding. ABS nets the 23000 phantom against it and reports 11000,
        a hold that does not exist
post-fix:
  PASS  reserved 12000, over-released 23000 across 1 reservation
```

Both are `HIVE_TEST_DB_URL`-gated, and `./internal/ledger/...` is added to the live-Postgres CI step so the second one actually runs; `tools/lint-go-db-test-wiring.mjs` passes with 20 wired pairs.

Full `go test ./apps/control-plane/... ./apps/edge-api/...` green, `go vet` clean.

## Also fixed here: the expand path had the identical defect

`ExpandReservation` committed the raised `reserved_credits` and then posted its hold in a second transaction, the same shape as create, with worse consequences: settlement releases what the ROW says is held, so a row raised from 500 to 2500 whose hold post failed would hand back 2500 against 500 ever taken. Nothing calls it in production today, which is why no row shows that shape, but it is exported and one caller away from reproducing this issue. It is now atomic, with its own live proof.

Both write paths go through one helper, which also refuses a deduplicated hold entry that belongs to a different reservation: `PostEntryTx` returns the stored entry when a key was already used, and an entry carrying another reservation's id would leave this row committed with no hold of its own.

## Test fakes that had to move with the code

The hold is no longer posted through the ledger service on create, so two fakes were changed rather than left asserting a call the system no longer makes:

- `repoStub` records the hold and, when a test's subject is the balance, posts it to that test's fake ledger, because otherwise those fixtures would silently start from an unreserved balance and their assertions would stop meaning anything.
- `concurrentRepo` applies the hold to the shared balance itself, since the TOCTOU race test (#106) is about the balance read and the hold write being serialized, and the hold write now lives in the repository.

Three assertions that counted `ledgerSvc.reserveCalls` on create now assert on the recorded hold, and one asserts that no second-transaction reserve happens at all.

## Cleanup for the 25,000, enumerated and NOT run

Not bundled into this change, and not executed. After this deploys the phantom stops affecting any balance, because a per-reservation clamp ignores it; the cleanup exists to make the rows self-consistent so that future queries, and any tool that nets per account, do not trip over them.

The truthful repair is a compensating `reservation_hold` for the hold that never posted, which makes each of the ten reservations net to zero. Not an `adjustment`: posted was never inflated by this defect, so moving posted would take real credits away.

```sql
-- PREVIEW, read only. Ten rows, 25000 credits as of 2026-08-17.
WITH orphan AS (
  SELECT r.id AS reservation_id,
         r.account_id,
         COALESCE(sum(e.credits_delta) FILTER (
            WHERE e.entry_type IN ('reservation_hold','reservation_release')), 0) AS net
  FROM public.credit_reservations r
  LEFT JOIN public.credit_ledger_entries e ON e.reservation_id = r.id
  GROUP BY r.id, r.account_id
)
SELECT count(*) AS reservations, sum(net) AS credits_released_without_a_hold
FROM orphan
WHERE net > 0;

-- REPAIR, to run only with the owner's go-ahead:
-- INSERT INTO public.credit_ledger_entries
--   (account_id, entry_type, credits_delta, idempotency_key, reservation_id, metadata)
-- SELECT account_id, 'reservation_hold', -net,
--        'reservation:' || reservation_id || ':reserve-missing',
--        reservation_id,
--        jsonb_build_object('reason', 'issue-918 hold that never posted', 'pr', '<this PR>')
-- FROM orphan
-- WHERE net > 0
-- ON CONFLICT DO NOTHING;
```

Append-only, no UPDATE, no DELETE, re-runnable, and the key shape `reserve-missing` cannot collide with a real hold's `reserve` key. Expect it to change no displayed balance after this PR ships, which is the point: it repairs the record, and the fix already stopped the record from lying.

## Buglog entries

```json
{"id":"bug-mhold918-4c7e21","timestamp":"2026-08-17T09:20:00.000Z","related_bugs":[],"occurrences":10,"last_seen":"2026-07-26T12:51:04.000Z","title":"Reservation row and its ledger hold were written in two transactions, so the reaper released credits that were never held","error_message":"Ten reservations carry a reservation_release entry with no matching reservation_hold, 25000 credits, and GetBalance rendered the resulting positive sum as a phantom hold","root_cause":"accounting.Service.CreateReservation committed the credit_reservations row and its reserved event in one transaction, then posted the ledger hold in a second. A transient failure on the second left a committed row claiming credits the ledger never held; the stranded-hold reaper later released that row and credited the account for credits never taken. Compounded by ledger.GetBalance computing reserved as ABS of one account-wide sum, which turned the resulting positive sum into a plausible-looking hold and netted it against genuine in-flight holds, making customer-facing available wrong in both directions.","fix":"The accounting repository writes the row, its reserved event and the hold in one transaction via the new ledger.PostEntryTx, so a failed hold rolls the row back with it. GetBalance sums reserved per reservation clamped at zero instead of taking ABS of the account-wide sum, and reports a positive per-reservation net separately as OverReleasedCredits, logged at error level on every read.","tags":["billing","ledger","credit-reservation","money-path","atomicity","issue-918","issue-616","control-plane","D-034"]}
{"id":"bug-mpool-ro-9a13f2","timestamp":"2026-08-16T17:04:00.000Z","related_bugs":[],"occurrences":1,"last_seen":"2026-08-16T17:04:00.000Z","title":"SET default_transaction_read_only outside a transaction persists on the shared Supavisor backend and makes later clients read-only","error_message":"cannot execute INSERT in a read-only transaction, on a connection that had issued no such SET, with pg_settings reporting source=session and the role config carrying no such setting","root_cause":"A read-only guard was issued as a bare `SET default_transaction_read_only = on` at the top of psql query files run through the transaction-mode pooler on port 6543. A SET outside a transaction applies to the pooled SERVER connection, not to the client session, and Supavisor does not reset it between clients, so every later client handed that backend inherits it. Three distinct backends were left read-only this way.","fix":"Reset the setting on each affected backend with RESET default_transaction_read_only, repeated until every backend the pooler hands out reports off, and never issue a bare SET on 6543 again. The correct read-only guard on a transaction-mode pooler is BEGIN TRANSACTION READ ONLY ... COMMIT, which is scoped to the transaction and cannot leak to the next client.","tags":["postgres","supabase","pooler","supavisor","transaction-mode","port-6543","read-only","tooling","gotcha"]}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -436,7 +436,11 @@ jobs:
             # ledger on real Postgres rather than through a fake, so leaving it
             # unlisted is the same never-runs trap as issues #701 and #708, on
             # the money path.
-            go test -tags integration -count=1 -p 1 -v ./internal/accounts/... ./internal/accounting/... ./internal/agenttask/... ./internal/egress/... ./internal/payments/... ./internal/profiles/... ./internal/rag/... ./internal/tenants/... ./internal/signup/... ./internal/routing/... ./internal/litellmconfig/... ./internal/catalog/... ./internal/providers/... ./internal/platform ./internal/platform/db/...
+            # ./internal/ledger/... joins the list with issue #918: its balance
+            # query is the only place an over-released reservation can be caught,
+            # and the proof that it no longer renders one as a hold has to run
+            # against real Postgres.
+            go test -tags integration -count=1 -p 1 -v ./internal/accounts/... ./internal/accounting/... ./internal/ledger/... ./internal/agenttask/... ./internal/egress/... ./internal/payments/... ./internal/profiles/... ./internal/rag/... ./internal/tenants/... ./internal/signup/... ./internal/routing/... ./internal/litellmconfig/... ./internal/catalog/... ./internal/providers/... ./internal/platform ./internal/platform/db/...
           else
             # -tags integration additionally compiles in
             # inference/chat_completions_integration_test.go (issue #708),
--- a/apps/control-plane/internal/accounting/repository.go
+++ b/apps/control-plane/internal/accounting/repository.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -43,6 +44,18 @@ func NewPgxRepository(pool *pgxpool.Pool) Repository {
 }
 
 func (r *pgxRepository) CreateReservation(ctx context.Context, reservation Reservation, reason string, hold ReservationHold) (Reservation, error) {
+	// Boundary guard, before any database work. A zero-value hold would post a
+	// 0-credit entry under an empty idempotency key and commit, leaving a row
+	// that looks authorized against a ledger holding nothing: the very state
+	// this signature exists to prevent. The service validates its own input,
+	// but the next caller of this method is not bound by that.
+	if hold.Credits <= 0 {
+		return Reservation{}, fmt.Errorf("accounting: reservation hold must be greater than zero, got %d", hold.Credits)
+	}
+	if strings.TrimSpace(hold.IdempotencyKey) == "" {
+		return Reservation{}, fmt.Errorf("accounting: reservation hold requires an idempotency key")
+	}
+
 	metadata, err := json.Marshal(map[string]any{
 		"policy_mode":    reservation.PolicyMode,
 		"request_id":     reservation.RequestID,
--- a/apps/control-plane/internal/accounting/repository.go
+++ b/apps/control-plane/internal/accounting/repository.go
@@ -10,10 +10,22 @@ import (
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/ledger"
 )
 
+// ReservationHold is the ledger hold that must land in the same transaction as
+// the reservation row it authorizes (issue #918). Passing it into
+// CreateReservation rather than posting it separately afterwards is what makes
+// "a reservation row exists if and only if its hold exists" a database
+// guarantee instead of an ordering convention.
+type ReservationHold struct {
+	IdempotencyKey string
+	Credits        int64
+	Metadata       map[string]any
+}
+
 type Repository interface {
-	CreateReservation(ctx context.Context, reservation Reservation, reason string) (Reservation, error)
+	CreateReservation(ctx context.Context, reservation Reservation, reason string, hold ReservationHold) (Reservation, error)
 	GetReservation(ctx context.Context, accountID, reservationID uuid.UUID) (Reservation, error)
 	ExpandReservation(ctx context.Context, accountID, reservationID uuid.UUID, additionalCredits int64, reason string) (Reservation, error)
 	FinalizeReservation(ctx context.Context, accountID, reservationID uuid.UUID, consumedCredits, releasedCredits int64, terminalUsageConfirmed bool, status ReservationStatus, reason string) (Reservation, error)
@@ -30,7 +42,7 @@ func NewPgxRepository(pool *pgxpool.Pool) Repository {
 	return &pgxRepository{pool: pool}
 }
 
-func (r *pgxRepository) CreateReservation(ctx context.Context, reservation Reservation, reason string) (Reservation, error) {
+func (r *pgxRepository) CreateReservation(ctx context.Context, reservation Reservation, reason string, hold ReservationHold) (Reservation, error) {
 	metadata, err := json.Marshal(map[string]any{
 		"policy_mode":    reservation.PolicyMode,
 		"request_id":     reservation.RequestID,
@@ -69,6 +81,24 @@ func (r *pgxRepository) CreateReservation(ctx context.Context, reservation Reser
 		VALUES ($1, 'reserved', $2, $3, $4::jsonb)
 	`, created.ID, created.ReservedCredits, reason, metadata); err != nil {
 		return Reservation{}, fmt.Errorf("accounting: insert reserved event: %w", err)
+	}
+
+	// The hold goes in THIS transaction (issue #918). It used to be posted by
+	// the service afterwards, through the ledger's own transaction, so a hold
+	// that failed to post left a committed reservation row claiming credits the
+	// ledger never held; the reaper then released that row and credited the
+	// account for credits it never had taken. Ten production reservations, 25000
+	// credits, are in that state. Here a failed hold rolls the row back with it.
+	if _, err := ledger.PostEntryTx(ctx, tx, created.AccountID, ledger.PostEntryInput{
+		EntryType:      ledger.EntryTypeReservationHold,
+		CreditsDelta:   -hold.Credits,
+		IdempotencyKey: hold.IdempotencyKey,
+		RequestID:      reservation.RequestID,
+		AttemptID:      &created.RequestAttemptID,
+		ReservationID:  &created.ID,
+		Metadata:       hold.Metadata,
+	}); err != nil {
+		return Reservation{}, fmt.Errorf("accounting: post reservation hold: %w", err)
 	}
 
 	if err := tx.Commit(ctx); err != nil {
--- a/apps/control-plane/internal/accounting/repository.go
+++ b/apps/control-plane/internal/accounting/repository.go
@@ -28,7 +28,7 @@ type ReservationHold struct {
 type Repository interface {
 	CreateReservation(ctx context.Context, reservation Reservation, reason string, hold ReservationHold) (Reservation, error)
 	GetReservation(ctx context.Context, accountID, reservationID uuid.UUID) (Reservation, error)
-	ExpandReservation(ctx context.Context, accountID, reservationID uuid.UUID, additionalCredits int64, reason string) (Reservation, error)
+	ExpandReservation(ctx context.Context, accountID, reservationID uuid.UUID, additionalCredits int64, reason string, hold ReservationHold) (Reservation, error)
 	FinalizeReservation(ctx context.Context, accountID, reservationID uuid.UUID, consumedCredits, releasedCredits int64, terminalUsageConfirmed bool, status ReservationStatus, reason string) (Reservation, error)
 	ReleaseReservation(ctx context.Context, accountID, reservationID uuid.UUID, releasedCredits int64, reason string) (Reservation, error)
 	CreateReconciliationJob(ctx context.Context, reservationID, requestAttemptID uuid.UUID, reason string) error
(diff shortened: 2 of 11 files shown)


sakibsadmanshajib:

Streams run against `b0e21d92`, fixes in `53b9983f` and `6d6e331f`.

| stream | status |
| --- | --- |
| CodeRabbit CLI (`coderabbit review --committed --base main`) | ran, 3 findings (1 major, 2 minor) |
| Independent security and correctness pass, separate model instance, read-only, money-path brief | ran, 7 findings (1 high, 4 medium, 2 low) |
| Database stream | run by the reviewing agent in role: schema, query plans on a scratch database with the real migration chain, index behaviour, and the live population read read-only through the pooler |
| `/codex:adversarial-review` | **SKIPPED**, CLI returned a usage limit and ran nothing. Not a pass |
| `ecc:code-review` skill | **SKIPPED**, structural capability gap: this agent has no Skill tool |

### The one that mattered: the same defect was live on the expand path

`ExpandReservation` had the identical two-transaction shape this PR was fixing for create, and worse consequences, because settlement releases what the ROW says is held: expand a reservation from 500 to 2500, lose the hold post, and the eventual release hands back 2500 against 500 ever taken. Same mechanism and same magnitude class as the ten production rows.

It is now atomic too, with its own live proof (`TestExpandReservationRaisesTheRowAndTheHoldTogether_Live`), and both write paths route through one helper rather than two copies of the same sequence. Nothing calls `ExpandReservation` in production today, which is why no row shows this shape, but it is exported, reachable, and was one caller away from reproducing the whole issue.

### Every other finding

**Deduplicated hold belonging to another reservation (medium).** `PostEntryTx` returns the stored entry when a key was already used, correct for a replay and wrong here: an entry carrying a different reservation's id would leave the row committed with no hold of its own. The helper now compares the returned entry's `reservation_id` and fails if it does not match. Unreachable from the service, which derives the key from a fresh UUID, but `Repository` is exported and its contract promises the row and hold exist together unconditionally.

**NULL `reservation_id` grouping (major, CodeRabbit).** SQL puts every NULL in one group, so hold and release rows without a reservation id would have netted against each other: account-wide netting reintroduced for exactly those rows. Verified live before fixing: only grants carry a NULL `reservation_id` (29 rows), no hold or release does. Grouping now falls back to the entry id for NULLs, so a stray release is reported as the anomaly it is.

**CTE materialization on the hot path (medium).** Postgres inlines a CTE only when referenced exactly once; the first version referenced `entries` twice and therefore materialized the account's whole ledger slice into a tuplestore on every call, inside the per-account critical section. Rewritten as two index-driven scans with no CTE. Measured on a scratch database with the real schema, one account holding 5000 entries across 1667 reservations among 55000 rows: **6.2 ms, no Materialize or CTE Scan node**, against the old ABS query's 3.4 ms. That is the honest price of the correct number, and the docstring now says so rather than claiming the change was free.

**Round trips, since it is being tracked separately: this PR removes 2 per reservation create and 2 per expand.** Not from the balance query, which was one round trip before and one now. They come from the atomicity change: posting the hold used to open its own transaction, so `BEGIN` and `COMMIT` were two extra round trips per create; the hold now rides the transaction that was already open. The two inserts it performs are unchanged. Nothing else in the metered path is touched here, so this is a small dent in that number, not a fix for it.

**Unbounded error logging (medium).** The anomaly condition is persistent by construction and `GetBalance` runs per request, so the line would repeat at whatever rate the tenant sends traffic. Throttled to once per account per fifteen minutes, in process, with the ceiling named in a comment.

**Two assertions that could not fail (medium and low).** The atomicity tests asserted only that an error occurred, which also holds for a rejection that returns before the transaction opens, so moving poisoned-key detection into a pre-flight check would have kept them green while proving nothing about the boundary they exist to pin; they now assert the failure came from the hold post. The lock assertion could not distinguish "inside the lock" from "called at all"; the fake locker now exposes whether it was held at call time, and the fake repository records it. Mutation-verified: flipping the fake to report "not held" turns it red with the #106 message.

**Seed failure treated as a skip (low, found by both streams).** The new ledger suite skipped when its account seed failed. The CI leg bootstraps the full migration chain, so a seed failure there is a schema regression and a skip would report it green: the same never-runs trap the ci.yml comment three lines above the edited line describes. Now fatal.

### Explicitly clean, from the security stream

No new deadlock cycle: the create transaction touches `credit_reservations`, `credit_reservation_events`, `credit_idempotency_keys`, `credit_ledger_entries` in one order, and no other path takes those in a different one. No fail-closed weakening: the balance check still precedes the hold, both still run inside the per-account advisory lock, and a failed hold now fails the whole create instead of leaving a row behind. No cross-tenant effect: every statement is filtered by `account_id`. No secret, token, or customer-facing detail in the new error paths; the anomaly log carries an internal account UUID, consistent with existing logging.

PR #936, opened 2026-08-17
fix: remove the unused Open WebUI chat surfaces from the image, not just from one compose file

sakibsadmanshajib:
Owner instruction, 2026-08-17: the chat interface is "overly future-proofed with unnecessary features" and the unused elements must go. This is the subtraction half only. The information architecture and visual redesign belong to the separate agent already working on them, so nothing here designs replacement navigation.

## What was actually wrong

The three surfaces the owner named (Notes, Calendar, Automations) were already off on the demo box, and only there. `docker-compose.yml` names three environment variables and `owui-patches/hive_rag_env_config.py` reconciles them onto that deployment's database. That is a complete control for one deployment and no control at all anywhere else: upstream's own defaults are all on, so a plain `docker run` of the Hive image rendered every one of them in full. The `before` capture in this pull request produces the owner's list word for word, which is what identifies the image, rather than the deployment, as what he was looking at. Off has to be a property of the product, not of one compose file that a bare run, an enterprise override or the next proof harness can fail to repeat.

The second half of the question was whether hiding a menu entry is honest in each case. For three of the four flag-backed surfaces it is: `calendar.py` checks its flag on all 13 of its routes, `automations.py` on all 8, `memories.py` on all 11. `notes.py` checks it on none of its 9, so with the entry gone and the page route already 404'd, `POST /api/v1/notes/create` still created a note for any signed-in user. That one gets a proxy rule; the other three deliberately do not, because a second control there would be decoration.

## Changes

| file | change |
| --- | --- |
| `Dockerfile.open-webui` | `ENV ENABLE_NOTES/ENABLE_CALENDAR/ENABLE_AUTOMATIONS/ENABLE_MEMORIES/ENABLE_VERSION_UPDATE_CHECK=false`, so the image itself ships the reduced product. Docker `ENV` rather than a rewrite of upstream's own `os.getenv` lines: no literal to drift on a digest bump, and compose still overrides it, so an enterprise deployment that wants Notes back sets the variable to true and the reconcile flips the persisted row |
| `docker-compose.yml` | adds `ENABLE_MEMORIES: "false"`. The compose values stay, all of them: on a deployment that has already booted they are still the only thing that reaches the database |
| `hive_rag_env_config.py` | adds `memories.enable` to the reconciled set |
| `hive_ui_surfaces.py` | two new bundle rewrites, `settings-admin-link` and `settings-vendor-translate-link` |
| `Caddyfile.owui` | `api/v*/notes` added to `@blocked` |
| `scripts/test_owui_ui_surfaces.py`, `scripts/test_caddy_owui_blocklist.py` | assertions for all of the above, written before the changes |

Settings > Personalization is the only tab removed. Its feature is upstream's own Memory: nobody chose it, `ENABLE_MEMORIES` defaults to true, and while it is on `Chat.svelte` adds `features.memory: true` to every turn, so each message also pays for a lookup against a second memory store that Open WebUI owns and Hive does not read. Hive's memory subsystem is one subsystem on Hive's own Postgres (`.wolf/decisions.md` D-020) and is not built yet, so this is a competing store rather than a head start on it.

The two new bundle rewrites were found by looking at the screen this change was already photographing, not by re-reading an issue. The Settings dialog's bottom-left "Admin Settings" link is the twin of the user menu's Admin Panel entry removed in #909 and points at the same `/admin` route Caddy already 404s, so every tenant owner sees a permanent control that always fails. The language picker still carried "Help us translate Open WebUI!", linking a Hive customer to the vendor's contributing guide by name, which is the same family as the Documentation and Releases links removed in #772.

## Two different promises, said plainly

Removing a surface from one compose file while the feature still ships in the image is a half-removal that reads as complete and is not. This pull request makes the promise explicit per surface, because they are not the same promise.

**Gone from the built image.** Cannot be brought back by configuration on any deployment, because the entry no longer exists in the client bundle. Workspace Models, Prompts, Tools and Skills; the Playground in both the sidebar and the user menu; the user menu's Admin Panel; the vendor's Documentation and Releases links; the release-notes dialog; the vendor branding on Settings > About; the Settings > Integrations tab; and now the Settings dialog's "Admin Settings" link and its "Help us translate Open WebUI!" link.

**Off in the image, end to end, and re-enablable on purpose.** Notes, Calendar, Automations, and the Memory feature behind Settings > Personalization. The image now ships all four false, so the reduced product is what any run gets, and the navigation entry, the page and the API go together because each router refuses on the same flag. A deployment that genuinely wants one back sets the variable to true and the reconcile flips the persisted row, which is why these stay flags rather than becoming bundle rewrites. Before this change all four were on in the image and off only where `docker-compose.yml` was present, which is exactly the half-removal the title names.

**One residual, stated rather than papered over.** `/api/v1/notes` is closed at the proxy, not in the image. `notes.py` is the one router in that set that checks its flag on none of its routes, so with `notes.enable` false its 9 endpoints still answer and `Caddyfile.owui` is what refuses them. Every deployment this repository ships puts that Caddy in front of Open WebUI, so there is no shipped configuration where those endpoints are reachable, but a bare `docker run` of the image is not covered. Closing it in the image would mean unmounting an upstream router in `main.py`, which would also make the feature un-re-enablable by configuration and so break the promise directly above. Deleting the upstream routers and Svelte pages outright is a fork-and-rebuild move that belongs to the redesign, not to a subtraction pass.

## Enumeration

Taken from the pinned image itself, `ghcr.io/open-webui/open-webui:v0.10.2@sha256:9fcea9c...`, not from the deployment and not from memory. Worth recording separately: that image ships JavaScript source maps whose `sourcesContent` carries the complete original Svelte source, 612 files. The repeated claim in `owui-patches/` that the image ships "no frontend source" is true only of the build tooling. Reading a gate no longer requires reverse-engineering minified identifiers.

### Sidebar

| entry | verdict | why |
| --- | --- | --- |
| New Chat | keep | the product |
| Search | keep | chat history search, used |
| Notes | remove | upstream note-taking, no Hive counterpart, no Hive code reads it |
| Workspace | keep | its only remaining tab is Knowledge, which is the RAG document upload |
| Folders | keep | chat organisation |
| Chats, pinned chats, archived | keep | the product |
| Channels | already off | upstream default is off; group chat, not a Hive feature |
| pinned Models section | keep | renders only when a user pins a model |

### User menu

| entry | verdict | why |
| --- | --- | --- |
| Settings | keep | reduced, see below |
| Archived Chats | keep | real, reachable, used |
| Workspace | keep | as above |
| Notes, Calendar, Automations | remove | as above |
| Sign Out | keep | the product |
| Admin Panel, Playground, Documentation, Releases, Keyboard Shortcuts help block | already removed | #772, #909 |
| Update your status, and the profile header it is wrapped in | design question, see below |
| Active Users count | design question, see below |

### Settings dialog

Nine tabs upstream, eight after #771 removed Integrations, six after this change.

| tab | verdict | why |
| --- | --- | --- |
| General | keep | theme, language, per-user system prompt, advanced parameters. Its vendor translation link is removed here |
| Interface | keep for now | see the design question below. Roughly forty upstream toggles, which is the "confusing" the owner named, but reducing it is composition, not subtraction |
| Personalization | remove | upstream Memory, nobody chose it, competes with D-020 |
| Audio | keep | speech-to-text and text-to-speech, a locked demo capability |
| Data Controls | keep | chat import, export, archive-all, delete-all |
| Account | keep, with a flag raised below | name and profile image are real. Its API keys, JWT token, notification webhook and password-change blocks are already hidden by upstream defaults this deployment does not change, correctly, since the Hive console owns keys and every account is SSO-only |
| About | keep | already de-branded by #784 |
| Connections | already off | `ENABLE_DIRECT_CONNECTIONS` defaults off; direct browser-to-provider connections would bypass the gateway |
| Integrations | already removed | #771 |

## Deferred as design questions

Each of these is a removal whose only available mechanism also takes something else, or whose answer is where the feature moves rather than whether it dies. Left alone deliberately, for the redesign owner.

1. **Update your status, and the user menu's profile header.** The presence feature is unambiguously unused. Its only flag, `ENABLE_USER_STATUS`, gates the whole `{#if profile}` block, so turning it off also removes the avatar, name and email from the top of the menu. Removing the status controls alone needs a bundle rewrite, and whether the identity header belongs in that menu at all is a navigation decision.
2. **Active Users count.** Its flag does not reach this deployment's audience: the gate is `enable_public_active_users_count || role === 'admin'`, and every tenant owner is an Open WebUI admin here, so hiding it needs a bundle rewrite. Small, but it is a decision about what the menu is for.
3. **Settings > Interface.** Around forty upstream toggles, most of which no Hive user will ever want. Reducing it is the settings rebuild, explicitly out of scope for this pull request.
4. **Settings > Account bio, gender and birth date.** Ungated personal-data collection with no consumer anywhere in Hive: nothing reads these fields. Flagged rather than removed because it sits inside a tab the redesign is about to rework, and because deleting input fields for data some accounts may already hold is a data question, not a navigation one.
5. **Per-message rating.** `ENABLE_MESSAGE_RATING` defaults on, so every assistant message carries thumbs up and down, and the feedback rows they write are only ever surfaced in Open WebUI's own admin Evaluations page, which is blocked. The data goes nowhere. Not reachable from the sidebar, the user menu or Settings, so it is outside this pull request's enumeration.
6. **Code interpreter and code execution.** Both default on and put a toggle in the chat input. Whether they overlap Cowork and the coding agent, or complement them, is a product call.
7. **Follow-up, tag and title auto-generation.** All default on and each spends a real model call per turn. A cost decision, not a surface one.

## Files touched, for collision checking

The navigation-shell work should not overlap any of these. Eight files outside `docs/proof/`:

```
deploy/docker/Caddyfile.owui
deploy/docker/Dockerfile.open-webui
deploy/docker/docker-compose.yml
deploy/docker/owui-patches/hive_rag_env_config.py
deploy/docker/owui-patches/hive_ui_surfaces.py
deploy/docker/owui-patches/pinned-bundle-excerpts.json
scripts/test_caddy_owui_blocklist.py
scripts/test_owui_ui_surfaces.py
```

`hive_ui_surfaces.py` and its `pinned-bundle-excerpts.json` fixture are the one real collision risk, since any other change that removes a surface by bundle rewrite lands in the same table and regenerates the same fixture. Nothing else here is frontend code at all.

## Test plan

- [x] `make test-scripts`, which is what CI runs for these files. New assertions were written first and observed failing: image ENV absent, `ENABLE_MEMORIES` absent from compose and from the reconcile, `/api/v1/notes` not blocked, memories missing from the coverage table.
- [x] Image built from this branch, the bundle rewrite pass reporting all 22 surfaces matched and both guards intact.
- [x] Before and after images built from the same pinned digest, menus and settings tabs read back out of the DOM.
- [x] Notes API probed with a real session token, direct and through a real Caddy on both configurations.
- [x] Whole interface run through the branch's own Caddy, recording every console error and every 4xx or 5xx: one pre-existing 404 on `/api/v1/terminals/` (#770) and no request to `/api/v1/notes` at all.
- [x] `npm run lint:proof-tokens`.
- Not run: the six locked demo capabilities are untouched by this diff. Chat, embeddings, voice, knowledge, Cowork and the coding agent have no path through Notes, Calendar, Automations, Memory, the Settings admin link or the vendor translation link, and the kept APIs were probed alongside the blocked one.

## Visual proof

`docs/proof/owui-surface-subtraction-2026-08-17/`, per the visual-proof rule. Attached below as well.

## Buglog entry

```json
{"id":"owui-surfaces-image-vs-compose-2026-08-17","date":"2026-08-17","title":"Removed Open WebUI surfaces were off in one compose file and on in the image","error_message":"Notes, Calendar and Automations rendered in the user menu and sidebar, and Settings > Personalization in the settings dialog, on any run of the Hive Open WebUI image that did not repeat docker-compose.yml's ENABLE_* lines, including the repo's own before/after proof builds. Separately, POST /api/v1/notes/create returned 200 and created a note with notes.enable false.","root_cause":"Two independent gaps with the same shape. First, the removal of a product surface was expressed only as environment variables in docker-compose.yml plus a persisted-config reconcile, so it reached exactly one deployment and nothing else; upstream's defaults for all of them are on, and PR CI never builds Dockerfile.open-webui, so image and compose could disagree indefinitely with no signal. Second, a feature flag was assumed to be a complete control without checking that the backend honours it: calendar.py, automations.py and memories.py check theirs on every route, notes.py checks its own on none of its 9, so the flag hid the navigation entry and left the API fully callable.","fix":"Dockerfile.open-webui sets ENABLE_NOTES, ENABLE_CALENDAR, ENABLE_AUTOMATIONS, ENABLE_MEMORIES and ENABLE_VERSION_UPDATE_CHECK false as image ENV defaults, so the reduced product is a property of the image while compose keeps overriding it and the reconcile keeps reaching already-booted databases. Caddyfile.owui adds api/v*/notes to @blocked. scripts/test_owui_ui_surfaces.py gains _image_env() and asserts the image, the compose value and the reconcile entry together for every flag-backed surface, plus that a non-persisted flag is deliberately absent from the reconcile.","tags":["open-webui","docker","feature-flags","caddy","persisted-config","dead-code","ui"]}
```

Closes nothing on its own; the redesign issue tracks the rest.


sakibsadmanshajib:
## Visual proof

Both images built from the same pinned upstream digest and the same `Dockerfile.open-webui`, differing only in the files this pull request touches, and run with no compose file at all. Full method, DOM readouts and API probe in `docs/proof/owui-surface-subtraction-2026-08-17/README.md`.

### User menu

Before, which is the owner's list word for word:

user menu before

After:

user menu after

### Settings dialog

Before, with Personalization in the tab list, "Admin Settings" in the bottom-left corner, the vendor's translation link under the language picker, and the update nag bottom right:

settings before

After:

settings after

### Sidebar

Before, with Notes:

sidebar before

After:

sidebar after

### Read out of the DOM, not off the pixels

```
[before] sidebar entries: New Chat, Search, Notes, Workspace, Folders, Chats
[after]  sidebar entries: New Chat, Search, Workspace, Folders, Chats

[before] user menu: Update your status, Settings, Archived Chats, Workspace, Notes, Calendar, Automations, Sign Out
[after]  user menu: Update your status, Settings, Archived Chats, Workspace, Sign Out

[before] settings tabs: General, Interface, Personalization, Audio, Data Controls, Account, About
[after]  settings tabs: General, Interface, Audio, Data Controls, Account, About

[before] settings admin links: 1        [after] settings admin links: 0
[before] settings vendor links: 1       [after] settings vendor links: 0
```

### The notes API, probed with a real session token

```
notes.enable in /api/config:
  enable_notes = False  enable_calendar = False  enable_automations = False  enable_memories = False

authenticated, straight at the after image (no proxy):
  GET  /api/v1/notes/        -> 200
  POST /api/v1/notes/create  -> 200
  GET  /api/v1/calendars/    -> 403
  GET  /api/v1/memories/     -> 404

authenticated, through Caddy as configured on main:
  GET  /api/v1/notes/        -> 200
  POST /api/v1/notes/create  -> 200

authenticated, through Caddy as configured on this branch:
  GET  /api/v1/notes/        -> 404
  POST /api/v1/notes/create  -> 404
  GET  /api/config           -> 200
  GET  /api/v1/chats/        -> 200
  GET  /api/v1/knowledge/    -> 200
```

Sessions are throwaway local `WEBUI_AUTH=false` ones. No real account, no password, and no URL in any capture carries a credential.

sakibsadmanshajib:
## Adversarial review, stage 6

| stream | status | outcome |
| --- | --- | --- |
| CodeRabbit CLI (`coderabbit review --agent --base main`) | ran, three passes | 2 findings, both accepted and fixed, third pass clean at `02dc643bb` |
| CodeRabbit GitHub app | SKIPPED, not a clean pass | the check reports "Review rate limited" on this pull request. The CLI covered the same diff |
| Plain adversarial pass | ran | notes below, one real gap found and closed |
| `ecc:code-review`, `/codex:adversarial-review` | SKIPPED, structural | this builder has no Skill or slash-command tooling. Stating it rather than treating the absence as a pass |
| Domain specialists | not selected | the diff is a Dockerfile `ENV`, a compose value, a proxy regex, two bundle literals and their tests. No Go, no TypeScript, no SQL, no auth, money or input-parsing path |

### CodeRabbit findings

1. **major, `capture.mjs`.** The proof script imported `playwright-core` by one machine's absolute path, so nobody else could re-run it. Fixed at `4bdc3c62f`: resolved relative to the script itself, with `PLAYWRIGHT_CORE` as the override a git worktree needs, since a worktree has no `node_modules` of its own. The suggested plain relative import was taken as the primary path but it cannot be the only one for that reason.
2. **major, `capture.mjs`.** The console-error and failed-request recorders were reset *after* `page.reload()` resolved, so they discarded everything the reload produced and the empty result proved nothing about the load it claimed to measure. This one mattered: it was a check that could not go red. Fixed at `02dc643bb`, and the proxied run's honest result is now one pre-existing 404 on `/api/v1/terminals/` (#770's block, whose only caller returns an empty list on any non-2xx) rather than a clean sheet.

### Plain adversarial pass

What I went looking for, and what I found.

- **Does the image `ENV` affect the build itself?** No. Every `RUN` in this Dockerfile is a text patch or an `ast.parse`; none imports `open_webui.config`, so none reads these variables at build time.
- **Does an enterprise deployment lose the override?** No. `docker-compose.enterprise.yml` adds no `open-webui` environment of its own, so it inherits the same service block, and compose still beats the image `ENV` in both directions. A deployment that wants Notes back sets the variable to true and the reconcile flips the persisted row.
- **Does the reconcile now write a key nothing reads?** No for `memories.enable`, which is in `DEFAULT_CONFIG` and read by `/api/config` through `config.get`. Deliberately no entry for `ENABLE_VERSION_UPDATE_CHECK`, which is a plain environment read, and a test asserts that absence, because a reconcile entry for it would be a no-op shaped like a control. That is the `ENABLE_ADMIN_PANEL` mistake this repo already made once.
- **Are the two new bundle literals unique?** Both counted exactly once across every `.js` in the pinned bundle before being written, and `apply_ui_surfaces_patch.py` asserts the same at build time. The build log for this branch reports all 22 rewrites matched and both guards intact.
- **Does blocking `/api/v1/notes` break anything the interface still does?** Every front-end caller renders only under `$config.features.enable_notes`: the sidebar's pinned note list, the message input's Attach Notes tab and its note-from-prompt action, the note link token, and `KnowledgeSelector`, which lives inside the already-removed Workspace > Models. Verified rather than argued: the `after-proxied` capture runs the whole interface through the branch's own Caddy and records no request to that router at all, which the block would have surfaced as a 404 if there had been one.
- **Do the six locked demo capabilities have a path through any of this?** No. The kept APIs were probed alongside the blocked one in `api-notes-block.log`, and no repository test, spec or workflow references Notes, Calendar, Automations or Personalization.
- **Can any new test pass while the thing it measures is broken?** The five new assertions were each observed failing before the change, which is in the test plan. The one I had written badly was the capture reset above, and CodeRabbit caught it; that is exactly the shape this repo has been bitten by before, a green signal that cannot go red.

### Not fixed here, deliberately

The seven items in the pull request body's "deferred as design questions" section. Each is either a removal whose only available mechanism also takes something else with it (the profile header, the Active Users count), or a question of where a feature moves rather than whether it dies (the Interface tab, the Account tab's personal-data fields), or outside the enumeration this pull request was scoped to (message rating, the code interpreter toggle, follow-up generation). They belong to the redesign, not to a subtraction pass.

PR #938, opened 2026-08-17
feat: build the chat frontend from a forked source tree and give it one Hive navigation

sakibsadmanshajib:
Wave one of `spec-2026-08-16-hive-ui-redesign` (Obsidian vault), which is the fork plus the navigation shell. It answers the three things the owner named: the left navigation carries neither the agent workspace nor the chat interface, the product still looks like Open WebUI, and the agent is reached through a button in the top right corner that opens a separate page.

All three were properties of shipping a prebuilt vendor bundle, so this stops doing that.

## What lands

**The fork.** `vendor/open-webui` is upstream at `v0.10.2`, added as a squashed git subtree. `Dockerfile.open-webui` gains a first stage that compiles its frontend the way upstream's own Dockerfile does (`node:22-alpine3.20`, `npm ci --force`, `npm run build`) and copies the result over `/app/build` in the pinned image. The Python backend, its dependency set and every existing backend patch are untouched, so a digest bump keeps delivering backend fixes for free. The stage asserts the vendored `package.json` version equals the version inside the pinned image, so a bump without a matching subtree pull fails the build rather than a browser.

Measured, cold, no cache: `npm ci` about 90 seconds, `npm run build` about 420 seconds. Warm, with the vendored tree untouched, both layers are cache hits. `owui-nightly.yml` boots the stack with `--build`, so its timeout moves from 30 to 45 minutes; `docker-bake.hcl` gains an `open-webui` target so the image is one command anywhere.

**The identity.** `packages/hive-tokens/tokens.css` carries the design system (`spec-2026-08-11-hive-design-system`) as plain CSS custom properties, outside the vendored tree on purpose so it survives a change of chat engine. Open WebUI funnels its entire surface and ink scale through fourteen values in one `@theme` block; remapping those to the brand's warm ramp is what turns the product from grey to cream and charcoal in one place rather than a hundred. Lightness is held within a step or two of upstream's own ramp, so no contrast pair inverts. Hanken Grotesk and Geist Mono are served from the deployment, not from a font CDN.

**The shell.** `vendor/open-webui/src/lib/hive/` holds every Hive authored piece: the navigation as data, a row component, and the stylesheet. Chats, Agents and Knowledge are labelled destinations, present in both sidebar states, with the current row carrying a coral bar as well as `aria-current`. Upstream's own files carry four insertion points a rebase can replay, listed in `docs/owui-fork.md`.

**The agent inside the shell.** `/agents` is a route in the chat application that renders the agent workspace over the same origin, so reaching the agent no longer leaves the product. The injected launcher overlay, `loader.js`, and the measured `right: 120px` apparatus in `custom.css` are deleted with it.

**Surfaces removed in source rather than in a bundle.** The Workspace tabs with no Hive counterpart and their routes, the Playground, the Admin Panel entry, the vendor's documentation, releases, changelog dialogue, social badges and copyright, the Settings Integrations tab, and the vendor links on the backend-required error screen. The accessible names the patch layer used to add are attributes in source now.

## Visual proof

`docs/proof/shell-nav-2026-08-17/` and the comment below. Two images built from the same pinned digest, differing only in this diff, captured in both themes with the DOM read as well as the pixels, the way PR #909 did it. Reproducible with the `run.sh` and `capture.mjs` committed beside the captures.

Checked against the design system's own floors, from `dom.json` rather than by eye: nav row height 32 matching the token, 44 under a coarse pointer, the two ring focus treatment (2px canvas then 2px coral) so the ring stays visible on the accent itself, transition 120ms which is `--hv-duration-fast`, and 0s under `prefers-reduced-motion`.

## What is deliberately not here

**The bundle rewrite layer is not retired.** It still runs, against the bundle it was written for, and the frontend built from source then replaces that bundle. That reads as waste and it is deliberate for one release. Those rewrites are exact literals over minified output and the identifiers in them are allocated per chunk, so any edit to our own source renames them: the first build of this stage failed on `sidebar-playground-item`, `changelog-modal` and `about-vendor-social-badges`, three surfaces that had not actually come back, and an identifier-tolerant pattern for two of them then matched sibling components that must not be touched. Every surface those rewrites remove is removed in source here, so nothing regresses; what they still buy is the digest-drift guard.

Retiring them properly is the natural next change and it is out of this one on purpose, because #936 was in flight against `hive_ui_surfaces.py` and two edits to that file from different directions is how a surface silently returns. **`hive_ui_surfaces.py`, `apply_ui_surfaces_patch.py` and `remove_integrations_tab.py` are byte identical to `main` in this branch.**

**Wave two and beyond.** The composer mode control, the greeting, the suggestion chips, the settings reduction from nine sections to five, the console group in the same sidebar, Artifacts as a nav row (it needs its host origin injected into the chat frontend, which is a compose change), and everything in section 4 of the spec: the agent as a conversation, activity rows, approval, steering. The agent panel is framed rather than ported, because a native port needs a token bridge: this frontend holds Open WebUI's own session token while `/v1/agent/*` authenticates the Supabase bearer that only the embedded application carries.

## What must not regress

| Capability | Status |
|---|---|
| Chat | The transcript, composer and model picker are upstream's, unchanged in behaviour. The nightly Open WebUI e2e suite is the gate; label this PR `run-owui-e2e` to run it |
| Embeddings | Untouched. No API surface changes here |
| Voice to text | Untouched. The microphone is upstream's own control |
| Knowledge work | Knowledge is promoted to a top level row; `/workspace/knowledge` is unchanged and `/workspace` now redirects straight to it |
| Cowork and the coding agent | The `/v1/agent/*` API is untouched. The workspace moves from a corner link to a sidebar row; the application behind it is the same one |

## Security note

`apps/agent-console/middleware.ts` moves from `frame-ancestors 'none'` to `frame-ancestors 'self'`, and `X-Frame-Options` from `DENY` to `SAMEORIGIN`, because the chat shell now frames it from the same origin through the same Caddy listener. Everything the previous value defended against, a third-party page framing this one to harvest a click or a session, is still refused. `apps/web-console` keeps `'none'`.

## Buglog entry

```json
{"id":"owui-fork-literal-rewrites","date":"2026-08-17","area":"deploy/docker","error_message":"open-webui bundle no longer matches hive_ui_surfaces.py; a removed surface may have come back: sidebar-playground-item, changelog-modal, about-vendor-social-badges","root_cause":"The exact-literal bundle rewrites match on minified identifiers, which the minifier allocates per chunk. Building the frontend from vendored source renames them, so three of nineteen rewrites stopped matching even though every surface was still present. An identifier-tolerant regex is not a fix: for changelog-modal and about-vendor-social-badges it matched sibling components with identical compiled shape, which would have neutered Settings or the admin dialogue.","fix":"Remove those surfaces in the vendored source instead, and order the Dockerfile so the rewrite layer still runs against the upstream bundle before the source build replaces it. Retiring the layer entirely is the follow-up.","tags":["owui","fork","bundle-patch","minifier","docker"]}
```


sakibsadmanshajib:
## Visual proof

Two images built from the same pinned Open WebUI digest, differing only in this branch's diff, captured at 1440x900 in both themes with the DOM read as well as the pixels. Method as PR #909. Everything below is reproducible with `docs/proof/shell-nav-2026-08-17/run.sh`, which needs no credential: both containers run with `WEBUI_AUTH=False` and the agent workspace is built against placeholder public values.

### The shell as a user meets it, light

`main`, with the agent behind a corner link and no agent or chat destination in the navigation:



This branch. Chats, Agents and Knowledge as labelled rows, the current row carrying a coral bar, the corner link gone, cream and charcoal in place of grey:



### The same pair, dark





### The agent, reached from the sidebar rather than from a corner

The sidebar is still there, Agents is the current row, and the panel fills the region in the shell's own theme:





The panel here is the agent workspace's own sign-in screen, because this stack has no Supabase session and that application still has a login of its own. That is today's behaviour on the live box too, and it is the third login that single sign-on removes. What the capture proves about the part this change owns: the panel is same origin, it is inside the shell, it knows it is embedded, and it follows the shell's theme rather than the desktop's.

### Collapsed, and focused

Every destination survives the collapse; only the labels go. Keyboard focus shows the two ring treatment, which is what keeps a focus ring visible on a coral ground:





### What the DOM says, from the same runs

Read from `docs/proof/shell-nav-2026-08-17/dom.json`, not from the pixels.

| Claim | Evidence |
|---|---|
| The floating launcher is on `main` and gone here | `before-light.launcherCount` 1, `after-light.launcherCount` 0 |
| Three labelled destinations, with real hrefs | Chats `/`, Agents `/agents`, Knowledge `/workspace/knowledge` |
| Current destination is not colour alone | `aria-current="page"` moves from Chats to Agents between `/` and `/agents` |
| Rows are the size the token specifies | height 32 expanded, 36 square on the rail, 44 under a coarse pointer |
| The brand face actually loads | `hankenLoaded` false on `main`, true here |
| The panel is same origin, embedded, themed | `sameOrigin` true, `embeddedFlag` 1, `themeFlag` `dark` |
| Motion follows the catalogue | row transition `0.12s`, and `0s` under `prefers-reduced-motion` |
| Focus ring survives on any surface | `box-shadow: 0 0 0 2px canvas, 0 0 0 4px coral` |

The authenticated panel with a real task list in it is captured against the demo box once this merges and deploys.

sakibsadmanshajib:
## Adversarial review

Streams that ran, and streams that did not. An absent stream is not a pass.

| Stream | Result |
|---|---|
| Plain adversarial pass, no specialised tool | **RAN**, below |
| Design checklists (emil-design-eng, frontend-ui-engineering, frontend-design, ui-ux-pro-max) | **RAN AS CHECKLISTS, NOT AS SKILLS.** This agent has no Skill or SlashCommand tool, so none of the four could be invoked. Their rule sets were read from their own sources on disk and applied by hand. Reported as a capability gap rather than worked around |
| CodeRabbit CLI | **SKIPPED.** `coderabbit review --agent --committed --base main` failed twice with `Connection failed: WebSocket closed` (TRPCWebSocketClosedError), once in full mode and once with `--light`. Not a clean pass, an unavailable stream |
| `ecc:code-review`, `adversarial-pr-review`, `/codex:adversarial-review` | **SKIPPED**, same capability gap: they are Skill and slash-command entry points and this agent has neither tool |
| CI | Two failures on this head are GitHub infrastructure, not this diff: `Go tests (storage)` failed in `Set up job` with `Failed to download archive 'actions/setup-go' ... 429 (Too Many Requests)` after three retries, and the API is intermittently answering 503. Re-run when GitHub is healthy |

Findings on this diff, then findings on the product it lands in. The second set is written up in full at `plan-2026-08-17-ui-ux-flow-audit` in the vault.

### Found and fixed before this comment

| Finding | Fix |
|---|---|
| `git subtree add` writes a tree directly and bypassed `/vendor/*` in `.gitignore`, so every new file under the vendored tree was silently unstageable. The entire `src/lib/hive/` shell, the `/agents` route and the three font files were absent from the first pushed commit while the image still built green, because the build reads the working tree and not the index | `.gitignore` re-includes `/vendor/open-webui/` explicitly, and the pushed tree was rebuilt from `git archive` of the exact commit rather than from the working tree, which is the only way that class of defect surfaces |
| The demo box deploy job recreates the stack with `--build` under a 15 minute ceiling, and the chat image now compiles a frontend first. A deploy that trips its own timeout mid recreate is the worst outcome available | Ceiling raised to 30 with the measured reason, and the stale-lock sweep documented to match it moved in the same step |
| The panel was hidden until its `load` event fired, so a load event that never arrives would leave a permanent "Opening the agent workspace" over a working panel | The line sits behind the frame instead. An unpainted frame is transparent; the panel's own ground covers the line the moment it paints |
| Nav rows had no press feedback | `transform: scale(0.98)` on `:active` at 100ms, removed under `prefers-reduced-motion` |

### Open on this diff, accepted with reasons

| Finding | Why it is accepted here |
|---|---|
| The bundle rewrite layer still runs and its output is then discarded | Measured: those literals match on minified identifiers, so any source edit renames them and three of nineteen stop matching surfaces that have not come back. Retiring the layer is the follow-up, kept out because #936 was in flight against `hive_ui_surfaces.py`. Those three files are byte identical to `main` here |
| `Chats` and `New Chat` both target `/` | Chats is the label that makes chat a named destination; the list of conversations is directly beneath it. A real Chats destination with an archived filter is wave two of the spec |
| Changing the theme remounts the agent panel | `{#key}` on the resolved theme, so in-frame state is lost. A theme change is rare and deliberate; the alternative is writing into another document from the parent |
| The palette remap touches every surface in the application | Verified in both themes on the chat home, the agents route and the collapsed rail. Not verified on every modal, the settings dialogue or the knowledge screens. That pass is named as wave seven in the vault plan rather than claimed here |
| The embedded panel shows the agent console's own sign-in | That is today's behaviour and it is the third login. It is the first item in the flow plan |

### Design checklist findings, Emil format

Two of these are ours and are fixed above; the rest are inherited chrome this change does not touch, and they are wave seven.

| Before | After | Why |
| --- | --- | --- |
| No `:active` on a nav row | `transform: scale(0.98)`, 100ms | Fixed here. A control that does not respond to a press feels dead |
| Panel hidden until `load` | Loading line behind the frame | Fixed here. A hidden panel plus a permanent line is the failure that looks like a hang |
| `transition-all duration-300` on the sidebar container (`Sidebar.svelte`) | `transition: width var(--hv-duration-base) var(--hv-ease-out)` | Inherited. `all` animates layout properties, and 300ms is past the catalogue's 260ms ceiling |
| `transition:slide` 250ms on sidebar open | The width transition alone | Inherited. Two motions describing one act |
| Upstream header buttons carry `focus:outline-hidden` | The two ring focus treatment from the token file | Inherited, and it is the accessibility floor rather than a style preference |
| Stock suggestion chips, "Tell me a fun fact about the Roman Empire" | Four Hive chips | Inherited. In a product sold to regulated buyers this is the clearest signal nobody has been through the screen. Wave 0 of the spec, needs no fork |
| Greeting prints "Hello, User" | The person's name and one line naming what the surface does | Inherited, wave 0 |
| Model selector as a page title with "Set as default" under it | Model selector in the composer rail | A model is a property of the turn, not of the page. Wave 0 and 2 |

### Security

`apps/agent-console/middleware.ts` moves from `frame-ancestors 'none'` to `'self'` and `X-Frame-Options` from `DENY` to `SAMEORIGIN`, because the shell frames it from the same origin behind the same listener. A third-party page still cannot frame it. Two tests cover it, plus two more asserting the embed and theme parameters survive the sign-in redirect and that an unknown theme value is dropped rather than reflected into a redirect target. `apps/web-console` keeps `'none'`.

### Not merging

GitHub is degraded (`429` on action downloads, intermittent `503` on the API), so this stays open. The two red checks on this head are that outage, not this diff.

.github/workflows/deploy-demo-box.yml
```diff
@@ -283,9 +291,11 @@ jobs:
           # checkout, other writers (git gc --auto, a human on the box) can
           # legitimately hold this lock, and the workflow-level concurrency
           # group only serializes this workflow's own runs against each
-          # other. 15 minutes matches this job's own timeout-minutes below,
+          # other. 30 minutes matches this job's own timeout-minutes above,
           # so a lock younger than that is still someone's live operation.
-          find .git -maxdepth 1 -name index.lock -mmin +15 -delete
+          # Both moved together when the chat frontend build landed; they are
+          # one invariant and must not drift apart.
+          find .git -maxdepth 1 -name index.lock -mmin +30 -delete
```
sakibsadmanshajib, line 298: Verified, no race introduced. `timeout-minutes` moved from 15 to 30 and the stale-lock sweep's `-mmin` moved from +15 to +30 in the same commit, keeping them equal. Sweep threshold equal to (never shorter than) the job's own worst-case timeout is the correct invariant: a lock younger than the sweep threshold can only belong to a run that is still within its own allowed timeout window, so a live deploy cannot have its own lock swept out from under it. No drift between the two values in this diff.

vendor/open-webui/package.json
```diff
@@ -0,0 +1,162 @@
+{
+	"name": "open-webui",
+	"version": "0.10.2",
```
sakibsadmanshajib, line 3: Supply-chain finding (npm audit, not one of the six requested checks but squarely in scope for what this fork adds): `npm audit --audit-level=high` against the vendored `package-lock.json` reports 29 advisories: 1 critical, 15 high, 12 moderate, 1 low.

The critical (arbitrary file read/execute via the Vitest UI server, `vite`/`vite-node`/`vitest`) and most of the high-severity cluster (`esbuild`, `extract-zip`, `@cypress/request`, `uuid`) sit under `cypress` and `vitest`, which are devDependencies exercised only by local dev tooling, not shipped in `npm run build`'s output, so they are not part of this image's runtime attack surface as built here.

Worth a look regardless, because these are direct `dependencies` (shipped in the built frontend): `xlsx` (high, prototype pollution and ReDoS, no fix available upstream, this is the long-known unpatched SheetJS package), `undici` (high, several advisories, fix available), `dompurify` (moderate, XSS-relevant since it is the sanitizer this app trusts for rendered content, fix available), and `@huggingface/transformers`/`kokoro-js` pulling in a vulnerable `sharp` transitively (high, libvips CVEs, no fix available).

`docs/owui-fork.md` says frontend npm advisories are now Hive's to own and that `vendor/open-webui` belongs in Dependabot's scope. There is no `.github/dependabot.yml` anywhere in this repo today (pre-existing gap, not introduced by this PR), so nothing is actually watching this tree yet. Given this PR roughly triples the repository's third-party JS dependency surface in one commit, wiring that up is worth prioritizing rather than deferring.

deploy/docker/Dockerfile.open-webui
```diff
@@ -175,3 +218,57 @@ RUN sed -i "s/or 'native'/or os.environ.get('HIVE_DEFAULT_FUNCTION_CALLING', 'na
...
+    pinned=$(python3 -c "import json;print(json.load(open('/app/package.json'))['version'])"); \
+    test "$vendored" = "$pinned" || { \
+      echo "vendor/open-webui is $vendored but this image's backend is $pinned" >&2; \
+      echo "run: git subtree pull --prefix vendor/open-webui <remote> v$pinned --squash" >&2; \
+      exit 1; }; \
+    rm /tmp/hive-frontend-package.json; \
+    echo "hive: frontend and backend agree on $pinned"
+
+RUN rm -rf /app/build
+COPY --from=frontend /build/vendor/open-webui/build /app/build
```
sakibsadmanshajib, line 260: Security review, build integrity (verified). Checked the ordering claim in the buglog entry against the actual stage sequence: the sed/python bundle rewrites (`hive_ui_surfaces.py`, `remove_integrations_tab.py`, the title and manifest patches) all run earlier in this stage, against `/app/build` as it exists in the pinned upstream image, then `RUN rm -rf /app/build` plus this `COPY --from=frontend` line discard that bundle and replace it with the Hive source build, last. That is the correct order: the exact-literal rewrites still see the bundle they were written against, so they keep failing the build on drift, and the source build wins the final content. Also confirmed present: the backend patches (env.py suffix removal, tenant-role patch, RAG env reconcile, function-calling default) are all untouched by this change, and the version-match assertion between the vendored `package.json` and the pinned image's `package.json` is in place and fails closed. No build-integrity issue found here.

docs/owui-fork.md
```diff
@@ -0,0 +1,62 @@
...
+
+Frozen frontend, tracked backend: upstream frontend features are not taken, backend fixes
+arrive by bumping the image digest, and when a digest bump crosses a release the subtree is
+pulled to the matching tag and the four edits above are replayed.
+
+Frontend npm advisories are ours now. `vendor/open-webui` belongs in Dependabot's scope.
+
+## What has not been retired yet
+
+The bundle rewrites in `deploy/docker/owui-patches/` still run, against our own build rather
```
sakibsadmanshajib: Minor wording issue. This sentence reads as "the bundle rewrites run against our own [source-built] output," but that is not what the Dockerfile does: the rewrites run against the stock bundle from the pinned upstream image, and the source build then replaces that bundle afterward (see `Dockerfile.open-webui` around the `COPY --from=frontend ... /app/build` step). The rewrites never see the Hive-built output at all in this image. Worth rewording so a future reader planning the retirement of this layer does not draw the wrong conclusion about what the rewrites are currently protecting.

.gitignore
```diff
@@ -24,6 +24,13 @@
 # trimmed, patched copy of OpenHands's software-agent-sdk; see
 # vendor/openhands/VENDORING.md.
 !/vendor/openhands/
+# vendor/open-webui is the forked chat frontend, added as a squashed git
+# subtree and built from source by deploy/docker/Dockerfile.open-webui; see
+# docs/owui-fork.md. `git subtree add` writes a tree directly and so ignored
+# this exclusion, which left the vendored tree tracked but every NEW file
+# under it silently unstageable, including the whole Hive shell. Re-included
+# explicitly so a `git add` of a new component cannot pass in silence.
+!/vendor/open-webui/
```
sakibsadmanshajib, line 33: Verified. This is the highest-value check for this review, per the review brief: confirmed the final pushed commit (`cf79a0956`) contains `vendor/open-webui/src/lib/hive/ShellNav.svelte`, `ShellNavIcon.svelte`, `hive.css`, `nav.ts`, `nav.test.ts`, the `/agents` route, and all three brand `.woff2` font files. The fix commit (`b09a8ad1a`) correctly diagnoses the root cause (`git subtree add` writes a tree directly and bypasses `.gitignore`, so the new files under a `/vendor/*`-ignored path were silently unstageable even though the directory itself was tracked) and this negation closes it. No trace of the missing-shell defect in the commit actually up for review.

vendor/open-webui/src/lib/components/chat/Settings/About.svelte
```diff
@@ -0,0 +1,136 @@
...
+			<div class="mb-2 text-xs">
+				{#if !$WEBUI_NAME.includes('Open WebUI')}
+					<span class=" text-gray-500 dark:text-gray-300 font-medium">{$WEBUI_NAME}</span> -
+				{/if}
+
+				<span class=" capitalize">{$config?.license_metadata?.type}</span> license purchased by
+				<span class=" capitalize">{$config?.license_metadata?.organization_name}</span>
+			</div>
+		{:else}
+			<!-- Hive: vendor badges removed, see docs/owui-fork.md -->
```
sakibsadmanshajib, line 125: License note, not a blocker. `LICENSE`, `LICENSE_HISTORY`, `LICENSE_NOTICE` and `CONTRIBUTOR_LICENSE_AGREEMENT` were compared byte for byte against the upstream v0.10.2 tag and are unmodified, so this is UI-surface removal, not license-file tampering. What is actually deleted here is the rendered "Copyright (c) ... Open WebUI Inc. All rights reserved," the "Created by Timothy J. Baek" line, and the Discord and Twitter and GitHub-star badges from the Settings > About screen. That plausibly falls under the same LICENSE clause 4 branding carve-out the Dockerfile already cites for stripping the "(Open WebUI)" name suffix (deployment under 50 end users per rolling 30 days), since these are also textual identifiers in the interface rather than the source-form or distributed-materials copyright notice that clauses 1 and 2 separately require and which the retained LICENSE file already satisfies. Recommend giving this removal the same explicit citation and the same "re-evaluate before exceeding 50 users" trigger comment the Dockerfile gives its own branding removal, since right now only the Dockerfile's copy carries that reminder and this one has no equivalent guard if the deployment scales past the carve-out.

sakibsadmanshajib:
## Security and supply-chain review, PR #938 (`vendor/open-webui` fork)

Scope was the six risks named in the review brief: provenance, malicious content, license integrity, build integrity, the `.gitignore` bug, and the deploy-workflow timeout change. Detailed findings on items 3, 4, 5 and 6 are posted inline. Summary of everything, including the items that came back clean:

**1. Provenance: verified clean, independently.** The squash commit's trailer records `git-subtree-split: ecd48e2f718220a6400ecf49eafd4867a38feb10`. I shallow-fetched that exact commit from `github.com/open-webui/open-webui` directly and confirmed two things myself, not from the PR's own claims: that commit is `refs/tags/v0.10.2` upstream, and the root tree hash of the subtree-add commit in this repo is byte-for-byte identical to that tag's tree. The squash destroyed per-file history as expected, but the subtree-split trailer is exactly the re-derivation record the brief asked whether one exists, and it holds up. `docs/owui-fork.md` also documents the tag and the `git subtree pull` command for future updates.

**2. Malicious or surprising content: clean.** No `preinstall`/`postinstall`/`prepare` lifecycle scripts in the vendored `package.json` (17 scripts total, none of those three). `.npmrc` is one line, `engine-strict=true`, identical to upstream. No `.patch` files anywhere in the vendored tree. One WASM binary (`static/sql.js/sql-wasm.wasm`, upstream's own client-side SQLite); hash matches upstream exactly. No `curl`/`wget`/network calls in any vendored build script beyond upstream's own pre-existing `backend/start.sh` localhost health check, which this PR does not touch or execute.

**3. License integrity: see inline comment on `About.svelte`.** `LICENSE`, `LICENSE_HISTORY`, `LICENSE_NOTICE` and `CONTRIBUTOR_LICENSE_AGREEMENT` are present and byte-for-byte identical to upstream, confirmed by hash. What was removed is UI-rendered text (a copyright line, an author-attribution line, three vendor badges) from the Settings > About screen, not a license file. Flagged as a documentation-parity suggestion, not a violation: this removal likely rests on the same LICENSE clause 4 branding carve-out (under 50 end users) the Dockerfile already cites explicitly for the WEBUI_NAME suffix, but it does not carry the same explicit citation or the same "re-evaluate before scaling past 50 users" reminder that the Dockerfile gives its own removal.

**4. Build integrity: verified correct**, inline comment on `Dockerfile.open-webui`. The exact-literal bundle rewrites run against the pinned image's stock bundle, then that bundle is discarded and replaced by the Hive source build last, matching the ordering the PR body and buglog entry claim. Backend patches are untouched. The frontend/backend version-match assertion is present and fails the build on mismatch.

**5. `.gitignore` bug: verified fixed.** Confirmed directly against the pushed head commit (`cf79a0956`), not against the working tree: `vendor/open-webui/src/lib/hive/` (all 5 files), the `/agents` route, and all three brand `.woff2` fonts are present in the final tree. This was the single highest-value check per the brief, and it holds.

**6. Deploy workflow timeout: verified, no race.** `timeout-minutes` and the stale-lock sweep's `-mmin` threshold both moved from 15 to 30 together, keeping them equal. That is the correct invariant (sweep threshold must never be shorter than the job's own worst-case runtime); this PR preserves it rather than letting it drift.

**Bonus finding, outside the six listed risks but in scope for a fork this size:** ran `npm audit --audit-level=high` against the vendored `package-lock.json`. 29 advisories total: 1 critical, 15 high, 12 moderate, 1 low. The critical and most of the high-severity cluster are in `vitest`/`vite`/`cypress`, which are devDependencies not shipped in the production build. Four direct runtime dependencies are worth triage regardless: `xlsx` (high, prototype pollution and ReDoS, no fix available upstream), `undici` (high, several advisories, fix available), `dompurify` (moderate, XSS-relevant, fix available), and `@huggingface/transformers`/`kokoro-js` pulling a vulnerable `sharp` transitively (high, no fix available). Also confirmed: no `.github/dependabot.yml` exists anywhere in this repo (a pre-existing gap, not introduced here), despite `docs/owui-fork.md` stating the vendored tree now belongs in Dependabot's scope. Worth wiring up given this PR roughly triples the repo's third-party JS surface in one commit.

**Not run:** `eslint-plugin-security` static scan was not run against the vendored tree; not requested in the review brief and 4,986 files makes a full lint pass a separate, larger task. Flagging as SKIPPED rather than implying a clean pass on that specific tool.

No CRITICAL or HIGH finding blocks this PR on the six requested risks. The npm audit results are a follow-up item, not a block on this PR's own changes (the vulnerable packages are upstream's existing dependency choices, inherited by the fork decision itself, not introduced by this diff).

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Agreed, and thank you for checking the invariant rather than the number. Both values moved together in the same step precisely so they cannot drift; the comment above the sweep now says they are one invariant and must not be separated. No change needed here.

vendor/open-webui/package.json
sakibsadmanshajib: Acted on, without upgrading anything in this PR.

`docs/owui-fork.md` gained a triage section in 909542b93, so the next reader does not repeat the analysis. It records that the critical advisory and most of the high cluster are devDependencies (`vitest`, `vite`, `vite-node`, `cypress`, `esbuild`, `extract-zip`, `@cypress/request`, `uuid`) that never reach the emitted bundle, and it answers the four direct dependencies by reachability read in this tree rather than assumed:

* `xlsx` **ships and is reachable**: `src/lib/utils/excelToTable.ts` dynamically imports it to parse a spreadsheet the user uploaded, and there is no fix upstream. This is the one that needs a real decision (replace the parser, or move parsing server side), which is more than a dependency bump.
* `dompurify` **ships and is reachable**, and it is the sanitiser the application trusts for rendered content. Fix available, and it deserves its own change with the XSS-relevant call sites exercised.
* `undici` **does not ship**: its only importer is `scripts/prepare-pyodide.js`, a build time Node script.
* `sharp` **does not ship**: verified against the built image, `/app/build` contains no import of it. It is a native Node module behind the transformers package's Node backend and cannot load in a browser build.

On Dependabot: you are right that the doc claimed a scope nothing enforces. The doc now says plainly that no `.github/dependabot.yml` exists in this repository, that the gap predates the fork, and that wiring it at `vendor/open-webui` is the follow-up. Kept out of this PR rather than bolted on, since a dependency automation config wants its own review.

deploy/docker/Dockerfile.open-webui
sakibsadmanshajib: Thank you for verifying the ordering against the stage sequence rather than against the claim. One consequence worth recording: `docs/owui-fork.md` described that same ordering backwards, and 909542b93 corrects it, including what the rewrites are actually protecting now (a drift check on upstream's own bundle, not the bundle we ship, because every surface they remove is removed again in source).

docs/owui-fork.md
sakibsadmanshajib: Fixed in 909542b93. You are right and the wording would have misled exactly the reader it was written for.

The section now states that the rewrites run against the stock bundle inside the pinned upstream image, that `rm -rf /app/build` plus `COPY --from=frontend` replace that bundle afterward, and that the rewrites therefore never see the Hive built output. It also spells out the consequence for retiring the layer: what those rewrites protect today is not the shipped bundle but a drift check on upstream, since every surface they remove is removed again independently in the vendored source, and that source removal is the one a user meets.

.gitignore
sakibsadmanshajib: Thank you for checking the pushed commit rather than the working tree. That is exactly how the defect was found in the first place: the image build reads the working tree and not the index, so the branch built green while the shell was absent from it. The build now also asserts `data-hive-nav` and the brand typeface are present in `/app/build`, which fails loudly on that class of miss rather than shipping a frontend with no navigation in it.

vendor/open-webui/src/lib/components/chat/Settings/About.svelte
sakibsadmanshajib: Out of scope for this PR by owner decision. The licensing and branding question, including the user count carve out, is settled in `.wolf/decisions.md` D-041, which landed on `main` after this branch was cut, which is why it was not visible to this review. No citation is being added here and no tracking issue is being opened.

The part of your comment that is not the licensing question stands verified and is appreciated: the four licence files are byte identical to upstream, and what changed is rendered interface text only.

sakibsadmanshajib:
## Response to the security and supply-chain review

Thank you for verifying provenance against upstream directly rather than against this PR's own claims, and for checking the `.gitignore` fix on the pushed commit rather than on a working tree. The second one is the exact reading that would have caught the defect in the first place.

Dispositions, one per finding. Two produced code, four were verifications with nothing to fix, one is out of scope.

| Finding | Disposition |
|---|---|
| 1. Provenance | Verified clean. No action |
| 2. Malicious or surprising content | Verified clean. No action |
| 3. License integrity | Out of scope for this PR by owner decision. Settled in `.wolf/decisions.md` D-041, which landed on `main` after this branch was cut, which is why it was not visible to you. No citation added, no tracking issue opened. The non-licensing half of your finding stands and is appreciated: the four licence files are byte identical to upstream and only rendered interface text changed |
| 4. Build integrity | Verified correct. It also surfaced a real documentation defect, below |
| 5. `.gitignore` bug | Verified fixed |
| 6. Deploy workflow timeout | Verified, invariant preserved |
| Bonus: npm audit | **Acted on.** Triage written, no upgrades |

### The two that produced work

**`docs/owui-fork.md` stated the patch ordering backwards.** Corrected in `909542b93`. The rewrites run against the stock bundle inside the pinned upstream image; `rm -rf /app/build` plus `COPY --from=frontend` replace that bundle afterward; the rewrites never see the Hive built output. The section now also says what that means for retiring the layer, which is the thing the wrong wording would have cost: those rewrites protect a drift check on upstream's own bundle, not the bundle we ship, because every surface they remove is removed again independently in the vendored source and that source removal is the one a user meets.

**Advisory triage, written down so it is not redone.** Same commit. The critical advisory and most of the high cluster are `devDependencies` that never reach the emitted bundle. The four direct dependencies were answered by reachability read in this tree:

* `xlsx` ships and is reachable, via a dynamic import in `src/lib/utils/excelToTable.ts` that parses a spreadsheet the user uploaded, with no fix available upstream. This is the one that needs a decision rather than a bump, and it is named as such.
* `dompurify` ships and is reachable, and it is the sanitiser the application trusts. Fix available; it wants its own change with the XSS-relevant call sites exercised.
* `undici` does not ship. Its only importer is `scripts/prepare-pyodide.js`, a build time Node script.
* `sharp` does not ship. Verified against the built image: `/app/build` contains no import of it.

No upgrades in this PR, deliberately. A dependency bump inside the same change that introduces 4,900 files is unreviewable, and three of the four need their own verification.

On Dependabot you were right that the doc asserted a scope nothing enforces. It now says plainly that no `.github/dependabot.yml` exists in this repository, that the gap predates the fork, and that wiring it at `vendor/open-webui` is the follow-up.

### Noted as skipped

`eslint-plugin-security` was not run. Recorded rather than treated as a clean pass, same posture this PR takes for the CodeRabbit CLI stream that failed to connect.

Issue #873, opened 2026-08-11
Bug entries on feature branches make .wolf/buglog.jsonl conflict on every parallel PR, which silently suppresses CI


sakibsadmanshajib:
Split out of #872, which was filed as a CI trigger defect and turned out not to be one. The real cause there is a recurring hazard worth removing.

## The pattern

`.wolf/buglog.jsonl` is tracked, and `.claude/rules/openwolf.md` asks every bug-fix branch to append a line to it. `.gitattributes` declares `merge=union` so those appends merge automatically. That works locally. It does not work on GitHub, whose server-side merge ignores the union driver, so concurrent branches that each appended a line are in hard conflict with each other.

## Two failure modes, both already observed

1. **Serial conflicts.** Every open fix branch that appended a line conflicts with every other one, so they have to be rebased and merged one at a time.
2. **Silent CI suppression.** While a branch is unmergeable, GitHub cannot build `refs/pull/N/merge` and creates no `pull_request` run at all. The pull request shows zero checks. This is what happened on PR #870, and it cost an investigation into workflow triggers that were never broken (#872). The required-status gate then blocks the merge for a reason that is never stated on the page.

The second one is the expensive mode, because the symptom points away from the cause.

## Proposal

Stop writing to the tracked file from feature branches. Bug entries land on `main` only, either through a small post-merge step or by batching them, so no two branches ever hold a competing version of the file.

If the append-on-fix protocol is worth keeping exactly as written, the alternative is to untrack `.wolf/buglog.jsonl` and treat it like the other hook-owned `.wolf/` telemetry, accepting the loss of a reviewable bug ledger in git.

## Why not a CI-side detector

A workflow that flags "this pull request has zero `pull_request` runs" would report a state GitHub already renders on the pull request page ("This branch has conflicts that must be resolved"), and it would leave the conflict source in place. That converts a recurring cause into a recurring notification. One file produces both failure modes, so removing the concurrent-write pattern removes both.

sakibsadmanshajib:
Reopening. This was closed on 2026-08-11 with no comment, but the automation it tracks does not exist.

Verified just now against the current tree:

- No workflow appends buglog entries after a merge. The only occurrence of "buglog" under `.github/workflows/` is a comment in `ci.yml` describing the inert-path allowlist, which is a different mechanism entirely (it lets a buglog-only pull request merge cheaply, it does not create one).
- No script under `scripts/` mentions buglog.
- Nothing anywhere references this issue number.

[...]

Issue #540, opened 2026-07-27
agent-console runs its own separate login session instead of sharing the chat session


sakibsadmanshajib:
**What happens**: agent-console maintains its own independent Supabase auth session rather than a
cookie handoff from Open WebUI, even though both sit behind the same Caddy listener on the same
host with `/agent-workspace/*` path-matched ahead of OWUI's catch-all. The app's own
`middleware.ts` states this is an intentional architecture decision ("this app is a standalone
console with its own Supabase session, not a cookie handoff from Open WebUI"). The practical
result: a signed-in chat user who clicks the agent-workspace launcher lands on a full second
sign-in page, mid-conversation. Live recon independently confirmed this: the same browser session
that had just completed chat OIDC login still hit agent-console's own separate sign-in wall.

**Evidence**: CODE — `apps/agent-console/middleware.ts:16-19` (documented as intentional). LIVE —
recon-findings.md finding #3 confirms no cookie/session carries over even from the first-party
in-app "Agent workspace" launcher link in the chat header.

**Why it matters**: Ranked the single worst product-shape problem in the competitive parity review
(parity-matrix.md section 4a). It directly contradicts the "one org equals one tenant, departments
via RBAC" architectural claim the moment the product shows two identity front doors, and an
unexpected credential prompt reached by clicking a link inside chat content is exactly the shape
security-aware buyers are trained to treat as phishing.

**Smallest fix**: Same origin, same cookie -- have agent-console read the existing Supabase session
instead of initiating its own sign-in. If the cookie scope genuinely cannot be shared, have the
OWUI launcher carry a one-time exchange token to `/agent-workspace/auth/callback`, which already
exists.

**Source**: ux-workflow-review.md finding W9, parity-matrix.md section 4a, recon-findings.md
finding #3.

sakibsadmanshajib:
Open PR #577 (fix: agent-workspace launcher navigates without polluting the transcript) appears to address this. Not yet merged.

PR #951, opened 2026-08-17
feat: render the agent surface natively instead of framing a second application

sakibsadmanshajib:
Owner directive, 2026-08-17: "stop using iframes and keep it native inside OpenWebUI."

`https://chat-hive.scubed.co/agents` rendered exactly one `<iframe>`, `loading="eager"`, pointed at `/agent-workspace/tasks`, which booted `apps/agent-console` (a second whole Next.js application) inside the page. That is why the agent surface still looked exactly as it did before the shell landed, why it was slow, and why a task could never become part of the conversation. The frame is gone.

Scope is bounded deliberately: visual and functional parity with what the frame showed, natively, which is the task composer and the task list. No tool call cards, no progress panel, no transcript. All three are specified in `spec-2026-08-17-agent-run-surface` and blocked on the event relay (its step S6), and a tool card with nothing to render is worse than today's status row.

## Two premises in the brief that were already stale

Both checked in the tree at `1dc67ee69`, both reported before building anything.

1. **`apps/agent-console/middleware.ts` does not send `X-Frame-Options: DENY` or `frame-ancestors 'none'`.** It sends `SAMEORIGIN` and `frame-ancestors 'self'` (`middleware.ts:80-84`), changed in #938 with a comment naming the shell as the reason. Nothing was working around anything: the frame was permitted because one Caddy listener serves both applications on one origin, so `'self'` is a real origin check that matches.
2. **The frame's `src` carried no credential.** It was `embed=1` and `theme=<light|dark>`, both re-validated and rebuilt rather than reflected (`middleware.ts:25-31`, `:93-100`).

What the credential path actually was: `apps/agent-console` holds its own first-party Supabase session in cookies on the shared origin, and its browser code calls `/v1/agent/*` directly with `session.access_token`. That token comes from a direct Supabase sign-in, so it carries the `tenant_id` claim the `custom_access_token_hook` mints, and `JWTMiddleware` accepts it with no fallback. The frame existed because a second application had a second, better token.

## The problem, and why the recommended fix needed one correction

The chat frontend holds Open WebUI's own session token, which edge-api has never heard of, plus a Supabase access token minted through Supabase's **OAuth-server** grant, which does not run `custom_access_token_hook` and therefore carries **no `tenant_id`**. `JWTMiddleware` rejects that (`middleware.go:123-129`). That is D-023 working, pinned by `inert_token_test.go`, and it is not relaxed here.

Worse for any browser-side fix: **that OAuth token is not in the browser at all.** Open WebUI keeps it server side and resolves it per request through `get_system_oauth_token`, refreshing it when expired (`utils/middleware.py:3012-3045`). Handing it to page JavaScript would be a downgrade, not a fix.

So the recommendation to reuse the mechanism `hive_jwt_forward.py` plus `owui_unwrap.go` already run in production is adopted. The correction: **`__metadata.upstream_auth` cannot serve this surface**, because three of the four agent-task calls have no JSON body to hide a token in.

| Call | Method | Body |
|---|---|---|
| `GET /v1/agent/tasks` | GET | none |
| `POST /v1/agent/tasks` | POST | JSON |
| `GET /v1/agent/tasks/{id}` | GET | none |
| `POST /v1/agent/tasks/{id}/cancel` | POST | none |

(`apps/edge-api/internal/agenttask/handler.go:38-72`.)

### 1. edge-api: a second carrier on the same boundary

`X-Hive-Upstream-Auth` carries the per-user token for requests with no body to carry it in. **This is not a new auth boundary.** The trust decision is unchanged and unchanged in strength: the header is honoured only when `Authorization` is exactly the shim key, which is the identical gate the body carrier already sits behind. What changes is the carrier, not who is allowed to use it.

Properties, each with a test:

* Honoured only under the shim key. Under a real API key, a real JWT, or no credential at all, it is ignored.
* **Stripped from the forwarded request on every branch**, including the ignored one and including when the middleware is disabled. A header a client controls must never be readable by a handler, a log or an audit sink, because a handler that can read it is one that could later be taught to trust it.
* Same 8 KiB cap as the body carrier, shared through one normaliser so the two cannot drift into accepting different shapes of the same credential.
* Fails closed when present but unusable, rather than forwarding with the shim key still on `Authorization`.

`requiresPerUserAuth` grows the agent-task paths. It previously covered `/v1/chat/completions` only, which meant a shim-key request to `/v1/agent/tasks` with no user token would pass through and **bind to the shim's own principal**, listing and cancelling the shim account's tasks instead of the user's. Nothing sent that request; this change is what keeps the new proxy from being one bug away from sending it. It is a tightening, not a widening: a silent mis-attribution becomes a 401.

One latent defect fixed while in the same function: a present-but-empty `__metadata.upstream_auth` previously returned `unwrapOK` with an empty token, which skipped the fail-closed arm and forwarded to `/v1/chat/completions` with the shim key intact. It now takes the missing-carrier arm, so it 401s and logs like every other missing token.

Not touched, deliberately: the tenant fallback is still gated on `IsOWUIUnwrapped` and is not widened by one line; `Selector` is unchanged; `inert_token_test.go` passes unmodified.

### 2. The chat container: a server-side proxy

Only the **frontend** of `hive-open-webui` is built from the fork; the Python backend stays upstream's pinned image on purpose (Dockerfile header, D-036). So this arrives as a copied module plus one asserted splice, the same shape as `hive_model_picker` and `hive_rag_env_config`. The frontend builds with `adapter-static`, so there is no SvelteKit server to put it in instead.

`deploy/docker/owui-patches/hive_agent_proxy.py`, mounted at `/api/v1/hive/agent`:

* Every route depends on `get_verified_user`, and **no route reads a user id, a tenant id, an email or a token from the request**, so a caller cannot influence which principal edge-api resolves.
* Presents the shim key on `Authorization` and the user's token on the carrier header. Neither is ever returned to the browser, logged, or included in an error.
* Four named operations with the task id validated as a UUID before it is interpolated. Deliberately not a general-purpose proxy, which is the shape #770 removed from this image.
* The create body is **rebuilt from two named fields** rather than passed through, so nothing a caller invents (a `__metadata` block, for one) reaches edge-api.
* Adds no configuration: `OPENAI_API_BASE_URL` and `OPENAI_API_KEY` are already set on this container, so the shim key gains no new home.

### 3. The frontend

`src/lib/hive/agentTasks.ts` and `AgentTasks.svelte`, plus the `/agents` route. Ported from the console, including two decisions that are easy to mistake for accidents and are not: the `unknown` status sentinel (a status this build cannot name still gets a row, because dropping it made a submitted task vanish) and the two engine sentinels (a deployment with no runtime reads "Blocked" in the warning register, never "Failed" in red next to a real failure).

## What happened to `/agent-workspace`, stated plainly

It still exists and it is still served. Nothing in this PR removes it, and the PR should not be read as removing it.

Three separate things get confused under that name, so all three, precisely:

* **The `agent-hive.scubed.co` subdomain is already gone**, and was before this PR. It is NXDOMAIN and decommissioned (D-026, verified 2026-07-28). There is no separate agent host and has not been one for some time.
* **`/agent-workspace` is a path on the chat host**, proxied to the `agent-console` container by `deploy/docker/Caddyfile.owui:161-168`. That rule is untouched here, so the path still resolves and still serves `apps/agent-console`.
* **Nothing in the product points at it any more.** The nav row goes to `/agents`, and `/agents` no longer loads it. A user reaches it only by typing the URL.

So after this PR the old surface is unreachable through the interface but still reachable by URL. Making it actually stop being a thing is the retirement described in the next section, and it is deliberately not in this PR: the console is the only surface that has ever launched a task against the live engine, so it is the control if the native path misbehaves on the box. Recommendation is to retire it once this has run there for a day.

## On the composer: the extraction was taken, not the fallback

Answering the directive's point 1 plainly.

`MessageInput.svelte` cannot be consumed wholesale: 2222 lines, 29 exported props, several required and chat-specific (`history`, `selectedModels`, `createMessagePair`, `stopResponse`, `taskIds`, `messageQueue`), and its container's classes are computed from chat stores. Cloning its CSS was banned, correctly.

So the **presentational shell was extracted**, which was the preferred option:

* `src/lib/hive/ComposerShell.svelte` is `#message-input-container` lifted verbatim: the rounded surface, the border, the hover and focus-within treatment, the inner padding. It reads no store; the two values that used to compute its classes arrive as props.
* `src/lib/hive/ComposerSendButton.svelte` is `#send-message-button` lifted verbatim, classes and glyph unchanged.
* `MessageInput.svelte` now renders both. **No behavioural edit was needed**, which was the stated condition for taking this option: the diff there is two tag swaps, one component swap and one import.

What is deliberately **not** shared is the row of controls inside the container. Chat's carries attach, tools, skills, web search, voice and the model picker; the agent composer's carries the pack toggle. Sharing that row would mean sharing chat state with a surface that has none of it, which is the coupling the extraction exists to avoid.

Per the directive: the heading block and the helper paragraph are deleted. The pack choice is a segmented **Knowledge work** / **Coding** toggle inside the composer. The one line describing the selected mode survives as a single muted line under the composer, because it is genuinely useful and it does not turn the composer back into a form. `Enter` sends and `Shift+Enter` is a newline, which is the chat composer's own behaviour. The two packs and their semantics are unchanged, and so is the default (`coding-pack`): this is a presentation change, not a capability change.

Chat's own composer is proved unchanged two ways, per the condition set: the vitest suite is green, and before and after screenshots of the **chat** composer are posted below alongside the agent one.

## Does `apps/agent-console` still need to exist?

Honest assessment, not acted on here.

**No route in it is reachable from the product any more.** Nothing frames it, and the only two links that ever pointed at it are gone: the nav row goes to `/agents`, and `/agents` no longer loads `/agent-workspace/tasks`. It is now a signed-out URL that a user reaches only by typing it.

What retiring it would take, in order:

1. Delete the `@agentConsole` matcher and its `reverse_proxy` from `deploy/docker/Caddyfile.owui` (lines 145 to 168), which is what makes `/agent-workspace` resolve at all.
2. Remove the `agent-console` service from `deploy/docker/docker-compose.yml` and its `Dockerfile.agent-console`, plus the `NEXT_PUBLIC_EDGE_API_BASE_URL` and `EDGE_API_INTERNAL_BASE_URL` wiring that exists only for it.
3. Remove the `agent-console-unit` job from `.github/workflows/ci.yml`, which is a required check, so branch protection needs updating in the same change or the merge gate waits forever on a job that no longer runs.
4. Delete `apps/agent-console/`.

Two things worth keeping first, and neither is code: its `proof/` harness, and the accessibility reasoning in `task-console.tsx` (it refuses `--color-ink-3` for body text on a measured contrast argument). The second is already carried across into `AgentTasks.svelte`'s comments and CSS.

One caveat against deleting it immediately: it is the only surface that has ever launched a task against the live engine, so it is the control if the native path misbehaves on the box. Recommendation is to retire it in a follow-up once this has run on the demo box for a day, not in this PR.

## Testing

* `go test ./apps/edge-api/... -count=1 -short` in the toolchain container: green, whole module, including nine new cases for the carrier and the fail-closed paths.
* `python3 scripts/test_owui_agent_proxy.py`, wired into `make test-scripts` (a required CI check). Pure standard library, no framework, matching the other `scripts/test_owui_*.py` self-checks. It is mutation tested: swapping the two credentials onto the wrong headers, deleting the missing-token guard, and forwarding the whole request body each make it fail.
* `npm run test:frontend` for the fork, covering the decoder, the status mapping and the four calls.

**A gap named rather than hidden.** `ci.yml` has vitest lanes for `apps/desktop`, `apps/web-console` and `apps/agent-console`, and none for `vendor/open-webui`, so every test ever written under `src/lib/hive` was coverage on paper only. This adds `npm run test:frontend -- --run` to the image's frontend build stage, which means the fork's tests gate the nightly build and the demo deploy. That is not the same as gating a pull request and is not claimed as such; a PR-time lane for this tree is a follow-up.

## Buglog entry

```json
{"id":"owui-agent-shim-principal-gap","date":"2026-08-17","title":"OWUI shim-key requests to /v1/agent/tasks would have bound to the shim's own principal","error_message":"none observed; latent","root_cause":"requiresPerUserAuth in apps/edge-api/internal/auth/owui_unwrap.go returned true only for /v1/chat/completions, so a shim-key request to any other path fell through with the shim key still on Authorization and resolved as the shim account. Correct for the paths that existed then (embeddings and text-to-speech authenticate as the shim by design), wrong the moment a second per-user OWUI path appeared. A present-but-empty __metadata.upstream_auth had the same effect on the chat path itself, because it returned unwrapOK with an empty token and skipped the fail-closed arm.","fix":"Extended requiresPerUserAuth to /v1/agent/tasks and its subtree, and made an empty upstream_auth report as a missing carrier so it takes the 401 arm and the warn log.","tags":["auth","edge-api","owui","fail-closed","tenancy"]}
```

## Visual proof: not posted yet, and exactly what exists today

Stated precisely, because an earlier version of this section said screenshots were "committed under `docs/proof/` and posted in a comment". Neither was true. Nothing is committed under `docs/proof/` in this diff, no image is on this PR, and `owui-nightly.yml` has no step that commits or comments; it only uploads a transient CI artifact. That sentence described intent before the run finished and read as completed work. It will not be restated until an image is actually visible on this page.

**What has been verified on a real stack with a real signed-in session**, run 32066161156, the OWUI end-to-end job on this branch:

- `the agent workspace opens inside the shell and frames nothing` passed. That is the load-bearing one: an `iframe` count of zero on `/agents`, the composer present and accepting typed input, both toggle options clickable, the list painting exactly the rows the API returned matched against that response's own text, and the call to the proxy answering something other than 401, which is the single failure mode this design has.
- `proof capture: the agent surface, natively, in both palettes` passed, so the images were produced.
- Every chat spec passed on that same build: send and stream, multi-turn, model switch, tenant model visibility, sign-out. That is evidence the composer extraction did not break chat, and it does not depend on anyone reading a screenshot.

**Why no image was attached at first, and a finding worth its own paragraph.** The captures were being written into `playwright-report-owui/proof`, inside the HTML reporter's own output folder. That reporter clears its folder before writing the report, so **every capture was produced and then deleted**, and the result looked exactly like a capture step that had never run. Nothing failed, nothing warned, and the test that wrote them passed. Confirmed by downloading the artifact and finding no `proof/` directory rather than by inferring it.

Anyone adding a capture step to a Playwright suite in this repository will walk into the same trap, which is why it is recorded here rather than only fixed: proof must never be written inside a reporter's output directory. The captures now go to a sibling directory with its own upload step. This is the same silent-absence shape that let the end-to-end gate rot unnoticed, where an artifact that is missing and an artifact that was never asked for are indistinguishable after the fact.

**Two attempts to re-capture since then both failed outside this branch**, and neither is counted as anything:

- Run 32112158077: the sign-in helper's consent hop exceeded its 30 second budget. The container log shows the token exchange completing 28 seconds after its redirect and the OAuth session being stored, so the login worked and was simply late.
- Runs 32113599893 and 32114089597: the seeder's first write returned `POST /tenants -> 522` in both. Without fixture credentials the OWUI project matches zero spec files, so nothing in this PR ran at all.

**Correction to an earlier version of this section**, which called that 522 an external Supabase problem. It is very probably ours. The same job log carries `(ECHECKOUTTIMEOUT) unable to check out connection from the pool after 15000ms in Session mode`, which is the documented 15-client session-mode pooler ceiling being hit, and that single cause explains both symptoms: a PostgREST request that cannot get a database connection hangs until Cloudflare gives up with a 522, while a direct `pg` client gets the checkout timeout by name. Blaming the provider was the comfortable read and the evidence does not support it.

I also want to retract a piece of reasoning rather than quietly drop it: I probed `/rest/v1/` unauthenticated, got a 401, and treated that as evidence the origin had recovered. An unauthenticated request never reaches the connection pool, so that probe could not have detected the condition that matters and proved nothing.

Neither failure touches a file in this diff, and a run that never reached the suite is an absence of information rather than a result. The captures are outstanding for that reason and for no other.

**How they will be posted.** Through `scripts/post-pr-visual-proof.sh`, which uploads to a release asset. A raw link pinned to this branch 404s the moment the branch is deleted, and this repo squash-merges and deletes branches, so that mechanism produces proof which expires exactly when the PR it proves gets merged.

## A third thing, and the budget I did not raise

Investigating why the proof runs kept failing turned up a cause that is not a test problem. The consent and sign-in pages are web-console's, on a different origin, and web-console is served by a development build, so it compiles each page the first time anyone requests it. On run 32112158077 the first two login attempts produced no page within 30 seconds and the third completed the whole exchange in 21 seconds once warm. That is a route compiling on demand, not a slow protocol.

The harness now waits for the consent app to be serving before it starts timing the login. **The 30 second login budget is deliberately unchanged.** Raising it was the obvious move and the wrong one: it would have buried a cold compile inside a per-login timeout, and every future slow login would then look exactly like this one, which is how the next real regression gets absorbed instead of noticed. The new wait is a separate budget for a separate thing, the app becoming reachable at all. It also adds no new failure mode, because the assertion it precedes already requires a DOM that exists only on the consent origin.

Why that app is served in dev mode is a product question, it is deliberate per D-022, and it is filed as #967 rather than decided here. It is a plausible cause of the slow sign-in being reported in real use.

## Two things a reviewer should know

**The OWUI end-to-end harness was red before this branch and is fixed here.** It still is on `main`, at that same readiness check, which is independent evidence this fix is real and not a symptom of something else on this branch. It is also a different defect from the provisioning race PR #963 is separately fixing, so the two should not be read as one problem. Its sign-in readiness signal waited for a `button` named "New chat", and Open WebUI renders both New Chat controls as anchors carrying an `aria-label`, so that query could never match. The sign-in it guards actually succeeds; the downloaded failure screenshot shows a working signed-in chat page. A sibling file already matched either role for exactly this reason. That fix is why any of the verification above exists, and it is separate from this feature.

**Two nav tests from #938 went red the first time they ever ran**, which was in this branch after that fix. Neither is a regression here and neither was weakened. Both assumed a sidebar that starts expanded while this fixture's starts collapsed, so a text assertion resolved to the icon-only rail and read "", and the collapsed-rail test waited for a control that only exists while the sidebar is open. Each test now drives the sidebar into the state it asserts about.

---

## Rebase onto current main, 2026-08-22, and what this closes

Rebased onto `main` at `c30882491`. The base this branch was validated against, `1dc67ee69`, predates the migration of the entire database off hosted Supabase onto the self-hosted instance on the demo box, the move of the public auth origin to a same-origin `/auth/v1` route on the console, and the switch to admin-provisioned-only signup. One conflict, in `Makefile`, where `main` had added seven self-checks to `test-scripts` and this branch adds one; both are kept.

### This is the fix for the top demo blocker, issue #540

Re-confirmed live on the box on 2026-08-22: a user signs into `chat-hive.scubed.co` normally, clicks **Agents**, and is presented with an email and password form captioned that the workspace is separate from chat. Proved there by difference rather than guessed: chat's OAuth handshake mints an Open WebUI token and never writes the `sb-...-auth-token` cookie that the embedded `apps/agent-console` reads, and injecting the same session's Supabase cookies on the `chat-hive` origin makes the real workspace render immediately.

That is the second-application problem this branch removes. The credential the embedded application was looking for is no longer needed by anything, because the surface is native and authenticates with the Open WebUI session the user already has: the panel calls `/api/v1/hive/agent/*` on the chat origin, and `deploy/docker/owui-patches/hive_agent_proxy.py` brokers it server side through `get_system_oauth_token`, which is the one place the user's Supabase token is reachable. #540's own analysis named that as the only possible broker.

### Does a second credential prompt remain reachable

**Yes, by one route, and this branch does not close it.** `/agent-workspace/*` is still proxied to `apps/agent-console` by `deploy/docker/Caddyfile.owui`, and that application still renders its own email and password form, so a typed or bookmarked `https://chat-hive.scubed.co/agent-workspace` still reaches one. This branch's only edit to that file is a comment; it changes no route.

No route inside the chat interface leads there any more. `vendor/open-webui/src/lib/hive/nav.ts` points the sidebar entry at `/agents`, and the only two remaining mentions of the old path anywhere in the front end are historical comments. So the demo path is fixed and the residual is a URL nobody is shown.

Answering that path 404 was considered and rejected here rather than skipped: the Tauri desktop app targets `/agent-workspace` as its console base path (`apps/desktop/src/settings.ts`, `apps/desktop/src-tauri/src/settings.rs`), and `apps/web-console/tests/e2e/_probe/agent-workspace-flows.spec.ts` is a coverage ledger over that surface. Retiring `apps/agent-console`, or moving the desktop app onto the native surface first, is a separate decision with a far larger blast radius than this pull request.

### One coupling worth knowing before this is demoed

The proxy reads the user's Supabase token through `get_system_oauth_token`, which goes through `OAuthManager.get_oauth_token`, which refreshes five minutes before expiry and **deletes the OAuth session outright when that refresh fails**. That is issue #782, still unfixed on `main` and fixed by PR #787. Until #787 lands, this agent surface stops working roughly 55 minutes after sign-in for the same reason chat does, and the panel will correctly report "Your Hive sign-in could not be confirmed. Sign in again and retry." A demo that runs longer than that from a single sign-in needs #787 merged first.

### Review findings addressed on the rebased head

Two threads, both fixed rather than argued, plus one defect found while verifying them.

A list refresh overlapping a create or a cancel replaced the whole task array with its older answer, dropping the row the user had just submitted or reverting one they had just cancelled, and when that stale answer held no in-flight task the poll loop stopped too and the screen did not recover without a reload. A refresh now captures a mutation counter before its request goes out and discards its own answer if a create or a cancel landed while the request was open.

The third entry in the identity smell tuple in `scripts/test_owui_agent_proxy.py` could not fail: for `headers.get('Authorization')` the accessor templates expanded to `request.query_params.get('headers.get('Authorization')')`, a string no Python source can contain, so that iteration reported a pass over the header read it was named after. The header read now has its own direct test against the whole `request.headers` attribute, and it fails on purpose when the string is planted on a line that never executes.

Found while running the above: `agentTasks.test.ts`, 203 lines of assertions, was running in no job at all. The module imported `$lib/constants`, which reaches `$app/environment`, and the only runner that covers this front end runs plain vitest with no alias resolution, so the file could not be loaded. It would have turned that required check red on whichever of #951 and #952 merged second. The API base is now a parameter with a production default and the component passes the dev-aware value, so a built bundle and `npm run dev` both behave exactly as before, and the tests load with no configuration: 16 of them, one of which fails when `encodeURIComponent` is dropped from the cancel path.

### Verification on the rebased head

- 31 vendored front end tests pass across three files, 16 of them in `agentTasks.test.ts`, which had never executed before this rebase.
- `make test-scripts` green, including `test_owui_agent_proxy.py` and the seven self-checks `main` added.
- `go test ./apps/edge-api/internal/auth/... -short` green, which covers the `owui_unwrap.go` change against `main`'s newer `x-api-key` normalisation (#954).
- `Caddyfile.owui` validates against the pinned Caddy image, using `main`'s new "Every Caddyfile adapts" CI step.

### Visual proof

Posted on this pull request as release assets, with the full method, the artefacts of the stub, and the `/agent-workspace` residual in `docs/proof/agents-native-no-iframe-2026-08-22/README.md`. Measured in the page rather than asserted: zero password inputs, zero iframes, and no "Sign in" text on `/agents`.

The capture uses the real bundle from `docker build --target frontend` on this branch with a stubbed backend on a loopback origin, and says so on every artefact. A live capture is not available before merge for two reasons that are not properties of this branch: the panel needs the `owui-patches` router this branch adds, which exists only in a rebuilt image, so bundle interception against the deployed origin would 404; and the development box's `.env` still points `SUPABASE_URL` at the hosted Supabase project deleted in the migration, which now returns `ENOTFOUND`, so no session can be minted. No password was set, reset or rotated to work around that.

### Buglog entry

To be appended to `.wolf/buglog.jsonl` on `main` in a separate buglog-only pull request after this merges, per the openwolf protocol.

```json
{"id":"owui-agents-second-credential-prompt","date":"2026-08-22","error_message":"Clicking Agents in the chat sidebar presented its own email and password form captioned that the workspace is separate from chat, one click after a successful chat sign-in","root_cause":"The Agents route embedded apps/agent-console in an iframe, and that application authenticates from a Supabase SSR cookie on the chat origin which chat's OAuth handshake never writes: the handshake mints an Open WebUI token only, and the user's Supabase token is reachable only server-side inside the chat container as the stored OAuth token","fix":"Removed the iframe and rendered the task surface natively in the chat application, authenticating with the Open WebUI session the user already holds and brokering the Supabase token server-side through a FastAPI router added by owui-patches/hive_agent_proxy.py","tags":["owui","auth","agents","iframe","session","demo-blocker","issue-540"]}
{"id":"agent-tasks-stale-poll-overwrites-mutation","date":"2026-08-22","error_message":"A newly created agent task row disappeared, or a cancelled row reverted, and polling sometimes stopped entirely until the page was reloaded","root_cause":"refresh assigned the fetched list over the whole tasks array, so a poll already in flight when a create or cancel completed landed afterwards and overwrote it, and schedulePoll then decided from that stale array and stopped when it held no in-flight task","fix":"A mutation counter captured before the request goes out; refresh discards its own answer when a create or cancel landed while the request was open","tags":["svelte","race","polling","agents"]}
{"id":"agenttasks-unit-tests-never-ran","date":"2026-08-22","error_message":"agentTasks.test.ts reported no failures because it was never loaded by any job","root_cause":"The module imported $lib/constants, which reaches $app/environment, and scripts/test-owui-hive-frontend.sh runs plain vitest over copied files with no SvelteKit alias resolution, so the test file could not be loaded at all","fix":"The API base is a parameter with a production default and the component passes the dev-aware value, so the module imports nothing and the tests load with no configuration","tags":["tests","unfailable-check","vitest","sveltekit"]}
{"id":"owui-agent-proxy-smell-check-unfailable","date":"2026-08-22","error_message":"The identity smell loop in test_owui_agent_proxy.py reported a pass over a header read it was named after","root_cause":"The tuple entry headers.get('Authorization') was expanded through accessor templates into request.query_params.get('headers.get('Authorization')'), a string no Python source can contain, so that iteration asserted nothing","fix":"Separated the header smell into a direct substring test against the whole request.headers attribute, demonstrated failing on purpose","tags":["tests","unfailable-check","security"]}
```

The PR replaces the iframe-hosted agent workspace with a native Open WebUI task surface and brokers per-user agent API credentials through a constrained server-side proxy.
- Adds a guarded upstream-auth header carrier and fail-closed agent-task authentication.
- Adds native task composition, listing, cancellation, polling, and shared composer presentation.
- Updates frontend, proxy, authentication, and end-to-end verification coverage.
</details>

The PR appears safe to merge.

No blocking failure remains; the mutation-generation guard prevents stale refreshes from overwriting completed creates or cancellations, and ID filtering prevents the create/refresh duplicate-row race.
</details>

| Filename | Overview |
|----------|----------|
| apps/edge-api/internal/auth/owui_unwrap.go | Adds a stripped, shim-gated header carrier for per-user agent credentials and extends fail-closed authentication to agent-task paths. |
| deploy/docker/owui-patches/hive_agent_proxy.py | Adds the authenticated server-side broker that resolves each user’s OAuth token and forwards only the four supported task operations. |
| vendor/open-webui/src/lib/hive/AgentTasks.svelte | Implements native task composition, polling, cancellation, stale-refresh suppression, and create-row deduplication; both previously reported races are fixed. |
| vendor/open-webui/src/lib/hive/agentTasks.ts | Defines the typed agent-task client, decoding, status handling, and constrained proxy calls used by the native surface. |
| vendor/open-webui/src/routes/(app)/agents/+page.svelte | Replaces the framed agent application with the native AgentTasks component. |
| vendor/open-webui/src/lib/components/chat/MessageInput.svelte | Adopts extracted composer shell and send-button components without changing the chat composer’s behavior. |
| apps/web-console/e2e/phase-19/owui/09-agent-workspace-nav.spec.ts | Verifies native rendering, absence of iframes, proxy authentication behavior, navigation, task rows, and visual-proof capture. |

</details>

```mermaid
sequenceDiagram
  participant U as Signed-in user
  participant UI as Native OWUI agent surface
  participant P as OWUI agent proxy
  participant A as Edge API auth middleware
  participant T as Agent-task API
  U->>UI: Open /agents
  UI->>P: "GET/POST /api/v1/hive/agent/*"
  P->>P: Resolve server-side OAuth token
  P->>A: Shim authorization + upstream-auth carrier
  A->>A: Validate shim gate, strip carrier, unwrap user token
  A->>T: Request authorized as signed-in user
  T-->>A: Task response
  A-->>P: Task response
  P-->>UI: Sanitized response
  UI-->>U: Native task list and composer
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix: deduplicate the created task row ag..."](https://github.com/sakibsadmanshajib/hive/commit/f3df569f5f19901e988ac0e2209af477185bfb15) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54086925)</sub>

**Context used:**

- Knowledge Base — [Control-plane marketplace-backed agent task workflow](https://app.greptile.com/scubed/-/custom-context/knowledge-base/sakibsadmanshajib/hive/-/docs/control-plane-agent-workflows.md)

---

## Buglog entry, third (added 2026-08-23)

Alongside the two above. To be appended to `.wolf/buglog.jsonl` on `main` in
the same buglog-only pull request after this merges, per the openwolf
protocol. Not appended on this branch: a branch that appends a line conflicts
with every other branch that appended one, and an unmergeable pull request
gets no CI run at all (issue #873).

```json
{"id":"owui-unwrap-carrier-strip-unobservable","date":"2026-08-23","title":"Header-carrier strip asserted on a path no test could observe","error_message":"TestOWUIUnwrap_HeaderCarrierPresentButBlank_StrippedAndRejected asserted only the Rejected half of its own name; the Stripped half was unobservable because next never runs on a 401","root_cause":"The strip was applied to a clone of the request, so the only observation point was the downstream handler. Every rejection branch answers without calling next, so nothing on those branches could be checked, and an outer middleware still holding the pre-clone pointer would also have kept seeing a live per-user token. Moving Header.Del into the forwarding branches alone would have left the whole test file green.","fix":"Strip the carrier from the inbound request in place instead of from a clone, making the invariant one fact rather than one fact per branch, and assert in both rejection tests that the header is gone from the request the middleware was handed. Safe because net/http never re-reads request headers after the handler returns and the header is ours alone.","verification":"Green confirmed on unmodified code first. The production strip was then narrowed to fire only for a usable carrier, which is the exact regression described; all four blank sub-cases and the over-long case went red naming the new assertion, and the pass-through case went red too. Restoring the unconditional strip returned ./apps/edge-api/... to green.","tags":["auth","edge-api","owui-unwrap","test-quality","guard-cannot-fail","security","pr-951"]}
```

## Visual proof: now posted, correcting the section above

The section headed "Visual proof: not posted yet" is out of date and its own
condition has been met. Images are now visible on this page, posted as
permanent release assets through `scripts/post-pr-visual-proof.sh`, from a
live signed-in capture against a complete local stack rather than a bundle or
a CI artifact.

The capture was taken after rebasing onto `main` at `82375d07f`, where #952
and #956 have landed, so it exercises the real post-merge state. Measured in
the live DOM at both hops: zero password inputs, zero iframes, zero child
browsing contexts, zero anchors to `/agent-workspace`. `GET
/api/v1/agent/tasks` answered 200 through the running proxy chain, so the
authenticated data path is exercised and not only the rendering.
`/agent-workspace` still answers 307 to `/agent-workspace/tasks`, verified
after the rebase and deliberately unchanged.

Method, substrate and credential handling:
`docs/proof/agents-native-live-2026-08-23/README.md`.

### DIFF
--- a/.github/workflows/owui-nightly.yml
+++ b/.github/workflows/owui-nightly.yml
@@ -470,6 +470,18 @@ jobs:
             !apps/web-console/playwright-report-owui/index.html
           retention-days: 5
       - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          # Proof captures, kept out of the HTML report's own folder because
+          # that folder is cleared when the report is written. Page screenshots
+          # only: Playwright captures the page, not browser chrome, and the
+          # routes captured carry no credential in their URLs.
+          name: owui-proof-captures
+          path: apps/web-console/playwright-report-owui-proof
+          if-no-files-found: warn
+          retention-days: 5
+
+      - uses: actions/upload-artifact@v4
         if: failure()
         with:
           name: owui-compose-logs
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ test-scripts:
 	python3 scripts/test_caddy_console_auth_origin.py --self-check
 	python3 scripts/restore-storage-objects.py --self-check
 	python3 scripts/test_owui_oauth_client_auth.py
+	python3 scripts/test_owui_agent_proxy.py
 
 # Node, not python, and it downloads a pinned vitest, so it is deliberately not
 # folded into test-scripts, which is pure python with no network.
--- a/apps/edge-api/internal/auth/owui_unwrap.go
+++ b/apps/edge-api/internal/auth/owui_unwrap.go
@@ -12,6 +12,16 @@
 // OWUI would carry the shim key in Authorization, route through the
 // API-key path, and bind to the shim's principal — defeating per-user
 // audit attribution, RLS, and tenant scoping.
+//
+// The same middleware accepts that token on the UpstreamAuthHeader request
+// header, because a JSON body is not a carrier every request has. Three of
+// the four agent-task calls the chat container proxies are bodyless (GET
+// /v1/agent/tasks, GET /v1/agent/tasks/{id}, and the bodyless POST
+// .../cancel — see apps/edge-api/internal/agenttask/handler.go), so
+// `__metadata.upstream_auth` cannot reach them at all. The header is a
+// second carrier on the same trust boundary, not a second boundary: it is
+// honoured only when Authorization is exactly the shim key, which is the
+// identical gate the body carrier already sits behind.
 package auth
 
 import (
@@ -55,6 +65,13 @@ const maxOWUIUnwrapBody = 2 << 20 // 2 MiB
 // anyway but failing early here keeps the JWT path cheap.
 const maxOWUIBearerToken = 8 << 10 // 8 KiB
 
+// UpstreamAuthHeader carries the signed-in user's token on requests that
+// have no JSON body to hide it in. It is meaningful to this middleware and
+// to nothing else, so it is removed from every request that passes through
+// here — honoured, ignored, or rejected — before any handler, log or audit
+// sink can read it. Exported for the tests that assert exactly that.
+const UpstreamAuthHeader = "X-Hive-Upstream-Auth"
+
 // OWUIUnwrapConfig configures the OWUI body-metadata Authorization
 // rewrite. ShimKey is the static OPENAI_API_KEY value Open WebUI sends
 // on every upstream call; when this exact token arrives, the body is
@@ -77,6 +94,12 @@ type OWUIUnwrapConfig struct {
 //
 // Behaviour matrix:
 //
+//   - UpstreamAuthHeader present                               → always removed
+//     from the forwarded request, on every branch below. Honoured
+//     as the per-user token only under the shim key, in which case
+//     it wins over the body carrier: it is read before the body so
+//     that a bodyless request, the case it exists for, is not
+//     rejected before the header is ever consulted.
 //   - Authorization != shim key                                → pass through unchanged.
 //   - ShimKey == "" (disabled)                                 → pass through unchanged.
 //   - Content-Type not application/json (multipart, audio,
@@ -106,12 +129,39 @@ type OWUIUnwrapConfig struct {
 func OWUIUnwrap(cfg OWUIUnwrapConfig) func(http.Handler) http.Handler {
 	shimKey := strings.TrimSpace(cfg.ShimKey)
 	return func(next http.Handler) http.Handler {
-		if shimKey == "" {
-			return next
-		}
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if !hasShimAuthorization(r.Header.Get("Authorization"), shimKey) {
+			// Take the carrier off the request before anything else runs,
+			// including when the middleware is disabled and when the
+			// credential is not the shim key. It is ours alone, so a handler
+			// that can read it is a handler that could later be taught to
+			// trust it, and a header a client controls must never become a
+			// principal by that route.
+			carrier := strings.TrimSpace(r.Header.Get(UpstreamAuthHeader))
+			if carrier != "" {
+				r = r.Clone(r.Context())
+				r.Header.Del(UpstreamAuthHeader)
+			}
+			if shimKey == "" || !hasShimAuthorization(r.Header.Get("Authorization"), shimKey) {
 				next.ServeHTTP(w, r)
+				return
+			}
+			if carrier != "" {
+				// Fail closed on a carrier that is present but unusable. The
+				// only thing that sets this header is our own forwarder, so a
+				// value it cannot mean is a broken forwarder, and forwarding
+				// with the shim key still on Authorization would bill and
+				// audit the call against the shim's principal.
+				token := ""
+				if len(carrier) <= maxOWUIBearerToken {
+					token = normalizeUpstreamAuth(carrier)
+				}
+				if token == "" {
+					writeAuthError(w, http.StatusUnauthorized, "UNAUTHENTICATED", "invalid token")
+					return
+				}
+				r2 := r.Clone(context.WithValue(r.Context(), owuiUnwrappedKey{}, true))
+				r2.Header.Set("Authorization", "Bearer "+token)
+				next.ServeHTTP(w, r2)
 				return
 			}
 			// This check has to come before the pass-through below, not
@@ -235,18 +285,40 @@ func hasShimAuthorization(header, shimKey string) bool {
 // carry a per-user token, i.e. whether a missing `__metadata.upstream_auth`
 // is a failure rather than the expected shape.
 //
-// Chat completions are the only path the `hive_jwt_forward` Function
-// decorates, and the only path where running under the shim principal
-// silently mis-attributes billing and audit. Open WebUI's other upstream
-// calls (document-RAG embeddings via RAG_OPENAI_API_KEY, and text-to-speech)
-// are never decorated and authenticate as the shim account by design, so
-// they must keep passing through.
+// Two families qualify, for the same reason. Chat completions are the path
+// the `hive_jwt_forward` Function decorates. The agent-task lifecycle is the
+// path the chat container's own agent proxy decorates
+// (deploy/docker/owui-patches/hive_agent_proxy.py). On both, running under
+// the shim principal silently mis-attributes billing and audit, and on the
+// agent path it would additionally list and cancel the shim account's tasks
+// rather than the signed-in user's. Open WebUI's other upstream calls
+// (document-RAG embeddings via RAG_OPENAI_API_KEY, and text-to-speech) are
+// never decorated and authenticate as the shim account by design, so they
+// must keep passing through.
 //
-// ponytail: exact-path list rather than a prefix rule, because /v1/chat is
-// the only prefix at stake today. If a second per-user OWUI path appears
-// (for example a future OWUI RAG chat route), add it here.
+// The agent arm is a prefix rather than an exact list because the subtree
+// carries a task id: /v1/agent/tasks/{id} and /v1/agent/tasks/{id}/cancel.
 func requiresPerUserAuth(path string) bool {
-	return path == "/v1/chat/completions"
+	return path == "/v1/chat/completions" ||
+		path == "/v1/agent/tasks" ||
+		strings.HasPrefix(path, "/v1/agent/tasks/")
+}
+
+// normalizeUpstreamAuth reduces a carried credential to a bare token,
+// tolerating both "Bearer <token>" (what the forwarders write today) and a
+// bare token (so a future revision that drops the scheme word is not a
+// silent auth failure). Returns "" for anything with no token in it, which
+// every caller treats as fail-closed.
+//
+// Shared by the header carrier and the body carrier so the two cannot drift
+// into accepting different shapes of the same credential.
+func normalizeUpstreamAuth(value string) string {
+	value = strings.TrimSpace(value)
+	scheme, rest, hasSpace := strings.Cut(value, " ")
+	if hasSpace && strings.EqualFold(scheme, "Bearer") {
+		return strings.TrimSpace(rest)
+	}
+	return value
 }
 
 // isJSONContent reports whether the Content-Type media type is
@@ -325,11 +397,13 @@ func unwrapOWUIBody(raw []byte) (rewritten []byte, token string, status unwrapSt
 	if len(authStr) > maxOWUIBearerToken {
 		return out, "", unwrapTokenTooLong
 	}
-	scheme, tokenPart, hasSpace := strings.Cut(authStr, " ")
-	if hasSpace && strings.EqualFold(scheme, "Bearer") {
-		return out, strings.TrimSpace(tokenPart), unwrapOK
+	token = normalizeUpstreamAuth(authStr)
+	if token == "" {
+		// A present-but-empty upstream_auth is a pipeline that stopped
+		// forwarding, not a request that meant to run as the shim. Reported
+		// as a missing carrier so it takes the fail-closed arm and the warn
+		// log, rather than forwarding with the shim key on Authorization.
+		return out, "", unwrapNoMetadata
 	}
-	// Tolerate raw token (no scheme word). The pipeline writes
-	// "Bearer <jwt>" today but a future revision may drop the prefix.
-	return out, authStr, unwrapOK
+	return out, token, unwrapOK
 }
--- a/apps/edge-api/internal/auth/owui_unwrap.go
+++ b/apps/edge-api/internal/auth/owui_unwrap.go
@@ -140,10 +140,23 @@ func OWUIUnwrap(cfg OWUIUnwrapConfig) func(http.Handler) http.Handler {
 			// value, so a whitespace-only or repeated header is removed too.
 			// Header.Del drops every value under the key, which is what makes
 			// "stripped on every branch" true rather than nearly true.
+			//
+			// The removal is in place, on the inbound request, and not on a
+			// clone. A clone would only protect code reachable through `next`,
+			// and this header has to disappear for everyone: an outer
+			// middleware still holding the original pointer would otherwise
+			// keep seeing a live per-user token, and no branch that answers
+			// without calling `next` at all, which is every rejection path
+			// below, would strip anything observable. In place makes the
+			// invariant one fact rather than one fact per branch, and lets a
+			// test assert it on the rejection paths too. Mutating the inbound
+			// header map is safe here because net/http never re-reads request
+			// headers after the handler returns, and this header is ours: no
+			// caller upstream of edge-api has any reason to send it, and none
+			// is permitted to.
 			_, carrierPresent := r.Header[http.CanonicalHeaderKey(UpstreamAuthHeader)]
 			carrier := strings.TrimSpace(r.Header.Get(UpstreamAuthHeader))
 			if carrierPresent {
-				r = r.Clone(r.Context())
 				r.Header.Del(UpstreamAuthHeader)
 			}
 			if shimKey == "" || !hasShimAuthorization(r.Header.Get("Authorization"), shimKey) {
--- a/apps/edge-api/internal/auth/owui_unwrap.go
+++ b/apps/edge-api/internal/auth/owui_unwrap.go
@@ -136,8 +136,13 @@ func OWUIUnwrap(cfg OWUIUnwrapConfig) func(http.Handler) http.Handler {
 			// that can read it is a handler that could later be taught to
 			// trust it, and a header a client controls must never become a
 			// principal by that route.
+			// Presence is tested on the header map rather than on the trimmed
+			// value, so a whitespace-only or repeated header is removed too.
+			// Header.Del drops every value under the key, which is what makes
+			// "stripped on every branch" true rather than nearly true.
+			_, carrierPresent := r.Header[http.CanonicalHeaderKey(UpstreamAuthHeader)]
 			carrier := strings.TrimSpace(r.Header.Get(UpstreamAuthHeader))
-			if carrier != "" {
+			if carrierPresent {
 				r = r.Clone(r.Context())
 				r.Header.Del(UpstreamAuthHeader)
 			}
@@ -145,24 +150,20 @@ func OWUIUnwrap(cfg OWUIUnwrapConfig) func(http.Handler) http.Handler {
 				next.ServeHTTP(w, r)
 				return
 			}
-			if carrier != "" {
-				// Fail closed on a carrier that is present but unusable. The
-				// only thing that sets this header is our own forwarder, so a
-				// value it cannot mean is a broken forwarder, and forwarding
-				// with the shim key still on Authorization would bill and
-				// audit the call against the shim's principal.
-				token := ""
+			// Fail closed on a carrier that is present but unusable. The only
+			// thing that sets this header is our own forwarder, so a value it
+			// cannot mean is a broken forwarder, and forwarding with the shim
+			// key still on Authorization would bill and audit the call against
+			// the shim's principal.
+			headerToken := ""
+			if carrierPresent {
 				if len(carrier) <= maxOWUIBearerToken {
-					token = normalizeUpstreamAuth(carrier)
+					headerToken = normalizeUpstreamAuth(carrier)
 				}
-				if token == "" {
+				if headerToken == "" {
 					writeAuthError(w, http.StatusUnauthorized, "UNAUTHENTICATED", "invalid token")
 					return
 				}
-				r2 := r.Clone(context.WithValue(r.Context(), owuiUnwrappedKey{}, true))
-				r2.Header.Set("Authorization", "Bearer "+token)
-				next.ServeHTTP(w, r2)
-				return
 			}
 			// This check has to come before the pass-through below, not
 			// after it. A per-user path can only receive its user token in a
@@ -172,6 +173,12 @@ func OWUIUnwrap(cfg OWUIUnwrapConfig) func(http.Handler) http.Handler {
 			// rejection further down simply by omitting Content-Type.
 			jsonBody := r.Body != nil && isJSONContent(r.Header.Get("Content-Type"))
 			if !jsonBody {
+				// The bodyless case the header carrier exists for. Nothing to
+				// read, nothing to strip, so forward immediately.
+				if headerToken != "" {
+					forwardUnwrapped(w, r, next, headerToken, nil)
+					return
+				}
 				if requiresPerUserAuth(r.URL.Path) {
 					warnMissingUpstreamAuth(r, 0, true)
 					writeAuthError(w, http.StatusUnauthorized, "UNAUTHENTICATED", missingUserTokenMessage)
@@ -199,7 +206,22 @@ func OWUIUnwrap(cfg OWUIUnwrapConfig) func(http.Handler) http.Handler {
 				writeAuthError(w, http.StatusRequestEntityTooLarge, "PAYLOAD_TOO_LARGE", "request body too large")
 				return
 			}
-			rewritten, token, status := unwrapOWUIBody(raw)
+			// The body is parsed even when the header already supplied the
+			// token, because __metadata must be stripped either way. Forwarding
+			// it would leak proxy-layer fields to downstream handlers, to the
+			// audit chain and on to a provider, which is the reason the strip
+			// is unconditional in unwrapOWUIBody.
+			rewritten, bodyToken, status := unwrapOWUIBody(raw)
+			token := bodyToken
+			if headerToken != "" {
+				// The header wins. It is the carrier our own forwarder sets on
+				// the request line, and a body that also carries one is either
+				// a second forwarder or a caller trying its luck; neither
+				// should be able to displace the credential already accepted.
+				token = headerToken
+				forwardUnwrapped(w, r, next, token, rewritten)
+				return
+			}
 			switch status {
 			case unwrapTokenTooLong:
 				writeAuthError(w, http.StatusUnauthorized, "UNAUTHENTICATED", "invalid token")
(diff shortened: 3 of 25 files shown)


sakibsadmanshajib:
## Review, and what changed because of it

Six independent passes over this diff, none of them sharing context: CodeRabbit's CLI, a Go pass over the middleware, a Python pass over the proxy, a security pass over the whole credential path, a frontend pass over the extraction and the port, and a plain read of the change against what was actually asked for. Everything below is either fixed on the branch or answered.

### The one that mattered

**The header carrier returned before the JSON body was parsed.** A request presenting both carriers had its `__metadata` block forwarded straight to the handler and into the audit chain. That strip is unconditional for a reason, and the early return had quietly reintroduced exactly what it exists to prevent. The body is now always parsed and rewritten and the header still wins as the credential, with a test that fails if `__metadata` or a body token survives.

### The rest, in the order they were found

- **The header was only stripped when its trimmed value was non-empty.** Go's wire parser trims ASCII space and tab; `strings.TrimSpace` also trims U+00A0. So a header carrying a non-breaking space arrived non-empty, trimmed to nothing, and survived to the handler. Presence is now tested on the header map and `Del` removes every value under the key. Pinned by a test covering spaces, a tab, a non-breaking space and a bare scheme word.
- **A scheme-only carrier became the literal token `Bearer`.** Found by that new test rather than by reasoning about the branch, which is the argument for writing it. Not a hole, since a nonsense token fails validation, but it reported itself to the customer as their session being invalid instead of as the forwarder sending nothing. Refused at the boundary now.
- **`schedulePoll` read reactive values Svelte had not flushed.** The poll scheduled immediately after a submit decided on the pre-submit list and stopped the loop that the newly queued task is the whole reason to keep running. It computes from the current values now.
- **The pack toggle lost its `aria-describedby`** in the port, so the line describing the selected mode was sighted-only. Restored.
- **The send button had no accessible name** on the agent surface, where it is not wrapped in chat's tooltip. It takes an optional label, left undefined on chat so the extraction stays invisible there.
- **A 401 or 403 was being reported with copy promising a retry.** Neither changes on the next attempt: 401 means the session cannot reach the agent service at all, 403 is the Cowork gate. Both now carry the server's own sentence and stop the poll instead of asking a settled question every few seconds.
- **Cancelling the last in-flight task** left one more request armed before the loop noticed. It stops immediately now.
- **A missing `tasks` array was reported as an empty list.** Taken, but not literally as suggested: edge-api answers `map[string]any{"tasks": tasks}` and Go marshals a nil slice as `null`, so `tasks: null` is a real empty list and raising on it would have shown a parse error to every user who has none. Absent, or present and neither null nor an array, raises; null and `[]` read as empty.
- Smaller ones: the upstream status was asserted nowhere in the proxy self-check; the error-vocabulary test checked the exception class but not the sentence; the page duplicated an `overflow-y` the stylesheet already sets.

### Two questions asked of the security pass, both answered

**Can `requiresPerUserAuth`'s new prefix be evaded by path shape?** No. `requiresPerUserAuth`, the `/v1/` gate in `authSelectorMiddleware` and `http.ServeMux` all read the same already-percent-decoded `r.URL.Path`; there is no second normalization for a mismatch to hide behind. A non-clean path gets a 301 from the mux rather than reaching the handler, and following it re-enters the whole middleware chain on the clean path. Paths are case-sensitive throughout, so a case variant 404s rather than being reclassified.

**Does Caddy need a block for the new `/api/v1/hive/*` prefix?** No, and it should not get one: it has to stay browser-reachable, that being the point. It matches none of the existing blocks, which is correct rather than accidental, and it is not the shape #770 and #771 blocked, those being stock routers that proxy to a caller-supplied destination with a stored credential. This is four named operations behind `get_verified_user` with a UUID-validated id and an outbound body rebuilt from two named fields. A comment in `Caddyfile.owui` now records that, so the next reviewer does not have to re-derive it from three regexes.

### Verified rather than assumed

- **The extraction is invisible on chat.** The send button's class logic was rewritten from one inline expression into a `disabled`/`pending` pair; with `E = prompt === '' && files.length === 0` and `U = uploadPending`, the old flag `!E || U` and the new `!(E||U) || U` are the same function for every input. Svelte omits an attribute bound to `undefined`, so the new optional label adds nothing to chat's DOM.
- **The aiohttp exception surface**, against the pinned `aiohttp==3.13.5` on Python 3.11.15 in the actual image, driven through real sockets rather than read from docs. A total-timeout breach raises `asyncio.TimeoutError` and is not a `ClientError`; everything else aiohttp can raise here is a `ClientError`; decode failures are `ValueError`. The three arms are exhaustive, and the order matters because `ServerTimeoutError` inherits from both.

### One stream still open

The cross-model pass has not reported. If it comes back with anything, it goes here; if its tooling is unavailable it will be recorded as skipped rather than as a clean pass.

### Not taken

The double `r.Clone` on a request that carries a header and is then forwarded is an extra allocation on a low-volume path, not a correctness problem. Left alone.

sakibsadmanshajib:
## Visual proof

Infrastructure note, not a review of this PR's feature. Orchestrator rule 8's visual-proof mechanism changed (see PR #961, .wolf/decisions.md D-042): a docs/proof/ image linked by a raw.githubusercontent.com URL pinned to this branch's name would 404 the instant this branch is deleted on merge. This comment demonstrates the replacement, scripts/post-pr-visual-proof.sh, working on this PR rather than only on a scratch one. The image below is a generic placeholder, not a capture of this PR's actual change. Whoever finishes this PR should capture the real agent-console-native screenshot this PR needs and post it the same way: scripts/post-pr-visual-proof.sh 951 <path-to-real-screenshot> --caption "...". Nothing was pushed to this PR's branch.

PR #954, opened 2026-08-17
fix: normalize x-api-key before the JWT/API-key auth selector

sakibsadmanshajib:
## Summary

A real Anthropic SDK client constructed the documented way (`Anthropic(api_key=...)`) sends its credential on `x-api-key`, never `Authorization`. Confirmed directly from the installed SDK's own source: `_api_key_auth` returns `{"X-Api-Key": api_key}`; `Authorization` only appears when the caller passes `auth_token=` instead.

`apps/edge-api/internal/anthropic/handler.go` already ships `APIKeyNormalizer`, unit-tested in isolation, which rewrites `x-api-key` to `Authorization: Bearer`. The bug was entirely in where it was wired: `main.go` applied it only at the mux leaf (`mux.Handle("/v1/messages", anthropic.APIKeyNormalizer(anthropicHandler))`), but `authSelectorMiddleware` (which decides API-key vs JWT path by inspecting `Authorization`) wraps the **entire mux** and runs first. Every `x-api-key`-only request had no `Authorization` header at the point the selector decided, fell through to the JWT path unconditionally, and 401'd with `"missing bearer"` regardless of key validity.

**Reproduced live** against the deployed gateway (`https://api-hive.scubed.co`) with the real `anthropic` Python SDK (0.122.0): the default `api_key=` construction always 401s with `missing bearer`; the same call with `auth_token=` (forcing an `Authorization` header directly) reaches the real API-key path instead. This is the empirical confirmation that the owner's literal ask — point a real SDK at Hive by changing only `base_url` and the key — could not have worked on any key before this fix.

## Fix

- `authSelectorMiddleware` (`apps/edge-api/cmd/server/main.go`): apply `anthropic.APIKeyNormalizer` around the selector itself, so `x-api-key` is normalized before `auth.Selector` inspects `Authorization`, for every `/v1/*` route. No-op wherever `Authorization` is already set.
- New `apps/edge-api/internal/anthropic/errors.go`: reshapes every `/v1/messages` error this surface owns (its own validation refusals, and delegated-chain errors it forwards) into Anthropic's real error envelope (`{"type":"error","error":{"type","message"}}`) instead of the OpenAI-shaped one used unconditionally before. Verified live against `api.anthropic.com` and the installed SDK's `_exceptions.py`: the SDK's exception **class** is driven by HTTP status, not body shape, so this never crashed a client, but `err.type` read values outside Anthropic's documented enum and `err.request_id` was always empty. Reuses the existing provider-blind sanitizer verbatim (through a small local `http.ResponseWriter` capture) rather than duplicating its regexes; no sanitization logic changed.
- `docs/anthropic-sdk-integration.md`: exact base URL, auth header, copy-pasteable Python/TypeScript snippets, an OpenCode-style config, the model aliases that actually answer `/v1/messages` (`hive-default`, `hive-fast`, `hive-auto`, live from `GET /catalog/models`), and every known limitation found during this pass, each one verified rather than assumed.

## Documented, not fixed here (filed as follow-up, out of this pass's scope)

- Errors emitted **before** the request reaches the anthropic package at all (the selector's own 401, `budgetGate`'s 429) still use the generic envelope: fixing those touches shared `internal/middleware`/`internal/errors` code used by every other surface, a materially larger blast radius, deliberately deferred.
- `middleware.CompatHeaders()` (global, unconditional) stamps `x-request-id`/`openai-version`/`openai-processing-ms` on every response including `/v1/messages` — confirmed live. Harmless to the SDK (unknown headers ignored) but means `.request_id` stays empty and leaks OpenAI-family headers onto an Anthropic-shaped endpoint.
- `top_k`, extended thinking (`thinking`/`redacted_thinking` blocks), prompt caching (`cache_control`), `container`/server tools, `service_tier`, `inference_geo`, and stop-sequence identification in the response are not implemented. Neither of Hive's two configured providers (OpenRouter, Groq) supports Anthropic's native versions of these, so these are provider-capability gaps, not oversights.

## Test plan

- [x] `TestAuthSelectorMiddlewareAcceptsXAPIKeyHeader` — RED before the fix (confirmed), GREEN after
- [x] `TestAuthSelectorMiddlewareStillRoutesJWTBearerToJWTPath` — regression guard, a non-hk_ session bearer still routes to the JWT path only
- [x] `TestHandler_ValidationErrorUsesAnthropicEnvelope`, `TestHandler_DownstreamErrorUsesAnthropicEnvelope`, `TestHandler_UpstreamErrorUsesAnthropicEnvelope` — new envelope shape assertions
- [x] Full existing `apps/edge-api/internal/anthropic/...` suite passes unchanged (message/status assertions matched verbatim; only the envelope changed)
- [x] Full `apps/edge-api/...` suite green (`go test ./apps/edge-api/... -count=1 -short`)
- [x] `go vet ./apps/edge-api/...` clean
- [x] Live reproduction of the bug and the intended fix behavior against the deployed box, with the real Anthropic Python SDK, both pre- and post-fix code paths (not deployed yet; deploy is the orchestrator's step)

## Buglog entry

```json
{"date":"2026-08-17","error_message":"real Anthropic SDK client 401s with 'missing bearer' regardless of API key validity when pointed at Hive with only base_url and api_key set","root_cause":"authSelectorMiddleware inspects only the Authorization header and runs outside the mux; anthropic.APIKeyNormalizer (x-api-key -> Authorization: Bearer) was wired only at the mux leaf for /v1/messages, so it never ran before the selector's routing decision","fix":"apply anthropic.APIKeyNormalizer around the selector itself in authSelectorMiddleware (apps/edge-api/cmd/server/main.go), before auth.Selector inspects Authorization","tags":["anthropic","auth","edge-api","x-api-key","routing"]}
```

## Boundary honored

No billing, reservation, ledger, or metering code touched (a separate agent owns that concurrently). No API key, token, or credential value printed anywhere in this PR or its tests (all test keys are obvious literals: `hk_test_key`, `hk_bogus_test_key_...`).

### DIFF
--- a/apps/edge-api/internal/anthropic/errors.go
+++ b/apps/edge-api/internal/anthropic/errors.go
@@ -1,0 +1,118 @@
+package anthropic
+
+import (
+	"encoding/json"
+	"net/http"
+
+	apierr "github.com/sakibsadmanshajib/hive/apps/edge-api/internal/errors"
+)
+
+// anthropicErrorEnvelope is the wire shape every real Anthropic Messages error
+// response uses: a top-level "type":"error" plus a nested error object, e.g.
+//
+//	{"type":"error","error":{"type":"authentication_error","message":"..."}}
+//
+// The rest of edge-api uses the OpenAI envelope ({"error":{"message","type",
+// "param","code"}}, no top-level "type"), which this surface used
+// unconditionally before this file existed. The Anthropic SDK's exception
+// CLASS selection keys off HTTP status, not body shape, so that half kept
+// working either way; but anthropic._exceptions.APIStatusError reads
+// body["error"]["type"] into its own .type attribute, and the OpenAI envelope
+// both lacks the wrapping "type":"error" and carries the wrong enum member
+// there (e.g. "UNAUTHORIZED", which is not one of Anthropic's documented
+// error types). A caller inspecting either field got nothing usable.
+type anthropicErrorEnvelope struct {
+	Type  string             `json:"type"`
+	Error anthropicErrorBody `json:"error"`
+}
+
+type anthropicErrorBody struct {
+	Type    string `json:"type"`
+	Message string `json:"message"`
+	// Code is a Hive extension carrying the finer-grained OpenAI-style error
+	// code where the delegated chain supplied one (e.g. "invalid_api_key").
+	// The real Anthropic API never sends this field; an Anthropic SDK ignores
+	// unknown JSON fields, so surfacing it here is additive, not a compliance
+	// risk, and keeps the detail available for anyone inspecting the raw body.
+	Code string `json:"code,omitempty"`
+}
+
+// anthropicErrorType maps an HTTP status to the closest member of Anthropic's
+// documented error-type enum: invalid_request_error, authentication_error,
+// permission_error, not_found_error, rate_limit_error, api_error,
+// overloaded_error, request_too_large. Status is authoritative here, not
+// whatever "type" string a delegated OpenAI-shaped body carried: status is
+// what actually drives the real SDK's exception class, so it is the one
+// value this surface must never get wrong.
+func anthropicErrorType(status int) string {
+	switch status {
+	case http.StatusBadRequest:
+		return "invalid_request_error"
+	case http.StatusUnauthorized:
+		return "authentication_error"
+	case http.StatusForbidden:
+		return "permission_error"
+	case http.StatusNotFound:
+		return "not_found_error"
+	case http.StatusRequestEntityTooLarge:
+		return "request_too_large"
+	case http.StatusTooManyRequests:
+		return "rate_limit_error"
+	case 529: // Anthropic's own "overloaded" status; we do not emit it today.
+		return "overloaded_error"
+	default:
+		return "api_error"
+	}
+}
+
+// writeAnthropicError writes the Anthropic-shaped error envelope. code is the
+// optional Hive-extension field (see anthropicErrorBody.Code) and may be "".
+func writeAnthropicError(w http.ResponseWriter, status int, message string, code string) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	_ = json.NewEncoder(w).Encode(anthropicErrorEnvelope{
+		Type: "error",
+		Error: anthropicErrorBody{
+			Type:    anthropicErrorType(status),
+			Message: message,
+			Code:    code,
+		},
+	})
+}
+
+// reshapeToAnthropicError re-emits an already-sanitized OpenAI-shaped error
+// body (the delegated chat/inference chain's own refusal, which already ran
+// through the provider-blind sanitizer at the upstream boundary) in
+// Anthropic's envelope. message and code are extracted best-effort from raw;
+// raw may be non-JSON, carry no error object, or be empty, in which case a
+// generic message for the status is used instead. This never re-sanitizes:
+// it only reshapes an envelope that was already safe to return.
+func reshapeToAnthropicError(w http.ResponseWriter, status int, raw []byte) {
+	message, code := extractOpenAIErrorFields(raw)
+	if message == "" {
+		message = genericMessageForStatus(status)
+	}
+	writeAnthropicError(w, status, message, code)
+}
+
+func extractOpenAIErrorFields(raw []byte) (message string, code string) {
+	var body apierr.OpenAIError
+	if err := json.Unmarshal(raw, &body); err != nil {
+		return "", ""
+	}
+	if body.Error.Code != nil {
+		code = *body.Error.Code
+	}
+	return body.Error.Message, code
+}
+
+func genericMessageForStatus(status int) string {
+	switch status {
+	case http.StatusTooManyRequests:
+		return "the request was rate limited."
+	case http.StatusServiceUnavailable, http.StatusGatewayTimeout, http.StatusBadGateway:
+		return "the request could not be completed."
+	default:
+		return "the request failed."
+	}
+}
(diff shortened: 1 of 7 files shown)


apps/edge-api/internal/anthropic/errors.go
```diff
@@ -0,0 +1,118 @@
...
+func genericMessageForStatus(status int) string {
+	switch status {
+	case http.StatusTooManyRequests:
+		return "the request was rate limited."
+	case http.StatusServiceUnavailable, http.StatusGatewayTimeout, http.StatusBadGateway:
+		return "the request could not be completed."
+	default:
+		return "the request failed."
+	}
+}
```
sakibsadmanshajib, line 118: **[CodeRabbit CLI stream]** major: Remove the Anthropic-specific envelope and translation helpers (`anthropicErrorEnvelope`, `anthropicErrorBody`, `anthropicErrorType`, `writeAnthropicError`, `reshapeToAnthropicError`). Update all POST /v1/messages failure paths to use the existing shared OpenAI-shaped error boundary consistently, including front-door and delegated failures.

---

**Rebuttal (not applying this suggestion):** this reads as a generic consistency heuristic ("every surface should use the one shared error shape") that does not hold here, and reverting would reintroduce the exact defect this PR fixes. Confirmed live against the real `api.anthropic.com`: every real Anthropic error response is `{"type":"error","error":{"type","message"}}`, never the OpenAI `{"error":{"message","type","code"}}` shape the rest of edge-api uses. A real Anthropic SDK client (confirmed from the installed SDK 0.122.0 source, `_exceptions.py`) reads `err.type` from `body["error"]["type"]` and expects the top-level `type:"error"` wrapper; the OpenAI envelope has neither. Consistency with the rest of edge-api is not the goal on this one surface: wire compatibility with a real, unmodified Anthropic SDK is the explicit task this PR exists for, and the two envelopes are mutually exclusive by construction (Anthropic's own documented wire format defines this shape; there is no third shape that satisfies both).

This is scoped correctly, not over-broad: only the errors this surface itself owns (its own validation refusals, and delegated-chain errors it forwards) are reshaped, via a single new file (`errors.go`) that reuses the existing shared sanitizer (`apierr.WriteProviderBlindUpstreamError`) verbatim rather than duplicating its logic. Pre-dispatch middleware errors are deliberately left on the shared shape (documented in the PR body and the vault spec) precisely to keep the blast radius this suggestion would spread even further contained to shared code nobody asked to change this pass, which is the actual place where "reuse the shared boundary as-is" is the right call, not this surface's own error paths.

Resolving this thread on that basis.

sakibsadmanshajib:
## Adversarial review summary (dynamic selector, D-038)

This PR touches an auth-routing boundary (`authSelectorMiddleware`) and an
input-parsing/error-serialization boundary (`/v1/messages` error envelope),
so the mandatory security review applies.

**CodeRabbit CLI** — ran (`coderabbit review --base main --agent`). 1 finding
(major, `errors.go`: suggested removing the Anthropic-specific envelope in
favor of the shared OpenAI-shaped one). Addressed with a reasoned rebuttal
inline on the finding and resolved: reverting would reintroduce the exact
defect this PR fixes, confirmed live against `api.anthropic.com` and the
installed Anthropic SDK source that the two envelopes are mutually exclusive
by construction, and the change is scoped to only the errors this surface
owns, reusing the shared sanitizer verbatim rather than duplicating it. See
the resolved thread on `errors.go`.

**ecc:code-review** — SKIPPED. Structural capability gap: this task was
dispatched to a subagent whose toolset is Read/Write/Edit/Bash only, with no
Skill-invocation tool available to run a plugin skill. Reporting this
precisely per the Gate Compliance rule rather than attempting a workaround.

**`/codex:adversarial-review`** — SKIPPED. Attempted (`codex review --base
main`); the CLI reported `ERROR: You've hit your usage limit... try again at
Sep 10th, 2026`. A genuinely unavailable stream, not a clean pass — recorded
as SKIPPED per the pipeline's fallback rule, not treated as passing review.

**Plain adversarial pass (ticket-first) and manual security pass** — run
directly, in place of the two tool-gated streams above, since the ticket
itself (owner: point a real Anthropic SDK at Hive by changing only base_url
and the key) is squarely a correctness-and-security question this pass could
still answer without those tools:

- Auth-bypass check: `anthropic.APIKeyNormalizer` only rewrites `x-api-key` to
  `Authorization: Bearer` when `Authorization` is **not already set**
  (`normalizeAPIKeyHeader`), so an existing credential can never be
  overridden, and `auth.Selector`'s own anti-smuggling check (any Authorization
  value containing a comma or whitespace is refused as a possible
  credential-smuggling attempt, `selector.go:33-40`) still runs downstream of
  the normalizer unchanged, so a multi-value or malformed `x-api-key` still
  gets refused rather than routed. No new bypass.
- Broadening the normalizer from the `/v1/messages` mux leaf to every `/v1/*`
  route (via `authSelectorMiddleware`) grants no new capability to an invalid
  credential; it only lets an already-valid `hk_` key authenticate via an
  additional, standard header name on more routes, matching
  `APIKeyNormalizer`'s own pre-existing doc comment ("normalising Anthropic
  x-api-key credentials... before dispatching", never scoped to one route in
  its own text).
- Information disclosure check: `reshapeToAnthropicError` extracts only
  `.error.message`/`.error.code` from an already-sanitized body (either the
  delegated chain's own provider-blind refusal, or the output of calling the
  existing `apierr.WriteProviderBlindUpstreamError` sanitizer verbatim into a
  local, in-memory `http.ResponseWriter` capture) and never introduces a new
  source of raw upstream text. The `error.type` field is remapped from a
  fixed, hardcoded status-to-enum table (`anthropicErrorType`), with no
  attacker-controlled input reaching it. No new leak.
- Pre-existing, unchanged precedence question (not introduced by this PR,
  noted for completeness): if a caller sends both `x-api-key` and
  `Authorization`, the existing `Authorization` always wins and `x-api-key` is
  silently ignored. This was true before this PR (already the behavior at the
  old mux-leaf wiring) and is unchanged by moving where it fires.
- No credential, token, or key value appears anywhere in the diff; all test
  literals are obviously fake (`hk_test_key`, `hk_bogus_test_key_...`,
  `hk_live_test_key`).

Verdict: no security or correctness findings beyond the one CodeRabbit
finding above, which is resolved.

PR #955, opened 2026-08-17
test: prove reaper SQL and panic-survives-release live, full billing audit

sakibsadmanshajib:
## Summary

Owner asked for a full audit of the billing/metering/reservation/ledger path: reservations released on every terminal path, no path serves a request unbilled, both directions billed correctly, math/big everywhere, Anthropic surface metered identically to OpenAI. This PR is the audit's code output: two new regression tests proving the two real gaps found. No production code changed.

## Audit method

1. Read `.wolf/decisions.md` D-031 through D-034 (credit unit, one-model-one-price, non-token modality rejection, fail-closed money path) and the extensive in-code documentation of prior incidents (#616 stranded holds, #626 three days unbilled, #652/#663/#672 idempotency key shape, #688 catalog-price settlement, #746 session chat free-serve, #918 reservation/hold atomicity).
2. Read every dispatch site end to end: `apps/edge-api/internal/chat/dispatch.go` and `billing.go` (session chat), `apps/edge-api/internal/inference/orchestrator.go` and `stream.go` (API-key sync and streaming), `apps/edge-api/internal/anthropic/handler.go` (Anthropic Messages surface), and the control-plane side: `apps/control-plane/internal/accounting/service.go`, `repository.go`, `reaper.go`, `apps/control-plane/internal/ledger/service.go`.
3. Enumerated the existing test suite by name (`grep '^func Test'` across every `*_test.go` in `accounting`, `ledger`, `chat`, `inference`, `anthropic`) before writing anything new, to avoid duplicating coverage. Result: every terminal path the owner named (success, upstream error, client disconnect mid-stream, context cancellation, timeout via finalize failure, no-usage-block stream) already has a named regression test (`TestExecuteStreaming_ClientDisconnect_*`, `TestExecuteStreaming_FinalizeError_ReleasesReservationInstead`, `TestSessionChatReleasesTheHoldWhenSettlementFails`, `TestFinalizeReservationTwiceChargesOnce`, `TestMessages_APIKeyStreamingPathReservesAndSettlesCredits`, etc).
4. Confirmed `math/big` is the only arithmetic path: `apps/edge-api/internal/metering/precedence.go` (`ChargeCredits`), never float64.
5. Confirmed the Anthropic surface is a pure translator that always forces `stream=true` and `stream_options.include_usage=true` before delegating to the exact same `/v1/chat/completions` handler chain used by the OpenAI surface (`anthropic/handler.go`), so it cannot drift from that surface's metering. `routing_metering_test.go` already asserts reservation and settlement on both the session and API-key paths, streaming included.
6. Ran the full `accounting`, `ledger`, `chat`, `inference`, `anthropic`, `audio`, `images`, `batches`, `usage` test suites (`go test -short ./apps/edge-api/... ./apps/control-plane/...`): all green. Two unrelated failures surfaced (`auditworker` serialization retry, `marketplace` RLS permission) from gaps in my locally-bootstrapped ephemeral schema versus the full CI bootstrap; neither package is in this audit's scope (billing/metering/reservation/ledger) and neither was touched.

## Live evidence

Read the real demo Supabase project (pooler, read-only) directly:

- Reservation status counts, all time: 1964 `finalized`, 149 `needs_reconciliation`, 65 `released`. Zero rows in `active`/`expanded` right now.
- Stranded holds (active/expanded, >1h old, no in-flight batch): **0**.
- Over-released reservations (`released_credits > reserved_credits`, the D-918 corruption signal): **0**.

Then stood up an ephemeral `pgvector/pgvector:pg17` database, applied the full `supabase/migrations/*.sql` chain plus `.github/ci/test-db-bootstrap.sql` (the same recipe `ci.yml` uses), and ran every `HIVE_TEST_DB_URL`-gated live test in `internal/accounting` against it for real: `TestReservationRowAndHoldAreWrittenAtomically_Live`, `TestExpandReservationRaisesTheRowAndTheHoldTogether_Live`, `TestSettlementLedgerMagnitude_Live`, plus the two new tests below. All pass against genuine Postgres, not mocks.

## What's new

**`TestReaperRunOnce_Live`** (`apps/control-plane/internal/accounting/reaper_live_test.go`). The existing reaper suite only ever drives `RunOnce` against a hand-written in-memory ledger and lister stub, so `ListStaleReservations`' actual SQL, including the batch-exclusion subquery that joins `credit_reservations.id` (uuid) against `batches.reservation_id` (text) via an explicit `::text` cast, had never run against real Postgres. This test backdates a real reservation past the TTL, seeds real `batches` rows in both an in-flight and a terminal status, and asserts the reaper releases exactly the eligible rows and leaves the rest alone, checked against `GetBalance`'s own arithmetic. Mutation-checked: replacing the cast comparison with a value that can never match turned the test red (confirmed locally, reverted before commit).

**`TestDispatchReleasesReservationEvenWhenTheResponseBodyPanics`** (`apps/edge-api/internal/chat/dispatch_panic_test.go`). Proves the session chat path's deferred reservation release survives a literal panic mid-stream (a panicking `io.Reader` standing in for a bug in response parsing), not just an error return. Go's `defer` semantics already guarantee this, but nothing in the existing suite exercised it. Mutation-checked: disabling the `defer` turned the test red (confirmed locally, reverted before commit).

## Test plan

- [x] `go vet ./apps/edge-api/... ./apps/control-plane/...`
- [x] `go test -count=1 -short ./apps/edge-api/... ./apps/control-plane/...` (all packages touched by this audit green; two unrelated pre-existing environment gaps noted above, out of scope)
- [x] Both new tests pass against a real, migrated Postgres instance
- [x] Both new tests mutation-checked (proven to fail when the invariant they guard is broken)
- [x] Live read of the real demo ledger: zero stranded holds, zero over-released reservations

## Buglog entry

No bug found in production code; this PR is regression-test coverage only. No buglog entry.

### DIFF
--- a/apps/control-plane/internal/accounting/reaper_live_test.go
+++ b/apps/control-plane/internal/accounting/reaper_live_test.go
@@ -1,0 +1,174 @@
+package accounting
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/ledger"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/usage"
+)
+
+// The reaper's own regression suite (reaper_test.go) proves RunOnce's control
+// flow against a hand-written reaperLedger and an in-memory staleReservationLister
+// stub: neither ever executes ListStaleReservations' real SQL. That query is the
+// entire reason the reaper exists (issue #616): a hold the money path failed to
+// release is only ever found by it, and it is the one piece of this package
+// that has never run against a real Postgres before this file. A wrong column,
+// an inverted cutoff comparison, or a type mismatch between
+// credit_reservations.id and batches.reservation_id (uuid vs text, cast with
+// ::text in the query) would pass every mocked test in this package and still
+// leave every stranded hold on the demo box un-found.
+//
+// Gated on HIVE_TEST_DB_URL, same convention as the package's other _live tests.
+
+func TestReaperRunOnce_Live(t *testing.T) {
+	pool := newAccountingTestPool(t)
+	ctx := context.Background()
+
+	accountID := seedReleaseIdempotencyAccount(t, pool)
+	ledgerSvc := ledger.NewService(ledger.NewPgxRepository(pool))
+	usageSvc := usage.NewService(usage.NewPgxRepository(pool))
+	repo := NewPgxRepository(pool)
+	svc := NewService(repo, ledgerSvc, usageSvc)
+
+	if _, err := ledgerSvc.GrantCredits(ctx, accountID, uuid.NewString(), 100000, map[string]any{"reason": "reaper live test grant"}); err != nil {
+		t.Fatalf("grant credits: %v", err)
+	}
+
+	// Genuinely stale: a hold nothing ever settled, backdated past the TTL.
+	stale := createTestReservation(t, ctx, svc, accountID, "/v1/chat/completions", "hive-fast", 4000)
+	backdateReservation(t, ctx, pool, stale.ID, 2*time.Hour)
+
+	// Positive control: a reservation created moments ago must not be treated
+	// as abandoned. Without this, a reaper that released EVERY active/expanded
+	// row regardless of age would still pass the assertion on `stale` alone.
+	fresh := createTestReservation(t, ctx, svc, accountID, "/v1/chat/completions", "hive-fast", 1500)
+
+	// Stale AND attached to a batch still in flight: ListStaleReservations must
+	// exclude it structurally (a batch legitimately holds credits for its whole
+	// 24h completion window), never by the age check alone.
+	inFlightBatch := createTestReservation(t, ctx, svc, accountID, "/v1/batches", "hive-fast", 9000)
+	backdateReservation(t, ctx, pool, inFlightBatch.ID, 2*time.Hour)
+	seedBatchRow(t, ctx, pool, accountID, inFlightBatch.ID, "in_progress")
+
+	// Stale AND attached to a batch that has already reached a terminal status:
+	// the batch link must not exclude it forever, only while the batch is
+	// genuinely running. This is the case that would silently strand a hold
+	// permanently if the NOT IN status list in the query were wrong (e.g. it
+	// listed no terminal statuses at all, or listed the wrong ones).
+	doneBatch := createTestReservation(t, ctx, svc, accountID, "/v1/batches", "hive-fast", 2500)
+	backdateReservation(t, ctx, pool, doneBatch.ID, 2*time.Hour)
+	seedBatchRow(t, ctx, pool, accountID, doneBatch.ID, "completed")
+
+	reaper := NewReaper(repo, svc, ReaperConfig{TTL: time.Hour})
+	result, err := reaper.RunOnce(ctx)
+	if err != nil {
+		t.Fatalf("RunOnce: %v", err)
+	}
+
+	released := map[uuid.UUID]bool{}
+	for _, r := range []Reservation{stale, inFlightBatch, doneBatch, fresh} {
+		row, err := repo.GetReservation(ctx, accountID, r.ID)
+		if err != nil {
+			t.Fatalf("reload reservation %s: %v", r.ID, err)
+		}
+		released[r.ID] = row.Status == ReservationStatusReleased
+	}
+
+	if !released[stale.ID] {
+		t.Fatalf("expected the genuinely stale reservation to be released by RunOnce, status stayed unreleased")
+	}
+	if released[fresh.ID] {
+		t.Fatalf("RunOnce released a reservation created moments ago; ListStaleReservations must only match rows older than the TTL cutoff")
+	}
+	if released[inFlightBatch.ID] {
+		t.Fatalf("RunOnce released a hold whose batch is still in_progress; the batch-exclusion subquery must keep this reservation out of the candidate scan")
+	}
+	if !released[doneBatch.ID] {
+		t.Fatalf("RunOnce left a stale hold alone because its batch was 'completed'; a terminal batch status must not exclude the reservation from reaping")
+	}
+
+	if result.Scanned < 2 {
+		t.Fatalf("expected RunOnce to have scanned at least the two eligible stale reservations (stale, doneBatch), got Scanned=%d", result.Scanned)
+	}
+	if result.Released < 2 {
+		t.Fatalf("expected RunOnce to report at least 2 released, got %d", result.Released)
+	}
+
+	// The released hold's credits must actually be back on the books, read
+	// straight off GetBalance rather than trusting the reservation row alone.
+	balance, err := ledgerSvc.GetBalance(ctx, accountID)
+	if err != nil {
+		t.Fatalf("GetBalance: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


sakibsadmanshajib:
## Adversarial review (dynamic selector)

Diff is two new Go test files, no production code changed. Streams run:

**CodeRabbit CLI: SKIPPED.** Rate-limited (3 included reviews used up on this account, 31 min reset window). Not treated as a clean pass.

**`ecc:code-review` / `adversarial-pr-review` skill streams: could not be invoked.** This agent's toolset for this task has no Skill-invocation tool (Read/Write/Edit/Bash only), which is the structural capability gap the Gate Compliance rule names as the one legitimate reason to substitute rather than route around a stream. Substituted with a manual line-by-line pass below covering the same three angles (language/quality, database correctness, security), since this diff touches the money path and both the security and database reviews are mandatory here regardless.

### Go quality
- `gofmt -l` clean on both new files.
- `go vet ./apps/edge-api/... ./apps/control-plane/...` clean.
- Both new tests reuse existing package fixtures (`newAccountingTestPool`, `seedReleaseIdempotencyAccount`, `fakeAccounting`, `pricedRouting`, `stubBilling`) rather than duplicating harness code.

### Database correctness (`reaper_live_test.go`)
- `seedBatchRow` inserts real `public.files` + `public.batches` rows matching the actual column types (`batches.reservation_id` is `text`, cast from `reservationID.String()`, matching the production query's `cr.id::text` cast) rather than asserting against a mock.
- Cleanup order is child-then-parent (`batches` before `files`), correct against the `input_file_id` FK.
- `ListStaleReservations` has no `account_id` filter (scans the whole table), so the test asserts `Scanned >= 2` / `Released >= 2` rather than exact counts, and confirms the real invariant on `GetBalance` (scoped to the test account) rather than trusting the reaper's own counters.
- Minor, non-blocking, pre-existing (not introduced here): `TestReservationRowAndHoldAreWrittenAtomically_Live`'s control reservation and `TestExpandReservationRaisesTheRowAndTheHoldTogether_Live`'s reservation are never released or finalized, so they persist as permanent `active` rows in whatever ephemeral test DB runs the suite. Harmless within one `go test` invocation (nothing crosses the reaper's 1h TTL inside a single run) and CI provisions a fresh ephemeral DB per job, so this doesn't accumulate across runs either. Flagging for visibility, not fixing here (out of this PR's scope, and touching those tests wasn't requested).

### Security
- No secrets, tokens, or credentials appear anywhere in either file.
- `dispatch_panic_test.go`'s injected panic is fully contained inside the test's own `defer recover()`; it cannot escape the test process.
- Neither test mutates or reads the live/demo database; both run only against the local ephemeral Postgres instance the audit stood up for this PR.

### Mutation checks (both confirmed locally, not committed)
- `reaper_live_test.go`: replacing the batch-exclusion cast comparison (`b.reservation_id = cr.id::text`) with a value that can never match turned `TestReaperRunOnce_Live` red.
- `dispatch_panic_test.go`: disabling the `defer` that releases the reservation turned `TestDispatchReleasesReservationEvenWhenTheResponseBodyPanics` red.

**Verdict: no blocking findings.** Ready for orchestrator verification and merge.

sakibsadmanshajib:
## Follow-up: needs_reconciliation and live pricing (per coordinator request)

### 1. The 149 `needs_reconciliation` rows

Queried live, read-only:

- **count=149**, sum(consumed_credits)=**70,217**, sum(reserved_credits)=1,490,000 (149 × the flat 10,000-credit session-chat hold — these are all session-chat requests).
- **oldest=2026-07-17, newest=2026-08-17 (today)**. Still actively happening: **14 in the last 24h, 66 in the last 7 days**. Not historical debris, an ongoing rate.
- **Zero of the 149 rows have consumed_credits = 0.** This is the actual free-serve check, not an assumption: every single row was charged a nonzero amount. Sampled the 20 oldest by hand too: an early batch (7/17-7/19) charged the full 10,000-credit hold (content estimate exceeded the hold, clamped per `finalizeLocked`'s unconfirmed-clamp rule), a later batch (8/01, same-day cluster, likely that day's incident-verification traffic rather than steady organic load) charged 5 credits each with 9,995 released back.

**What puts a row there:** `finalizeLocked` in `apps/control-plane/internal/accounting/service.go` sets status to `needs_reconciliation` (instead of `finalized`) whenever `TerminalUsageConfirmed=false` — i.e. the upstream provider never sent a real usage frame, so the charge is a content-length estimate rather than a measured token count. The charge still lands (via `ChargeUsage`) either way; the status is a confidence label on the number, not a "did we bill" flag.

**What's supposed to drain it, and does anything:** `finalizeLocked` also calls `CreateReconciliationJob`, writing a row to `public.credit_reconciliation_jobs`, meant to later true the estimate up to a real measured figure. Grepped the entire `apps/` and `supabase/` tree: the only reference to that table outside its own migration is the single `INSERT` in `repository.go`. **Nothing reads it, nothing ever has.** 148 rows sitting there since 2026-07-17, growing, with a `(status, run_after)` index built for a worker that was never written (issue #600, already known and named in the code's own comments — confirmed here, not merely assumed). So: real money charged, correctly, at an estimate; that estimate is never corrected to the true figure. Not "served free," but a real, ongoing, unaddressed reconciliation debt worth its own tracked fix (separate from this PR, which was a test-only change).

### 2. Live catalog pricing, every alias currently routable

| alias | visibility | input price/M | output price/M | unit | verdict |
|---|---|---|---|---|---|
| hive-auto | preview | 56,000 | 224,000 | tokens | priced |
| hive-default | public | 21,000 | 84,000 | tokens | priced |
| hive-fast | public | 7,000 | 11,200 | tokens | priced |
| hive-embedding-default | public | 1 | 0 | tokens | input-only by design (embeddings) |
| hive-stt | public | 0 | 4,316,667 | seconds | output-only by design (audio duration) |
| hive-tts | public | 0 | 3,080,000 | characters | output-only by design (character count) |

**No alias has both sides zero. No alias with a live (non-disabled) route is missing a `model_aliases` row at all** (joined `provider_routes` against `model_aliases`, checked explicitly for orphans). The single-sided zeros are the legitimate embeddings/voice shape, not a pricing gap: `routing.Service.SelectRoute` refuses (`ErrAliasNotPriced`) any alias where both `input_price_credits` and `output_price_credits` are ≤ 0, before a route is ever handed back — confirmed this is a hard `return SelectionResult{}, err` before dispatch, not a soft warning. Text metering (`inference.CanPriceTokens`/`requireTokenPricing`) enforces the identical rule a second time at the endpoint boundary. **Fails closed, not serves free — and today, closed is unreachable because no live alias trips it.**

Bonus finding while checking this: `apps/edge-api/internal/audio/pricing.go`'s `creditsForQuantity` derives from `route.UnitPriceCredits` (the real catalog price) and `canPrice` fails closed on `UnitPriceCredits <= 0` — this is the CURRENT code, and it contradicts `CLAUDE.md`'s "Known Issues" #3/#627 text describing audio as charging hardcoded flat literals disconnected from the catalog. That description appears to be stale (the fix seems to have landed after the doc was last touched); flagging for whoever owns that doc, not fixing here, out of this PR's scope.

### Bottom line

Neither concern turns up a hole in the money path. Both are confirmed clean by direct query, not inference from comments. `needs_reconciliation` is real but boring: a permanent-estimate state, always nonzero, not a free-serve; the drain that never got built is the actual gap, tracked separately (following up as its own issue). Pricing has no zero/missing rate live today; the fail-closed guard exists and would catch one if it ever appeared.

Filed the CI-bootstrap environment gap noted in the original PR body as its own issue rather than leaving it here: #958.

Issue #953, opened 2026-08-17
Add regression coverage for BYPASS_ADMIN_ACCESS_CONTROL / ENABLE_ADMIN_CHAT_ACCESS staying off


sakibsadmanshajib:
Follow-up to #947. The fix that sets `BYPASS_ADMIN_ACCESS_CONTROL: "false"` and `ENABLE_ADMIN_CHAT_ACCESS: "false"` in `deploy/docker/docker-compose.yml`'s `open-webui` service was verified manually against a running stack (two demo-owned accounts, screenshot in the fix PR), not with an automated test, per the owner's standing instruction that day (demo first, tests later).

Needed: an automated regression guard so a future edit cannot silently drop or flip these two lines back to the upstream default without a build failure. Two options, either is acceptable:

1. A live-stack integration test (in the `apps/web-console/tests/e2e` suite or a small Go/Python script under `scripts/`) that signs in as a second-tenant Open WebUI admin account and asserts `GET /api/v1/knowledge/` does not return a first tenant's collection, and that `GET /api/v1/chats/list/user/{id}` for another user's id returns 401.
2. A cheap static check (grep or a small parser) run in CI that fails if `deploy/docker/docker-compose.yml`'s `open-webui` service does not explicitly set both variables to `"false"`, so at minimum an accidental deletion of the lines is caught even without a live stack.

Either needs two demo-owned test accounts with distinct tenant OWNER memberships; do not use the owner's personal account for this. See `docs/live-test-auth.md` for how to mint a session without touching a shared account's password.

Related: #947, #948, #949, #950.

sakibsadmanshajib:
Open PR #960 (fix: stop Open WebUI admin sessions reading across tenants) appears to address this. Not yet merged.

Issue #950, opened 2026-08-17
Every OIDC-provisioned chat account has its display name set to its own email address


sakibsadmanshajib:
Every account provisioned through the Hive OIDC login has its Open WebUI display name set to its own email address, so the chat interface prints full email addresses wherever a name belongs, including to other users.

## Evidence

Live on the demo box, 2026-08-17: 5 of the 6 accounts have `name` set literally equal to their email address.

## Cause

Supabase's OAuth authorization server issues a minimal third-party OIDC id_token for external relying parties, standard claims only (this is the same constraint recorded at length in `deploy/docker/owui-patches/tenant_role_from_db.py` and in #457). There is no `name` claim in it, and `OAUTH_USERNAME_CLAIM` defaults to `name`, so Open WebUI takes its documented fallback:

```python
# vendor/open-webui/backend/open_webui/utils/oauth.py:1900-1905
name = user_data.get(username_claim)
if not name:
    log.warning('Username claim is missing, using email as name')
    name = email
```

`OAUTH_UPDATE_NAME_ON_LOGIN` then leaves it alone on subsequent logins, because the claim is still absent (`utils/oauth.py:1846-1851`).

## Why it matters beyond looks

* The Users page, any shared conversation, and any future multi-user surface print a full email address as a display name, to people in other tenants on this shared instance.
* #940 already records that the chat home greets the user with an internal routing alias and that there is no chosen name to greet with. This is the other half of that: there is no name because nothing ever supplies one.

## Remediation

Three candidates, in increasing order of cost:

1. Set `OAUTH_USERNAME_CLAIM` to a claim Supabase does send, if one carries something better than the email local part. Needs a live check of the actual token contents; the `preferred_username` and `name` claims are both believed absent, so this may resolve to nothing.
2. Extend the existing tenant lookup patch. It already connects to our Postgres by email on every login to resolve the role, so it can read a display name from the same query and set it. One extra column on the select in `tenant_role_from_db.py`, guarded the same way, and it fixes the greeting in #940 too.
3. Ask for a display name during signup in the console and store it where option 2 reads it. Correct, and the largest.

Option 2 is the shortest path that produces a real name rather than a truncated email.

sakibsadmanshajib:
Open PR #960 (fix: stop Open WebUI admin sessions reading across tenants) appears to address this. Not yet merged.

Issue #949, opened 2026-08-17
Open WebUI Admin Settings is reachable from the Settings dialog and most of its writes land, contradicting D-014


sakibsadmanshajib:
D-014 (2026-07-27) records that the web console and control plane are the sole source of truth for admin settings and that "OWUI's own admin panel stays disabled (env flags off), Caddy blocks mutation verbs". The owner opened Settings then Admin Settings on the demo box on 2026-08-17 and found a full panel of options. This issue is the gap between the recorded decision and the deployed reality, kept separate from the mechanical matcher defect in #437 and the link defect in #846.

## One: the panel is reachable, by design of our own front end

`vendor/open-webui/src/lib/components/chat/SettingsModal.svelte:843-861` still renders the Admin Settings link for `$user?.role === 'admin'` and navigates with `goto('/admin/settings')`. That is client-side routing, so the blocked address is never requested and the panel renders from the already-loaded bundle. Refreshing on it gives Caddy's 404, which is why it reads as half-broken.

The removal for that link exists only in `deploy/docker/owui-patches/hive_ui_surfaces.py`, which `Dockerfile.open-webui` runs against the upstream prebuilt bundle and then discards when it swaps in our own build from `vendor/open-webui`. The Dockerfile is explicit that every such removal has to be repeated in the vendored source; this one was not. The user-menu twin (#909) was.

## Two: most of the panel's writes land

Measured on 2026-08-17 by running the exact `@blocked`, `@removedSurfaces` and `@adminMutation` matchers from `deploy/docker/Caddyfile.owui` against a throwaway `caddy:2-alpine` container with a 200 responder behind them.

Blocked, working as intended:

| Address | Result |
| --- | --- |
| `/admin/*` | 404 |
| `/api/v1/auths/admin/config`, `/api/v1/auths/admin/details` | 404 |
| `/api/v1/configs/export` | 404 |
| `/openai/config/update` | 404 |
| `/api/v1/terminals/*`, `/api/v1/configs/tool_servers`, `/api/v1/configs/terminal_servers` | 404 |

Not blocked, all reaching Open WebUI, all `get_admin_user` and nothing more:

| Address | What it changes |
| --- | --- |
| `POST /api/v1/retrieval/config/update` | embedding model, dimension, chunking, and the base URL plus key every uploaded document is sent to |
| `POST /api/v1/audio/config/update` | speech to text and text to speech engines, endpoints and keys |
[...]

PR #960, opened 2026-08-17
fix: stop Open WebUI admin sessions reading across tenants

### LINKED ISSUE #947: Shared Open WebUI instance exposes every tenant's documents and every user's conversations to any administrator, by two upstream defaults we never set
Two Open WebUI defaults we never set turn every administrator of the shared chat instance into a reader of every tenant's data. Since the tenant-role patch promotes every sole tenant OWNER to administrator (#748), this is cross-tenant exposure that is live today, not a hypothetical.

## Evidence, live, 2026-08-17

Signed in as the owner on chat-hive.scubed.co:

* `GET /api/v1/users/` returns 6 accounts, 5 of them role `admin`, 1 role `user`.
* `/workspace/knowledge` shows a collection owned by the demo account while signed in as a different, personal account belonging to a different tenant.

## Cause one: documents

`BYPASS_ADMIN_ACCESS_CONTROL` defaults to true (`vendor/open-webui/backend/open_webui/config.py:2029-2035`, which falls back to `ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS`, also defaulting to `True`). `deploy/docker/docker-compose.yml` never sets either.

With it true, the knowledge listing skips its ownership filter entirely for an administrator:

```python
# vendor/open-webui/backend/open_webui/routers/knowledge.py:143-148
if not user.role == 'admin' or not BYPASS_ADMIN_ACCESS_CONTROL:
    if groups:
        filter['group_ids'] = [group.id for group in groups]
    filter['user_id'] = user.id
```

An administrator therefore receives every knowledge base on the instance from `GET /api/v1/knowledge/` and `GET /api/v1/knowledge/search`, with `write_access` true on all of them. The same bypass applies to models, prompts, tools and files. This is one shared instance for every Hive tenant (see the header comment in `deploy/docker/owui-patches/tenant_role_from_db.py`), so the listing crosses tenant boundaries by construction.

## Cause two: conversations

`ENABLE_ADMIN_CHAT_ACCESS` defaults to true (`config.py:2037`) and is likewise never set. It enables:

* `GET /api/v1/chats/list/user/{user_id}` (`routers/chats.py:600-612`), a per-user chat listing gated on `get_admin_user` plus this flag
* the administrator fallback in `GET /api/v1/chats/share/{share_id}` (`routers/chats.py:1056-1058`), which resolves any chat by its own id when the share lookup misses, and skips the access-grant check entirely at `routers/chats.py:1063`

Neither address is covered by `deploy/docker/Caddyfile.owui`. Confirmed by running the real `@blocked` and `@removedSurfaces` matchers from that file against a throwaway Caddy container: `/api/v1/chats/all` and `/api/v1/chats/list/user/x` both pass through.

So five of the six accounts on the demo box can read the sixth's conversations, and any future sole-OWNER account can read all of them.

## Why this is not the same issue as #748

#748 is about how the administrator bit is granted. This is about what that bit is worth. Even if #748 were fixed so that only one person held it, that person would still hold cross-tenant read of documents and conversations on a shared instance, which contradicts the single org equals single tenant posture (D-007) and the data-sovereign positioning of Hive Enterprise. Both need fixing, in either order.

## Remediation

Minimum, and it is two environment lines in the `open-webui` service, neither of which is persisted config, so they take effect on the next recreate with no reconcile entry needed:

```yaml
BYPASS_ADMIN_ACCESS_CONTROL: "false"
ENABLE_ADMIN_CHAT_ACCESS: "false"
```

Verify before merging that `BYPASS_ADMIN_ACCESS_CONTROL: "false"` does not regress the model picker. `deploy/docker/Dockerfile.open-webui` records a measurement that with `BYPASS_MODEL_ACCESS_CONTROL` off, an administrator still saw all six aliases because of this admin bypass while a member saw an empty picker. `BYPASS_MODEL_ACCESS_CONTROL: "true"` stays set independently (`docker-compose.yml:762`) and is what actually governs the picker, so the expectation is no change there, but it must be checked on a booted container rather than assumed.

Real fix, larger: the shared-instance model itself. Either scope Open WebUI content by tenant, or stop deriving an instance-wide administrator from a tenant-scoped role. That decision belongs with the fork work under D-036 and D-040.

## Visual proof

Not attached: this is a read-only audit and reproducing it requires signing in as two accounts in two tenants on the live box. The reproduction is one signed-in `GET /api/v1/knowledge/` as any administrator, compared with the same call as the `user` account.

sakibsadmanshajib:
## Summary

Fixes #947. `BYPASS_ADMIN_ACCESS_CONTROL` and `ENABLE_ADMIN_CHAT_ACCESS` both default to `true` upstream (`vendor/open-webui/backend/open_webui/config.py:2029`, `:2037`) and were never set in `deploy/docker/docker-compose.yml`. Because every sole tenant OWNER on this deployment is promoted to Open WebUI `admin` (56 live accounts hold that shape), any admin session could list every other tenant's Knowledge collections (`routers/knowledge.py:143`) and read any user's chats (`routers/chats.py:600`, `:1056`). The owner observed this live: a demo account's Knowledge collection visible from an unrelated personal account.

This sets both to `"false"` on the `open-webui` service, next to the other hardcoded security defaults already there (`ENABLE_ADMIN_EXPORT`, `ENABLE_COMMUNITY_SHARING`). One service block covers every compose profile that starts Open WebUI (`local`, `enterprise`, `chat`). No `.env.example` entry: these are security defaults, not deployment-tunable knobs, matching the sibling flags in the same block.

Both flags are plain `os.getenv(...)` reads at Python import time, not `PersistentConfig`, so unlike the RAG/audio/login-form keys `owui-patches/hive_rag_env_config.py` reconciles from the environment on every boot, these two have no database-persisted override to fight: the environment value is authoritative on the next container start.

## Explicitly out of scope (per brief)

- Who holds the `admin` role (`tenant_role_from_db.py`, issue #748). Product decision, not this fix.
- The Caddy `@adminMutation` matcher gap (issue #949).
- The `/api/v1/configs/namespace/oauth` secret-exposure gap (issue #950).
- Demoting or deleting any account.
- An automated test suite. Owner's standing instruction: demo first, tests later. Follow-up regression-test issue: #953.

## Blast-radius check (what else this flag gates, and why the demo is unaffected)

`BYPASS_ADMIN_ACCESS_CONTROL` is also read in `routers/models.py`, `files.py`, `prompts.py`, `tools.py`, `skills.py`, `notes.py` (the same admin-sees-everyone bypass, repeated). Two things matter for the demo:

- **Knowledge** (`DEMO.md`: create own collection, upload own document, attach to own chat). Turning the bypass off only removes the branch that skips the `user_id`/group filter for admins; an admin's own collections stay visible (the filter becomes `user_id == own`, not deny-all). The demo only ever touches the signed-in account's own collection, so it is unaffected, and now correctly scoped besides.
- **Model list** (chat picker, Workspace > Models). Gated by a *different* flag, `BYPASS_MODEL_ACCESS_CONTROL`, left untouched here. In the pinned v0.10.2 image that flag alone skips `get_filtered_models` for every role, independent of `BYPASS_ADMIN_ACCESS_CONTROL`, and the picker's own unconditional Hive filter (`owui-patches/hive_model_picker.py`) runs regardless of both flags. Model listing is identical before and after this change.

`ENABLE_ADMIN_CHAT_ACCESS` only gates the admin-only `/list/user/{user_id}` endpoint (Users page's per-user chat listing) and the share-chat by-ID fallback. Neither is on the path a user takes to open their own chats.

Files/prompts/tools/skills/notes admin-sees-everyone visibility also narrows to own+group under this same flag. Nothing in `DEMO.md` exercises an admin viewing another account's prompts, tools, skills or notes, so this is accepted collateral narrowing consistent with the fix's intent.

## Verification

Manual, against an isolated local compose stack built from this branch (fresh Open WebUI volume, `local` + `chat` profiles). Two Open WebUI accounts created directly via the app's own internal APIs (no password set/read/rotated anywhere, neither the owner's account nor the shared demo account), both independently tenant-OWNER (so both admin), one owning a Knowledge collection the other has no relationship to.

**Control, addressing review feedback that the original screenshots could not distinguish "the fix works" from "the test account was never admin":** same two accounts, same collection, captured twice against the same container/volume, flipping only the two flags between captures. Both accounts' `role=admin` confirmed via `GET /api/v1/auths/` at the moment of each capture. Proof posted as a release-asset comment (not `docs/proof/`, which 404s once this branch is deleted at merge): https://github.com/sakibsadmanshajib/hive/pull/960#issuecomment-5325135646

- **Before** (flags forced back to upstream's default `true`): fixture B, an admin in a different tenant, sees fixture A's collection.
- **After** (this branch's actual `docker-compose.yml`, both `false`): same fixture B sees "No knowledge found".

## Buglog entry

For the follow-up buglog-only PR (per `.claude/rules/openwolf.md`, never appended directly on this branch):

```json
{"error_message":"Open WebUI admin sessions could read every other tenant's Knowledge collections and any user's chats","root_cause":"BYPASS_ADMIN_ACCESS_CONTROL and ENABLE_ADMIN_CHAT_ACCESS default to true upstream and were never set in deploy/docker/docker-compose.yml, and every sole tenant OWNER on this deployment is promoted to Open WebUI admin","fix":"set both to \"false\" on the open-webui service in deploy/docker/docker-compose.yml, covering local/enterprise/chat profiles; no DB reconciliation needed since both are plain os.getenv reads at import time","tags":["security","open-webui","rbac","cross-tenant"]}
```

A second entry, for the branch contamination described at the bottom of this description:

```json
{"error_message": "PR #960 carried five vendor/open-webui/src/lib/hive files and a hive.css hunk from the rejected #951 agent-surface design, under a PR titled for a tenant-isolation fix", "root_cause": "git checkout <ref> -- <path> stages what it checks out, so a later path-limited git commit -- docs/ committed the whole index including those staged vendor files rather than only the named paths", "fix": "rebuilt the branch from origin/main and checked out only the eight wanted paths instead of cherry-picking; verified the three source files byte-identical between merge base and main first, then confirmed with git status --short after the checkout and gh pr diff --name-only after the push", "tags": ["git", "process", "contamination", "open-webui"]}
```

## Follow-up

- #953: automated regression coverage for these two flags.

## Test plan

- [x] Manual verification against a running stack, before/after control, both accounts' admin role confirmed via API (see Verification section)
- [x] Confirmed via source read that the demo's Knowledge and model-list surfaces are unaffected
- [x] Confirmed `/v1/rag/chat` enforces tenant scoping independently (Postgres RLS + explicit `tenant_id` filter in `apps/edge-api/internal/rag/repository.go`), not through the Open WebUI layer this PR changes; not a second door to the same data
- [ ] CI (Go tests, web-console build). This PR touches only `deploy/docker/docker-compose.yml`, `deploy/docker/Caddyfile.owui`, `scripts/test_caddy_owui_blocklist.py` and `docs/`, no application code

This PR closes Open WebUI’s administrator cross-tenant read paths by explicitly disabling the upstream admin access-control bypass and cross-user chat access defaults.

- Applies both security defaults to the shared Open WebUI service used by the local, enterprise, and chat profiles.
- Documents controlled before-and-after verification with two independently confirmed administrator accounts from different tenants.
- Keeps image evidence in permanent release assets while committing the sanitized textual proof record.
</details>

The PR appears safe to merge and correctly narrows Open WebUI administrator reads to authorized user and group scope.

The shared Compose service supplies both flags to every profile that starts Open WebUI, the pinned implementation parses the quoted false values correctly at startup, and the remaining router branches preserve access to owned or explicitly granted resources.
</details>

| Filename | Overview |
|----------|----------|
| deploy/docker/docker-compose.yml | Adds correctly parsed, environment-authoritative security defaults to the shared Open WebUI service, disabling administrator-wide content and chat access across all applicable profiles. |
| docs/proof/owui-admin-bypass-fix-2026-08-17/README.md | Records a controlled before-and-after tenant-isolation demonstration while following the repository’s text-log, release-asset, and credential-handling requirements. |

</details>

```mermaid
flowchart LR
  A[Open WebUI admin session] --> B{Access-control flags}
  B -->|Previous upstream defaults: true| C[Admin-wide Knowledge and chat reads]
  B -->|Compose defaults: false| D[User and group scoped access]
  D --> E[Own or explicitly granted resources]
  D -. blocks .-> F[Unrelated tenant resources]
```
</details>

<sub>Reviews (1): Last reviewed commit: ["docs: commit the proof text log for hive..."](https://github.com/sakibsadmanshajib/hive/commit/74811ed5536f57dbada1618e1c49eac8c3a61322) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54104603)</sub>

---

## Rebase onto current main, 2026-08-22

Rebased onto `main` at `c30882491` and force pushed, so CI has now run against the current base rather than reporting a stale green. Branch protection on this repository has `strict=false`, so GitHub does not require an up-to-date branch and the previous CLEAN was measured against `1dc67ee69`, before the database moved off hosted Supabase and before `docker-compose.yml` was substantially rewritten around the `selfhost` profile and the layered override and enterprise files. No conflicts: the two lines land in the `open-webui` service's environment block, which none of that work touched.

### Still needed, verified rather than assumed

- `main` sets neither variable: `git show origin/main:deploy/docker/docker-compose.yml | grep -E 'BYPASS_ADMIN|ENABLE_ADMIN_CHAT'` returns only the one pre-existing comment that mentions the default. So no later fix has superseded this.
- Both are still plain `os.getenv` reads at import time in the vendored `config.py`, not `PersistentConfig`, so no reconciler entry is needed and a container recreate is sufficient. `BYPASS_ADMIN_ACCESS_CONTROL` still falls back to `ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS`, also defaulting true, which is why the line sets it directly.
- Nothing about this depends on where Supabase runs. These flags gate Open WebUI's own authorization checks over its own tables, inside the pinned upstream backend image, and the migration changed neither.

### Nothing here can delete the data plane

Recording this because compose is the one file where a rebase mistake is destructive. This branch adds two environment lines inside an existing service and changes no service list, no profile, no volume and no file layering. It does not touch `HIVE_COMPOSE_FLAGS`, does not remove `docker-compose.enterprise.yml` from the invocation, and does not alter the `selfhost` profile, so the `--remove-orphans` in the deploy recreate step has nothing new to orphan.

### One stale comment left in place, deliberately

An unrelated comment further down the same service still reads that the per-model access control fix "could never fire" because `BYPASS_ADMIN_ACCESS_CONTROL` defaults true. After this change that premise no longer holds. It is left alone rather than rewritten here: it is the historical record of why #776 was abandoned, and rewriting history in a comment to match a later change is how the reasoning behind an abandoned approach gets lost. Worth a separate docs pass over that block.

### Visual proof

The existing capture stands on its own terms and is not re-run. Its record is in `docs/proof/owui-admin-bypass-fix-2026-08-17/README.md`: two Open WebUI accounts, both confirmed `admin` via `GET /api/v1/auths/` at the moment of each capture rather than asserted in prose, one Knowledge collection, captured twice against the same container and volume with only these two flags flipped between captures, going from `bSeesFixtureACollection: true` to `false`.

The reason it is not re-run is worth stating rather than leaving as an omission. That capture ran in an isolated local compose project against a fresh Open WebUI volume, using the container's own model layer and `create_token`; it never touched hosted Supabase, so the migration did not invalidate it, and the flags it exercises are read by the pinned upstream backend image which has not changed. Re-running it would produce the same two screenshots from the same inputs. If a fresh pair is wanted before merge regardless, say so and it can be re-captured.

No credential appears in that capture: the session tokens were injected via `localStorage`, never a URL query string or fragment, so there was nothing to redact in either the text or the pixels.

## Summary by CodeRabbit

- **Security**
  - Restricted administrators from accessing other tenants’ Knowledge collections and chats by default.
  - Preserved administrators’ access to content within their own tenant.
  - Changes take effect after the service restarts.

- **Documentation**
  - Added verification records covering the updated cross-tenant access behavior, test setup, and scope limitations.

---

## Branch rebuilt from `main`, 2026-08-23

Force pushed a clean rebuild. The previous head `b6c2850e5` also carried five files under `vendor/open-webui/src/lib/hive/` (`AgentTasks.svelte`, `ComposerSendButton.svelte`, `ComposerShell.svelte`, `agentTasks.ts`, `agentTasks.test.ts`) plus a `hive.css` hunk that existed only to style them. Those belong to the separate agent-surface change in #951, whose design was rejected, and none of them exists on `main`. Merging as-is would have shipped a rejected design under a PR titled for a tenant-isolation fix.

How they got there, since the mechanism is not obvious: `git checkout <ref> -- <path>` stages what it checks out, and a later path-limited `git commit -- docs/` still commits the whole index rather than only the named paths. The staged vendor files rode along silently. The lesson is to run `git status --short` after any path-limited checkout, not before.

The rebuild branched from `origin/main` and checked out only the eight wanted paths, rather than cherry-picking, which would have dragged the same contamination forward. `deploy/docker/Caddyfile.owui`, `deploy/docker/docker-compose.yml` and `scripts/test_caddy_owui_blocklist.py` were first verified byte-identical between the branch's merge base and current `origin/main`, so taking them from the old branch head reproduces exactly the intended edit and reverts no later work on those files.

The diff is now eight files, 326 insertions and 1 deletion, with zero `vendor/open-webui` paths. `python3 scripts/test_caddy_owui_blocklist.py` reports 43 blocked, 40 open.

Nothing load-bearing was lost. The tenant-isolation fix is the two compose flags plus the Caddy block, and none of the five removed files is referenced by either. The spec's section 8 records that `/v1/agent/*` is untouched, so the agent plumbing and the security fix were never coupled.

### The proof predates the strip and still holds

The before-and-after capture was taken against the pre-strip branch and is deliberately not re-run. Everything it exercises survived the strip unchanged: the two compose flags, the Caddy block and both fixture tenants are all still present and byte-identical. The strip removed only frontend Svelte and TypeScript files that no part of the captured path references. The capture therefore still describes the behaviour this PR ships.

One detail from that capture is worth stating plainly, because it is the most valuable line in the whole proof. The first attempt at the after arm ran before the container was healthy and answered 401 to everything. That would have read as a clean pass on the strength of nobody being signed in: no collection visible, because no session. The script now aborts unless both roles verify as `admin` through `GET /api/v1/auths/` before either arm is captured, which is exactly what lets the recorded run distinguish "the fix works" from "the test was never authenticated".

### DIFF
--- a/deploy/docker/docker-compose.yml
+++ b/deploy/docker/docker-compose.yml
@@ -883,6 +883,46 @@ services:
       DEFAULT_USER_ROLE: "pending"
       ENABLE_ADMIN_EXPORT: "false"
       ENABLE_COMMUNITY_SHARING: "false"
+      # #947: both flags below default to true upstream (config.py:2005 and
+      # config.py:2013) and we never set them, so every Open WebUI
+      # administrator -- which is every sole tenant OWNER, per the
+      # tenant-role patch below -- could list every other tenant's Knowledge
+      # collections (routers/knowledge.py:143, :202) and read any user's chats
+      # (routers/chats.py:611, :1056). Confirmed live: a demo account's
+      # Knowledge collection was visible from an unrelated personal account.
+      # Both are plain os.getenv reads at import time, not PersistentConfig,
+      # and neither name appears in hive_rag_env_config.py's reconciler, so
+      # setting them here takes effect on the next container start and the
+      # database cannot quietly put them back.
+      #
+      # What this closes, stated precisely, because the two are not the same
+      # thing. Chats are closed: every route that can read another user's
+      # chat is gated on ENABLE_ADMIN_CHAT_ACCESS (chats.py:611, :1056,
+      # :1063, :1174, :1592), and the bulk route at :890 was already behind
+      # ENABLE_ADMIN_EXPORT above. Knowledge is closed for DISCOVERY only:
+      # the listing routes honour BYPASS_ADMIN_ACCESS_CONTROL, so an admin no
+      # longer sees another tenant's collections, but three by-id routes
+      # short-circuit on `user.role == 'admin'` without consulting either
+      # flag -- knowledge.py:1039 (GET /{id}), :1289 (GET /{id}/files, which
+      # takes include_content) and :2096 (GET /{id}/export, which checks not
+      # even ENABLE_ADMIN_EXPORT). The export route is blocked at the proxy
+      # instead, in Caddyfile.owui, since a backend edit here would be inert:
+      # Dockerfile.open-webui builds only the frontend from vendor/. The
+      # other two need a real ownership check through owui-patches and are
+      # tracked separately; they are reachable only by someone who already
+      # knows a collection id.
+      #
+      # Scope: this does not change who holds the admin role (#748) or the
+      # proxy write-verb gap (#949).
+      #
+      # BYPASS_ADMIN_ACCESS_CONTROL has an undocumented second name: unset,
+      # it falls back to ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS (also default
+      # true, config.py:2003). Irrelevant here because this line sets
+      # BYPASS_ADMIN_ACCESS_CONTROL directly and os.getenv's own default
+      # never triggers, but named so a future edit auditing this block does
+      # not have to rediscover the alias by reading the vendored source.
+      BYPASS_ADMIN_ACCESS_CONTROL: "false"
+      ENABLE_ADMIN_CHAT_ACCESS: "false"
       # #772: stock Open WebUI surfaces with no Hive counterpart. These three
       # are the ones upstream actually gates on a feature flag rather than on a
       # permission, and the flag is checked before the role, so turning them
--- a/docs/proof/owui-admin-bypass-fix-2026-08-17/README.md
+++ b/docs/proof/owui-admin-bypass-fix-2026-08-17/README.md
@@ -1,0 +1,60 @@
+# Proof: BYPASS_ADMIN_ACCESS_CONTROL / ENABLE_ADMIN_CHAT_ACCESS fix
+
+Fixes hive#947, PR #960. This directory carries the capture's text log only,
+per `.claude/rules/orchestrator.md` rule 8: the images themselves are posted
+as permanent GitHub Release assets (`scripts/post-pr-visual-proof.sh`), not
+committed here, because a `raw.githubusercontent.com` link pinned to this
+branch's name would 404 the instant the branch is deleted at squash-merge.
+`npm run lint:proof-tokens` scans this directory for credential-shaped text,
+which is the reason the text log lives here rather than only in a PR comment.
+
+## Method
+
+Ran the fixed `deploy/docker/docker-compose.yml` in an isolated local compose
+project, `local` + `chat` profiles, against a fresh Open WebUI volume.
+Created two Open WebUI accounts directly through the running container's own
+internal APIs (`open_webui.models.users.Users`,
+`open_webui.models.knowledge.Knowledges`), each minted a normal Open WebUI
+session token with the container's own `create_token`/`WEBUI_SECRET_KEY` (the
+same primitive every real login produces). No password was set, read, reset,
+or rotated on any account, and neither account is the owner's personal
+account or the shared demo account.
+
+**Control, addressing review feedback that the first round of screenshots
+could not distinguish "the fix works" from "the test account was never
+admin":** same two accounts, same Knowledge collection, captured twice
+against the same container/volume, with only the two flags flipped between
+captures, and each account's role independently confirmed via
+`GET /api/v1/auths/` at the moment of each capture (not asserted in prose).
+
+- `tenant-a-owner@hive-verify-947.invalid` (admin, tenant A): owns one
+  Knowledge collection, `Verify-947-Tenant-A-Secret-Kb`.
+- `tenant-b-owner@hive-verify-947.invalid` (admin, tenant B): no relationship
+  to fixture A or its collection.
+
+Role check JSON, captured via the same bearer tokens used for each
+screenshot:
+
+```text
+before: {"roleA":{"role":"admin","email":"tenant-a-owner@hive-verify-947.invalid"},"roleB":{"role":"admin","email":"tenant-b-owner@hive-verify-947.invalid"},"bSeesFixtureACollection":true}
+after:  {"roleA":{"role":"admin","email":"tenant-a-owner@hive-verify-947.invalid"},"roleB":{"role":"admin","email":"tenant-b-owner@hive-verify-947.invalid"},"bSeesFixtureACollection":false}
+```
+
+- **BEFORE** (`BYPASS_ADMIN_ACCESS_CONTROL`/`ENABLE_ADMIN_CHAT_ACCESS` forced
+  `true`, simulating the upstream default this PR closes): fixture B, an
+  admin in a different tenant, sees fixture A's Knowledge collection.
+- **AFTER** (this branch's actual `docker-compose.yml`, both `false`; Open
+  WebUI container recreated, same DB volume, same accounts, same
+  collection): fixture B sees "No knowledge found".
+
+## Images
+
+Posted as release assets, linked from the PR comment (not hotlinked here, so
+this file carries no branch-pinned image reference):
+
+- https://github.com/sakibsadmanshajib/hive/pull/960#issuecomment-5325135646
+
+Local URL: `http://localhost:3003` (an isolated local compose project, never
+a public address). No credential appears in any captured URL for this flow
+(session tokens were injected via `localStorage`, not a URL query string or
+fragment), so no redaction was needed in either the text or the pixels.
--- a/deploy/docker/Caddyfile.owui
+++ b/deploy/docker/Caddyfile.owui
@@ -100,8 +100,34 @@
   # front end (the sidebar's pinned note list, the message input's Attach
   # Notes tab and its note-from-prompt action, the note link token) renders
   # only under `$config.features.enable_notes`.
+  # `knowledge/<id>/export` closes the residual half of #947 that the two
+  # environment flags below in docker-compose.yml cannot reach.
+  #
+  # Those flags gate the LISTING routes (routers/knowledge.py `GET /` and
+  # `GET /search`), so with them off an admin no longer discovers other
+  # tenants' collections. They do not gate this one. `export_knowledge_by_id`
+  # depends on `get_admin_user` and nothing else: no ownership check, no
+  # BYPASS_ADMIN_ACCESS_CONTROL check, and, unlike every sibling export route
+  # (`chats.py` and `routers/utils.py` both check it), no ENABLE_ADMIN_EXPORT
+  # check either. So a tenant OWNER, who is an instance admin here, could
+  # still download another tenant's entire collection as a zip given only its
+  # id, which is not secret in the way a credential is: it travels in shared
+  # links, support tickets, logs and screenshots.
+  #
+  # Blocked here rather than patched in Python because Dockerfile.open-webui
+  # builds only the frontend from vendor/open-webui, so an edit to
+  # routers/knowledge.py is inert unless it goes through owui-patches. This
+  # proxy already owns exactly this kind of defence-in-depth block, and the
+  # route has no legitimate caller: the product surfaces no Knowledge export,
+  # and ENABLE_ADMIN_EXPORT is already false for the routes that honour it.
+  #
+  # Deliberately NOT blocking `GET /api/v1/knowledge/<id>` or its `/files`
+  # child, which have the same admin short-circuit but ARE the routes an owner
+  # uses to open their own collection. Those need a real ownership check in
+  # Python through owui-patches, tracked separately; blocking them here would
+  # be an outage, not a fix.
   @blocked {
-    path_regexp blocked (?i)^/+(?:admin|signup|signin|signup\.html|register|create-account|users/new|invite|auth/signup|auth/signin|auth/register|auths/signup|auths/signin|openai/config|api(?:/v[0-9]+)?/(?:admin|signup|register|invite|notes|terminals|users/admin|users/new|auths/(?:signup|admin)|configs/(?:export|terminal_servers|tool_servers)))/?(?:/.*)?$
+    path_regexp blocked (?i)^/+(?:admin|signup|signin|signup\.html|register|create-account|users/new|invite|auth/signup|auth/signin|auth/register|auths/signup|auths/signin|openai/config|api(?:/v[0-9]+)?/(?:admin|signup|register|invite|notes|terminals|users/admin|users/new|auths/(?:signup|admin)|configs/(?:export|terminal_servers|tool_servers)|knowledge/[^/]+/export))/?(?:/.*)?$
   }
   respond @blocked 404
 
--- a/docs/proof/owui-admin-cross-tenant-live-2026-08-23/README.md
+++ b/docs/proof/owui-admin-cross-tenant-live-2026-08-23/README.md
@@ -1,0 +1,107 @@
+# Proof: an admin in one tenant stops reading another tenant's Knowledge
+
+Fixes hive#947. This is a security claim, so what has to be shown is the
+**negative** case, against a control that rules out the two ways such a claim
+usually lies: that the test account was never an admin, and that the "fix"
+simply broke the feature for everyone.
+
+Supplements the 2026-08-18 capture already on this PR with a re-run on the
+current stack. This branch is rebased onto `main` at `82375d07f`, where #952
+and #956 have already landed, and the capture stack additionally carries #951,
+which merges ahead of this one. So the substrate is the state that will
+actually exist when this PR lands, and `main` at that point also carries
+#1043's LiteLLM digest pin and #1012's actual-cost billing. The migration
+chain was brought current against the local database first.
+
+Both arms were re-run from scratch after the rebase. The after arm reproduced
+byte for byte; the before arm differs only in its timestamps.
+
+Text log only here; images are posted as permanent release assets through
+`scripts/post-pr-visual-proof.sh`. `npm run lint:proof-tokens` scans this
+directory and nothing else.
+
+## Method
+
+One local stack. Same two accounts, same Knowledge collection, same
+containers, same database volume across both arms. The **only** thing that
+changes between them is the two environment variables this PR sets, applied by
+recreating the `open-webui` container and confirmed live with
+`docker inspect` before the second arm ran.
+
+- **before** — the flags at their upstream default, `true`. This is not a
+  simulation of the bug: it is what every deployment runs today, because these
+  variables were never set, and upstream defaults both to `true` in
+  `config.py`. On a branch that has already fixed them, forcing them back
+  explicitly is the only way to reproduce that state, so the before arm adds a
+  scratch compose overlay that sets exactly those two variables and nothing
+  else.
+- **after** — the flags as this PR sets them, `false`, straight from the
+  branch's own `docker-compose.yml` with no overlay.
+
+The values in force were read back from the running container with
+`docker inspect` at the start of each arm, so which arm was which is a
+recorded fact rather than an intention.
+
+Two independent tenants, each provisioned through `scripts/seed-demo-owner.py`
+with its own tenant and account slug:
+
+- `owner@hive-verify-952.invalid`, tenant A, owns the collection
+  `Verify-960-Tenant-A-Secret-Kb`
+- `owner-b@hive-verify-960.invalid`, tenant B, no relationship to tenant A or
+  its collection
+
+Both signed in through a real OAuth 2.1 authorization-code round trip. Both
+accounts' Open WebUI role was read from the running server with
+`GET /api/v1/auths/` **at the moment of each capture**, so "B is an admin" is
+observed rather than asserted in prose.
+
+## Result
+
+| | before (flags `true`) | after (flags `false`) |
+|---|---|---|
+| Tenant A role | `admin` | `admin` |
+| Tenant B role | `admin` | `admin` |
+| Distinct accounts | yes | yes |
+| B's `GET /api/v1/knowledge/` | HTTP 200, **1 entry** | HTTP 200, **0 entries** |
+| A's collection in B's API response | **true** | **false** |
+| A's collection on B's Knowledge page | **true** | **false** |
+| A still sees its own collection | true | **true** |
+
+The last row is the control that matters most. If the flags had merely broken
+Knowledge, tenant A would have lost sight of its own collection too. A still
+sees it after the fix, so what changed is cross-tenant reach and not the
+feature.
+
+In the before screenshot the leak is visible with attribution: signed in as
+tenant B, the page reads "Knowledge 1" and lists
+`Verify-960-Tenant-A-...` with the byline "By Owner@hive-verify-952.invalid",
+an account in a different tenant. After, the same page for the same account
+reads "Knowledge 0" and "No knowledge found".
+
+## A failure this capture nearly recorded as a pass
+
+The first attempt at the after arm ran before the recreated `open-webui`
+container was healthy. Every request answered 401, so tenant B "saw nothing"
+and the arm would have read as a clean pass while proving only that nobody was
+signed in.
+
+The capture script now aborts outright when either account's role is not
+`admin` at capture time, rather than warning and continuing. A signed-out arm
+is indistinguishable from a working fix by the metric under test, so it must
+never be allowed to produce a result at all.
+
+## Scope
+
+This closes the read path only. It does not change who holds the Open WebUI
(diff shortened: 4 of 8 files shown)


sakibsadmanshajib:

**CodeRabbit CLI**: SKIPPED. `coderabbit review --base main --agent` returned `rate_limit` ("You've used all 3 included reviews currently available under your plan," 16 minute wait, no assigned seat for usage-based reviews). Not treated as a pass.

**ecc:code-review / `/codex:adversarial-review`**: SKIPPED. This review was run by a security-reviewer subagent with Read/Edit/Bash tools only, no Agent/Task-spawning or Skill-invocation tool in its toolset, so these two streams could not be dispatched. Structural capability gap, not a judgment call to skip; flagging for the orchestrator to run separately if the full dynamic-selector set is wanted before merge.

**Security review (mandatory for this diff, run directly)**:

- Diff is exactly two new lines plus a comment in `deploy/docker/docker-compose.yml`'s `open-webui` service (`BYPASS_ADMIN_ACCESS_CONTROL: "false"`, `ENABLE_ADMIN_CHAT_ACCESS: "false"`). No new code path, no secrets, no input handling.
- Confirmed by source read that `BYPASS_ADMIN_ACCESS_CONTROL`'s upstream fallback (`ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS`, also default `true`) only applies when the primary env var is *unset*. Since this PR sets it explicitly, that fallback is never consulted, and grepped the backend to confirm `ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS` has no other independent call site, so there is no parallel side-door left open.
- `docker compose config` on the edited file parses clean (exit 0).
- `npm run lint:proof-tokens` passes clean against the new `docs/proof/owui-admin-bypass-fix-2026-08-17/` folder (83 files scanned, no credential-shaped text).
- Verified live: both env vars read `false` inside a running container built from this branch, and the fix behaves as intended (admin sees own Knowledge, not another tenant's) — see screenshots in the PR description.
- Confirmed the three explicitly out-of-scope issues (#748 role-grant, #949 proxy write gate, #950 config-secret exposure) remain open and are named in the PR body, not silently left implying "fixed."

No blocking findings. Recommend merge once CI is green.

deploy/docker/docker-compose.yml
```diff
@@ -709,6 +709,21 @@ services:
...
+      # (routers/chats.py:600, :1056). Confirmed live: a demo account's
+      # Knowledge collection was visible from an unrelated personal account.
+      # Both are plain os.getenv reads at import time, not PersistentConfig,
+      # so setting them here takes effect on the next container start with
+      # no database reconciliation needed. Scope: this does not change who
+      # holds the admin role (#748) or the proxy write-verb gap (#949); it
+      # only stops an admin session from reading across tenants once it
+      # exists.
+      BYPASS_ADMIN_ACCESS_CONTROL: "false"
+      ENABLE_ADMIN_CHAT_ACCESS: "false"
```
sakibsadmanshajib, line 733: Confirmed both coverage and precedence for this change. `docker-compose.enterprise.yml` never defines an `open-webui` service, so this single block is the only place Open WebUI starts across `local`, `enterprise`, and `chat`. No `env_file` exists anywhere in this file, and both values here are quoted literal strings rather than `${VAR}` interpolation, so nothing in `.env` or a shell export can flip them back. Checked the OWUI parser directly: `config.py:2029` and `:2037` do `os.getenv(KEY, 'True').lower() == 'true'`, so the literal `"false"` reads as `False`, matching how `ENABLE_ADMIN_EXPORT` already behaves two lines above.

One nuance worth a one-line comment addition here, not a blocker: `BYPASS_ADMIN_ACCESS_CONTROL`'s upstream default, when unset, falls back to reading `ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS` (also default `true`), an undocumented second name for the same knob. It doesn't matter today since this line sets `BYPASS_ADMIN_ACCESS_CONTROL` directly and the outer `os.getenv` wins, but a future editor auditing this block for completeness has no way to know that alias exists unless it's named here.

docs/proof/owui-admin-bypass-fix-2026-08-17/README.md
```diff
@@ -0,0 +1,49 @@
...
+same mechanism every real login produces. No password was set, read, reset,
+or rotated on any account anywhere in this exercise, and neither account is
+the owner's personal account or the shared demo account:
+
+* `tenant-a-owner@hive-verify-947.invalid`, role `admin`, owns one Knowledge
+  collection, `Verify-947-Tenant-A-Secret-Kb`.
+* `tenant-b-owner@hive-verify-947.invalid`, role `admin`, a different tenant,
+  no relationship to the first account or its collection.
+
+Both accounts are `admin` (as every sole tenant OWNER is on this deployment,
```
sakibsadmanshajib: This is the one fact the entire proof rests on, and it isn't independently verifiable from what's posted. The two screenshots show a Knowledge list with one item versus zero, which is consistent with both fixtures being `admin` and the fix working, but it's equally consistent with fixture B simply being a non-admin `user` account, since non-admins were already correctly scoped to their own `user_id` before this fix and would show the identical empty-list screenshot regardless of whether `BYPASS_ADMIN_ACCESS_CONTROL` is true or false. The role claim here is asserted in prose, not shown.

Suggest adding one more artifact: a screenshot or API response (Settings > Users, or `GET /api/v1/users/`) showing `role: admin` for both fixture accounts, so the proof is self-contained rather than resting on the author's account of how they were created. Not asking for a pre-fix negative-control screenshot since the underlying leak was already independently confirmed live by the owner, but the admin-role gap is worth closing before merge given this is a security fix.

sakibsadmanshajib (commented):
Independent review, substituting for the two review streams that could not run. Verified against the pushed diff and the vendored Open WebUI source (v0.10.2), not just the PR description.

Coverage: confirmed clean. `docker-compose.enterprise.yml` defines no `open-webui` service at all (it only overlays the self-hosted Supabase data plane), so the single service block this PR edits is the only place Open WebUI ever starts, across `local`, `enterprise`, and `chat` profiles. `deploy-demo-box.yml` brings the stack up with `--profile local --profile chat --profile monitoring`, which is covered.

Precedence: confirmed clean. No `env_file` directive exists anywhere in `docker-compose.yml` (grepped, zero hits), and both new lines are quoted literal strings (`"false"`), not `${VAR}` interpolation, so nothing in `.env` or the shell environment can override them at the Compose interpolation stage. Checked the parser too: `open_webui/config.py:2029` and `:2037` do `os.getenv(KEY, 'True').lower() == 'true'`, so the literal string `"false"` evaluates to Python `False` as intended, the same pattern the sibling `ENABLE_ADMIN_EXPORT` flag already relies on two lines above.

Completeness: grepped every call site of both `BYPASS_ADMIN_ACCESS_CONTROL` and `ENABLE_ADMIN_CHAT_ACCESS` across the whole vendored backend (knowledge.py, chats.py, files.py, models.py, prompts.py, tools.py, skills.py, notes.py, functions.py, access_control/__init__.py, main.py). Every occurrence gates through one of these two flags with no independent bypass. I could not find a third door inside Open WebUI's own backend. I did not check whether Hive's own `/v1/rag/chat` path in control-plane/edge-api reads the same Postgres-backed Knowledge tables directly, bypassing Open WebUI's access-control layer entirely; that's outside this diff and outside what I had time to verify, so treat completeness as confirmed for Open WebUI's own routes only, not for the whole system.

Blast radius: confirmed accurate. `BYPASS_MODEL_ACCESS_CONTROL: "true"` is already set at `docker-compose.yml:762`, untouched by this PR, and `main.py:1032` short-circuits the entire model-access check whenever that flag is true, independent of `BYPASS_ADMIN_ACCESS_CONTROL`. The model list claim holds.

Proof quality: pulled both PNGs out of the branch and looked at them directly rather than trusting the description. They match the claims exactly (`Verify-947-Tenant-A-Secret-Kb` visible in fixture A, `No knowledge found` in fixture B), and neither frame shows a URL bar, cookie, or token. See the inline comment on the proof README for the one gap that matters: nothing in the posted evidence independently shows both fixtures actually hold the `admin` role, which is the one fact the whole proof depends on.

sakibsadmanshajib:
## Visual proof

Control for #960 (same account, same collection, before/after). Both fixture accounts confirmed `role=admin` via `GET /api/v1/auths/` at the moment of each capture (JSON below, not a separate screenshot): `tenant-a-owner@hive-verify-947.invalid` and `tenant-b-owner@hive-verify-947.invalid`, same account ids across the flag flip.

- **BEFORE** (`BYPASS_ADMIN_ACCESS_CONTROL`/`ENABLE_ADMIN_CHAT_ACCESS` forced `true`, simulating the upstream default this PR closes): fixture B, an admin in a different tenant, sees fixture A's Knowledge collection `Verify-947-Tenant-A-Secret-Kb`.
- **AFTER** (this branch's actual `docker-compose.yml`, both `false`; container recreated, same DB volume, same accounts, same collection): fixture B sees "No knowledge found".

Role check JSON, captured at each phase via the same bearer tokens used for the screenshots:

```
before: {"roleA":{"role":"admin","email":"tenant-a-owner@hive-verify-947.invalid"},"roleB":{"role":"admin","email":"tenant-b-owner@hive-verify-947.invalid"},"bSeesFixtureACollection":true}
after:  {"roleA":{"role":"admin","email":"tenant-a-owner@hive-verify-947.invalid"},"roleB":{"role":"admin","email":"tenant-b-owner@hive-verify-947.invalid"},"bSeesFixtureACollection":false}
```

sakibsadmanshajib:
## `/v1/rag/chat` tenant-scoping check (not a fix, reporting only)

Read `apps/edge-api/internal/rag/chat_handler.go` and `apps/edge-api/internal/rag/repository.go` end to end.

**Plain answer: it enforces tenant scoping in its own right. It is not a second door to the data this PR closes, and does not rely on the Open WebUI layer these two flags gate.**

`/v1/rag/chat` and Open WebUI's own Knowledge feature are two entirely separate stores (also noted in `DEMO.md`): this endpoint is Hive's own Go code in `edge-api`, backed by its own Postgres tables (`rag_documents`, `rag_chunks`), never touching Open WebUI's database or its `BYPASS_ADMIN_ACCESS_CONTROL`/`ENABLE_ADMIN_CHAT_ACCESS` checks at all.

Tracing the actual query path, `handleChat` (`chat_handler.go:68`) resolves `user.TenantID` from the request's own auth context (`auth.UserFrom`, populated from the caller's JWT/API key, independent of any Open WebUI role) and passes it straight into `h.store.SearchChunks(ctx, user.TenantID, vec, topK)`. `SearchChunks` (`repository.go:200`) enforces the tenant boundary twice, not once:

1. `withTenantTx` (`repository.go:71`) opens the transaction with `SELECT set_config('app.current_tenant_id', $1, true)`, which is a genuine Postgres session-scoped setting a Row Level Security policy on `rag_chunks`/`rag_documents` reads from — this is DB-level enforcement, not application-level.
2. The query itself also binds `tenant_id` as an explicit `WHERE` parameter (`repository.go`'s comment names this directly: "Explicit tenant_id filter is defense-in-depth alongside RLS: protects against SECURITY DEFINER / superuser-bypass scenarios").

There is a dedicated live test for exactly this: `TestRepo_RLS_SearchChunksCannotReadOtherTenantChunks` (`repository_rls_test.go:221`).

So: no action taken here, this is a report as asked. If anything about tenant resolution in `auth.UserFrom` itself is ever suspected of being foolable, that would be a separate, unrelated finding from #947 (Open WebUI's admin bypass) — nothing in this PR's diff touches this path, and nothing in this path reads the two flags this PR sets.

deploy/docker/docker-compose.yml
sakibsadmanshajib: Fixed in 2f982b34d: added a one-line comment naming `ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS` as `BYPASS_ADMIN_ACCESS_CONTROL`'s undocumented fallback alias (default `true`, `config.py:2032`), right in this block, so a future edit auditing it does not have to read the vendored source to rediscover it. Confirmed it changes no behavior here: `docker compose config` still parses clean and the explicit `BYPASS_ADMIN_ACCESS_CONTROL: "false"` still wins regardless of the alias's value.

docs/proof/owui-admin-bypass-fix-2026-08-17/README.md
sakibsadmanshajib: Agreed, and fixed: the original two screenshots could not distinguish "the fix works" from "the test account was never admin." Replaced with a real control in commit 6cd1d7ef4 (proof text log now at `docs/proof/owui-admin-bypass-fix-2026-08-17/README.md`, images posted via `scripts/post-pr-visual-proof.sh`, comment: https://github.com/sakibsadmanshajib/hive/pull/960#issuecomment-5325135646).

Same two accounts, same Knowledge collection, captured twice against the same container and volume, with only `BYPASS_ADMIN_ACCESS_CONTROL`/`ENABLE_ADMIN_CHAT_ACCESS` flipped between captures:

- Both accounts' `role=admin` confirmed via `GET /api/v1/auths/` at the moment of each capture (not asserted in prose): `tenant-a-owner@hive-verify-947.invalid`, `tenant-b-owner@hive-verify-947.invalid`.
- Before (flags forced back to upstream's `true` default): fixture B, admin in a different tenant, sees fixture A's collection.
- After (this branch's actual `docker-compose.yml`, both `false`): same fixture B sees "No knowledge found".

That is a proof that would look different if the fix did not work, which the original screenshots were not. Resolving this thread.

PR #959, opened 2026-08-17
test: scratch PR to empirically race proof-posting mechanisms

sakibsadmanshajib:
Scratch PR, not for merge. Testing which mechanism renders an inline image in a PR comment on github.com, and which survives branch deletion. Will be closed and the branch deleted once the test is done. See internal task: image-proof mechanism selection.

PR #961, opened 2026-08-17
fix: post PR visual proof via a permanent release asset, not a branch-pinned link

sakibsadmanshajib:
## What

The owner reported that the current visual-proof mechanism (commit a
screenshot to `docs/proof/` and link it from a PR comment) does not display
for him. This establishes, empirically, which posting mechanism actually
renders inline on github.com and survives this repo's merge policy, then
makes it the standard.

## How this was established

Four candidates were raced on a real scratch PR (#959, closed, scratch
branch already deleted), each posted as its own comment, then screenshotted
with an actual Playwright-driven Chromium browser (not assumed from memory
of how GitHub is supposed to behave):

| Candidate | Renders inline? | Survives branch deletion? |
|---|---|---|
| `raw.githubusercontent.com` pinned to the PR's own branch **name** | Yes, until the branch is deleted | **No.** `curl` against that exact URL returned a 404 with `x-cache: MISS` (straight from origin, not a stale edge cache) seconds after `git push origin --delete` on the branch it was pinned to. |
| `blob/...` link, as a markdown image or a bare link | **No.** Broken-image icon either way; a blob URL serves an HTML page, not raw bytes. | n/a |
| `data:` URI embedded directly in the comment | **No.** GitHub's markdown sanitizer strips it; broken-image icon. | n/a |
| Release asset (`releases/download/...`) | **Yes.** | **Yes.** Not reachable through any branch at all. |

This reproduces, on a fresh test, exactly the failure already recorded for
PR #867: proof read once at review time looked fine, then rotted the moment
the PR's branch was deleted, which is what this repo's squash-and-delete
merge policy does to every PR on merge.

A `raw.githubusercontent.com` link pinned to the commit **SHA** instead of
the branch name also survives branch deletion (GitHub retains a PR's head
commit via its own internal ref indefinitely, confirmed both for an open PR
and for an already squash-merged, closed one). It was rejected as the
standard anyway: it depends on undocumented retention behavior, and is one
easy slip away from silently repeating PR #867 (pinning the branch name
instead of the SHA is exactly the mistake that PR made). The release asset
has no branch-vs-SHA distinction for anyone to get wrong.

The true drag-and-drop attachment upload (the one producing
`github.com/user-attachments/assets/...` URLs) was confirmed unreachable by
script: a 404 on the plausible REST path, and no matching field on GraphQL's
`Issue` type. It is web-UI only, authenticated by browser session cookies
and a CSRF token. Per the task's constraint, this was not attempted with
anyone's browser session; the release-asset mechanism was adopted instead.

Full narrative and the four rendered/broken screenshots referenced above:
`.wolf/decisions.md` D-042.

## What ships

- `scripts/post-pr-visual-proof.sh`: uploads one or more local images to a
  permanent, never-deleted GitHub Release (`visual-proof-assets`, created
  once and reused by every future proof) and posts one PR comment with the
  resulting `releases/download/` markdown images.
- `.claude/rules/orchestrator.md` rule 8: names the script explicitly
  instead of the vague "posted in the PR" requirement. The existing
  credential-masking obligation (mask any URL carrying a token in both the
  text log and the screenshot pixels before upload) is unchanged.
- `.wolf/decisions.md` D-042: the full empirical record.

### Not shipped here: the skill file

The task asked for a skill at `.claude/skills/pr-visual-proof.md`. The
skill-router hook blocks any edit under `.claude/skills/*.md` until
`skill-creator:skill-creator` or `superpowers:writing-skills` has been
invoked this session, and this agent's toolset (Read/Write/Edit/Bash) has no
Skill tool to satisfy that gate. Per the Gate Compliance rule this is
reported as a genuine capability gap rather than routed around. The
finished file content is ready and was handed to the dispatching agent to
land with one Skill-enabled call.

## Proof this PR's own claim renders

Posted below via `scripts/post-pr-visual-proof.sh` itself, so this PR is its
own demonstration.

## Buglog entry

```json
{"id":"bug-2026-08-17-proof-link-dies-on-branch-delete","date":"2026-08-17","title":"PR visual proof rendered fine at review time and then rotted the moment the PR merged","error_message":"raw.githubusercontent.com/<owner>/<repo>/<branch>/docs/proof/... 404s once <branch> is deleted","root_cause":"Proof images were linked by a raw.githubusercontent.com URL pinned to the PR's own branch name. This repo's merge policy is squash merge with branch deletion, so the branch ref the URL depends on stops existing the moment the PR merges, and the image silently starts 404ing. Nobody notices because proof is read once, at review time, before the branch is gone. Confirmed empirically on scratch PR #959: curl showed a cache MISS 404 straight from origin seconds after git push origin --delete on the branch a raw link had been pinned to.","fix":"scripts/post-pr-visual-proof.sh uploads every proof image to one permanent, never-deleted GitHub Release (visual-proof-assets) and posts the PR comment with its releases/download/ URL, which is not reachable through any branch at all. Orchestrator rule 8 updated to require this script by name. A commit-SHA-pinned raw link also survives branch deletion but was rejected as the standard since it depends on undocumented GitHub retention behavior and repeats the branch-name mistake one slip away.","tags":["visual-proof","github","raw-githubusercontent","release-asset","link-rot","pr-867","d-042"]}
```


sakibsadmanshajib:
## Visual proof

First screenshot: the existing agent-visual-proof.yml CI job already used this same class of mechanism (a dedicated, never-deleted branch plus a raw.githubusercontent.com link) and it renders inline, confirmed live on PR #891. Second screenshot: this PR's own script (scripts/post-pr-visual-proof.sh) posting to the new permanent release, confirmed rendering inline seconds ago.

sakibsadmanshajib:
## Adversarial review (stage 6), run late

Stage 6 was skipped when this PR was opened, because the agent that wrote it had no Skill tool. This is that review, run properly. The mechanism this PR chooses (release asset over branch pinned raw link) is correct and the empirical record behind it holds up. The defects are all in the script that this PR is about to make mandatory for every agent, and which publishes to a public repository permanently.

Two of them were confirmed by experiment rather than by reasoning, and both produced a success exit code while the proof did not actually exist. Fixes are pushed in 3af5a6b.

### Streams

| Stream | Status |
|---|---|
| Plain adversarial pass | Ran (findings below) |
| Security review | Ran (findings below) |
| shellcheck | Ran, clean (exit 0) after fixes |
| Runtime failure path testing | Ran (transcript below) |
| CodeRabbit CLI | **SKIPPED**, not a pass |
| `ecc:code-review` | Ran (folded into the findings below) |

CodeRabbit CLI returned `rate_limit`: "You have used all 3 included reviews currently available under your plan", and this Git provider account is not linked to an assigned seat. Several agents are live in this repo and share that allowance. Declared SKIPPED rather than treated as a clean pass.

### HIGH: the asset name and the URL could diverge, silently

`gh release upload` does not publish the name it is given. GitHub rewrites it. Verified against this repo's own release rather than assumed: a file uploaded as `name space test.png` was published as `name.space.test.png`. The test asset was removed immediately afterwards.

The script built the markdown URL from the name it chose locally, so any name GitHub would rewrite yielded a URL pointing at nothing. Markdown compounds it by terminating a URL at the first space and treating `#` as a fragment. The script then exited 0 and reported success. Screenshot tools produce spaced filenames routinely ("Screen Shot 2026-08-17 at ....png"), so this was not a corner case, and it is the same class of failure this PR exists to eliminate: proof that looks posted and does not render.

Fixed by folding every name into `[A-Za-z0-9._-]` before staging, so the published name and the URL match by construction. Names already inside that set publish verbatim, which the release's existing assets confirm. A preflight self test asserts the sanitizer on every invocation, mirroring the MUST_CATCH pattern this repo's `tools/lint-*.mjs` guards already use.

### HIGH: images were published before the pull request was validated

Uploads ran in the loop and `gh pr comment` ran last, so a mistyped or non existent PR number left every image public and permanent with no comment referencing it, and no way to unpublish.

The obvious guard is subtly wrong, which is worth recording. `gh pr view <n> --json number` answers from the argument itself and exits 0 for a PR that does not exist:

```
$ gh pr view 99999999 --repo sakibsadmanshajib/hive --json number
{"number":99999999}
$ echo $?
0
```

The first version of this fix used exactly that field, which made the guard a no op. It was caught by running it rather than reading it: the test published two `pr99999999-*` assets to the public release, which were then deleted. `--json state` forces the API round trip and correctly exits 1. The script now carries a comment saying not to simplify that field.

### MEDIUM: rule 8 kept citing a CI guard it had just cut off

`tools/lint-no-token-in-proof-captures.mjs` scans `docs/proof/` and nothing else. The rewritten rule 8 forbids `docs/proof/` for the image, mandates the script, and never says where the capture text log goes, while still naming `npm run lint:proof-tokens` as the guard for the text half.

The linter would not have failed. It would have kept scanning the 82 legacy files, kept passing, and kept reporting green over a frozen corpus that no new proof would ever join. That is the failure mode this repo has already written down for itself: a loud CI failure traded for a quiet absence, and worse than a deleted check because it still looks alive. The asymmetry that makes it a real regression is that the linter never inspected pixels under either mechanism, so the text half is the only thing it ever caught, and the text half is exactly what leaked in PR #578.

Rule 8 now requires the capture text log to stay committed under `docs/proof/<slug>/` in the same PR. Only the image moves to the release.

### MEDIUM: nothing scans an uploaded asset, and deletion is not revocation

A release asset is blob storage attached to a tag, not a git object, so GitHub secret scanning, push protection and GitGuardian never see it, and neither does our linter. Under the old committed file mechanism a leaked credential had two independent chances to be caught after the fact. Under this one it has none, and depends entirely on the capturing agent's eyes.

Separately, both the script comment and the release notes presented `gh release delete-asset` as the remedy. Removal is genuinely easier here than a history rewrite, which is a real improvement worth keeping, but it does not close an incident: the credential stays valid until revoked, the PR comment and its download URL remain in the timeline and the API, the notification emails GitHub already sent are not retractable, and release asset downloads are not logged anywhere readable, so exposure cannot be bounded afterwards.

Rule 8, the script header and the release notes now say revoke first, then delete, and state that nothing downstream will catch a credential in pixels.

### MEDIUM: the script would publish whatever it was handed

Validation was `[ -f "$img" ]` only. A `.txt`, a `.pem`, a heap dump or a dotenv all uploaded fine, and `post-pr-visual-proof.sh 961 docs/proof/run-1/*` is a natural thing to type that would have published every text log in that directory to a place nothing scans. `cp` also dereferences symlinks by default, so a link in a capture directory would publish its target's bytes rather than the file the caller reviewed.

Now refuses anything that is not `png/jpg/jpeg/gif/webp`, and refuses symlinks outright rather than resolving them, so what gets reviewed is what gets uploaded.

### MEDIUM: rule 8 accepted a screen recording and mandated a script that cannot post one

Rule 8 says "screenshot or screen recording" and then requires the artifact to render inline via this script. Both cannot hold. A markdown image tag pointing at an `.mp4` renders a broken image icon; GitHub plays video only for its own web UI attachment host, which the script's own header documents as unreachable. That was a false green available on the one rule this repo calls non negotiable. Rule 8 now says a recording cannot render inline, and the script refuses non image files rather than accepting an `.mp4` and exiting 0.

### LOW, fixed: obligations dropped in the rewrite

The old rule 8 forbade proof "held in a scratch dir"; the rewrite dropped the clause. Restored.

### LOW, fixed: no chance to review what was about to become public

The header comment carried the permanence warning but the runtime said nothing. A blocking prompt is the wrong answer because agents run unattended, so the script now prints a manifest of repo, release, PR and every file to stderr before uploading, which also lands in the transcript.

### LOW, fixed: concurrent release creation

Two agents could both see no release and both try to create it; the loser aborted and lost the proof it was posting. The create now falls back to a re check.

### Verified, not asserted

```
$ shellcheck scripts/post-pr-visual-proof.sh          # exit 0

$ post-pr-visual-proof.sh 961 <missing>
::error:: no such file: .../nope.png                                     exit 1
$ post-pr-visual-proof.sh 961 notes.txt
::error:: not an image (png/jpg/jpeg/gif/webp): .../notes.txt            exit 1
::error:: GitHub renders no other type inline, and an upload cannot be unpublished.
$ post-pr-visual-proof.sh 961 link.png
::error:: refusing a symlink (publish the real file you reviewed): ...   exit 1
$ post-pr-visual-proof.sh 961 --caption "only a caption"
Usage: ...                                                               exit 1
$ post-pr-visual-proof.sh 99999999 real.png
GraphQL: Could not resolve to a PullRequest with the number of 99999999
::error:: no such pull request in sakibsadmanshajib/hive: 99999999 (nothing was uploaded)
                                                                         exit 1
```

Every one fails before any network call except the last, which fails before any upload.

### Checked and correct as written

- Not passing `--clobber` to `gh release upload` is right. A name collision fails loudly instead of silently overwriting another agent's proof.
- An unauthenticated `gh` now fails at PR resolution, before anything is published.
- D-042 does not collide: `origin/main` is at D-041, and no other remote branch claims D-042.
- No workflow triggers on `release`, so creating the release fires nothing.
- The `mktemp`/`trap` cleanup is correct under `set -euo pipefail`.

### Also landed here

`.claude/skills/pr-visual-proof.md`, which rule 8 already cited and which did not exist, so the rule was a dangling pointer to the one step that has no automated backstop. The staged draft needed two corrections before landing: `name` carried spaces where the spec allows only letters, digits and hyphens, and `description` described what the skill does rather than when to use it, which is the documented way to get an agent to follow the description and skip the body.

### Not done, deliberately

No fresh end to end proof comment was posted from the hardened script. The upload and comment path is byte identical to the version that already rendered twice on this PR, every change is a validation gate in front of it, and the one changed line in that path is covered by a self test that runs on every invocation. Re posting would have added a redundant comment for no new evidence.

`Web E2E (full stack)` is red for unrelated reasons and is another agent's gate. Not touched.

sakibsadmanshajib:
## Visual proof

Posted by the hardened script itself, so this comment is its own end-to-end test: upload, published-name readback, and comment. Verification detail in the review comment above.

sakibsadmanshajib:
## Second adversarial pass, and end to end verification

A second independent adversarial pass ran against the hardened script and found seven more issues, all probe verified rather than asserted. Fixes in `fa377bb3` and `2c340fbe`.

### Fixed

**The markdown label used the raw basename while only the URL was sanitized.** A basename containing `]` closes the label early, so the comment renders no image and the script still exits 0. That is the same silent-success class this PR exists to remove, reproduced inside the fix for it. The label is now the sanitized name; alt text is cosmetic, so nothing is lost.

**A mid-loop upload failure left assets public and permanent that the caller could not name.** Uploads that already succeeded stayed published, no comment was posted, and the published names were never printed because upload output went to `/dev/null` and the name embeds `$RANDOM`. The header comment tells the operator to remove a leaked asset by name, and on that path the script had given them no name. Every published URL is now echoed to stderr as it succeeds, which also gives the happy path a transcript record of exactly what went public.

**A leading-dash filename passed every gate and then died inside the loop**, with `basename` or `cp` parsing it as an option. On a multi-image call that is the orphan-asset case above. Rejected in the parser now, and `basename` and `cp` are given `--` as well.

**An unknown flag was silently reclassified as an image**, surfacing later as a misleading "no such file: --verbose".

**The temp body file was created before the trap that removes it**, so a failure of the following `mktemp` leaked it. One temp dir, one trap, body file inside it: fewer lines and no window.

**The PR-resolution error claimed the PR did not exist**, when the same failure also covers an unauthenticated `gh` and a network error. It now says what it actually knows.

### The published name is now read back, not assumed

Taking the reviewer's closing suggestion. The script sanitized a name so GitHub would not rewrite it, then built the URL from that name, which is a correct-by-construction claim of exactly the kind this script exists to stop us making. GitHub owns the published name, we do not. After every upload the release is queried and the published name matched exactly against the one sent; a mismatch aborts rather than posting a comment whose image would 404. This would have caught the original space-rewrite defect automatically.

Matching uses a herestring rather than a pipe on purpose: `grep -q` exits early, and under `pipefail` the resulting SIGPIPE on `gh` would fail a lookup that had actually succeeded. Match semantics were checked for both false negatives and substring false positives.

### Not fixed, deliberately

A 255-byte basename still fails at `cp` with ENAMETOOLONG mid-loop. It fails safely, nothing wrong is published, and the published-URL line makes it recoverable.

`--clobber` is still absent, and deliberately so. Failing loudly on a name collision prevents one agent silently overwriting another agent's proof, which with several agents posting concurrently is a live scenario.

### Verified end to end

The comment below this one was posted by the hardened script itself, so it is its own test: manifest, upload, published-name readback, and comment. The resulting URL was then fetched independently.

```
published: .../releases/download/visual-proof-assets/pr961-20260818073622-8180-proof-script-verification.png
curl -L -o /dev/null  ->  http=200  bytes=50920
```

`shellcheck` is clean (exit 0) on the final script.

### Also folded in

The skill now says a capture must never be written into a test reporter's own output directory. Playwright's HTML reporter clears its output folder before writing, so proof captured into `playwright-report/` is created and then deleted and the run looks like it produced nothing. That cost a full capture pass on PR #951.

`.claude/skills/pr-visual-proof.md` has landed on this branch, so rule 8 no longer cites a file that does not exist.

PR #963, opened 2026-08-18
fix: recover default-workspace provisioning from a slug collision race

sakibsadmanshajib:
## Summary

Unblocks the required `Web E2E (full stack)` check on `main`, which was intermittently red and hard-blocking every queued PR (including #960, a live cross-tenant data exposure fix).

- Fixes a genuine, intermittent backend race condition, not a test-selector or fixture drift.
- `provisionDefaultWorkspace` (default-workspace auto-provisioning on first sign-in) had no idempotency guard. Next.js Server Components call `getViewer()` unmemoized per component (`layout.tsx` and each `page.tsx` independently), so a brand-new user's first console page load can fire two concurrent `GET /api/v1/viewer` requests. Both see zero active memberships and both try to create a personal workspace with the same deterministic slug (`buildSlug` derives it purely from the viewer's own display name). The loser hits `accounts.slug`'s unique constraint, and `CreateAccount` let that raw pg error escape unhandled as an opaque 500 ("could not load your workspace"), tripping the console's generic error boundary.
- E2E mints a fresh test user per run, so its very first page load reliably lands in this race window on a real, unlucky-timing basis, hence flaky (fails once, passes on Playwright's retry) rather than a hard, permanent break. This is exactly what the suite's flake-rate gate ("A test that only passes on a retry is not a passing test") is there to catch.

## Root cause investigation

- Walked `Web E2E (full stack)` job history on `main`: last clean pass was `1dc67ee6` (2026-08-17 16:56). PR #955's own pre-merge CI run hit this exact race on two specs (`console-spend-alerts.spec.ts`, `profile-completion.spec.ts`), both reported "2 flaky" (failed once, passed on retry) by the job's flake gate, which is what failed the job, not a hard 0-passed failure.
- Downloaded that run's `nextjs-log-web-e2e` artifact: the underlying server error was `Error: could not load your workspace` with digests matching the two `error-context.md` snapshots' "quote reference" numbers exactly (`270132166`, `3259196144`). Traced that message to `writeInternal(w, r, "could not load your workspace", err)` in `accounts/http.go`'s `handleGetViewer`, then to `EnsureViewerContext` → `provisionDefaultWorkspace` → `CreateAccount`'s unhandled `INSERT ... accounts (slug ...)` unique-violation.
- This is unrelated to the separate, already-known-stale `OWUI nightly e2e` selector issue (`button` named "New chat" vs the `<a>` OWUI actually renders) that's in flight on #951 — different harness, different spec, not touched here.
- Neither #954 nor #955 introduced this: #955 is a test-only Go change and could not plausibly affect the web console; #954 touches auth-header normalization, unrelated to workspace provisioning. Both merged while this check was still pending (a known process gap, flagged separately), but neither is the cause.
- This gate has **not** been silently ignored: the two prior real (non-skipped, non-cancelled) job completions on `main` before this PR were both green (`1dc67ee6` success, `ef06d693` success). PR #955 merged with its own pre-merge run showing this failure, which is the actual process gap worth naming; it did not merge past a run that had already gone green.

## Fix

- New sentinel `accounts.ErrSlugTaken`, returned by `CreateAccount` on a unique violation on `slug` (mirrors the existing `CreateMembership` → `ErrAlreadyMember` translation already in the same file).
- `provisionDefaultWorkspace` now distinguishes the two situations a slug collision can mean:
  1. A concurrent request for the **same** viewer already won the provisioning race — re-checking that viewer's own memberships finds the winner's row, and this call recovers silently (no error; the caller re-lists memberships right after, same as the success path).
  2. A genuine collision between two **different** viewers whose display names happen to produce the same slug (`buildSlug` has no per-user uniqueifier) — this viewer still holds no membership after the collision, so it retries with a de-duplicated slug (bounded, 3 attempts) instead of failing that viewer's first sign-in outright.

## Test plan

- [x] `go build ./apps/control-plane/...` — clean.
- [x] `go vet ./apps/control-plane/...` — clean.
- [x] `gofmt -l apps/control-plane/internal/accounts/` — no unformatted files.
- [x] `go test ./apps/control-plane/internal/accounts/... -count=1` — all pass, including two new regression tests:
  - `TestEnsureDefaultAccount_SlugCollisionFromConcurrentSelf_Recovers`
  - `TestEnsureDefaultAccount_SlugCollisionFromDifferentViewer_RetriesWithSuffix`
- [ ] `Web E2E (full stack)` on this PR — will confirm the spend-alerts and profile-completion specs no longer flake once CI runs.

## Buglog entry

Per `.wolf/` policy, not appended to `.wolf/buglog.jsonl` on this branch. To be carried into a separate buglog-only PR off `main` once this merges:

```json
{"error_message": "could not load your workspace (500) on GET /api/v1/viewer, surfacing as the web console's generic error boundary on the spend-alerts and billing-settings pages", "root_cause": "provisionDefaultWorkspace has no idempotency guard: two concurrent Server Component calls to getViewer() for the same brand-new user both attempt to create a personal workspace with the same deterministic slug, and CreateAccount let the resulting pg unique-violation on accounts.slug escape as a raw 500 instead of recovering the race", "fix": "translate the unique violation into accounts.ErrSlugTaken and have provisionDefaultWorkspace re-check the viewer's own memberships on collision: recover silently if a concurrent request for the same viewer already won, otherwise retry once with a de-duplicated slug for the genuine different-viewer collision case", "tags": ["accounts", "web-e2e", "race-condition", "ci-gate", "flaky-test"]}
```

The PR makes first-login workspace provisioning atomic and resilient to both same-viewer races and cross-viewer slug collisions.
- Adds a per-viewer PostgreSQL advisory transaction lock and lock-scoped membership re-check.
- Atomically inserts the account, owner membership, and profile.
- Translates slug uniqueness violations into a sentinel and retries cross-viewer collisions with a deduplicated slug.
- Adds regression coverage and updates repository test doubles for the expanded interface.
</details>

The PR appears safe to merge, with no actionable changed-code defect identified.

The transaction serializes provisioning by viewer, re-checks the same active-membership invariant used by the service, atomically persists all required workspace rows, and correctly handles the current slug uniqueness constraint.
</details>

| Filename | Overview |
|----------|----------|
| apps/control-plane/internal/accounts/repository.go | Adds atomic, transaction-scoped workspace provisioning with a per-viewer advisory lock, active-membership re-check, and precise slug-conflict translation. |
| apps/control-plane/internal/accounts/service.go | Routes default provisioning through the atomic repository method and retries genuine cross-viewer slug collisions with bounded randomized suffixes. |
| apps/control-plane/internal/accounts/service_test.go | Adds focused regression tests for concurrent same-viewer recovery and different-viewer slug deduplication. |
| apps/control-plane/internal/accounts/types.go | Introduces the ErrSlugTaken sentinel used to distinguish retryable slug collisions. |
| apps/control-plane/internal/accounting/http_test.go | Updates the account repository test double for the new atomic provisioning interface. |
| apps/control-plane/internal/budgets/http_test.go | Updates the account repository test double for the new atomic provisioning interface. |
| apps/control-plane/internal/ledger/service_test.go | Updates the account repository test double for the new atomic provisioning interface. |
| apps/control-plane/internal/profiles/service_test.go | Updates the profile-aware repository test double for atomic provisioning. |
| apps/control-plane/internal/usage/service_test.go | Updates the account repository test double for the new atomic provisioning interface. |

</details>

```mermaid
sequenceDiagram
  participant A as Viewer request A
  participant B as Viewer request B
  participant S as Accounts service
  participant DB as PostgreSQL
  A->>S: EnsureViewerContext
  B->>S: EnsureViewerContext
  A->>DB: Begin + lock(viewer)
  B->>DB: Begin + wait for lock(viewer)
  A->>DB: Re-check membership
  A->>DB: Insert account, membership, profile
  A->>DB: Commit and release lock
  B->>DB: Acquire lock and re-check
  DB-->>B: Existing active membership
  B-->>S: "wonElsewhere = true"
  S-->>B: Re-list and use winner workspace
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix: close the workspace-provisioning ra..."](https://github.com/sakibsadmanshajib/hive/commit/e08b57f7b408d9dcd03222acf14b221260e88865) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54270438)</sub>

**Context used:**

- Knowledge Base — [Control-plane identity and access](https://app.greptile.com/scubed/-/custom-context/knowledge-base/sakibsadmanshajib/hive/-/docs/control-plane-identity-access.md)

### DIFF
--- a/apps/control-plane/internal/accounts/repository.go
+++ b/apps/control-plane/internal/accounts/repository.go
@@ -115,11 +115,22 @@ func (r *pgxRepository) ActiveTenantID(ctx context.Context, userID uuid.UUID) (u
 	return tenantID, true, nil
 }
 
+// CreateAccount inserts one account row.
+//
+// UNIQUE (slug) is what surfaces as ErrSlugTaken. See its doc comment in
+// types.go for the two situations that trips it, and provisionDefaultWorkspace
+// in service.go for how each is told apart and recovered — without this
+// translation the raw pgx unique-violation escaped as an opaque
+// "could not load your workspace" 500.
 func (r *pgxRepository) CreateAccount(ctx context.Context, acct Account) error {
 	_, err := r.pool.Exec(ctx, `
 		INSERT INTO public.accounts (id, slug, display_name, account_type, owner_user_id)
 		VALUES ($1, $2, $3, $4, $5)
 	`, acct.ID, acct.Slug, acct.DisplayName, acct.AccountType, acct.OwnerUserID)
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation {
+		return ErrSlugTaken
+	}
 	return err
 }
 
--- a/apps/control-plane/internal/accounts/repository.go
+++ b/apps/control-plane/internal/accounts/repository.go
@@ -21,6 +21,14 @@ type Repository interface {
 	CreateAccount(ctx context.Context, acct Account) error
 	CreateMembership(ctx context.Context, m Membership) error
 	CreateProfile(ctx context.Context, p AccountProfile) error
+	// ProvisionDefaultWorkspace atomically provisions acct/membership/profile
+	// for a first-time viewer, serialized against every other concurrent call
+	// for the same viewer. See provisionDefaultWorkspace in service.go for why
+	// this must be atomic rather than three independent Create* calls.
+	// wonElsewhere is true when a concurrent call for this same viewer had
+	// already committed a workspace by the time this call took its lock: in
+	// that case nothing is inserted and existingAccountID is the winner's.
+	ProvisionDefaultWorkspace(ctx context.Context, acct Account, membership Membership, profile AccountProfile) (existingAccountID uuid.UUID, wonElsewhere bool, err error)
 	GetAccountByID(ctx context.Context, id uuid.UUID) (*Account, error)
 	CreateInvitation(ctx context.Context, inv Invitation) error
 	FindInvitationByTokenHash(ctx context.Context, tokenHash string) (*Invitation, error)
@@ -115,20 +123,26 @@ func (r *pgxRepository) ActiveTenantID(ctx context.Context, userID uuid.UUID) (u
 	return tenantID, true, nil
 }
 
-// CreateAccount inserts one account row.
-//
-// UNIQUE (slug) is what surfaces as ErrSlugTaken. See its doc comment in
-// types.go for the two situations that trips it, and provisionDefaultWorkspace
-// in service.go for how each is told apart and recovered — without this
-// translation the raw pgx unique-violation escaped as an opaque
-// "could not load your workspace" 500.
+// accountsSlugUniqueConstraint is the Postgres-assigned name for accounts'
+// column-level `slug ... unique` constraint (supabase/migrations/
+// 20260328_01_identity_foundation.sql), i.e. Postgres's default
+// <table>_<column>_key naming, never given an explicit name in the
+// migration. Matched by name, not just by SQLSTATE 23505, so a future
+// unique constraint added to this table is never misreported as
+// ErrSlugTaken.
+const accountsSlugUniqueConstraint = "accounts_slug_key"
+
+// CreateAccount inserts one account row. provisionDefaultWorkspace does not
+// call this directly — see ProvisionDefaultWorkspace below, which needs the
+// insert inside its own locked transaction. UNIQUE (slug) still surfaces
+// here as ErrSlugTaken for any other caller; see its doc comment in types.go.
 func (r *pgxRepository) CreateAccount(ctx context.Context, acct Account) error {
 	_, err := r.pool.Exec(ctx, `
 		INSERT INTO public.accounts (id, slug, display_name, account_type, owner_user_id)
 		VALUES ($1, $2, $3, $4, $5)
 	`, acct.ID, acct.Slug, acct.DisplayName, acct.AccountType, acct.OwnerUserID)
 	var pgErr *pgconn.PgError
-	if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation {
+	if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation && pgErr.ConstraintName == accountsSlugUniqueConstraint {
 		return ErrSlugTaken
 	}
 	return err
@@ -159,6 +173,77 @@ func (r *pgxRepository) CreateProfile(ctx context.Context, p AccountProfile) err
 		VALUES ($1, $2, $3, $4)
 	`, p.AccountID, p.OwnerName, p.LoginEmail, p.ProfileSetupComplete)
 	return err
+}
+
+// ProvisionDefaultWorkspace creates acct, membership, and profile inside one
+// transaction guarded by a pg_advisory_xact_lock keyed on membership.UserID.
+// Two concurrent calls for the same viewer serialize on that lock — the
+// second does not even start its re-check until the first has committed or
+// rolled back — so the TOCTOU window a bare "list memberships, then insert"
+// leaves open (a second Server Component request for a brand-new viewer
+// deciding independently that no workspace exists yet) cannot produce two
+// personal workspaces for one viewer. It can still return ErrSlugTaken: that
+// signals a *different* viewer's display name collapsed to the same slug
+// (buildSlug has no per-user uniqueifier), which this lock does not
+// serialize against (different key) and provisionDefaultWorkspace in
+// service.go handles by retrying with a de-duplicated slug.
+func (r *pgxRepository) ProvisionDefaultWorkspace(ctx context.Context, acct Account, membership Membership, profile AccountProfile) (existingAccountID uuid.UUID, wonElsewhere bool, err error) {
+	tx, err := r.pool.Begin(ctx)
+	if err != nil {
+		return uuid.Nil, false, fmt.Errorf("accounts: begin provisioning tx: %w", err)
+	}
+	defer tx.Rollback(ctx)
+
+	if _, err := tx.Exec(ctx, `SELECT pg_advisory_xact_lock(hashtext($1)::int8)`, membership.UserID.String()); err != nil {
+		return uuid.Nil, false, fmt.Errorf("accounts: acquire provisioning lock: %w", err)
+	}
+
+	// Re-check now that the lock is held: a concurrent transaction for this
+	// same viewer may have committed and released the lock while this one was
+	// waiting to acquire it.
+	var winnerAccountID uuid.UUID
+	err = tx.QueryRow(ctx, `
+		SELECT account_id FROM public.account_memberships
+		WHERE user_id = $1 AND status = $2
+		ORDER BY created_at ASC, id ASC
+		LIMIT 1
+	`, membership.UserID, StatusActive).Scan(&winnerAccountID)
+	if err == nil {
+		return winnerAccountID, true, nil
+	}
+	if !errors.Is(err, pgx.ErrNoRows) {
+		return uuid.Nil, false, fmt.Errorf("accounts: check active membership under lock: %w", err)
+	}
+
+	if _, err := tx.Exec(ctx, `
+		INSERT INTO public.accounts (id, slug, display_name, account_type, owner_user_id)
+		VALUES ($1, $2, $3, $4, $5)
+	`, acct.ID, acct.Slug, acct.DisplayName, acct.AccountType, acct.OwnerUserID); err != nil {
+		var pgErr *pgconn.PgError
+		if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation && pgErr.ConstraintName == accountsSlugUniqueConstraint {
+			return uuid.Nil, false, ErrSlugTaken
+		}
+		return uuid.Nil, false, fmt.Errorf("accounts: create default account: %w", err)
+	}
+
+	if _, err := tx.Exec(ctx, `
+		INSERT INTO public.account_memberships (id, account_id, user_id, role, status)
+		VALUES ($1, $2, $3, $4, $5)
+	`, membership.ID, membership.AccountID, membership.UserID, membership.Role, membership.Status); err != nil {
+		return uuid.Nil, false, fmt.Errorf("accounts: create owner membership: %w", err)
+	}
+
+	if _, err := tx.Exec(ctx, `
+		INSERT INTO public.account_profiles (account_id, owner_name, login_email, profile_setup_complete)
+		VALUES ($1, $2, $3, $4)
+	`, profile.AccountID, profile.OwnerName, profile.LoginEmail, profile.ProfileSetupComplete); err != nil {
+		return uuid.Nil, false, fmt.Errorf("accounts: create account profile: %w", err)
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		return uuid.Nil, false, fmt.Errorf("accounts: commit provisioning tx: %w", err)
+	}
+	return acct.ID, false, nil
 }
 
 func (r *pgxRepository) GetAccountByID(ctx context.Context, id uuid.UUID) (*Account, error) {
--- a/apps/control-plane/internal/accounts/service.go
+++ b/apps/control-plane/internal/accounts/service.go
@@ -166,22 +166,65 @@ func (s *Service) EnsureViewerContext(ctx context.Context, viewer auth.Viewer, r
 	}, nil
 }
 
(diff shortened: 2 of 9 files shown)


apps/control-plane/internal/accounts/service.go
```diff
@@ -166,22 +166,65 @@ func (s *Service) EnsureViewerContext(ctx context.Context, viewer auth.Viewer, r
...
+		//     caller re-lists again right after this returns and picks it up.
+		//  2. A genuinely different viewer's display name collapsed to the
+		//     same slug. This viewer still has no active membership, so fall
+		//     through and retry with a de-duplicated slug instead of failing
+		//     their first-ever sign-in.
+		memberships, listErr := s.repo.ListMembershipsByUserID(ctx, viewer.UserID)
+		if listErr != nil {
+			return fmt.Errorf("accounts: list memberships after slug collision: %w", listErr)
+		}
+		if len(activeMemberships(memberships)) > 0 {
```
sakibsadmanshajib: **Race is narrowed, not closed.** This recovery check cannot tell apart the two collision causes it claims to distinguish, in one specific timing window, and confirmed this is not just theoretical by inspection of the surrounding code (not just the PR's own tests).

`CreateAccount` and `CreateMembership` are two separate, non-transactional statements (lines 196 and the post-loop call at ~236). Consider two concurrent requests for the *same* new viewer, A and B, both racing on `provisionDefaultWorkspace`:

1. A's `CreateAccount` commits first (wins the slug).
2. Before A calls `CreateMembership`, B's `CreateAccount` fails with `ErrSlugTaken` and reaches this line.
3. B calls `ListMembershipsByUserID(viewer.UserID)` here. A hasn't created the membership row yet, so this returns zero active memberships.
4. B cannot distinguish this from a genuine different-viewer collision, so it falls through to the retry branch and creates a **second** personal account + membership + profile for the same viewer with a suffixed slug.
5. A then proceeds and successfully creates its own membership on the *original* account.

End state: one viewer, two personal workspaces, both active. That's a data-integrity bug the PR's own regression tests can't catch, because `simulateSelfRace` in the test (service_test.go:59-65) hard-codes the winner's account *and* membership as already committed atomically as a side effect of the failed `CreateAccount` call — which is not how the two real statements actually race.

The window here is small (two back-to-back local INSERTs vs. a full HTTP round trip for the Next.js double-call), so this is a real improvement over the status quo, just not a complete fix of the class of bug the PR title/summary claims ("idempotency guard"). Wrapping `CreateAccount` + `CreateMembership` in a single DB transaction would close this properly, since then a concurrent reader would see either both rows or neither. Worth a decision: land as a narrower-but-real improvement (documented explicitly in the PR description as such), or close the window with a transaction before merging, given this gates a security-fix PR (#960) and future readers will read the PR title's "recover ... from a race" as a complete fix.

apps/control-plane/internal/accounts/repository.go
```diff
@@ -115,11 +115,22 @@ func (r *pgxRepository) ActiveTenantID(ctx context.Context, userID uuid.UUID) (u
...
+// in service.go for how each is told apart and recovered — without this
+// translation the raw pgx unique-violation escaped as an opaque
+// "could not load your workspace" 500.
 func (r *pgxRepository) CreateAccount(ctx context.Context, acct Account) error {
 	_, err := r.pool.Exec(ctx, `
 		INSERT INTO public.accounts (id, slug, display_name, account_type, owner_user_id)
 		VALUES ($1, $2, $3, $4, $5)
 	`, acct.ID, acct.Slug, acct.DisplayName, acct.AccountType, acct.OwnerUserID)
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation {
```
sakibsadmanshajib: Minor, pre-existing pattern (mirrors `CreateMembership` below) rather than new in this PR, but worth tightening while this exact translation is being touched: this checks `pgErr.Code == pgerrcode.UniqueViolation` but not `pgErr.ConstraintName`. Any unique violation on the `accounts` insert (e.g. a future unique index, or the `id` PK in the near-impossible case of a UUID collision) gets mapped to `ErrSlugTaken` and silently routed into the slug-retry loop in `service.go`, which is the wrong recovery for a non-slug violation and would mask what's actually wrong. Suggest `pgErr.ConstraintName == "accounts_slug_key"` (or whatever the real constraint name is) alongside the code check, same as I'd flag on `CreateMembership` if it were in scope here.

apps/control-plane/internal/accounts/service_test.go
```diff
@@ -48,6 +57,17 @@ func (s *stubRepo) ActiveTenantID(_ context.Context, userID uuid.UUID) (uuid.UUI
 }
 
 func (s *stubRepo) CreateAccount(_ context.Context, acct accounts.Account) error {
+	if s.simulateSelfRace {
+		s.simulateSelfRace = false
+		s.accountsMap[s.raceAccount.ID] = s.raceAccount
+		s.memberships = append(s.memberships, *s.raceMembership)
```
sakibsadmanshajib: Verified independently (reverted the `provisionDefaultWorkspace` retry/recovery logic back to the pre-fix single-attempt version in a throwaway worktree, kept `ErrSlugTaken`/`CreateAccount` translation): both new tests in this file correctly go red against the reverted code (`EnsureViewerContext error: accounts: create default account: accounts: workspace slug already taken`), so they are real regression tests, not tautologies. Good.

One gap though: this stub writes `raceAccount` **and** `raceMembership` together, atomically, as the side effect of the single failed `CreateAccount` call. That models the race as "the winner is always fully done (account + membership) by the time the loser collides," which is the easy case. It doesn't exercise the harder case flagged on service.go:222, where the winner's `CreateAccount` has committed but its `CreateMembership` hasn't yet — the actual timing window the real two-statement, non-transactional write pattern leaves open. A stub that lets `raceMembership` land on a *second*, later call (or omits it entirely to model "winner mid-flight") would be a stronger test and should currently fail against this PR's code, which is exactly the point.

sakibsadmanshajib:
## Adversarial review — provisioning race fix

Reviewed the pushed diff (`gh pr diff 963`, head `a2afb1a5`), not the author's summary. Three inline findings posted. Summary by axis:

**1. Race closed vs. narrowed: narrowed, not closed.** `CreateAccount` and `CreateMembership` are separate, non-transactional statements. If the collision is hit *between* the winner's `CreateAccount` commit and its `CreateMembership` commit, the loser's recovery check (`ListMembershipsByUserID`) still sees zero active memberships, cannot distinguish that from a genuine different-viewer collision, and creates a second personal workspace for the same viewer instead of recovering. See inline comment on service.go:222. Real improvement over the status quo, not a complete fix of the race class the title claims.

**2. Cross-tenant takeover shape: not present, checked closely since this was flagged as the most likely place for a real bug.** The self-race recovery resolves by re-querying `ListMembershipsByUserID(viewer.UserID)`, i.e. filtered on the *viewer's own* user id, never by slug or by the colliding account's identity. There is no code path where a viewer is attached to a different, unrelated tenant's account. The different-viewer collision branch correctly falls through to create a brand-new account with a de-duplicated slug rather than reusing the colliding one. Clean on this axis.

**3. Error handling breadth:** `CreateAccount`'s translation checks `pgErr.Code == UniqueViolation` but not `pgErr.ConstraintName`, so any future unique violation on `accounts` (not just the slug one) would be misrouted into the slug-retry loop. Pre-existing pattern (mirrors `CreateMembership`), not new here, but worth tightening while this exact spot is being touched. See inline comment on repository.go:131.

**4. Fails closed:** yes. After `maxSlugAttempts` (3) exhausted retries, provisioning returns a real error rather than a half-provisioned session; no silent wrong-state.

**5. Test quality — verified, not just read.** Checked out the PR branch into a throwaway worktree, reverted `provisionDefaultWorkspace` to the pre-fix single-attempt logic (kept `ErrSlugTaken`), and reran both new tests via the toolchain container: both go red as expected (`accounts: workspace slug already taken`), so they are genuine regression tests, not tautologies. Gap: the self-race stub (service_test.go:59-65) hard-codes the winner's account+membership as committed together atomically, so it cannot exercise the narrower race in finding 1. See inline comment on service_test.go:63.

Not fixing or merging per my brief; findings are for the author to address or rebut.

apps/control-plane/internal/accounts/service.go
sakibsadmanshajib: Fixed by wrapping the whole provisioning write in one transaction: new commit adds accounts.Repository.ProvisionDefaultWorkspace, which does the account + membership + profile inserts inside a single pg transaction guarded by pg_advisory_xact_lock(hashtext(userID)::int8) (same pattern already used in rag/provision.go). A concurrent request for the same viewer now blocks on the lock until the winner's transaction fully commits or rolls back, then re-checks account_memberships under the lock — there is no more window where the loser can observe an account row without its membership, because both only ever become visible together, atomically. provisionDefaultWorkspace in service.go now calls this single method instead of the three separate Create* calls. This is a real fix of the race class, not a documented narrower improvement.

apps/control-plane/internal/accounts/repository.go
sakibsadmanshajib: Tightened in the new commit: both CreateAccount and the new ProvisionDefaultWorkspace now check pgErr.ConstraintName == accountsSlugUniqueConstraint ("accounts_slug_key", the Postgres default name for the migration's column-level `slug ... unique`) in addition to the SQLSTATE 23505 code, so any other future unique violation on accounts is never misreported as a slug collision.

apps/control-plane/internal/accounts/service_test.go
sakibsadmanshajib: Addressed by construction, not by improving this test's fidelity to the old bug: since ProvisionDefaultWorkspace is now one atomic transaction (see reply on service.go), there is no longer an intermediate state where a winner's account exists without its membership — production itself never produces that state anymore, so a test modeling it would be testing something that can no longer happen. The rewritten stub instead models the correct atomic contract: the outer, unlocked ListMembershipsByUserID check in EnsureViewerContext still sees nothing for this viewer (matching the real pre-lock TOCTOU window), and the winner's account+membership become visible only as the return value of the single ProvisionDefaultWorkspace call, exactly mirroring what the real lock-scoped re-check would see. Also added an explicit len(repo.accountsMap) == 1 assertion after recovery, the direct anti-duplicate-workspace check you asked for.

sakibsadmanshajib:
## Update: race closed with an atomic transaction, not just narrowed

Follow-up commit b08aed5cf pushed. Answering the coordinator's three asks directly:

**Which path did I take, and why.** The repository layer already exposed a transaction boundary (`pgxpool.Pool.Begin`/`pgx.Tx`) and this exact pattern — a transaction-scoped `pg_advisory_xact_lock` guarding a critical section — was already precedented in this codebase (`rag/provision.go`, and the session-scoped sibling in `accounting/pglock.go`). So this stayed the "small, contained change" the time-box allowed, not a Repository interface restructuring. I did not need to land the narrower version and open a follow-up issue.

**What changed.** `accounts.Repository` gains `ProvisionDefaultWorkspace(ctx, acct, membership, profile) (existingAccountID uuid.UUID, wonElsewhere bool, err error)`. It creates all three rows (account, owner membership, profile) inside one transaction guarded by `pg_advisory_xact_lock(hashtext(userID)::int8)`. A second concurrent call for the same viewer blocks on the lock until the first transaction fully commits or rolls back, then re-checks `account_memberships` under the lock. There is no more window between "check" and "insert" for a second personal workspace to slip through, because the account and its membership only ever become visible together, atomically. `provisionDefaultWorkspace` in `service.go` now calls this one method instead of the three separate `Create*` calls.

**Cross-tenant confirmation, restated for the record.** Recovery resolves exclusively through `ListMembershipsByUserID(viewer.UserID)`, filtered on the viewer's own user ID. The slug is used only to *detect* a collision (via `ErrSlugTaken`), never to *resolve* one — a genuinely different viewer whose display name collapses to the same slug still gets their own account via the de-duplicated-slug retry, never the other viewer's.

**The two smaller items.** Both done: `CreateAccount` and the new `ProvisionDefaultWorkspace` now also match `pgErr.ConstraintName == "accounts_slug_key"`, not just the SQLSTATE code. The self-race regression test now exercises the stub's `ProvisionDefaultWorkspace` directly and asserts `len(accountsMap) == 1` after recovery — the direct anti-duplicate-workspace check. (The specific "account committed, membership not yet" window the earlier stub couldn't reach no longer exists in production either, now that the write is one transaction, so there's nothing left for a test to model there.)

**Verification.** `go build`, `go vet`, and `go test` all green across `apps/control-plane/internal/{accounts,ledger,usage,profiles,budgets,accounting}` (the last five needed a matching trivial `ProvisionDefaultWorkspace` on their local `accounts.Repository` fakes to keep compiling; none of them exercise the race). All three prior review threads replied to and resolved.

Issue #969, opened 2026-08-18
No live model-availability canary: dead alias found by someone editing a README


sakibsadmanshajib:
## What happened

`hive-fast` silently 404'd with `model_not_found` against every real customer
request for some period before anyone noticed (issue #965). It was found by
an agent editing `docs/anthropic-sdk-integration.md`/the README, who happened
to try the example command and hit the failure, not by any monitoring.

## Why nothing caught it

- `GET /health` on `api-hive.scubed.co` and `control-hive.scubed.co` returns
  200 regardless: it is a shallow liveness check and never exercises routing
  or reaches a provider. It stayed green for the entire time `hive-fast` was
  dead, and separately stayed green through an unrelated total gateway
  outage found in the same investigation (control-plane authz/DB-pool
  degradation, `/v1/chat/completions` and `/v1/models` failing for every
  alias including `hive-fast`, `hive-default`, `hive-auto`), so `/health`
  cannot distinguish "fully working" from "100% of real requests failing."
- No workflow under `.github/workflows/` runs a scheduled request against a
  real provider model for any alias. `deploy-demo-box.yml` has an "Assert
  model catalog prices agree with the model LiteLLM will call" step, but it
  only compares the DB's `provider_routes.provider_model` string against
  LiteLLM's own `/model/info`, both of which can agree with each other while
  both are stale relative to what the upstream provider (Groq, OpenRouter,
  etc.) currently actually serves. It cannot detect a provider decommissioning
  a model id, which is exactly what happened here (`llama-3.1-8b-instant`
  removed from Groq's live catalog between 2026-08-03 and 2026-08-18,
  confirmed against `https://api.groq.com/openai/v1/models` with a real key).
- Provider catalogs drift on the provider's own schedule, independent of any
  deploy or migration on this side, so nothing in this repo's CI or deploy
  pipeline would ever re-check a previously-working alias unless something
  in this repo changed.

## Suggested shape (not being built now; owner chose not to prioritize this pass)

A scheduled job (e.g. hourly/daily GitHub Actions cron) that sends one
minimal real completion through every enabled, token-priced alias against
the live gateway (or directly against each configured provider/model pair)
and alerts on the first non-2xx or on a response shape indicating
`model_not_found`/decommission, rather than relying on someone tripping over
a dead route while doing unrelated work.

## Evidence

- Issue #965 (this alias's actual repro).
- Live Groq catalog fetch 2026-08-18 (`GET /v1/models` with a real
[...]

Issue #713, opened 2026-08-04
deploy-demo-box's model-price assertion goes circular once an admin config sync targets a DB-managed route


sakibsadmanshajib:
## Context

PR #711 (fixing #689) added a deploy-time assertion in `deploy-demo-box.yml`, "Assert model catalog prices agree with the model LiteLLM will call". It compares `provider_routes.provider_model` (the DB side) against LiteLLM's live `GET /model/info` (`litellm_params.model`, the resolved value LiteLLM will actually send upstream), and fails the deploy on any mismatch.

That comparison is only real because, on a freshly recreated box, the live LiteLLM config is still the first-boot seed copied from `deploy/litellm/config.yaml`, which resolves each DB-managed route's model from an `os.environ/VAR` template. The env var and `provider_routes.provider_model` are two genuinely independent mechanisms today, which is exactly why comparing them catches drift.

## The gap

`apps/control-plane/internal/litellmconfig`'s config sync (issue #701, fixed by #705; #707 tracks remaining preservation gaps) writes a DB-managed route's `litellm_params.model` directly from `provider_routes.provider_model` -- the same column PR #711's assertion reads. After an admin triggers that sync against a DB-managed route (`route-openrouter-default`, `route-openrouter-auto`, the embedding routes), the live config and the DB value are no longer independent: both trace back to the identical `provider_routes` row. PR #711's assertion cannot fail after that point even if the underlying env-var mechanism it was designed to catch is wrong, because there is no longer a second, independently-sourced value to disagree with it.

The `api_base` canonicalization PR #711 added (comparing `litellm_params.api_base` in addition to `litellm_params.model`, so an OpenRouter-adapter swap to a different endpoint under the same model name still fails) narrows this, but does not restore independence to the model-string comparison itself: a sync-driven config still gets both its model and its `api_base` from `provider_routes`, so both sides of the assertion read from the same row.

## Why this is not fixed in #711

Restoring independence needs either:
- a check against something outside `provider_routes` entirely (e.g. the actual container environment via `docker exec ... printenv`, cross-checked against the `os.environ/VAR` name each `deploy/litellm/config.yaml` route names, for routes that are still env-var-driven), or
- tracking whether the box has ever run a sync against a given route, and switching that route's assertion mode (or skipping it) once it has, or
- a design decision on whether the sync-driven path should be trusted as its own source of truth going forward, separate from the deploy-time assertion's job.

[...]

Issue #684, opened 2026-08-01
LiteLLM's repo config seed never reaches the running container after first boot, and the deploy still passes


sakibsadmanshajib:
## Summary

`deploy/litellm/config.yaml` in the repo is not the file the running LiteLLM reads, and nothing in the deploy makes an edit to it take effect. It is the same silent-staleness family as the Caddyfile defect in #619, found while fixing that one, but the mechanism and the fix are different, so it is filed separately rather than folded in.

## Evidence, measured on the demo box

`deploy/docker/docker-compose.yml` mounts the repo file read-only at `/seed/config.yaml` and copies it into the shared `litellm-config` volume only when the volume has no config yet:

```
if [ ! -f /etc/litellm/config.yaml ]; then
  cp /seed/config.yaml /etc/litellm/config.yaml
fi
exec litellm --config=/etc/litellm/config.yaml
```

On the box right now:

| file | md5 |
| --- | --- |
| `/etc/litellm/config.yaml` (what LiteLLM reads) | `cc8a9e2072813e8cc200da604451e3e9` |
| `/seed/config.yaml` (the repo file, mounted) | `e9fcfed9397f209998de05d0d31e5705` |
| `deploy/litellm/config.yaml` in the box checkout | `e9fcfed9397f209998de05d0d31e5705` |

The volume copy has diverged from the seed, as designed: control-plane generates the live config and writes it into that volume via `POST /internal/litellm/sync`. The consequence is that once the volume exists, which it has since the box was first brought up, an edit to `deploy/litellm/config.yaml` on `main` reaches nothing. The deploy still reports success.

Second, smaller hazard from the same family: `/seed/config.yaml` is a single-FILE bind mount, so it is pinned to the inode it had at container start, and `git pull` replaces files by rename. The seed the container can see is therefore not guaranteed to be the seed in the checkout either. They happen to match today (both `e9fcfed9`), because the litellm container was recreated after the last change to that file, but that is luck rather than a property of the deploy.

## Why this is not fixed in the #619 change

The Caddy fix is a validate plus reload or recreate of a proxy whose bind-mounted file is authoritative. Here the bind-mounted file is not authoritative: control-plane's generated config is. A correct fix has to decide what the intended source of truth is per key, and probably means one of:

* deploy triggers control-plane's LiteLLM sync so the generated config is regenerated from the database after every deploy, and the repo seed is documented as bootstrap-only, or
[...]

PR #970, opened 2026-08-18
fix: revert hive-fast to a live Groq model, collapse routing-internals leak

### LINKED ISSUE #965: hive-fast routes to a deprecated Groq model and leaks routing internals on failure
## What's broken

`hive-fast` currently 404s on both `/v1/messages` and `/v1/chat/completions`.
Reproduced live 2026-08-18 against `api-hive.scubed.co` with a real API key:

```
curl https://api-hive.scubed.co/v1/chat/completions \
  -H "Authorization: Bearer <API_KEY>" -H "content-type: application/json" \
  -d '{"model":"hive-fast","messages":[{"role":"user","content":"hi"}],"max_tokens":8}'
```

Response body (truncated):

```
{"error":{"message":"upstream error - {\"error\":{\"message\":\"The model
`llama-3.1-8b-instant` does not exist or you do not have access to it.\",
\"type\":\"invalid_request_error\",\"code\":\"model_not_found\"}} No fallback
model group found for original model_group=hive-fast.
Fallbacks=[{'hive-fast': ['hive-fast']}, {'hive-fast': ['hive-fast']},
{'hive-fast': ['hive-fast']}]. ... upstream provider Retried: 3 times"}}
```

`hive-default` and `hive-auto` both work correctly against the same key at
the same time, so this is isolated to `hive-fast`'s routing.

## Root cause

`hive-fast`'s live route maps to Groq's `llama-3.1-8b-instant`
(`supabase/migrations/20260801_14_route_groq_fast_cheapest_model.sql` set
this mapping on 2026-08-01 as the then-cheapest Groq model). Groq now answers
that model id with `model_not_found`, i.e. Groq has deprecated or removed it
since that migration ran. This is a live provider-catalog drift, not a code
bug: the routing table needs a DB update to point `hive-fast` at a Groq
model that currently exists, or the model-picker logic needs to re-derive
"cheapest" dynamically instead of pinning a specific model id in a
migration.

## Secondary issue: the failure leaks internal routing detail

The forwarded error message includes LiteLLM's internal fallback-group
bookkeeping (`Fallbacks=[{'hive-fast': ['hive-fast']}, ...]`, retry counts)
verbatim. No literal provider name appears (so the existing provider-name
sanitizer doesn't flag it), but "LiteLLM fallback group" and "Retried: 3
times" are still internal implementation detail no customer-facing error
should carry. Worth deciding whether the provider-blind sanitizer should
also collapse this shape (any upstream_error / model_not_found body) to a
generic message, independent of fixing the routing itself.

## Impact

- `docs/anthropic-sdk-integration.md` and the open README refresh (PR #962)
  both currently show `hive-fast` as the example model in copy-pasteable
  code; anyone following either literally hits this. PR #964 corrects the
  doc to use `hive-default` instead and documents this as a known
  limitation; #962 needs the same fix, flagged there directly.

## Suggested fix

Update the live routing row (or whatever admin/catalog path manages it) to
point `hive-fast` at a Groq model that currently exists, and re-verify with
a real request on both `/v1/messages` and `/v1/chat/completions`. Separately,
consider whether `anthropicErrorType`/the OpenAI-side sanitizer should
collapse any `upstream_error`/`model_not_found` body to a fixed message
rather than forwarding LiteLLM's raw fallback-group text.

sakibsadmanshajib:
Fixes #965.

## What's broken

`hive-fast` 404s on every request with `model_not_found`. Groq removed
`llama-3.1-8b-instant` from its live catalog after
`20260801_14_route_groq_fast_cheapest_model.sql` pinned `route-groq-fast` to
it on 2026-08-01. Confirmed live 2026-08-18 against
`GET https://api.groq.com/openai/v1/models` with a real `GROQ_API_KEY`: the
model id is absent from the response entirely. Groq's current chat-capable
catalog has no small Llama model at all.

Secondary issue #965 also flagged: the failure forwards LiteLLM's own
fallback-group bookkeeping verbatim to the customer (`No fallback model
group found... Fallbacks=[{'hive-fast': ['hive-fast']}, ...]... Retried: 3
times`). No literal provider name appears, so the existing provider-name
sanitizer (`apps/edge-api/internal/errors/provider_blind.go`) never caught
it, but it's still internal routing detail with no place in a customer-facing
error.

## Fix

**1. Repoint + reprice (`supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql`)**

One alias maps to exactly one route and one price; both are updated
together so nothing silently changes what customers are charged.

| | Before (dead) | After (live, confirmed 2026-08-18) |
|---|---|---|
| `route-groq-fast.provider_model` | `groq/llama-3.1-8b-instant` | `groq/openai/gpt-oss-20b` |
| `hive-fast` input credits/M | 7000 | 10500 |
| `hive-fast` output credits/M | 11200 | 42000 |

`openai/gpt-oss-20b` is the model `route-groq-fast` pointed at *before* the
2026-08-01 cost migration, and the closest fast/cheap chat model Groq still
serves live today (`groq/compound[-mini]` are agentic search products,
`gpt-oss-120b` is the flagship reasoning tier, `qwen/qwen3.6-27b` is a
general-purpose model with no established fast-tier track record here).
Its published rate ($0.075 in / $0.30 out per million, re-checked against
`groq.com/pricing` and `console.groq.com/docs/models` 2026-08-18) is
unchanged since this repo last priced it, so the reprice is a straight
revert to the figures `20260801_01_alias_pricing_correction.sql` already
derived, not a fresh guess. Full derivation and a documented, deliberately
accepted trade-off (gpt-oss-20b supports tool use but not *parallel* tool
calls, unlike the now-dead llama-3.1-8b-instant; no currently-live Groq
model in this tier has both) are in the migration file's own comment block.

**Also reverted:** `.env.example` and `.github/workflows/owui-nightly.yml`'s
hardcoded `GROQ_FAST_MODEL` default back to `groq/openai/gpt-oss-20b`. The
GitHub Actions repo variable `GROQ_FAST_MODEL` used by `ci.yml` and
`agent-visual-proof.yml` was already `groq/openai/gpt-oss-20b` (unchanged
since 2026-04-24) and needed no change.

**Operational step required on the demo box, not done by this PR** (same
shape as the original 2026-08-01 migration's own PR): the running gateway
resolves `route-groq-fast`'s upstream model from the box's own `.env`
`GROQ_FAST_MODEL` value, not from this catalog
(`deploy/litellm/config.yaml`'s `model: os.environ/GROQ_FAST_MODEL`; issue
#684). This migration alone does not make the box serve the new model.
After merge, the box's `.env` needs
`GROQ_FAST_MODEL=groq/openai/gpt-oss-20b` and the `litellm` container needs
a **recreate** (not restart) for the change to take effect, or the next
`deploy-demo-box.yml` run needs to already carry that `.env` value. Note:
`apps/control-plane/internal/litellmconfig` can now generate LiteLLM's
config directly from `provider_routes.provider_model` (issues #701/#705),
but per `deploy-demo-box.yml`'s own "Assert model catalog prices agree"
step comment, that sync is not yet wired into the automatic deploy path
(issue #713 tracks the remaining gap) — today the box still boots from the
static `os.environ/GROQ_FAST_MODEL` template, genuinely independent of this
migration. If the box's `.env` is not updated, `deploy-demo-box.yml`'s own
"Assert model catalog prices agree with the model LiteLLM will call" step
will fail loudly and block the deploy rather than silently serve a mismatch.

**2. Provider-blind leak fix (`apps/edge-api/internal/errors/provider_blind.go`)**

Added `providerBlindLooksLikeRoutingInternals()`, checked in
`sanitizeProviderBlindMessage` alongside the existing transport/auth/rate-limit
checks. On any message containing LiteLLM's fallback-group bookkeeping
(`fallback model group`, `model_group=`, `fallbacks=[`, `retried:`), the
whole message collapses to a fixed `"<resource> is not available."` rather
than trying to regex-scrub an open-ended, LiteLLM-versioned bookkeeping
format field by field. This is the single shared function every chat
completions/streaming/dispatch call site already routes error text through
(`inference/orchestrator.go`, `inference/stream.go`,
`inference/stream_responses.go`, `chat/dispatch.go`, plus the images/audio/RAG
handlers), so the fix applies to every caller at once, not just hive-fast's
path.

## Sibling alias check (per instruction, before fixing)

Enumerated the live catalog's aliases from migrations:
`hive-fast, hive-default, hive-auto, hive-embedding-default, hive-stt, hive-tts`.
`hive-default` and `hive-auto` were confirmed working at the time issue #965
was filed (2026-08-18), isolating the failure to `hive-fast`'s routing only.
**Separately**, at the time of my own verification pass a few hours later,
`/v1/chat/completions` and `/v1/models` were timing out for every alias
including `hive-default` and `hive-auto`, consistent with a known,
already-logged shared-DB-pool/authz degradation (unrelated to this fix,
not touched by this PR, and the reason live proof-of-fix below is pending).
Filed the "why did nothing catch this" gap as #969 (evidence only, not
built, per explicit instruction not to add a catalog health check in this
pass).

## Test plan

- [x] `TestWriteProviderBlindUpstreamErrorHidesFallbackRoutingInternals`
      (new): reproduces issue #965's exact leaked shape, confirmed RED
      before the fix, GREEN after.
- [x] Full `apps/edge-api/internal/errors` suite green, no regressions
      (`go test ./apps/edge-api/internal/errors/... -v -count=1`).
- [x] `apps/control-plane/internal/routing/catalog_pricing_integration_test.go`'s
      `TestHiveFastIsPinnedToGroqAtCorrectedPrice` updated to the reverted
      route/price and confirmed green in this PR's own CI run (it was red
      against the old expectations before this fix, see resolved review
      thread).
- [x] All CI checks green: `Go tests (control-plane)`, `Go tests (edge-api)`,
      `apply-migrations.sh against a throwaway Postgres` (validates the new
      migration's transaction wrapping applies cleanly), and every other
      required check.
- [x] Database review (mandatory: money/routing path). Ran, one real finding
      (missing transaction wrap) fixed in follow-up commit, thread resolved.
- [x] Security review (mandatory: customer-facing error content). Ran, one
      real finding (`retried:` false-positive risk) fixed in follow-up
      commit, thread resolved. Overall verdict: diff is security sound.
- [x] CodeRabbit CLI and `ecc:code-reviewer` streams also ran; all findings
      addressed, all 8 review threads on this PR resolved.
- [ ] Live proof that `hive-fast` answers correctly against the deployed
      gateway: blocked on the operational box step above (box's own
      `GROQ_FAST_MODEL` needs updating and `litellm` needs a recreate,
      neither of which this PR's merge does automatically). Also, the test
      API key used for this investigation was found revoked partway through
      this session (`invalid_api_key: API key is revoked`), independent of
      the box step, so no live call could be made with it regardless. Will
      post once merged, the box step is done, and a valid key is available,
      per the visual-proof-before-merge rule (no UI surface here, but the
      same "prove it against the real running system" standard applies to
      the alias actually answering).

Buglog entry (queued for the separate buglog-only PR per repo convention,
never appended on this branch):
```json
{"error_message":"hive-fast 404s with model_not_found; failure also leaked LiteLLM fallback-group bookkeeping","root_cause":"Groq decommissioned llama-3.1-8b-instant after the 2026-08-01 cost migration pinned route-groq-fast to it; separately, WriteProviderBlindUpstreamError's sanitizer only stripped literal provider names/classes, not LiteLLM's routing bookkeeping phrases (fallback model group, Fallbacks=[...], Retried: N times)","fix":"Reverted route-groq-fast to groq/openai/gpt-oss-20b (confirmed live) and hive-fast pricing to 10500/42000 credits; added providerBlindLooksLikeRoutingInternals() to collapse any message carrying LiteLLM fallback-group bookkeeping to a fixed generic message","tags":["routing","pricing","provider-blind","groq","catalog-drift"]}
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -158,7 +158,7 @@ OPENROUTER_EMBEDDING_FALLBACK_MODEL=openrouter/qwen/qwen3-embedding-8b
 # Groq free chat/reasoning models (llama-3.1-8b-instant primary, cheapest
 # Groq model as of 2026-08-03: $0.05 in / $0.08 out per million tokens vs
 # gpt-oss-20b's $0.075 / $0.30; 120b reasoning unchanged).
-GROQ_FAST_MODEL=groq/llama-3.1-8b-instant
+GROQ_FAST_MODEL=groq/openai/gpt-oss-20b
 GROQ_REASONING_MODEL=groq/openai/gpt-oss-120b
 
 # ─── RAG query embeddings (issue #232) ──────────────────────────────
--- a/.env.example
+++ b/.env.example
@@ -155,9 +155,13 @@ OPENROUTER_FAST_FALLBACK_MODEL=openrouter/free
 OPENROUTER_EMBEDDING_MODEL=openrouter/nvidia/llama-nemotron-embed-vl-1b-v2:free
 OPENROUTER_EMBEDDING_FALLBACK_MODEL=openrouter/qwen/qwen3-embedding-8b
 
-# Groq free chat/reasoning models (llama-3.1-8b-instant primary, cheapest
-# Groq model as of 2026-08-03: $0.05 in / $0.08 out per million tokens vs
-# gpt-oss-20b's $0.075 / $0.30; 120b reasoning unchanged).
+# Groq free chat/reasoning models (gpt-oss-20b primary, 120b reasoning).
+# Reverted 2026-08-18 (issue #965): the previously-pinned llama-3.1-8b-instant
+# was cheaper as of 2026-08-03 but Groq has since decommissioned it entirely,
+# so hive-fast 404'd on every request. gpt-oss-20b is the closest fast/cheap
+# model Groq still serves live; do not repoint this to a specific low-cost
+# model id again without re-checking it against Groq's own /v1/models first,
+# that is exactly what broke last time.
 GROQ_FAST_MODEL=groq/openai/gpt-oss-20b
 GROQ_REASONING_MODEL=groq/openai/gpt-oss-120b
 
--- a/apps/edge-api/internal/errors/provider_blind.go
+++ b/apps/edge-api/internal/errors/provider_blind.go
@@ -80,6 +80,16 @@ func sanitizeProviderBlindMessage(alias string, httpStatus int, raw string) stri
 	if providerBlindLooksLikeTransportFailure(lowerRaw) || providerBlindLooksLikeTransportFailure(lowerMessage) {
 		return fmt.Sprintf("%s is temporarily unavailable.", resourceLabel)
 	}
+	if providerBlindLooksLikeRoutingInternals(lowerRaw) || providerBlindLooksLikeRoutingInternals(lowerMessage) {
+		// LiteLLM's own fallback-group bookkeeping (issue #965): "No fallback
+		// model group found for original model_group=X. Fallbacks=[...].
+		// Retried: N times." Carries no provider name, so the regexes above
+		// never touch it, but it is still internal routing detail no
+		// customer-facing error should carry. Collapse the whole message
+		// rather than trying to scrub an open-ended, LiteLLM-versioned
+		// bookkeeping format field by field.
+		return fmt.Sprintf("%s is not available.", resourceLabel)
+	}
 	if providerBlindLooksLikeAuthFailure(httpStatus, lowerRaw) || providerBlindLooksLikeAuthFailure(httpStatus, lowerMessage) {
 		return fmt.Sprintf("%s request was rejected by the upstream provider.", resourceLabel)
 	}
@@ -197,6 +207,20 @@ func providerBlindLooksLikeTransportFailure(message string) bool {
 		"temporarily unavailable",
 		"no such host",
 		"econnrefused",
+	} {
+		if strings.Contains(message, token) {
+			return true
+		}
+	}
+	return false
+}
+
+func providerBlindLooksLikeRoutingInternals(message string) bool {
+	for _, token := range []string{
+		"fallback model group",
+		"model_group=",
+		"fallbacks=[",
+		"retried:",
 	} {
 		if strings.Contains(message, token) {
 			return true
--- a/apps/edge-api/internal/errors/provider_blind.go
+++ b/apps/edge-api/internal/errors/provider_blind.go
@@ -216,12 +216,13 @@ func providerBlindLooksLikeTransportFailure(message string) bool {
 }
 
 func providerBlindLooksLikeRoutingInternals(message string) bool {
-	for _, token := range []string{
-		"fallback model group",
-		"model_group=",
-		"fallbacks=[",
-		"retried:",
-	} {
+	// "retried: N times" deliberately excluded: it is ordinary English a
+	// legitimate upstream message could use on its own, so alone it risked
+	// collapsing safe error text (security review finding on this PR). The
+	// three tokens below are specific to LiteLLM's own bookkeeping
+	// vocabulary and already fully cover issue #965's reported leak shape
+	// without it.
+	for _, token := range []string{"fallback model group", "model_group=", "fallbacks=["} {
 		if strings.Contains(message, token) {
 			return true
 		}
--- a/apps/edge-api/internal/errors/provider_blind_test.go
+++ b/apps/edge-api/internal/errors/provider_blind_test.go
@@ -137,6 +137,36 @@ func TestWriteProviderBlindUpstreamErrorHidesInternalTransportDetails(t *testing
 	assertNoProviderLeak(t, resp.Error.Message)
 }
 
+func TestWriteProviderBlindUpstreamErrorHidesFallbackRoutingInternals(t *testing.T) {
+	w := httptest.NewRecorder()
+
+	// Reproduces issue #965: a model_not_found upstream response wrapped in
+	// LiteLLM's own fallback-group bookkeeping. No literal provider name
+	// appears in "No fallback model group found ... Fallbacks=[...] ...
+	// Retried: 3 times", so the provider-name regex alone never catches it;
+	// the message must still not reach the customer.
+	raw := `upstream error - {"error":{"message":"The model ` + "`llama-3.1-8b-instant`" + ` does not exist or you do not have access to it.","type":"invalid_request_error","code":"model_not_found"}} No fallback model group found for original model_group=hive-fast. Fallbacks=[{'hive-fast': ['hive-fast']}, {'hive-fast': ['hive-fast']}, {'hive-fast': ['hive-fast']}]. Upstream provider Retried: 3 times`
+
+	WriteProviderBlindUpstreamError(w, "hive-fast", http.StatusNotFound, raw)
+
+	resp := decodeOpenAIError(t, w)
+	assertNoProviderLeak(t, resp.Error.Message)
+	for _, forbidden := range []string{
+		"fallback model group",
+		"model_group=",
+		"fallbacks=",
+		"retried:",
+		"llama-3.1-8b-instant",
+	} {
+		if strings.Contains(strings.ToLower(resp.Error.Message), forbidden) {
+			t.Fatalf("expected routing internals stripped, found %q in %q", forbidden, resp.Error.Message)
+		}
+	}
+	if !strings.Contains(resp.Error.Message, "hive-fast") {
+		t.Fatalf("expected alias in sanitized message, got %q", resp.Error.Message)
+	}
+}
+
 func decodeOpenAIError(t *testing.T, w *httptest.ResponseRecorder) OpenAIError {
 	t.Helper()
 
--- a/apps/edge-api/internal/errors/provider_blind_test.go
+++ b/apps/edge-api/internal/errors/provider_blind_test.go
@@ -149,6 +149,10 @@ func TestWriteProviderBlindUpstreamErrorHidesFallbackRoutingInternals(t *testing
 
 	WriteProviderBlindUpstreamError(w, "hive-fast", http.StatusNotFound, raw)
 
+	if w.Code != http.StatusNotFound {
+		t.Fatalf("status = %d, want %d", w.Code, http.StatusNotFound)
+	}
+
 	resp := decodeOpenAIError(t, w)
 	assertNoProviderLeak(t, resp.Error.Message)
 	for _, forbidden := range []string{
@@ -162,8 +166,24 @@ func TestWriteProviderBlindUpstreamErrorHidesFallbackRoutingInternals(t *testing
 			t.Fatalf("expected routing internals stripped, found %q in %q", forbidden, resp.Error.Message)
 		}
 	}
-	if !strings.Contains(resp.Error.Message, "hive-fast") {
-		t.Fatalf("expected alias in sanitized message, got %q", resp.Error.Message)
+	if resp.Error.Message != "hive-fast is not available." {
+		t.Fatalf("message = %q, want %q", resp.Error.Message, "hive-fast is not available.")
+	}
+}
+
+func TestProviderBlindLooksLikeRoutingInternalsRequiresSpecificToken(t *testing.T) {
+	w := httptest.NewRecorder()
+
+	// "retried:" alone, with none of LiteLLM's specific bookkeeping tokens,
+	// must NOT trigger the routing-internals collapse: a legitimate upstream
+	// message can plausibly use ordinary English like this, and over-collapsing
+	// it would throw away real detail for no safety gain (security review
+	// finding on this PR).
+	WriteProviderBlindUpstreamError(w, "hive-default", http.StatusBadGateway, "request failed, retried: 3 times before giving up")
+
+	resp := decodeOpenAIError(t, w)
+	if resp.Error.Message != "request failed, retried: 3 times before giving up" {
+		t.Fatalf("expected message to pass through unchanged, got %q", resp.Error.Message)
 	}
 }
 
--- a/supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
+++ b/supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
@@ -68,6 +68,15 @@
 --   second run of this file affects zero rows and errors on nothing.
 -- =============================================================================
 
+-- Wrapped in a transaction (database review finding on this PR): these two
+-- UPDATEs enforce the "one route, one price" invariant together. Without
+-- BEGIN/COMMIT, apply-migrations.sh commits each statement independently, and
+-- a request landing on ListRouteCandidates/LoadAliasPricing between them
+-- could read the new route with the old price, or vice versa. Same gap
+-- pre-existed in 20260801_01/20260801_14; closed here since this migration
+-- is specifically the one guarding that invariant.
+BEGIN;
+
 -- 1. Point route-groq-fast back at the model Groq currently serves.
 UPDATE public.provider_routes
    SET provider_model = 'groq/openai/gpt-oss-20b'
@@ -81,3 +90,5 @@ UPDATE public.model_aliases
        updated_at           = now()
  WHERE alias_id = 'hive-fast'
    AND (input_price_credits <> 10500 OR output_price_credits <> 42000);
+
+COMMIT;
--- a/supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
+++ b/supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
@@ -1,0 +1,83 @@
+-- =============================================================================
+-- Issue #965 (owner-ruled fix, 2026-08-18): hive-fast currently routes to
+-- Groq's llama-3.1-8b-instant, which Groq has removed from its live catalog
+-- since the 20260801_14 migration pinned it. Every hive-fast request 404s
+-- with model_not_found. Confirmed live 2026-08-18 against
+-- https://api.groq.com/openai/v1/models (Authorization: Bearer $GROQ_API_KEY):
+-- llama-3.1-8b-instant is absent from the response entirely. Groq's current
+-- catalog carries no small/fast Llama chat model at all; the closest fast,
+-- cheap, currently-live alternative is openai/gpt-oss-20b, which is also
+-- exactly what route-groq-fast pointed at before 20260801_14 repointed it
+-- for a (now moot) cost saving, so this is a revert to a config already
+-- proven live in production, not a fresh guess.
+--
+-- PROVIDER RATE, re-fetched 2026-08-18
+--   Groq openai/gpt-oss-20b   $0.075 in / $0.30 out USD per million tokens
+--     source: https://groq.com/pricing ("GPT OSS 20B" row) and
+--     https://console.groq.com/docs/models, both agree, unchanged since the
+--     20260801_01 migration derived hive-fast's price from the same figures.
+--
+-- FORMULA (same mechanical shape as 20260801_01 and 20260801_14)
+--   credits_per_million = ceil(provider_list_usd_per_million * MARGIN * CREDITS_PER_USD)
+--     MARGIN          = 1.4
+--     CREDITS_PER_USD = 100000   (apps/control-plane/internal/payments/types.go)
+--
+-- WORKED DERIVATION
+--   hive-fast  in  0.075 * 1.4 * 100000 = 10500
+--   hive-fast  out 0.300 * 1.4 * 100000 = 42000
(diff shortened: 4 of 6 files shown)


apps/edge-api/internal/errors/provider_blind.go
```diff
@@ -205,6 +215,20 @@ func providerBlindLooksLikeTransportFailure(message string) bool {
 	return false
 }
 
+func providerBlindLooksLikeRoutingInternals(message string) bool {
```
sakibsadmanshajib, line 218: This closes the exact #965 reproduction, and centralizing it inside sanitizeProviderBlindMessage means every WriteProviderBlindUpstreamError call site (orchestrator, rag chat handler, audio handler, images handler, stt client, inference stream, chat dispatch, anthropic translate writer) gets the fix for free, which is the right place to put it.

The match list itself is narrow though: fallback model group, model_group=, fallbacks=[, retried:. These four substrings describe one specific LiteLLM router message, the fallback-group exhaustion case from #965. Other LiteLLM router bookkeeping strings I would expect to see in production, for example a no-healthy-deployment message, a cooldown-deployment list, or a router exception repr that does not happen to say fallback or model_group, would still fall through every existing regex above (none of them require a literal provider name) and reach the customer verbatim. That is the same class of leak as #965, just different wording.

I would not block on this since it correctly fixes the reported shape, but I would file a follow up to capture LiteLLM's other internal router message shapes with a broader heuristic, for example detecting the Python list-of-dict literal syntax these bookkeeping messages tend to use, rather than four literal substrings tied to today's exact wording.

apps/edge-api/internal/errors/provider_blind.go
```diff
@@ -205,6 +215,20 @@ func providerBlindLooksLikeTransportFailure(message string) bool {
...
+	for _, token := range []string{
+		"fallback model group",
+		"model_group=",
+		"fallbacks=[",
+		"retried:",
```
sakibsadmanshajib: retried: is the least anchored of the four tokens here. fallback model group, model_group=, and fallbacks=[ are all tied to LiteLLM's internal vocabulary and very unlikely to show up in a legitimate upstream message a customer should actually see. retried: on its own is plain English a provider could use unprompted, for example a rate limit or timeout message phrased as request was retried: 3 times before failing.

If that ever happens, this collapses an otherwise safe, informative message down to the generic X is not available, losing detail with no security upside, since the earlier transport-failure and rate-limit branches already run first and would catch most of those cases anyway. This fails safe rather than leaking anything, so I would not block on it, but I would tighten the token to something like exhausted retries: or require it to co-occur with model_group= or fallbacks=[ in the same message, since the real #965 leak always carries more than one of these together, never retried: in isolation.

apps/edge-api/internal/errors/provider_blind_test.go
```diff
@@ -137,6 +137,36 @@ func TestWriteProviderBlindUpstreamErrorHidesInternalTransportDetails(t *testing
...
+		"retried:",
+		"llama-3.1-8b-instant",
+	} {
+		if strings.Contains(strings.ToLower(resp.Error.Message), forbidden) {
+			t.Fatalf("expected routing internals stripped, found %q in %q", forbidden, resp.Error.Message)
+		}
+	}
+	if !strings.Contains(resp.Error.Message, "hive-fast") {
+		t.Fatalf("expected alias in sanitized message, got %q", resp.Error.Message)
+	}
```
sakibsadmanshajib, line 187: TestWriteProviderBlindUpstreamErrorHidesFallbackRoutingInternals only checks that the sanitized message excludes forbidden substrings and contains the alias name. It never asserts the HTTP status code or the exact message contract the way the sibling test above it does (lines 129-136). Given sanitizeProviderBlindMessage returns "hive-fast is not available." for this input, I would tighten this test to assert w.Code == http.StatusNotFound and resp.Error.Message == "hive-fast is not available." directly, keeping the forbidden-substring checks as a belt-and-suspenders guard against future regressions in the sanitizer. As written, the test would still pass if the handler returned the wrong status code or a differently worded but still leak-free message, which is a weaker guarantee than the fix is meant to lock in.

sakibsadmanshajib (commented):
CodeRabbit CLI review (coderabbit review --agent -t all --base main). One finding, posted inline.

sakibsadmanshajib (commented):
Database review of the migration file (schema/column/idempotency checks against provider_routes, model_aliases, and the current row state pulled from the prior three migrations). One finding below; everything else I checked (WHERE guard idempotency, provider_model/route_id and input_price_credits/output_price_credits column names and types, the claim that provider_capabilities/alias_route_policies/price_unit need no change, and the credits-per-million-tokens unit match against apps/edge-api/internal/metering/precedence.go) came back correct as written.

supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
```diff
@@ -0,0 +1,83 @@
...
+ WHERE route_id = 'route-groq-fast'
+   AND provider_model <> 'groq/openai/gpt-oss-20b';
+
+-- 2. Re-derive hive-fast's price from that model's published rate.
+UPDATE public.model_aliases
+   SET input_price_credits  = 10500,
+       output_price_credits = 42000,
+       updated_at           = now()
+ WHERE alias_id = 'hive-fast'
+   AND (input_price_credits <> 10500 OR output_price_credits <> 42000);
```
sakibsadmanshajib, line 92: These two UPDATE statements are not wrapped in an explicit transaction, and scripts/apply-migrations.sh hands each migration file to psql with no --single-transaction wrapping, so each statement commits on its own. Separately, the two hot-path reads that consume these tables per chat request are two independent, uncached queries (apps/control-plane/internal/routing/repository.go: ListRouteCandidates reads provider_routes/provider_capabilities, LoadAliasPricing reads model_aliases, in two separate pool.Query calls with no shared transaction and no caching layer in front of either). That combination means there is a real, if narrow, window during rollout where a concurrent request can read the new provider_model (gpt-oss-20b) together with the still-old price (7000/11200 for llama-3.1-8b-instant), or the reverse, since the file's own two writes aren't atomic with each other either.

This PR's own framing is that route and price for one alias must change together, never separately, which is exactly the property that only holds at the row level right now, not across the two tables during the moment of migration. Wrapping the two UPDATEs in an explicit BEGIN/COMMIT (the pattern already used in 20260801_10_tenants_personal_owner.sql for a similar concurrency concern) would make the route change and price change land as one atomic unit as far as any reader is concerned, closing the one part of this gap the migration itself can control. It would not need any change to the WHERE guards; both already correctly no-op on a second run.

This same gap already exists in 20260801_01_alias_pricing_correction.sql and 20260801_14_route_groq_fast_cheapest_model.sql, which also update these two tables in separate unwrapped statements, so it is a pre-existing pattern rather than something this PR introduces. Flagging it here since this migration is exactly the kind of change (revert of a decommissioned, 404ing model, on the hot chat path) where a request landing in the wrong half-second would either get billed at the wrong rate or momentarily see a route/price pair that never should have coexisted.

sakibsadmanshajib (commented):
I reviewed the migration, the provider-blind sanitizer changes, and the .env.example/CI workflow reverts. Two inline comments below, plus one finding that isn't anchorable inline because it sits in a file this PR does not touch.

apps/control-plane/internal/routing/catalog_pricing_integration_test.go, TestHiveFastIsPinnedToGroqAtCorrectedPrice (around lines 144 to 165), still asserts that hive-fast's provider_model equals groq/llama-3.1-8b-instant and that its price is 7000 input / 11200 output. That suite is compiled in with -tags integration and run against ROUTING_TEST_DB_URL in .github/workflows/ci.yml line 443, on a database with the full migration chain applied. After this PR merges, that chain includes 20260818_01_revert_hive_fast_groq_model_decommissioned.sql, which sets provider_model to groq/openai/gpt-oss-20b and the price to 10500/42000. The test's hardcoded expectations were not updated to match, so the next CI run against a freshly migrated database will fail this test. It needs the same treatment 20260801_14 gave it last time this route changed: new expected values and an updated derivation comment, in this PR.

For completeness, apps/control-plane/internal/routing/service_test.go's TestSelectRouteHiveFastResolvesToGroqAtGroqPrice and apps/edge-api/internal/metering/precedence_test.go's correctedHiveFast fixture also reference the old llama-3.1-8b-instant numbers, but both are self-contained unit tests with their own stub data rather than a real database, so they will keep passing. Their doc comments are just stale now, not broken.

.env.example
```diff
@@ -158,7 +158,7 @@ OPENROUTER_EMBEDDING_FALLBACK_MODEL=openrouter/qwen/qwen3-embedding-8b
 # Groq free chat/reasoning models (llama-3.1-8b-instant primary, cheapest
 # Groq model as of 2026-08-03: $0.05 in / $0.08 out per million tokens vs
 # gpt-oss-20b's $0.075 / $0.30; 120b reasoning unchanged).
-GROQ_FAST_MODEL=groq/llama-3.1-8b-instant
+GROQ_FAST_MODEL=groq/openai/gpt-oss-20b
```
sakibsadmanshajib, line 165: The comment block directly above this line (158 to 160) still says llama-3.1-8b-instant is the primary model and calls it out as the cheapest option, which is the exact reasoning that led to the 20260801_14 migration pinning hive-fast to the model Groq has since decommissioned. Now that the value on this line is back to gpt-oss-20b, that comment is describing a decommissioned model as the current choice and could lead someone to revert this back to llama-3.1-8b-instant for the same cost reason, recreating issue #965. Worth updating the comment to match the reverted value and note that llama-3.1-8b-instant is gone from Groq's catalog.

apps/edge-api/internal/errors/provider_blind.go
sakibsadmanshajib: This substring list is built from the one bookkeeping string quoted in issue #965 (fallback model group, model_group=, fallbacks=[, retried:). Any other LiteLLM router-internal message that leaks through this same code path but doesn't literally contain one of these four substrings (for example a no-healthy-deployment or cooldown-list message, or a future LiteLLM version that phrases retries as "Number of retries: N" instead of "Retried: N times") still reaches the customer verbatim, since it also carries no literal provider name for the existing regexes above to catch. The test only reproduces the exact reported shape. Not asking to block on this since the reported leak is fixed and the comment already flags the scrubbing approach as a tradeoff, but the coverage here is narrower than the fix's framing ("routing internals" generally) suggests, and a slightly wider net or a comment noting the intended scope would help the next person who hits a different LiteLLM message shape.

sakibsadmanshajib (commented):
Plain adversarial pass (no review tool, manual read against issue #965). Verified the reprice math and the live-CI evidence below directly; posting two findings.

supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
```diff
@@ -0,0 +1,83 @@
...
+
+-- 1. Point route-groq-fast back at the model Groq currently serves.
+UPDATE public.provider_routes
+   SET provider_model = 'groq/openai/gpt-oss-20b'
+ WHERE route_id = 'route-groq-fast'
+   AND provider_model <> 'groq/openai/gpt-oss-20b';
+
+-- 2. Re-derive hive-fast's price from that model's published rate.
+UPDATE public.model_aliases
+   SET input_price_credits  = 10500,
```
sakibsadmanshajib, line 88: This migration is already failing CI on this PR's own run. `apps/control-plane/internal/routing/catalog_pricing_integration_test.go`'s `TestHiveFastIsPinnedToGroqAtCorrectedPrice` (added when 20260801_14 first pinned hive-fast to llama-3.1-8b-instant) asserts the pre-revert catalog values directly against a live, fully-migrated Postgres instance, and this PR does not update it. The "Go tests (control-plane)" check on this PR is red right now with exactly the mismatch this migration causes:

```
catalog_pricing_integration_test.go:150: hive-fast provider_model = "groq/openai/gpt-oss-20b", want groq/llama-3.1-8b-instant
catalog_pricing_integration_test.go:161: hive-fast input_price_credits = 10500, want 7000
catalog_pricing_integration_test.go:164: hive-fast output_price_credits = 42000, want 11200
--- FAIL: TestHiveFastIsPinnedToGroqAtCorrectedPrice (0.01s)
```

(from https://github.com/sakibsadmanshajib/hive/actions/runs/32116452470/job/95646984589)

The migration itself and the reprice math look correct to me (10500/42000 matches the worked derivation and matches what 20260801_01 originally derived for this same model), but `catalog_pricing_integration_test.go` needs its expected `provider_model`, `input_price_credits`, and `output_price_credits` updated to the new values (and its comment/doc block, which still cites the 0.05/0.08 llama-3.1-8b-instant rate) before this can merge under the repo's all-checks-green policy.

apps/edge-api/internal/errors/provider_blind.go
sakibsadmanshajib: The four tokens here are OR'd, so a message that hits only `retried:` is enough to collapse the whole customer-facing message to the generic "is not available", even without any of the more distinctive fallback-bookkeeping phrases (`fallback model group`, `model_group=`, `fallbacks=[`). `retried:` alone isn't unique to LiteLLM's fallback exhaustion text: it's plausible for a genuine upstream provider error body (passed straight through via `string(upstreamBody)` at the call sites in orchestrator.go, stream.go, dispatch.go, images/audio/rag handlers) to mention a retry count in an otherwise legitimate, non-leaky message, and that content would now be silently discarded in favor of the generic string.

I don't have a concrete example of a live provider doing this, so this is a design concern rather than a confirmed bug, but given the other three tokens are already specific enough to catch issue #965's actual shape, it seems safer to require `retried:` to co-occur with one of the other three (e.g. `strings.Contains(message, "fallback") && strings.Contains(message, "retried:")`) rather than firing on it alone.

apps/edge-api/internal/errors/provider_blind.go
sakibsadmanshajib: Agreed on scope, this fixes the exact #965 shape and nothing broader. Filed #972 to track widening coverage rather than doing it speculatively here. Thanks for calling out that the fix lands in the shared sanitizer, that was deliberate so every call site benefits without a separate patch each.

apps/edge-api/internal/errors/provider_blind.go
sakibsadmanshajib: Good catch, fixed in 99056b6e4. Dropped the standalone retried: token entirely rather than requiring co-occurrence: the three remaining tokens already fully cover the #965 shape on their own, so retried: alone was providing no additional coverage, only false-positive risk. Added TestProviderBlindLooksLikeRoutingInternalsRequiresSpecificToken to lock in that a retried: only message now passes through unchanged.

apps/edge-api/internal/errors/provider_blind_test.go
sakibsadmanshajib: Fixed in 99056b6e4, added the exact status and message assertions alongside the existing forbidden-substring checks.

supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
sakibsadmanshajib: Fixed in 99056b6e4, wrapped both UPDATEs in BEGIN/COMMIT following the 20260801_10 pattern you pointed to. Left the pre-existing gap in 20260801_01/20260801_14 alone since those are already merged and out of scope for this PR.

.env.example
sakibsadmanshajib: Fixed in 99056b6e4, updated the comment to describe gpt-oss-20b as the current choice and added a note against repointing to a cheaper model without checking Groq's live catalog first, since that reasoning is exactly what caused #965.

apps/edge-api/internal/errors/provider_blind.go
sakibsadmanshajib: Agreed, same scope note as the other thread on this file. Filed #972 to track it rather than widening the heuristic speculatively in this PR.

supabase/migrations/20260818_01_revert_hive_fast_groq_model_decommissioned.sql
sakibsadmanshajib: Confirmed and fixed in 99056b6e4, thanks for pulling the actual CI log. Updated TestHiveFastIsPinnedToGroqAtCorrectedPrice's expected provider_model/prices and its doc comment to the reverted values.

apps/edge-api/internal/errors/provider_blind.go
sakibsadmanshajib: Same finding as the other thread on this file, fixed in 99056b6e4 by dropping the standalone retried: token entirely (three remaining tokens already cover #965's real shape). Added a regression test for the retried:-only case.

PR #971, opened 2026-08-18
fix: drop vestigial Chats nav row, default sidebar to expanded

sakibsadmanshajib:
## Summary

Three small, high-value fixes to the chat left navigation (owner-reported).

1. **Removed the vestigial "Chats" nav row.** Its `href` was `/`, identical to New Chat, and no list view was ever built behind it (added in #938 purely for row parity with Agents/Knowledge). `HIVE_NAV` now ships two rows: Agents, Knowledge.
2. **Investigated the "Folders/Recents never populate" report live**, against the deployed demo box (`chat-hive.scubed.co`), signed in as a dedicated, run-scoped e2e fixture identity (`owui-e2e+<runkey>@hive-e2e.invalid`, minted via `scripts/seed-owui-e2e-user.py`) — never the shared `demo@hive-demo.invalid` account and never the owner's account. A clean run confirmed both sections are correctly wired (`getFolders`/`getChatList`) and render as soon as the sidebar is expanded. No code defect in Folders/Recents.
3. **Defaulted the sidebar to expanded** for a first-time visitor. `Sidebar.svelte` reads `localStorage.sidebar === 'true'` on mount, so an unset key (first visit) reads as collapsed, hiding Folders/Recents with no visible affordance to expand. New logic (`sidebarDefaultExpanded`, extracted to its own pure, unit-tested function since `Sidebar.svelte` is too large/stateful to unit test directly) treats anything other than the literal string `'false'` as expanded, so a first-time visitor gets expanded by default while an explicit prior choice (collapsed or expanded) is preserved on every later visit.

## Why not fix Folders/Recents directly

Ground truth (prior read-only investigation) said they were fully wired with no broken code, and this PR's own live investigation reconfirmed it end to end on the real deployed box: `FOLDERS_HEADING_VISIBLE=true`, `RECENTS_HEADING_VISIBLE=true` once the sidebar was toggled open, on a freshly created account with real DOM/localStorage checks (not just a visual glance). The collapsed-by-default state was the whole complaint, exactly as scoped.

## Visual proof — partial, live infra was unstable during this session

I made five separate live attempts (Playwright, headless Chromium) against `chat-hive.scubed.co` to capture full before/after screenshots with Folders and Recents populated (a real folder + two real chats). Only one attempt completed cleanly end-to-end; the DOM-level assertions from that clean run are quoted above and are real, direct evidence from the live, currently-deployed (pre-fix) build. The other four attempts each failed on a distinct, unrelated infrastructure error during this specific window:

- Supabase Auth `oauth/authorize` returned `504` once.
- Cloudflare returned `522` (origin connection timeout) to the demo box once.
- Caddy/edge returned `upstream request timeout` (crashed the SPA to SvelteKit's default error boundary) once.
- A session bounce landed back on the Hive sign-in page without completing OAuth once.

None of these touch the changed code (pure frontend, no auth/backend changes). I also spun up an isolated local verification stack (`docker compose -p navqw952`, same shared Supabase, separate from any other agent's stack) to get a controlled screenshot instead; `edge-api` stalled during startup on this host for several minutes with no log progress, consistent with heavy concurrent-agent load on this dev box observed throughout the session (docker builds and `docker ps` itself intermittently took 60-120s). I tore that stack down rather than keep contending for host resources.

I am not fabricating a "populated" screenshot. Given the required `Web E2E (full stack)` check is already known red for unrelated reasons and blocks merge regardless, I'm opening this PR now with the code, tests, and the real evidence gathered, and flagging that a clean before/after screenshot set (collapsed vs. expanded, with a real folder and two real chats visible) should be captured and attached here once live infra is stable, before merge.

## Tests

`vendor/open-webui/src/lib/hive/nav.test.ts` (8 tests, updated for the row removal) and new `vendor/open-webui/src/lib/hive/sidebar-default.test.ts` (5 tests, RED confirmed before the implementation existed, GREEN after). Both pass. Frontend production build (`npm run build`, includes this repo's own assertion step checking `data-hive-nav` and brand assets survive the bundle) also succeeds.

## Buglog entry

```json
{"date":"2026-08-18","error_message":"Folders and Recents sidebar sections appear to never populate","root_cause":"Sidebar.svelte defaults localStorage.sidebar to collapsed for a first-time visitor (localStorage.sidebar === 'true' check), hiding Folders/Recents with no visible affordance to expand; both sections were already correctly wired to getFolders/getChatList","fix":"Default sidebar to expanded when no explicit prior choice exists (sidebarDefaultExpanded in vendor/open-webui/src/lib/hive/sidebar-default.ts), preserving an explicit collapse/expand choice on later visits; also removed the vestigial 'Chats' nav row whose href duplicated New Chat","tags":["owui","sidebar","nav","frontend"]}
```
(To be appended to `main`'s `.wolf/buglog.jsonl` via a separate buglog-only PR once this merges, per `.claude/rules/openwolf.md`.)

## Out of scope (explicitly)

Agents row, a Projects section, and the history region restructure are all out of scope per the dispatching brief; not touched here.

## Test plan

- [x] `nav.test.ts` (8 tests) and `sidebar-default.test.ts` (5 tests) pass
- [x] Frontend production build succeeds, brand/nav assertions pass
- [x] Live DOM confirmation: Folders/Recents headings render once sidebar expanded (see above)
- [ ] Clean before/after screenshot set with populated Folders/Recents (blocked by live infra instability this session; retry before merge)

### DIFF
--- a/vendor/open-webui/src/lib/components/layout/Sidebar.svelte
+++ b/vendor/open-webui/src/lib/components/layout/Sidebar.svelte
@@ -4,7 +4,7 @@
 	import Sortable from 'sortablejs';
 
 	import { goto } from '$app/navigation';
-	import { sidebarDefaultExpanded } from '$lib/hive/sidebar-default';
+	import { sidebarDefaultExpanded, shouldPersistSidebarChoice } from '$lib/hive/sidebar-default';
 	import {
 		user,
 		chats,
@@ -583,7 +583,15 @@
 				}
 			}),
 			showSidebar.subscribe(async (value) => {
-				localStorage.sidebar = value;
+				// Mobile's forced collapse (the mobile.subscribe callback above)
+				// fires this same subscriber, and unconditionally persisting here
+				// would overwrite an explicit desktop choice with 'false' on a
+				// visitor's first mobile visit, silently reverting the sidebar
+				// default (or the user's own prior choice) the next time they are
+				// back on desktop. See shouldPersistSidebarChoice.
+				if (shouldPersistSidebarChoice($mobile)) {
+					localStorage.sidebar = value;
+				}
 
 				// nav element is not available on the first render
 				const navElement = document.getElementsByTagName('nav')[0];
--- a/vendor/open-webui/src/lib/components/layout/Sidebar.svelte
+++ b/vendor/open-webui/src/lib/components/layout/Sidebar.svelte
@@ -4,6 +4,7 @@
 	import Sortable from 'sortablejs';
 
 	import { goto } from '$app/navigation';
+	import { sidebarDefaultExpanded } from '$lib/hive/sidebar-default';
 	import {
 		user,
 		chats,
@@ -561,7 +562,12 @@
 			document.documentElement.style.setProperty('--sidebar-width', `${w}px`);
 		});
 
-		showSidebar.set(!$mobile ? localStorage.sidebar === 'true' : false);
+		// A first-time visitor (no stored choice yet) now defaults to expanded,
+		// so Folders and Recents are visible without an extra click -- both
+		// render correctly once expanded (confirmed live), the collapsed
+		// default was the whole complaint. An explicit prior choice, expanded
+		// or collapsed, is preserved either way. See sidebar-default.ts.
+		showSidebar.set(!$mobile ? sidebarDefaultExpanded(localStorage.sidebar) : false);
 
 		const unsubscribers = [
 			mobile.subscribe((value) => {
--- a/vendor/open-webui/src/lib/hive/ShellNavIcon.svelte
+++ b/vendor/open-webui/src/lib/hive/ShellNavIcon.svelte
@@ -26,9 +26,7 @@
 	aria-hidden="true"
 	focusable="false"
 >
-	{#if name === 'chats'}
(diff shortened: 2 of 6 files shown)


sakibsadmanshajib:
## Adversarial review: CodeRabbit CLI

Ran `coderabbit review --base origin/main` locally against this branch.

**1 major finding**, `Sidebar.svelte` (mobile-forced-collapse persistence):

> Line 570 sets showSidebar to false on mobile. The subscriber at Lines 585-587 then writes "false" to localStorage.sidebar. This overwrites an explicit desktop "true" choice and makes a mobile-first visitor stay collapsed on the first desktop visit. Persist the preference only for desktop user changes, and restore it when the viewport returns to desktop.

**Status: fixed**, in `3fdd664e8`. Added `shouldPersistSidebarChoice(mobile)` (`sidebar-default.ts`, unit tested) and gated the `localStorage.sidebar` write in `Sidebar.svelte` on it, so mobile's forced collapse (a transient drawer state, not a setting) can no longer overwrite an explicit desktop choice or the new expanded default.

Not implementing the finding's secondary suggestion ("restore it when the viewport returns to desktop"): that is a separate UX enhancement (auto re-expand on a live mobile-to-desktop viewport transition within one SPA session), not a fix for the reported data-integrity bug, and upstream has no equivalent auto-restore behavior for any other persisted preference today. A page reload already restores correctly once the corrupt-write bug above is fixed, since `localStorage.sidebar` is no longer clobbered.

Second CodeRabbit pass after the fix confirms no findings.

PR #973, opened 2026-08-18
feat: publish knowledge-work-pack deck output as a real artifact link

sakibsadmanshajib:
deckgen.Render and artifactsclient.Client both existed, both tested, and had zero non-test callers anywhere in the repo. An agent could produce a deck and nothing stored it, nothing could address it afterwards, and the user could not open it: `agent_tasks.result_summary_ref` was always the agent's raw final-response text, and the task console printed it as if it were a link (`task-console.tsx:621`).

This wires the two subsystems together in the process that actually executes on the demo box: the socket-arm host launcher (`apps/agent-engine/cmd/agent-engine -serve`), not the in-process arm, which cannot run under docker compose regardless of configuration.

- `SandboxEngine.Status` (`apps/agent-engine/internal/engine/engine.go`), on a knowledge-work-pack task's terminal success and before its `/workspace` bind mount is reaped, looks for exactly one declared output file: `.hive/deck.json`. Never a directory scan, never a glob. If present and valid, it is rendered with the already-tested `deckgen.Render` and published via `artifactsclient.Client.Create` (or `AddVersion` when the manifest names an existing artifact id), authenticated as the task's own user.
- That authentication needed a bearer JWT that never reached this process before. It is captured once, in `apps/edge-api/internal/agenttask/handler.go`'s `handleCreate` (the only place that ever sees the raw `Authorization` header), and threaded in memory only through `control-plane`'s `CreateTask` and the agent-engine daemon's `/launch` call. It is never persisted to `public.agent_tasks`, never exposed to the sandboxed agent, and an API-key-authenticated create (no Supabase JWT to forward) just skips publishing rather than forwarding an API key as if it were one.
- Ownership is enforced entirely by code this PR does not touch: edge-api's `/v1/artifacts` resolves `tenant_id` strictly from that same JWT's claims, and `AddVersion`'s tenant-scoped RLS lookup 404s outright on a manifest-supplied `artifact_id` from another tenant. No new internal-token bypass was introduced anywhere.
- A publish failure of any kind (no manifest, malformed or oversized manifest, edge-api unreachable or rejecting it) falls back to the agent's own final-response text and never turns a succeeded task into a failed one. A manifest over 1 MiB is rejected outright, never silently truncated.
- `apps/agent-console`'s task console renders `result_summary_ref` as a real `<a>` when it is an artifact-shaped path. The href points at this app's own deck proxy rather than the artifacts origin, because a deck is a private artifact and a browser navigation carries no `Authorization` header; see the update below for why the original artifacts-origin link could not open.
- `scripts/install-agent-engine-host.sh` and `deploy/docker/docker-compose.yml` wire the one new environment variable this needs (`EDGE_API_URL`, host-reachable, mirroring how `CONTROL_PLANE_URL` is already overridden there). Left unset, the whole feature is a no-op with zero behavior change (see the review-fix update below for why an earlier version of this claim was briefly false).
- `apps/agent-engine/packs/knowledge-work-pack/skills/deck-generation/AGENTS.md` is corrected to describe what is actually true: the sandbox has no Go toolchain and no network route to edge-api, so the agent's real job is writing the manifest, not calling `deckgen.Render`/`artifactsclient.Client` itself.

## Update: review findings fixed (a39d72cdc)

Two Go-review bugs and five security-review findings came back. All fixed, all replied to inline. Summary:

- **Kill switch (Go review, was a real bug):** `EDGE_API_URL` previously could not be turned off (`envOr`'s fallback masked "explicitly empty" as "use the default", and the default was a live, correct address under the demo box's deployment path). The claim above that this feature is opt-in and off-by-default was false at the time it was first written. `serve.go` now reads a plain `os.Getenv` with no fallback: unset means off, full stop. The install script's own default now uses `${VAR-x}` so an explicit empty override is honored.
- **Idempotent publish (Go review):** two concurrent `Status` calls for the same session could each independently publish an artifact for the same task, orphaning one of them. Fixed with a per-session mutex serializing the terminal-transition work, never the engine-wide lock. New test forces the interleaving deterministically.
- **TOCTOU (security review, HIGH, blocked merge):** the prior symlink-escape fix (`resolveWithinRoot`) checked a path string and reopened that same string fresh, an unsynchronized two-step sequence a live sandboxed process could race. Replaced with Go 1.24's `os.Root`, which resolves and opens in one syscall. New test hammers a real rename/symlink swap for 500 iterations.
- **FIFO hang (security review, MEDIUM-HIGH, blocked merge):** `mkfifo .hive/deck.json` would hang the open forever with no race required, leaking the goroutine, the quota slot, and the whole sandbox container. Fixed with `O_NONBLOCK` plus a regular-file check on the opened descriptor.
- **Credential lifetime (security review, MEDIUM):** the bearer JWT was never cleared from the long-retained session struct. `reap()` now zeroes it on any terminal transition.
- **Silent expiry (security review, LOW):** added a distinguishable log line (and an `ErrUnauthorized` sentinel in `artifactsclient`) for a 401/403 publish rejection, so an expired-token failure mode doesn't read the same as any other transient error.

## Update: the published link did not open, and now does (576292e8e onward)

The rebase pass turned up two defects and one design gap. The gap is the important one: **as previously written, this pull request's headline outcome did not work.**

### The gap: a private artifact behind a header-only route

A published deck is private (`artifacts.is_public` is `NOT NULL DEFAULT false`) and edge-api's serving route resolves a viewer only from an `Authorization: Bearer` header (`optionalViewerTenant`). There is no cookie path anywhere in `internal/artifacts` or `internal/auth`, and `Caddyfile.artifacts` is a bare reverse proxy that injects no credential. A browser navigation carries no such header, so the owner following their own deck link was an anonymous viewer and RLS answered 404.

This was proven rather than argued. `TestRouteServe_PrivateArtifact_NoAuth404` passes unmodified; flipping only the artifact's privacy, and nothing else, turns it red with `status = 200, want 404`; restoring returns it to green. That 200 under mutation is the same request the anchor issued.

It also mattered more than a missing link, because this change **replaces** `result_summary_ref` with an opaque `/artifacts/{id}` path where it previously held the agent's readable final-response text. Merging without a working link would have been a regression against main, not a partial landing.

Two shortcuts were considered and rejected on the record:

- **Publishing decks anonymously readable.** Rejected. It makes an unguessable UUID the access control, which is the pattern Hive Enterprise's data-sovereignty posture exists to reject, and it would contradict #960 tightening cross-tenant reads in the same week. It also runs against a recorded decision: `.wolf/decisions.md` D-015 specifies the public artifact link tier as **owner-toggled visibility**, so flipping a deck public automatically is precisely the thing that decision reserves for an explicit owner action.
- **A short-lived signed URL.** Priced, then rejected. No signing primitive exists to reuse: the only `crypto/hmac` in the tree is `signWebhook` (an outbound webhook body digest, no expiry, no resource scoping), and `S3Client.PresignedURL` is SigV4 against the S3 *object* URL, so it would serve tenant HTML from a shared storage origin carrying none of `writeArtifactHeaders`'s CSP. That is a worse posture than the option already rejected. Building it properly means a new secret, a mint endpoint, a verification branch and a credential permanently living in a query string, the shape that leaked four real invitation tokens on PR #578.

### The fix: a same-origin deck proxy

`GET /agent-workspace/api/deck/{id}[/v/{n}]` (`apps/agent-console/app/api/deck/[...ref]/route.ts`) resolves the caller's existing Supabase session server side and re-attaches that same access token on a server-to-server call to edge-api. Nothing is minted and nothing is loosened: the viewer must still be signed in as the owning tenant, and edge-api still decides. Decks stay private.

Same-origin is load bearing, not incidental: `/artifacts/*` is not on this app's Caddy route and edge-api emits no CORS headers, so a browser-side fetch could not reach it at all.

The response mirrors `writeArtifactHeaders` and adds `sandbox allow-scripts`. That is the security-critical line. Serving tenant-authored HTML from the app's own host would otherwise place it in the app's origin with reach into its cookies; `allow-scripts` with no `allow-same-origin` drops it into an opaque origin, so the deck's inline navigation JS still runs but can touch nothing of ours. It restores exactly the isolation the separate artifacts origin was providing.

`parseArtifactRef` is now the single place a ref is judged artifact-shaped, shared by the link builder and the route, so the two cannot drift into trusting different shapes. `BASE_PATH` is applied explicitly because Next does not prepend `basePath` to raw hrefs, and without it the link would fall through to Open WebUI's catch-all.

No Caddy or infrastructure change is required: `@agentConsole path /agent-workspace /agent-workspace/*` already routes it, and `EDGE_API_INTERNAL_BASE_URL` is a compose `environment:` entry with a literal `http://edge-api:8080` default, so a deploy picks it up with no operator `.env` edit.

`NEXT_PUBLIC_ARTIFACTS_BASE_URL`, added earlier in this pull request, now has no readers and is removed along with its Dockerfile build arg.

### Two defects fixed on the way

- **Data race on the task bearer JWT.** `publishDeckArtifact` read `sess.bearerJWT` without `e.mu` while `reap` clears it under `e.mu`, and `Cancel` reaches `reap` without ever taking `sess.finishMu`, so a cancel landing as a task finished raced the publish. The read now goes through an `e.mu`-guarded accessor. Confirmed by mutation: the unsynchronized read produces `WARNING: DATA RACE` naming `engine.go:824` against `engine.go:687`; the fix is green across twelve packages under `-race`, which CI runs.
- **Local artifact origin.** The fallback pointed at port 80 while `caddy-artifacts` publishes on 3004. Moot now that the variable is gone, but it was real.

### The middleware nearly deleted the fix's one security control

Recorded at length in the buglog entries below, and worth stating here because the next person adding a route will meet the same edge.

`apps/agent-console/middleware.ts` runs on every path except static assets and calls `res.headers.set("Content-Security-Policy", "frame-ancestors 'self'")`. **`set` replaces, it does not merge.** So the middleware overwrote the deck route's own policy wholesale, deleting `sandbox allow-scripts`, which is the single directive keeping tenant-authored HTML out of this app's origin. Without it the deck would have executed inline script in the console's origin with reach into its cookies and storage: a stored-XSS-shaped hole on the surface that holds admin sessions. Nothing failed and nothing logged.

The route's unit tests could not see it. They exercise the exported `GET` handler and observe the `Response` it returns; middleware runs outside that boundary and mutates the response afterwards. Every assertion about the route's CSP passed while the header actually served to a browser was a different, weaker one. It was caught by building the production image, running `next start` and curling it.

**This is header precedence, and it is reintroducible.** Any future route handler that writes its own `Content-Security-Policy` will have it silently replaced unless its path is exempted. The fix is an explicit path allowlist (`/api/deck/`) rather than a "skip if the response already has a CSP" check, and that choice is deliberate: the allowlist is fail-safe, hardened by default with an explicit opt-out, whereas a presence check is fail-open, letting any future handler that sets any policy silently opt itself out of clickjacking protection too. The trade is that a new route must remember to exempt itself; the middleware comment now says so at the point of the decision.

One broader note, since it bears on how this feature was reviewed. A was rejected for widening access through the URL, and B then nearly widened access through the origin instead. This class of feature, hosting untrusted tenant-authored HTML, leaks by default, and every path to it needs the same scrutiny rather than only the one that looked risky first.

One note on the regression guard, because it nearly shipped unable to fail: its first version waited on the publisher's `started` channel before cancelling, which ordered the read before the write and removed the race it claimed to test. It passed against deliberately broken code. The committed version parks the engine inside the final-response fetch, the last step before the read, and cancels from there.

## Test plan

- [x] `go build`/`go vet` clean across `apps/agent-engine`, `apps/control-plane`, `apps/edge-api`.
- [x] `go test` green: `apps/agent-engine/internal/engine` (26 tests now, including the concurrency, TOCTOU-swap, and FIFO regression tests added in the fix round), `.../deckgen`, `.../artifactsclient`, `apps/control-plane/internal/agenttask`, `.../agentengine`, `apps/edge-api/internal/agenttask`.
- [x] `apps/agent-console` vitest suite green (75 passed). `npx tsc --noEmit` clean.
- [x] `go test -short -race` green across the agent-engine, control-plane agenttask and edge-api agenttask packages (twelve packages, no warnings), including the new `TestSandboxEngine_Status_CancelDuringPublishIsRaceFree`.
- [x] `apps/agent-console` type check, unit suite and production build green, including the new deck-proxy route tests.
- [ ] **Live visual proof outstanding, and this pull request is held for it.** The capture must show a Cowork task producing a deck AND the link opening, which needs a real Apptainer sandbox. The only pre-merge mechanism for that is `.github/workflows/agent-visual-proof.yml`, which currently fails for every pull request, including unrelated dependabot branches, at `Start the stack` with control-plane reporting `database not available at startup: database unreachable after 13 attempt(s) over 1m15s`. That is environmental and tracked separately as issue #1050; this branch is not the cause. A capture showing only the link opening was deliberately not substituted: it would prove the smaller half and still leave the real one owed.

## Buglog entry

```json
{"date":"2026-08-18","error_message":"agent_tasks.result_summary_ref always held the agent's raw final-response text; a knowledge-work-pack deck was rendered nowhere and linked nowhere","root_cause":"deckgen.Render and artifactsclient.Client (Create/AddVersion) had zero non-test callers; no bearer JWT ever reached the agent-engine host process to authenticate a publish call, and no file-output convention existed between the sandboxed agent and the host process","fix":"added Task.BearerJWT threaded from edge-api's task-create handler through control-plane to the agent-engine daemon's /launch call; SandboxEngine.Status reads a single well-known .hive/deck.json manifest before reaping /workspace, renders it, and publishes via artifactsclient, overriding resultSummary with the artifact URL only on success; task-console now linkifies artifact-shaped refs via NEXT_PUBLIC_ARTIFACTS_BASE_URL; review fixes added a real off switch, idempotent publish via a per-session lock, os.Root-based TOCTOU-proof manifest reads, FIFO rejection, bearer-JWT lifetime bounding, and a distinguishable log line for expired-token publish failures","tags":["agent-engine","artifacts","cowork","knowledge-work-pack","result_summary_ref"]}
```

Additional entries from the rebase and deck-proxy pass:

```json
{"date":"2026-08-23","error_message":"data race between Cancel and a knowledge-work-pack publish on the agent-engine session bearer JWT","root_cause":"publishDeckArtifact read sess.bearerJWT without holding e.mu while reap clears that same field under e.mu; Cancel reaches reap without ever taking sess.finishMu, so a user cancelling a task at the moment it finishes puts the write concurrent with the read with no happens-before edge between them, which the race detector fails the build on and which could hand Create a half-cleared credential","fix":"read the credential once through a new bearerJWTOf accessor that takes e.mu and use that copy for the rest of publishDeckArtifact; regression guard TestSandboxEngine_Status_CancelDuringPublishIsRaceFree holds the engine inside the final-response fetch, the last step before the read, and launches Cancel from there","tags":["agent-engine","concurrency","data-race","cowork","artifacts"]}
```

```json
{"date":"2026-08-23","error_message":"a race regression test passed against deliberately unsynchronized code","root_cause":"the first version of TestSandboxEngine_Status_CancelDuringPublishIsRaceFree waited on the publisher's started channel before launching Cancel; Create runs after the bearer JWT is read, so that channel send established a happens-before edge that ordered the read before the write and removed the very race the test claimed to cover","fix":"gate on the fake agent server's final-response handler instead, which parks the engine immediately before the read, so the cancel is genuinely concurrent with it; verified by mutation, the unsynchronized read now produces WARNING: DATA RACE naming engine.go:824 against engine.go:687","tags":["testing","false-green","race-detector","guard-cannot-fail"]}
```

```json
{"date":"2026-08-23","error_message":"published deck artifact link returns 404 for the user who owns it","root_cause":"publishDeckArtifact calls Create and never marks the artifact public, artifacts.is_public defaults to false, and the serving route resolves a viewer only from an Authorization Bearer header (optionalViewerTenant) with no cookie path anywhere in internal/artifacts or internal/auth; a browser navigation sends no such header, so the viewer is anonymous and RLS admits only public rows","fix":"serve the deck through a same-origin server-side proxy in agent-console that resolves the caller's Supabase session server side and calls edge-api with that user's own bearer, returning the HTML under Content-Security-Policy sandbox allow-scripts for an opaque origin; rejected making decks public (data sovereignty) and a signed URL (no signing primitive, credential in query string)","tags":["artifacts","auth","cowork","rls","csp"]}
```

```json
{"date":"2026-08-23","error_message":"NEXT_PUBLIC_ARTIFACTS_BASE_URL fallback unreachable in local and enterprise profiles","root_cause":"the fallback resolved to http://artifacts.localhost, port 80, while caddy-artifacts publishes on host port 3004, so the browser was handed a link that could not connect before authentication was even reached","fix":"carry the published port in the fallback, http://artifacts.localhost:3004, leaving the configured-ARTIFACTS_DOMAIN path unchanged","tags":["docker-compose","artifacts","config"]}
```

```json
{"date":"2026-08-23","error_message":"agent-console middleware silently replaced the deck proxy's Content-Security-Policy, deleting the sandbox directive that keeps tenant HTML out of the app origin","root_cause":"middleware.ts runs on every path except static assets and calls res.headers.set(\"Content-Security-Policy\", \"frame-ancestors 'self'\"); set REPLACES rather than merges, so a route handler that writes its own policy has it overwritten wholesale after the fact. The deck proxy serves tenant-authored HTML from the console's own host and depends on sandbox allow-scripts for an opaque origin; with that directive stripped the deck would have executed inline script in the console's origin with access to its cookies and storage, a stored-XSS-shaped hole on the surface that holds admin sessions. Nothing failed and nothing logged","fix":"exempt /api/deck/ from the middleware's header stamping so the route's own strictly narrower policy survives; the route sets frame-ancestors 'none', so no protection is lost. Kept as an explicit path allowlist rather than a has-CSP presence check on purpose: the allowlist is fail-safe (hardened by default, opt out explicitly) while a presence check is fail-open (any future handler that sets any CSP silently opts itself out of clickjacking protection)","tags":["security","csp","xss","middleware","header-precedence","agent-console","false-green"]}
```

```json
{"date":"2026-08-23","error_message":"route unit tests could not detect a security header being stripped by middleware","root_cause":"vitest exercises the exported GET handler directly, so it observes the handler's own Response object; Next middleware runs outside that boundary and mutates the response afterwards, so every assertion about the route's Content-Security-Policy passed while the header actually served to a browser was a different, weaker one","fix":"verify security headers against a running production image (docker build, next start, curl) rather than against the handler in isolation; unit assertions on response headers are necessary but prove nothing about what the server emits once middleware composes the response","tags":["testing","false-green","middleware","integration-gap","security"]}
```

## Summary by CodeRabbit

- **New Features**
  - Generated knowledge-work decks are now published as accessible artifacts.
  - Task results display valid deck artifacts as secure links that open in a new tab.
  - Supports creating new artifact versions and viewing versioned decks.
  - Added an authenticated same-origin deck viewer with access validation and security protections.

- **Bug Fixes**
  - Invalid, missing, oversized, or unavailable deck manifests no longer replace successful task results with broken links.
  - Improved handling of concurrent task completion and artifact publishing failures.


sakibsadmanshajib:
## Adversarial review status

This subagent's toolset is Read/Write/Edit/Bash only (no Agent/Task spawning, no Skill invocation), so the full dynamic multi-stream pipeline from `adversarial-pr-review` cannot run from here. Status per stream:

- **CodeRabbit CLI**: attempted (`coderabbit review --base main`). Hit the account's review-limit rate cap immediately (3/3 reviews used, ~44 min reset at attempt time). **SKIPPED**, not a clean pass.
- **`ecc:code-review` / language reviewers (`go-reviewer`, `typescript-reviewer`, `security-reviewer`)**: require the Agent tool to dispatch as subagents. Not available to this session. **SKIPPED** — needs a dispatch with that tool.
- **Plain adversarial pass**: done manually by the author (self-review, not independent, flagged as such). Found and fixed one real issue before this comment: a symlink-following arbitrary-file-read path (`.hive/deck.json` could be replaced with a symlink to any host path the agent-engine daemon's own OS user can read, e.g. its own `HIVE_AGENT_ENGINE_LLM_API_KEY`/`CONTROL_PLANE_INTERNAL_TOKEN` runtime file). Fixed in `b5213bc68` (`resolveWithinRoot`), with a regression test that plants such a symlink and asserts it is never followed.

**This PR should not merge on my say-so alone.** Auth/money/input-parsing paths are touched (bearer JWT threading, tenant-scoped artifact publish), which per `.claude/rules/orchestrator.md` mandates the security-reviewer stream before merge, and that stream has not run. Requesting the orchestrator dispatch `go-reviewer`, `security-reviewer`, and (once the rate limit clears) CodeRabbit CLI against this diff before this is considered mergeable.

## Self-review notes for the reviewers (starting points, not a substitute)

- Tenant/user ownership of a published artifact rides entirely on edge-api's existing JWT-based tenant resolution (`apps/edge-api/internal/artifacts`, untouched by this PR) — worth independently confirming `AddVersion`'s RLS-scoped lookup (`apps/edge-api/internal/artifacts/repository.go`) actually 404s on a manifest-supplied `artifact_id` belonging to another tenant, rather than trusting my read of it.
- `Task.BearerJWT` is threaded as a plain string through `agenttask.Task` (control-plane), `engine.Task` (agent-engine), and a new `bearer_jwt` JSON field on the internal `/internal/agent-tasks` and daemon `/launch` request bodies. Neither is customer-reachable (internal-token-guarded / Unix-socket-guarded respectively), and the field is never logged and never persisted to `public.agent_tasks` — worth confirming independently rather than taking that on faith.
- `EDGE_API_URL` is a new env var read only by the socket-arm daemon (`apps/agent-engine/cmd/agent-engine -serve`); left unset it is a documented no-op. The in-process arm (`apps/control-plane/internal/agentengine/engine.go`) also forwards `BearerJWT` for interface consistency but that arm cannot run under docker compose regardless (pre-existing, unrelated to this change).

apps/agent-engine/cmd/agent-engine/serve.go
```diff
@@ -154,7 +160,23 @@ func serve(socketPath, controlPlaneURL, controlPlaneToken string) error {
...
+	// the same thing in Go, and the latter would panic the first time
+	// engine.SandboxEngine.Status tried to call through it), so
+	// knowledge-work-pack deck output just never publishes and Status keeps
+	// returning the agent's own final-response text exactly as it always
+	// has. "http://edge-api:8080" only resolves for this daemon's own host
+	// when it runs inside the compose network; the real host launcher
+	// (scripts/install-agent-engine-host.sh) always sets this explicitly to
+	// the box's published edge-api port instead, mirroring how that script
+	// already overrides CONTROL_PLANE_URL's own compose-shaped default.
+	if edgeAPIURL := envOr("EDGE_API_URL", "http://edge-api:8080"); edgeAPIURL != "" {
```
sakibsadmanshajib: **Bug: this "opt-in" wiring is always on; there is no environment-variable value that disables it.**

```go
if edgeAPIURL := envOr("EDGE_API_URL", "http://edge-api:8080"); edgeAPIURL != "" {
```

`envOr` (main.go:212) returns the fallback whenever `os.Getenv(key) == ""`, and that's true both when the var is unset *and* when it's explicitly set to the empty string -- Go's `os.Getenv` can't tell those apart, and this helper doesn't try. So `edgeAPIURL` is never `""` here, the guard is always true, and `engineCfg.Publisher` is never nil.

This contradicts the PR description ("left unset, the whole feature is a documented no-op with zero behavior change") and the comment two lines above this one ("an operator who sets it to the empty string leaves engineCfg.Publisher at its true zero value") -- that state is unreachable via this env var.

It's not just theoretical: `scripts/install-agent-engine-host.sh:133` writes `EDGE_API_URL="${EDGE_API_URL:-http://127.0.0.1:8080}"` into the daemon's env file unconditionally, and `docker-compose.yml` publishes edge-api on host port 8080 (`"8080:8080"`) -- so on the one deployment path that runs real tasks, `127.0.0.1:8080` is a live, correct address, not a dead one. Publishing is wired on by default from the moment this merges, for every knowledge-work-pack task that writes a deck manifest, with no supported way to opt out. That may be a fine product decision, but it isn't the decision this PR describes making, and nothing tests this wiring (`apps/agent-engine/cmd/agent-engine` has zero `_test.go` files).

Suggest `os.LookupEnv("EDGE_API_URL")` (distinguishing absent from empty) or an explicit boolean feature flag instead of overloading the URL string's emptiness.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -466,6 +560,17 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
 		status, resultSummary = StatusSucceeded, summary
+
+		// Runs before reap() below removes sess.workingDir: this is the last
+		// point anything can still read what the sandbox wrote to
+		// /workspace. A publish failure of any kind (no manifest, malformed
+		// manifest, edge-api rejected it, network error) intentionally falls
+		// back to the agent's own text summary above rather than touching
+		// the task's success/failure at all — see publishDeckArtifact's doc
+		// comment for the reasoning.
+		if url, ok := e.publishDeckArtifact(ctx, sess); ok {
```
sakibsadmanshajib, line 626: **Bug, empirically confirmed: two concurrent `Status()` calls on the same session both publish, producing two artifacts for one task.**

`Status` reads `sess.terminal` once under `e.mu` (line 545), releases the lock, and only then walks `GetConversation` -> `FinalResponse` -> here (`publishDeckArtifact`, a real network call) -> `reap()` (line 592), which is what finally sets `sess.terminal` under the lock again. Nothing serializes two goroutines that both observe `sess.terminal == nil` before either reaches `reap()`. `reap()`'s `sess.reaped` guard only protects bookkeeping/teardown; it runs *after* the publish already happened, so it can decide which outcome gets cached but cannot prevent the second network call.

This was harmless before this PR -- `GetConversation`/`FinalResponse` are read-only and idempotent, safe to race. This PR is the first thing to put a non-idempotent side effect inside what was previously a safe-to-poll path, with no added synchronization for it.

I reproduced this directly: checked out this branch, added a temporary test (not part of this PR) that launches a knowledge-work-pack session, writes a deck manifest, and fires two goroutines into `Status` concurrently against a `Publisher.Create` fake with a 200ms delay. Ran via `docker compose --profile tools run toolchain`. Result: `Create` called twice for one task. Whichever goroutine's `reap()` loses the race has its own artifact URL discarded from the cache, but its *caller* still receives that discarded URL as `Status`'s return value -- so two callers can walk away with two different, both real, artifact URLs for the same task, and only one is ever cached for replay or has a chance of being what control-plane persists.

Realistic trigger: `Remote.Status` (control-plane side) has a 5-minute client timeout against `r.Context()` server-side, so a client-side retry after a slow response lands as a brand-new goroutine racing whatever the first attempt is still doing. Also relevant if control-plane ever runs more than one replica -- `Poller` has no leader election or per-task locking against other instances. I did not additionally verify the HTTP-retry path end-to-end; the repro is at the `SandboxEngine.Status` level, which is enough to prove the race exists and where it lives.

Suggest a per-session mutex (or `sync.Once`) around the "check terminal -> do terminal work -> reap" critical section, not just around the reap bookkeeping.

apps/agent-engine/internal/engine/engine_test.go
```diff
@@ -681,3 +682,393 @@ func TestSandboxEngine_Launch_SendsProfileIDWhenNoLLMConfigured(t *testing.T) {
...
+	fake.setStatus(controlclient.StatusFinished)
+
+	_, summary, _, err := e.Status(context.Background(), sessionRef)
+	if err != nil {
+		t.Fatalf("Status: %v", err)
+	}
+	if summary != "fallback text" {
+		t.Fatalf("resultSummary = %q, want the agent's own final-response text", summary)
+	}
+}
```
sakibsadmanshajib, line 1272: None of the 9 new tests in this file cover repeat-call or concurrent-call safety for `Status`. Given the race in `publishDeckArtifact`'s inline comment above (see the `engine.go` comment on this review): a test polling an already-terminal session twice and asserting `Create`/`AddVersion` was called exactly once, or two goroutines calling `Status` concurrently on one session, would have caught the double-publish bug before merge. Worth adding either as a regression guard for whatever fix lands.

Separately, on the 9 tests that do exist: I mutation-tested one (`TestSandboxEngine_Status_PublishFailureFallsBackWithoutFailingTask`, by removing its guarded error-handling branch) and it correctly failed. I didn't have budget to mutation-test the other 8 individually -- they read as exercising real, distinct branches against the fake rather than tautologies, but I want to be explicit that's a read, not a verified mutation-kill, for those 8.

sakibsadmanshajib (commented):
Go correctness review (scope: Go logic and test quality only; symlink containment and the bearer-JWT credential path are covered by a separate security pass).

Two confirmed defects below, both verified against a real checkout of this branch via the repo's own `docker compose --profile tools run toolchain` path, not by inspection alone. Full writeup available on request; summarizing the load-bearing parts inline.

**Verified clean, for the record:** failure semantics never flip a succeeded task to failed (and the inverse -- a failed task carrying a fresh artifact URL -- is structurally impossible, since `publishDeckArtifact` only runs inside the `StatusFinished` branch); `BearerJWT` interface churn is complete and consistent across every real implementation and fake, with the repository layer confirmed (by grep) to never reference it; error wrapping never leaks the bearer JWT, rendered HTML, or a real host path; `e.mu` is never held across the network call. Spot-checked one new test (`TestSandboxEngine_Status_PublishFailureFallsBackWithoutFailingTask`) by mutating away its guarded branch -- it correctly went red. Did not mutation-test the other 8 new tests individually.

sakibsadmanshajib (commented):
Adversarial security review (independent pass, not the author). Read the diff directly, not the PR description. Overall: the symlink-escape fix (`resolveWithinRoot`) blocks the static case it was written for, but the containment check and the file open are still two separate, unsynchronized syscall sequences, and there's a second, unrelated, fully deterministic DoS at the same call site. Ownership/RLS, credential-logging, credential-persistence, and the frontend link-rendering path were all independently checked and hold up — see inline comments for exactly what and where. No fixes applied; findings only.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -488,6 +593,134 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
+		return "", false // the ordinary case: this task did not produce a deck
+	}
+	if err != nil {
+		// Includes a rejected symlink escape (resolveWithinRoot's own error).
+		// No detail from err ever names a real host path here, only "outside
+		// <root>", so this is safe to log as-is.
+		log.Printf("engine: deck manifest %s: %v", deckManifestRelPath, err)
+		return "", false
+	}
+	data, err := readCapped(realPath, maxDeckManifestBytes)
```
sakibsadmanshajib: **HIGH — TOCTOU survives the symlink-escape fix (CWE-367).**

`resolveWithinRoot` (line 685) returns a resolved path *string* via `filepath.EvalSymlinks`. This line then hands that string to `readCapped`, which does a brand-new `os.Open(path)` (line 708) — a second, independent kernel path resolution. Nothing pins the two together (no shared fd, no fstat/inode check, no O_NOFOLLOW).

This isn't hypothetical here: the attacker's own process is confirmed still alive at this exact point. `Status()`'s own comment states "the agent-server is a server and keeps running after its conversation finishes," and this whole function runs *before* `reap()` kills `sess.proc`. A knowledge-work-pack task (arbitrary shell access) can write a legitimate small `.hive/deck.json`, let it pass `resolveWithinRoot`, then race-swap it to a symlink pointing at the daemon's own secret file (e.g. the file `scripts/install-agent-engine-host.sh` writes `HIVE_AGENT_ENGINE_LLM_API_KEY`/`CONTROL_PLANE_INTERNAL_TOKEN` into, under the same OS account) before `readCapped`'s `os.Open` runs.

Each task is one shot at the window, but nothing rate-limits task creation, and an unsynchronized rename/relink race is routinely winnable with a handful of attempts (busy-loop swap, or timing it off the daemon's own access pattern).

Suggested fix: don't resolve-then-reopen-by-string. Use Go 1.24's `os.OpenInRoot`/`os.Root` (openat2 `RESOLVE_BENEATH`-backed, race-free root confinement — the actual primitive for this problem), or at minimum open first and fstat-verify the opened fd's device/inode, never re-derive trust from a path string after the fact.

Checked and ruled out as an easier parallel vector: a hard link doesn't need this race at all (a hard link's path is already inside root, so the path-based check can't see it) — but I confirmed via `apps/agent-engine/internal/sandbox/launcher.go`'s `BuildArgv` that the sandbox's only binds are workspace(rw)/pack-config(ro)/sockets, so the container has no path to the daemon's secret file to `link(2)` in the first place. That vector is closed by sandbox topology, not by this code — don't let a hard-link-focused fix miss the actual TOCTOU above.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -488,6 +593,134 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
+}
+
+// readCapped reads path fully only if it is at most limit bytes. A file
+// exactly at limit+1 bytes or larger is rejected outright rather than
+// silently truncated to limit bytes and parsed as if that were the whole
+// deck — a truncated manifest would otherwise either fail JSON decoding
+// (confusing) or, worse, decode successfully into a partial deck nobody
+// asked to publish.
+func readCapped(path string, limit int64) ([]byte, error) {
+	f, err := os.Open(path) // #nosec G304 -- path is resolveWithinRoot's output, already confirmed to resolve inside the session's own workspace
```
sakibsadmanshajib: **MEDIUM-HIGH — unbounded blocking open() on a non-regular file: trivial, deterministic, permanent DoS.**

`resolveWithinRoot` never checks file type — a FIFO isn't a symlink, so `EvalSymlinks` resolves it to itself with no error and containment passes clean. This `os.Open` is plain `O_RDONLY`, no `O_NONBLOCK`. Opening a FIFO for read-only blocks until a writer opens the other end; if the write end is opened and never closed (or just never opened), this call — and the `io.ReadAll` right after it — blocks forever.

A sandboxed task needs exactly one command, `mkfifo .hive/deck.json`, with no race and no timing luck required, 100% reliable every time.

Because this whole function runs before `reap()` in `Status()`, the hang is permanent for that session: the goroutine never returns, `sess.proc` (the whole Apptainer container) is never killed since `reap()` is unreachable, the tenant/user quota slot is never freed, and the session lives in `e.sessions` forever. A user can wedge their own concurrency quota permanently and leak a running sandbox container per hung task, recoverable only by restarting the daemon.

Suggested fix: `os.Lstat` the resolved path and reject anything that isn't a regular file before opening, or open with `O_NONBLOCK` and treat the immediate ENXIO/no-writer case the same as "no manifest."

apps/agent-engine/internal/engine/engine.go
```diff
@@ -211,6 +296,13 @@ type session struct {
 	workingDir     string // removed when reaped; the sandbox's /workspace bind mount
 	release        func() // frees this session's quota slot; safe to call repeatedly
 
+	// pack and bearerJWT are captured from Task at Launch and read back by
+	// publishDeckArtifact when this session's task succeeds. bearerJWT is
+	// never logged and never leaves this process except as the
+	// Authorization header on the one edge-api call it authenticates.
+	pack      string
+	bearerJWT string
```
sakibsadmanshajib, line 307: **MEDIUM — the bearer JWT's lifetime is not actually bounded.**

`reap()` (below, around line 728) sets `sess.reaped`/`sess.terminal` but deliberately never deletes the session from `e.sessions` — pre-existing behavior (see `Cancel`'s own comment: the poller retries `Status` and an unknown-session error there would leave a task active forever). That was harmless when `session` held no secret. This PR adds one: `bearerJWT` now sits in that permanently-retained struct for the rest of the daemon process's lifetime, for every knowledge-work-pack task ever run, not just in-flight ones.

That contradicts the doc comment two lines up ("this session only holds it in memory for the task's lifetime") — the actual lifetime is "until the daemon process exits," which on a long-running box accumulates one live-until-natural-expiry JWT per deck task ever launched. If this process's memory is ever inspected (core dump, attached debugger, or an unrelated future memory-disclosure bug in the same process), the blast radius is every historical user who ever ran a deck task.

Suggested fix: zero `sess.bearerJWT` inside `reap()`, or right after `publishDeckArtifact` consumes it in `Status()`, regardless of whether publish succeeded, failed, or was skipped.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -488,6 +593,134 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
+	if err != nil {
+		// Includes a rejected symlink escape (resolveWithinRoot's own error).
+		// No detail from err ever names a real host path here, only "outside
+		// <root>", so this is safe to log as-is.
+		log.Printf("engine: deck manifest %s: %v", deckManifestRelPath, err)
+		return "", false
+	}
+	data, err := readCapped(realPath, maxDeckManifestBytes)
+	if err != nil {
+		log.Printf("engine: deck manifest %s: %v", deckManifestRelPath, err)
```
sakibsadmanshajib, line 696: **LOW — silent publish failures have no operator-visible signal beyond a log line.**

Every failure path here (malformed manifest, oversized manifest, render error, and — per the review brief's specific concern — an expired JWT on a long-running task hitting edge-api's 401) falls through to `log.Printf` and silently degrades to the plain-text fallback. The fallback design itself is correct (never fail a succeeded task over a storage-side problem, verified via `TestSandboxEngine_Status_PublishFailureFallsBackWithoutFailingTask`), but there's no counter/metric on publish success vs. failure. A systemic failure — e.g. every deck task's JWT routinely outliving Supabase's TTL because deck generation typically runs longer than that — would silently regress every knowledge-work-pack task to "no artifact, ever," discoverable only by grepping this daemon's logs.

Suggest at minimum a counter metric (publish attempted / succeeded / failed-by-reason) so this is observable without log-diving.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -488,6 +593,134 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
+// owns its own runtime state (scripts/install-agent-engine-host.sh writes
+// HIVE_AGENT_ENGINE_LLM_API_KEY and CONTROL_PLANE_INTERNAL_TOKEN to a file
+// under that same account), and a symlink target is an ordinary string with
+// no container-namespace translation applied to it. A file the task could
+// coax into looking like a deck manifest would then have this process
+// publish its contents as a world-of-the-tenant-readable artifact; even one
+// that does not parse still leaks existence/size as a timing or error-shape
+// oracle. Resolving and bounds-checking before ever opening the file closes
+// both.
+func resolveWithinRoot(root, path string) (string, error) {
```
sakibsadmanshajib: Closing the loop on device nodes specifically (FIFO is covered in the comment on line 708; hard links and intermediate-symlinked directories are covered on line 636): confirmed via `apps/agent-engine/internal/sandbox/launcher.go`'s `BuildArgv` and `deploy/apptainer/agent-engine.def` that this sandbox runs with no `--fakeroot` and no added capabilities, i.e. Apptainer's default unprivileged mode. `mknod` for a character or block device requires `CAP_MKNOD`, which an unprivileged sandboxed process does not have, so `.hive/deck.json` cannot be replaced with a device node here (unlike a FIFO, which needs no capability at all). Not a live vector, verified rather than assumed.

sakibsadmanshajib (commented):
Follow-up to close out the last unaddressed sub-checks (non-regular-file types) from the containment review above.

apps/agent-engine/cmd/agent-engine/serve.go
sakibsadmanshajib: Fixed in a39d72cdc. Agreed on both points: envOr's fallback can't tell unset from empty, and the install script made the guessed default unconditional either way.

serve.go now reads `os.Getenv("EDGE_API_URL")` directly with no fallback at all: unset (the vast majority of deployments, and any deployment that predates this change) means the Publisher stays nil and Status behaves exactly as before. No LookupEnv needed — plain empty-string-means-off is simpler and this env var has no legitimate reason to ever be set to `""` on purpose except "disable this."

install-agent-engine-host.sh switched from `${EDGE_API_URL:-http://127.0.0.1:8080}` to `${EDGE_API_URL-http://127.0.0.1:8080}` (no colon): an operator's explicit `EDGE_API_URL=""` is now honored as "stay off" instead of silently replaced by the default, while a genuinely unset variable still gets the sensible host-reachable default for the demo box. That script is still the only place that turns this on by default, and it does so deliberately for the one deployment meant to exercise the feature.

PR description corrected in the same commit.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Confirmed and fixed in a39d72cdc. Your repro shape (two goroutines racing into `Status` against a delayed `Publisher.Create`) is exactly what the new regression test forces, via a gate channel instead of a timing delay so it's deterministic rather than probabilistic.

Fix: `Status` no longer does the terminal-state work inline. It now calls `finishTerminal`, which takes a new per-session `sess.finishMu` before touching `FinalResponse`/`publishDeckArtifact`/`reap`, and re-checks the cached `sess.terminal` immediately after acquiring the lock (double-checked locking) — a second caller that was blocked on the lock while the first one published gets the first one's cached result instead of doing its own `FinalResponse`+`Create` call. The lock is scoped to the session, never `SandboxEngine.mu`, so it never blocks any other session's `Status` call while held across the network call.

On your realistic-trigger note (control-plane's client timeout causing a retry that races the original attempt, or multiple control-plane replicas polling the same session): both are now safe at the `SandboxEngine.Status` layer regardless of who or how many callers invoke it concurrently, since the serialization lives inside the session itself rather than depending on any caller-side coordination.

apps/agent-engine/internal/engine/engine_test.go
sakibsadmanshajib: Added TestSandboxEngine_Status_ConcurrentCallsPublishExactlyOnce in a39d72cdc: two goroutines call Status concurrently on one session, with a gate channel forcing the second to actually block on the new per-session finishMu rather than possibly winning a race to run first. Asserts exactly one Create call and both callers receiving the same URL.

Also added two more regression tests as part of the same fix round (from the separate security-review thread on this PR): TestReadCapped_SwapDuringOpenNeverReadsOutsideRoot (hammers a real symlink swap for 500 iterations) and TestSandboxEngine_Status_RejectsFIFOManifestWithoutHanging (mkfifo, asserts Status returns within 2s). Engine package is now 26 tests total, up from 9 new + the pre-existing suite.

Thank you for the mutation-testing note on the one test — appreciated the explicit distinction between "verified" and "read as probably fine" on the other 8.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Fixed in a39d72cdc, exactly as suggested: `resolveWithinRoot` is deleted entirely, replaced with Go 1.24's `os.Root`/`os.Root.OpenFile`. Resolution and opening now happen in one syscall (openat2 `RESOLVE_BENEATH` on Linux), so there is no separate check step to swap into between — the class of bug, not just this instance of it, is closed.

Confirmed the toolchain image is `golang:1.24-alpine` (used both for `go build`/`go test` and the real binary build in `scripts/install-agent-engine-host.sh`) and `apps/agent-engine/go.mod` already declares `go 1.24.0`, so `os.Root` is available with no toolchain bump needed.

Added `TestReadCapped_SwapDuringOpenNeverReadsOutsideRoot`: a goroutine continuously does write-safe-file / remove / symlink-to-secret / remove in a tight loop while the main goroutine calls `os.OpenRoot` + `readCapped` 500 times, asserting the secret's content is never returned. This is a real concurrent filesystem race against real syscalls, not a static assertion — it's the kind of test that would have caught the original `resolveWithinRoot` bug (a resolve-then-reopen-by-string implementation has a real, if narrow, window for a rename/symlink swap to land in), and it passes reliably against the `os.Root` fix because that window no longer exists.

Thank you for chasing the hard-link vector down to sandbox topology rather than leaving it as an open question — confirmed the same via `BuildArgv`/`agent-engine.def` independently and left it as-is (no code change needed there).

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Fixed in a39d72cdc. `readCapped` now opens with `os.O_RDONLY|syscall.O_NONBLOCK` (so opening a FIFO returns immediately instead of blocking) and immediately calls `f.Stat()` on the already-open descriptor, rejecting anything whose `Mode().IsRegular()` is false — fstat on an open fd rather than a path-based `Lstat`, so the type check can't be raced against by a second swap the way a path-based check could (ties into the TOCTOU thread above: same principle, check the thing you're about to read, not a separate reference to it).

Added `TestSandboxEngine_Status_RejectsFIFOManifestWithoutHanging`: creates a real FIFO with `syscall.Mkfifo`, calls `Status` on a goroutine, and asserts it returns within 2 seconds with the fallback text rather than hanging — this is deterministic exactly as you said, no race needed, and the test fails loudly (rather than hanging the whole test binary for the package's full timeout) via a `select`/`time.After` bound.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Fixed in a39d72cdc. `reap()` now zeroes `sess.bearerJWT` in the same critical section where it sets `sess.reaped`/`sess.terminal`, right after the one call that could still need it (`publishDeckArtifact`, inside `finishTerminal`) has already run and before `sess.reaped` ever becomes visible to another goroutine. The session struct itself still lives in `e.sessions` forever, exactly as before for the reason you quoted (poller retries against an unknown-session error) — only the secret field is cleared, not the session's identity.

Corrected the doc comment on the field too: it previously said "this session only holds it in memory for the task's lifetime," which as you point out was not actually true before this fix.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Partial fix plus a scoped rebuttal on the rest, in a39d72cdc.

Fixed: added `artifactsclient.ErrUnauthorized` (wrapped when edge-api returns 401/403) and a distinct log line in `publishDeckArtifact` — "bearer JWT rejected, likely expired" instead of a bare status number an operator has to already know to grep for. This directly targets the specific scenario in your brief (a long-running deck task outliving its JWT's TTL) with a signal that's actually legible without already suspecting the cause.

Not built: a counter/metric. This package has zero existing metrics infrastructure (no prometheus import anywhere under apps/agent-engine), and adding one for a single counter, on a fallback path that's already correct and already tested, is more surface than this finding's severity (LOW, explicitly) justifies on its own. If publish-failure-rate needs to be actually monitored rather than log-grepped, that's a real ask but a bigger one — happy to open a follow-up issue for it rather than build it inline here. Flag if you'd rather block on it.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Acknowledged, thank you for closing that out explicitly rather than leaving it implicit. No code change needed for device nodes; noted alongside the hard-link confirmation in the TOCTOU thread above.

sakibsadmanshajib:
All two Go-review bugs and five security-review findings fixed and pushed (9b65daedb, on top of a39d72cdc). Replied on all 8 review threads individually with the specific fix or rebuttal for each. Summary:

- Kill switch: real now (plain `os.Getenv`, no fallback; install script uses `${VAR-x}`).
- Idempotent publish: per-session `finishMu` serializes the terminal-transition work; new deterministic concurrency test.
- TOCTOU (HIGH): `os.Root`/`os.Root.OpenFile` replaces the resolve-then-reopen-by-string pattern entirely; new 500-iteration swap-hammer test.
- FIFO hang (MEDIUM-HIGH): `O_NONBLOCK` + regular-file check on the opened descriptor; new mkfifo test with a 2s bound.
- Credential lifetime (MEDIUM): `reap()` now zeroes `sess.bearerJWT` on any terminal transition.
- Silent expiry (LOW): `artifactsclient.ErrUnauthorized` sentinel + distinct log line for 401/403; metrics explicitly not added (rebuttal posted, no existing metrics infra in this package, flagged as a possible follow-up issue instead).
- One stale doc comment (claimed a lifetime bound that wasn't true even before this round) corrected in a follow-up commit.

`go build`/`go vet`/`go test` all green across `apps/agent-engine`, `apps/control-plane`, `apps/edge-api` after every commit in this round (re-verified post-rebase before each push). Engine package test count: 26.

Still open from my side, unchanged from before this round: live visual proof (needs a running stack with the socket arm wired), and the orchestrator's own CI-green + zero-unresolved-threads verification before merge.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Follow-up commit 9b65daedb also corrects Task.BearerJWT's doc comment, which still claimed "this session only holds it in memory for the task's lifetime" after the a39d72cdc fix — that sentence was stale (untrue even before this round) and now points at reap() instead of restating a lifetime claim. Resolving this thread as of 9b65daedb.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -488,6 +644,144 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
+// chance to reject it; it has no effect on reads from a genuine regular
+// file, which is the only thing this is ever meant to succeed on.
+//
+// A file exactly at limit+1 bytes or larger is rejected outright rather
+// than silently truncated to limit bytes and parsed as if that were the
+// whole deck — a truncated manifest would otherwise either fail JSON
+// decoding (confusing) or, worse, decode successfully into a partial deck
+// nobody asked to publish.
+func readCapped(root *os.Root, name string, limit int64) ([]byte, error) {
+	f, err := root.OpenFile(name, os.O_RDONLY|syscall.O_NONBLOCK, 0)
```
sakibsadmanshajib, line 761: **VERIFIED FIXED — HIGH TOCTOU.** `os.Root.OpenFile` is the actual call that reads the file (not a separate lookup step): `readCapped` opens via `root.OpenFile`, stats the returned `*os.File` (fd-based, immune to a path swap), and reads from that same fd. No string path is reopened anywhere downstream — `resolveWithinRoot` is gone entirely.

I didn't take the containment claim on faith. Go's own doc for `os.Root` (`go doc os.Root`, checked via the toolchain container, go.mod pins 1.24.0) states symlinks may not reference a location outside the root and calls out that only `GOOS=js` is vulnerable to TOCTOU in symlink validation — implying the Linux path (this daemon's only real target, per the repo's own Apptainer notes) resolves and opens as one atomic `openat2(RESOLVE_BENEATH)`-backed step.

I also didn't take the new 500-iteration test on faith. I extracted the *old* `resolveWithinRoot`+`os.Open` implementation verbatim into a standalone program and ran the exact same swap-goroutine race (continuously write-safe/remove/symlink-secret/remove) against both implementations, 20,000 iterations each, for a wider margin than the checked-in test's 500:

```
OLD (resolveWithinRoot+os.Open): leaked=2 / 20000 iterations
NEW (os.Root.OpenFile):          leaked=0 / 20000 iterations
```

The old implementation genuinely leaked the secret content twice under real race pressure; the new one never did. The checked-in test is a real regression test for a real, empirically-reproducible bug, not a tautology that would pass against either version.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -488,6 +644,144 @@ func (e *SandboxEngine) Status(ctx context.Context, sessionRef string) (status S
...
+	if err != nil {
+		return nil, err // ordinary os.ErrNotExist included, unwrapped for errors.Is
+	}
+	defer func() { _ = f.Close() }()
+
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, fmt.Errorf("stat: %w", err)
+	}
+	if !fi.Mode().IsRegular() {
```
sakibsadmanshajib, line 771: **VERIFIED FIXED — MEDIUM-HIGH FIFO hang.** The regular-file check runs on `f.Stat()` — an `fstat` on the already-open descriptor returned by `root.OpenFile`, not a path-based `os.Stat`/`Lstat` call — so it can't be swapped out from under it between check and read; it's inherently tied to whatever `ReadAll` actually reads next. This is the right shape: a path-based check here would have just reintroduced Finding #1's race in miniature.

`O_NONBLOCK` is correctly scoped: it's a POSIX no-op on regular files (only pipes/FIFOs/sockets/some char devices honor it for read-readiness), so it has zero effect on the read semantics, buffering, or EOF handling for the actual regular-file case this is meant to succeed on. No short-read-as-EOF risk either: `io.ReadAll`/`io.LimitReader` already loop until a real `io.EOF` or the byte cap, regardless of how many underlying `Read` calls that takes. Its only job is making the FIFO case return immediately with a rejectable descriptor instead of blocking forever — confirmed by the new `TestSandboxEngine_Status_RejectsFIFOManifestWithoutHanging` test, which fails loudly (2s timeout) if this regresses, and passed on the real Linux GitHub Actions runner (`Go tests (agent-engine)`, not just locally).

apps/agent-engine/internal/engine/engine.go
```diff
@@ -500,6 +794,14 @@ func (e *SandboxEngine) reap(sess *session, outcome *terminalOutcome) error {
...
 	sess.reaped = true
 	sess.terminal = outcome
+	// The one call that could still need it (publishDeckArtifact, inside
+	// finishTerminal) always runs before reap: reap is either finishTerminal's
+	// own last step, or Cancel's, and Cancel never publishes. Nothing reads
+	// bearerJWT after this point, and reap() never deletes the session from
+	// e.sessions (poller retries need it), so without this a completed
+	// knowledge-work-pack task's JWT would otherwise sit in this process's
+	// memory for the rest of its life.
+	sess.bearerJWT = ""
```
sakibsadmanshajib, line 804: **VERIFIED FIXED — MEDIUM JWT lifetime.** `reap()` is the single choke point every terminal path already ran through before this line existed (finishTerminal's Finished/Errored/Stuck/Paused/Deleting cases all fall through to the same call at the bottom of the switch, and `Cancel()` calls `reap()` directly) — so clearing `bearerJWT` here, once, covers success, failure, and cancellation without needing a second copy of the clear anywhere else. That's the right place for it: root-caused at the shared function rather than patched into each caller.

One residual, pre-existing edge case worth naming rather than treating as newly introduced: a session whose underlying agent-server never reports a terminal `ExecutionStatus` at all (hung/crashed such that `GetConversation` never returns Finished/Errored/Stuck/Paused/Deleting) never reaches `finishTerminal` or `reap()`, so its JWT — like the rest of that session — would sit uncleared until the daemon restarts. That's an existing property of the whole session lifecycle (an orphaned session already leaked before this PR touched anything), not a gap in this specific fix, and out of scope for a credential-clearing patch to solve on its own.

apps/agent-engine/internal/artifactsclient/client.go
```diff
@@ -112,6 +117,14 @@ func (c *Client) post(ctx context.Context, bearerJWT, path string, body map[stri
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
```
sakibsadmanshajib, line 120: **VERIFIED FIXED (with rebuttal accepted) — LOW silent expiry.** `ErrUnauthorized` wraps both 401 and 403 with `%w`, so `errors.Is` at the call site in `publishDeckArtifact` correctly distinguishes an auth rejection from every other failure shape and logs it by name ("bearer JWT rejected, likely expired") instead of an unremarkable status number. That's a real, meaningful improvement in operator-visible signal for the exact scenario the original finding named (a task outliving its own JWT's TTL).

On the no-metrics-counter rebuttal: I checked rather than assumed. `grep -rln "prometheus\|/metrics\|promhttp\|expvar"` across `apps/agent-engine/` returns nothing, and `serve.go`'s mux only ever registers `/launch`, `/status`, `/cancel`, `/health` — this binary genuinely has zero metrics instrumentation today, not just a missing counter for this one path. Building a registry and scrape endpoint to house one counter would be new infrastructure for a single call site. Agreed the distinguishable log line is the proportionate fix here; a real metrics counter is legitimately a separate, larger piece of work if/when this daemon gets instrumented at all, not something this PR should carry.

apps/agent-engine/internal/engine/engine.go
```diff
@@ -211,6 +299,25 @@ type session struct {
...
+	// calls can each independently observe the same terminal execution
+	// status from the agent-server (asking does not change it), and without
+	// this both would fetch the final response, both would run
+	// publishDeckArtifact's network call, and both would return their own
+	// locally computed result even though only one outcome ever gets
+	// cached — two real artifacts created for one task, one of them
+	// orphaned forever. Scoped to this one session, never
+	// SandboxEngine.mu, so holding it across that network call never blocks
+	// any other session's Status call.
+	finishMu sync.Mutex
```
sakibsadmanshajib, line 319: **Concurrency fix (Go-review's finding, verified no new hazard) — the specific thing I was asked to check: does holding this lock across a network call stall the status endpoint or deadlock against the reaper?**

No deadlock: `reap()` never touches `finishMu`, only `e.mu` plus process/filesystem cleanup, so there's no lock-ordering cycle between `finishTerminal` (which calls `reap()` while still holding its own `finishMu` via defer) and anything else. `Cancel()` doesn't take `finishMu` either, so it's never blocked by a concurrent `finishTerminal`.

No unbounded stall: `finishMu` is per-session, not `e.mu` — it only serializes concurrent `Status()` calls for that *same* session, never other sessions' status checks, exactly as the comment claims. The worst case is bounded by the sum of `sess.client.FinalResponse`'s own RPC timeout and `artifactsclient`'s hardcoded 30s HTTP timeout, and both calls are threaded with the caller's own `ctx`, so an early-cancelled caller unblocks the lock sooner, not later. A second poll of the exact same task arriving mid-publish will see added latency up to that ceiling — a real, bounded tradeoff, not a hang, and clearly the lesser problem versus the double-publish bug this fixes (confirmed via the deterministic `TestSandboxEngine_Status_ConcurrentCallsPublishExactlyOnce`, which forces the interleaving with a gate channel rather than hoping timing cooperates).

One pre-existing race, unchanged by this fix, worth flagging separately rather than silently: `Cancel()` and `finishTerminal` can still race on `reap()`'s idempotency check the same way they could before this PR (finishMu only serializes concurrent *natural-finish* observers against each other, not against a concurrent `Cancel`). If a task finishes and gets cancelled at almost the same instant, the caller inside `finishTerminal` returns its own locally computed "succeeded, here's the artifact URL" while `sess.terminal` — what every later `Status()` replay sees — ends up `Cancelled` because `Cancel`'s `reap()` call won the race. That's not new (the pre-fix code had the identical exposure, since `reap()` was already idempotent-first-writer-wins), and not something this specific patch could reasonably be expected to close, but it means a task cancelled at the exact moment it finishes can still orphan a real, already-published artifact with no record of it in the persisted task state. Worth its own ticket if it matters, not a blocker on this PR.

sakibsadmanshajib (commented):
Verification pass on the fixes for the five findings above (head 9b65daed). Summary: all four of my findings are genuinely closed, not just reworded. Also checked the two Go-review fixes for regressions against my findings (none), and specifically audited whether the new per-session finishMu mutex introduces a lock-across-network-call hazard. Findings below, and full writeup at /tmp/claude-1001/-home-sakib-hive/47b1dc44-1a51-42c2-8e34-41475aa46c9d/scratchpad/review-973-security.md in my sandbox (not in-repo).

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Acknowledged — no action needed, this is your own independent verification of the TOCTOU fix (os.Root replacing resolveWithinRoot+os.Open), not a new finding.

Recording the numbers here per the coordinator's instruction, so the evidence lives on the thread rather than only in a session scratchpad: standalone extraction of the old resolveWithinRoot+os.Open implementation raced against the same swap harness the checked-in test uses, 20,000 iterations each — old implementation leaked the secret's content 2/20000, new os.Root-based implementation leaked 0/20000. That's empirical confirmation the vulnerability was real (not theoretical) and that the fix closes it, not just narrows it. Resolving.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Acknowledged — independent verification of the FIFO fix, no action needed. Confirms the two properties that mattered: the type check is fstat-on-open-fd (can't be raced the way a path-based check could), and O_NONBLOCK is a documented no-op on regular files so the happy-path read semantics are unchanged. Also noted it passed on the real Linux GitHub Actions runner, not just the local toolchain container. Resolving.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Acknowledged — independent verification of the token-clearing fix, no action needed on that part.

The residual edge you named (a session whose agent-server never reports any terminal ExecutionStatus never reaches reap(), so its JWT is never cleared) is the same one already filed: sakibsadmanshajib/hive#979, second bullet ("a session that never reaches any terminal status never zeroes its bearer JWT"). Filed with the reasoning that it's a pre-existing gap in the terminal-detection contract (the module doc already calls out the identical shape for e.sessions itself never being pruned), not something this PR's credential-clearing fix introduced or could reasonably close on its own. Not duplicating the issue, just confirming this thread's finding is already tracked there. Resolving.

apps/agent-engine/internal/artifactsclient/client.go
sakibsadmanshajib: Acknowledged — independent verification of both the ErrUnauthorized fix and the metrics rebuttal, no action needed. Appreciate that the rebuttal was checked (the prometheus/expvar/metrics grep) rather than taken on trust — same conclusion I reached, glad it's independently confirmed rather than just repeated back. Resolving.

apps/agent-engine/internal/engine/engine.go
sakibsadmanshajib: Acknowledged — independent verification that finishMu introduces no new deadlock or unbounded-stall hazard, no action needed on that part. The lock-ordering argument (reap() never touches finishMu, Cancel() never takes it) and the bounded-by-30s-HTTP-timeout argument match my own reasoning; glad to have it checked independently rather than just repeated back.

The pre-existing Cancel()/finishTerminal race you flagged (a task cancelled at the exact instant it finishes can have finishTerminal return "succeeded, here's the artifact" to its caller while sess.terminal ends up Cancelled, orphaning a real published artifact with no record pointing at it) is the same race already filed: sakibsadmanshajib/hive#979, first bullet. Filed with your reasoning as relayed: pre-existing (reap() was already idempotent-first-writer-wins before this PR), and this PR's actual effect is raising the impact from "a discarded string" to "an orphaned real stored object," not introducing the race itself. Not a new ticket, confirming this thread's finding is already tracked there. Resolving.

PR #974, opened 2026-08-18
fix: wire litellm's DB config sync into the deploy pipeline

### LINKED ISSUE #713: deploy-demo-box's model-price assertion goes circular once an admin config sync targets a DB-managed route

sakibsadmanshajib:
Closes #713. Follow-up to #965/#970, requested by the orchestrator: merging the catalog fix in #970 does not make the deployed box actually serve the new model, because the running gateway resolves its upstream model from an environment variable baked into the box's own `.env` at container boot, not from the database. This PR closes that gap at the mechanism level instead of patching the env var again, which is the same pattern that caused the outage in the first place.

**This is now demonstrated, not theoretical.** The deploy triggered by #970's merge failed at exactly the "Assert model catalog prices agree with the model LiteLLM will call" step: the catalogue named the corrected Groq model, the box's `GROQ_FAST_MODEL` still named the decommissioned one, and the assertion refused to ship a box whose two sources of truth disagreed. That is the guard working exactly as designed, and it will keep refusing every deploy until the two sides can agree, which requires this PR. The guard and this fix were built to fit each other: the guard is what proves the divergence is real, this PR is what makes the guard satisfiable again.

## The two-mechanisms problem this closes

`hive-fast`'s route lived in two places that could disagree: `provider_routes.provider_model` in Postgres, and `os.environ/GROQ_FAST_MODEL` templated into `deploy/litellm/config.yaml` at container boot. #970 corrected the first. Nothing corrected the second, because nothing in the deploy pipeline was calling the fix that already exists for exactly this: `apps/control-plane/internal/litellmconfig`, a service that generates LiteLLM's model list straight from `provider_routes.provider_model` and restarts the container, shipped under issues #701/#705/#707 as Phase 20 Plan 03, and never wired into `deploy-demo-box.yml`. Issue #713 tracked this remaining gap.

## What this PR does

Adds one step to `.github/workflows/deploy-demo-box.yml`, right after the existing public smoke test confirms control-plane is healthy:

```
POST http://localhost:8081/internal/litellm/sync
Header: X-Internal-Token: $CONTROL_PLANE_INTERNAL_TOKEN
```

This is the existing `litellmSyncHandler` (`apps/control-plane/internal/litellmconfig/http.go`), already registered at `/internal/litellm/sync` and already guarded by `RequireInternalToken`. Calling it regenerates `deploy/litellm/config.yaml`'s `model_list` directly from `provider_routes.provider_model` for every active, DB-managed route, and restarts the `litellm` container over the Docker socket control-plane already has mounted read-write (`deploy/docker/docker-compose.yml`, `litellm-config` volume + `/var/run/docker.sock`, both already there for exactly this).

No new secret, no new privilege, no new code path: everything this step calls already exists and already ships. `CONTROL_PLANE_INTERNAL_TOKEN` is read from the box's own `.env` using the same pattern the agent-engine launcher step above it already uses. Retried 3 times with a 5 second backoff since it runs immediately after the smoke test, so a persistent failure here is treated as real and fails the job loudly, not swallowed.

Also updated the following step's ("Assert model catalog prices agree with the model LiteLLM will call") own comment, which explicitly called this out as an unsolved circularity risk pending issue #713. Its scope narrows now that the sync runs: it can no longer catch a stale env-var template for a DB-managed route (that mechanism is out of the loop entirely), but it still catches the sync step failing, a route the sync's join unexpectedly excludes, or an api_base drift on a route the sync doesn't touch. Explained in the updated comment rather than left implicit.

## Why this and not another environment variable

The orchestrator's framing, which I agree with: an alias whose real model comes from an env var while its catalog says something else is exactly how #965 happened, and patching `GROQ_FAST_MODEL` again on the box would leave the identical class of bug in place for the next model this alias (or any DB-managed route) needs. This PR makes the database the single source of truth for every DB-managed route's upstream model on every deploy, going forward, not just for `hive-fast` today.

## What this does not do

- Does not touch `provider_routes`, `model_aliases`, or any catalog data.
- Does not change `LITELLM_CONFIG_MODE` (stays default/`file`; `db` mode is explicitly `log.Fatalf`'d as not-yet-implemented in `main.go` and this PR does not touch that).
- Does not remove the `os.environ/GROQ_FAST_MODEL` template from `deploy/litellm/config.yaml` itself; it becomes a fallback the sync's merge preserves for anything the DB doesn't manage, not a delete-on-sight.
- **Update:** `WriteAndRestart` now skips both the write and the restart when the generated config is byte-identical to what is already on disk (`apps/control-plane/internal/litellmconfig/generator.go`, `bytes.Equal` check added in this PR after review). This step runs after the public smoke test has already confirmed the box is serving real traffic, so an unconditional restart would have been a live chat-completions interruption on every single deploy, most of which touch nothing about model routing. The fix lives in the generator, not the deploy script, so every future caller of the sync gets the same behavior for free rather than each one needing its own diff check. New test: `TestWriteAndRestartSkipsRestartWhenConfigUnchanged`. The very first sync (replacing the hand-authored seed file's comments/formatting) still restarts once, correctly.

## Verification

This cannot be exercised in CI: `deploy-demo-box.yml` triggers only on `push` to `main` (self-hosted runner on the box itself, no PR-time run). Verification is the next real deploy actually succeeding, specifically the new step passing and the following "Assert model catalog prices agree" step staying green against whatever route the DB currently holds. To be watched via `gh run watch` on the triggered run after merge, per the orchestrator contract's deploy-verification stage.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/deploy-demo-box.yml'))"` parses without error.
- [x] Confirmed every component this step calls already exists and is already deployed: the handler, its route registration, its auth guard, the docker socket mount, the litellm-config volume, the `custom_providers` row the sync's query joins against for `groq`.
- [ ] Security review (mandatory: internal-auth-guarded endpoint, docker restart capability).
- [ ] Live verification against the box (post-merge, cannot run in CI, see above).

### DIFF
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -806,6 +806,60 @@ jobs:
           echo "SMOKE FAIL"
           exit 1
 
+      - name: Sync LiteLLM model list from the DB catalog (issue #713)
+        # control-plane's litellmconfig service (Phase 20 Plan 03) already
+        # generates deploy/litellm/config.yaml's model_list directly from
+        # provider_routes.provider_model and can trigger a LiteLLM restart via
+        # the Docker socket it already has mounted, but nothing in this
+        # pipeline ever called it, so every deploy kept running whatever
+        # env-var template (os.environ/GROQ_FAST_MODEL and siblings) was last
+        # baked into the box's own .env. That is exactly the two-mechanisms-
+        # one-checked shape that let hive-fast silently point at a Groq model
+        # the catalog had already been corrected away from (issue #965): a
+        # migration can update the DB and still never reach the running
+        # gateway. Calling this endpoint on every deploy makes the DB the
+        # only source of truth for every DB-managed route's upstream model,
+        # which is the fix issue #713 asked for rather than another env var.
+        #
+        # POST /internal/litellm/sync is guarded by RequireInternalToken
+        # (X-Internal-Token), the same internal auth edge-api already uses
+        # for service-to-service calls to control-plane, so no new secret is
+        # introduced here. CONTROL_PLANE_INTERNAL_TOKEN is read from the
+        # box's own .env, same pattern as the agent-engine launcher step
+        # above. Retried briefly since this runs immediately after the smoke
+        # test above already confirmed control-plane healthy over HTTPS, so a
+        # failure here past the first attempt or two is a real one, not a
+        # cold-start race, and is intentionally not swallowed: a sync failure
+        # (a genuine 5xx, not a transient connection reset while the
+        # container finishes booting) fails this job rather than silently
+        # leaving LiteLLM on a stale model list.
+        working-directory: /home/sakib/hive
+        run: |
+          set -euo pipefail
+
+          token=$(grep -E '^CONTROL_PLANE_INTERNAL_TOKEN=' .env | head -1 | cut -d= -f2-)
+          if [ -z "$token" ]; then
+            echo "::error::CONTROL_PLANE_INTERNAL_TOKEN is missing or empty in the box's .env"
+            exit 1
+          fi
+
+          status=""
+          for i in 1 2 3; do
+            status=$(curl -sS -o /tmp/litellm-sync-response.json -w '%{http_code}' \
+              --connect-timeout 2 --max-time 10 \
+              -X POST -H "X-Internal-Token: $token" \
+              http://localhost:8081/internal/litellm/sync) || status="000"
+            if [ "$status" = "200" ]; then
+              echo "LiteLLM config synced from provider_routes"
+              exit 0
+            fi
+            echo "attempt $i: sync returned $status, retry in 5s"
+            sleep 5
+          done
+          echo "::error::litellm sync failed after 3 attempts, last status $status"
+          cat /tmp/litellm-sync-response.json 2>/dev/null || true
+          exit 1
+
       - name: Assert model catalog prices agree with the model LiteLLM will call
         # Issue #689. provider_routes.provider_model is metadata nobody reads at
         # dispatch time: the model LiteLLM actually sends upstream resolves at
@@ -875,20 +929,27 @@ jobs:
         # two-mechanisms-one-checked shape as #689 itself, just moved one
         # field over.
         #
-        # Known circularity, not solved here: apps/control-plane/internal/
-        # litellmconfig's config sync (issue #701, fixed by #705; #707 tracks
-        # the remaining gaps in what it preserves) writes a DB-managed route's
-        # `litellm_params.model` straight from provider_routes.provider_model,
-        # the same column this step's DB side reads. After that sync runs
-        # against a DB-managed route, this comparison cannot fail even if the
-        # env-var mechanism the sync bypasses is wrong, because both sides
-        # would trace back to the identical database value. Today the check
-        # is real: the live config on a freshly recreated box is still the
-        # first-boot seed with `os.environ/...` templates, genuinely
-        # independent of the DB. The api_base comparison above only partially
-        # covers the gap (it still catches an api_base drift after a sync);
-        # it does not restore independence to the model-string comparison.
-        # See issue #713 for the actual fix; not attempted here.
+        # Known circularity (issue #713, wired in by the preceding step): the
+        # previous step now calls apps/control-plane/internal/litellmconfig's
+        # config sync (issue #701, fixed by #705; #707 tracks the remaining
+        # gaps in what it preserves) on every deploy, which writes a
+        # DB-managed route's `litellm_params.model` straight from
+        # provider_routes.provider_model, the same column this step's DB side
+        # reads. That means this comparison can no longer fail for a
+        # DB-managed route even if the sync's own query or the generator's
+        # adapter logic is wrong, because both sides trace back to the
+        # identical database read. What this step still catches: the sync
+        # step above failing outright (it fails the job itself before this
+        # one runs), a route this catalog thinks is DB-managed but the sync
+        # query's join excludes (a real drift the two steps would otherwise
+        # disagree on), or an api_base mismatch on a route the sync does not
+        # cover. It no longer catches a bad `provider_model` value reaching a
+        # stale env-var template, because that mechanism is no longer in the
+        # loop for any route the sync's query returns. This is an accepted,
+        # narrower scope for this step now that issue #713 is fixed, not a
+        # regression: the two-mechanisms-one-checked shape #689 and #965 both
+        # hit is what #713 removed, and a check that could still fail on it
+        # would mean the sync had not actually taken effect.
         working-directory: /home/sakib/hive/deploy/docker
         run: |
           set -euo pipefail
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -824,29 +824,39 @@ jobs:
         # POST /internal/litellm/sync is guarded by RequireInternalToken
         # (X-Internal-Token), the same internal auth edge-api already uses
         # for service-to-service calls to control-plane, so no new secret is
-        # introduced here. CONTROL_PLANE_INTERNAL_TOKEN is read from the
-        # box's own .env, same pattern as the agent-engine launcher step
-        # above. Retried briefly since this runs immediately after the smoke
-        # test above already confirmed control-plane healthy over HTTPS, so a
-        # failure here past the first attempt or two is a real one, not a
-        # cold-start race, and is intentionally not swallowed: a sync failure
-        # (a genuine 5xx, not a transient connection reset while the
-        # container finishes booting) fails this job rather than silently
-        # leaving LiteLLM on a stale model list.
+        # introduced here. CONTROL_PLANE_INTERNAL_TOKEN is read directly from
+        # the box's own .env: compose always sources control-plane's real
+        # token from there regardless of whether a GitHub secret is also set
+        # for other steps, so there is no fallback to try. Retried since this
+        # runs immediately after the smoke test above already confirmed
+        # control-plane healthy over HTTPS, so a failure here past the first
+        # attempt or two is a real one, not a cold-start race, and is
+        # intentionally not swallowed: a sync failure (a genuine non-200, not
+        # a transient connection reset while the container finishes booting)
+        # fails this job rather than silently leaving LiteLLM on a stale
+        # model list. --max-time is 35s, not the more typical 10s used
+        # elsewhere in this file: the endpoint's own work is a DB query, an
+        # atomic config write, and a Docker container restart bounded by its
+        # own 30s internal timeout (litellmconfig/restart.go), so a shorter
+        # client timeout would fire while the server side is still legitimately
+        # working and misreport a slow-but-successful sync as a failure.
         working-directory: /home/sakib/hive
         run: |
           set -euo pipefail
 
-          token=$(grep -E '^CONTROL_PLANE_INTERNAL_TOKEN=' .env | head -1 | cut -d= -f2-)
+          token=$(grep -E '^CONTROL_PLANE_INTERNAL_TOKEN=' .env | head -1 | cut -d= -f2- || true)
           if [ -z "$token" ]; then
             echo "::error::CONTROL_PLANE_INTERNAL_TOKEN is missing or empty in the box's .env"
             exit 1
           fi
 
+          response_file=$(mktemp)
+          trap 'rm -f "$response_file"' EXIT
+
           status=""
           for i in 1 2 3; do
-            status=$(curl -sS -o /tmp/litellm-sync-response.json -w '%{http_code}' \
-              --connect-timeout 2 --max-time 10 \
+            status=$(curl -sS -o "$response_file" -w '%{http_code}' \
+              --connect-timeout 2 --max-time 35 \
               -X POST -H "X-Internal-Token: $token" \
               http://localhost:8081/internal/litellm/sync) || status="000"
             if [ "$status" = "200" ]; then
@@ -857,7 +867,7 @@ jobs:
             sleep 5
           done
           echo "::error::litellm sync failed after 3 attempts, last status $status"
-          cat /tmp/litellm-sync-response.json 2>/dev/null || true
+          cat "$response_file" 2>/dev/null || true
           exit 1
 
       - name: Assert model catalog prices agree with the model LiteLLM will call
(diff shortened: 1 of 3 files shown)


sakibsadmanshajib (commented):
CodeRabbit CLI review (coderabbit review --agent -t all --base main).

.github/workflows/deploy-demo-box.yml
```diff
@@ -793,6 +793,60 @@ jobs:
...
+          if [ -z "$token" ]; then
+            echo "::error::CONTROL_PLANE_INTERNAL_TOKEN is missing or empty in the box's .env"
+            exit 1
+          fi
+
+          status=""
+          for i in 1 2 3; do
+            status=$(curl -sS -o /tmp/litellm-sync-response.json -w '%{http_code}' \
+              --connect-timeout 2 --max-time 10 \
+              -X POST -H "X-Internal-Token: $token" \
```
sakibsadmanshajib, line 873: This step passes CONTROL_PLANE_INTERNAL_TOKEN to curl as a literal command line argument (`-H "X-Internal-Token: $token"`), and writes the response to a fixed path (`/tmp/litellm-sync-response.json`). On a shared runner, the token is visible to any other process that can read this process's argv while curl is running (`ps aux`, `/proc/<pid>/cmdline`), and the fixed response path could be read or replaced by another process between writes.

I want to flag that this exact pattern (passing a bearer token straight into a curl `-H` argument) already exists twice elsewhere in this same file, for `master_key` at line 962 and `HIVE_API_KEY` at line 1076, so this step is consistent with existing style rather than introducing a new one. Given this runs on a single tenant demo box for a short lived call, I would not block merge on it. If it is worth closing, the fix would be to write the header to a `mktemp`'d, mode 600 file and pass it to curl via `-K`/`--config` (which supports a `header = "..."` line) instead of inline on argv, and use a `mktemp`'d path for the response file instead of the fixed `/tmp` one, cleaning both up on exit. That would be worth doing for all three call sites at once rather than just this one.

.github/workflows/deploy-demo-box.yml
```diff
@@ -793,6 +793,60 @@ jobs:
...
+        # failure here past the first attempt or two is a real one, not a
+        # cold-start race, and is intentionally not swallowed: a sync failure
+        # (a genuine 5xx, not a transient connection reset while the
+        # container finishes booting) fails this job rather than silently
+        # leaving LiteLLM on a stale model list.
+        working-directory: /home/sakib/hive
+        run: |
+          set -euo pipefail
+
+          token=$(grep -E '^CONTROL_PLANE_INTERNAL_TOKEN=' .env | head -1 | cut -d= -f2-)
```
sakibsadmanshajib: With `set -euo pipefail` active, this pipeline aborts the whole script when CONTROL_PLANE_INTERNAL_TOKEN is entirely missing from .env, not just when it's present but empty. I checked this locally: when grep finds no matching line, grep itself exits 1 even though head and cut still exit 0 on their empty input, pipefail turns that into the pipeline's own exit status, and `token=$(...)` as a plain assignment is not one of the exempted contexts for errexit, so the script dies right on this line. The `if [ -z "$token" ]; then echo "::error::..."; exit 1; fi` two lines down never runs in that case. The step still fails the job (this is not a silent success), but it loses the specific diagnostic message this step was written to produce, and shows up as a bare non-zero exit with no explanation of why. Something like `token=$(grep -E '^CONTROL_PLANE_INTERNAL_TOKEN=' .env | head -1 | cut -d= -f2- || true)` would let the existing empty-check handle both the missing-line and present-but-empty cases the same way.

.github/workflows/deploy-demo-box.yml
```diff
@@ -793,6 +793,60 @@ jobs:
...
+          if [ -z "$token" ]; then
+            echo "::error::CONTROL_PLANE_INTERNAL_TOKEN is missing or empty in the box's .env"
+            exit 1
+          fi
+
+          status=""
+          for i in 1 2 3; do
+            status=$(curl -sS -o /tmp/litellm-sync-response.json -w '%{http_code}' \
```
sakibsadmanshajib: This writes the sync response to a fixed path, /tmp/litellm-sync-response.json, with no mktemp and no cleanup afterward. This same file already has an established pattern for exactly this need, the featuregate check earlier in this job creates its curl scratch file with `body=$(mktemp)` plus `trap 'rm -f "$body"' EXIT`. Because this job runs on the self-hosted runner directly on the persistent demo box rather than an ephemeral hosted VM, /tmp here is long lived and shared with every other process on that box, so a fixed, predictable filename written with default permissions is readable by any other local account and is never removed, it just accumulates across every deploy. Worth following the same mktemp plus trap pattern already used a few hundred lines up in this file for consistency and to avoid leaving the response sitting around.

sakibsadmanshajib (commented):
Security review of the new deploy step. Token handling, quoting, and endpoint exposure all check out (the grep-from-.env pattern matches the existing agent-engine step, curl has no -v/--trace so the header value never reaches the log, and this step runs on the self-hosted demo-box runner so it only automates a call that was already reachable from the box's own network, it does not expose /internal/litellm/sync any more broadly). Two smaller issues inline below on the fail-loud design and on a leftover response file.

.github/workflows/deploy-demo-box.yml
```diff
@@ -793,6 +793,60 @@ jobs:
...
+              --connect-timeout 2 --max-time 10 \
```
sakibsadmanshajib: The sync step's curl call uses --max-time 10, but the operation it is timing includes a DB query, an atomic config write, and (via WriteAndRestart in apps/control-plane/internal/litellmconfig/generator.go) an unconditional Docker container restart. Docker's restart endpoint sends SIGTERM to litellm and waits up to the t=10 grace period before SIGKILL (restart.go's endpoint is /containers/<name>/restart?t=10, with its own 30 second internal timeout), so if litellm does not exit promptly on SIGTERM, the server side alone can take close to 10 seconds before the container even starts again, before the DB round trip and file write are counted at all. That makes a curl timeout on attempt 1 a plausible, non cold start outcome, not only the transient connection reset the step's comment anticipates. The retry loop still recovers by re-running Sync() on the next attempt, so this is not fatal, but the comment's framing (a failure here past the first attempt or two is a real one, not a cold start race) undersells how easily attempt 1, or all three attempts if litellm is consistently slow to stop, can time out purely on curl's own budget rather than a genuine sync failure. Worth loosening --max-time closer to restart.go's own 30 second budget, or noting in the comment that a curl timeout here does not necessarily mean the sync failed server side.

.github/workflows/deploy-demo-box.yml
```diff
@@ -793,6 +793,60 @@ jobs:
           echo "SMOKE FAIL"
           exit 1
 
+      - name: Sync LiteLLM model list from the DB catalog (issue #713)
```
sakibsadmanshajib, line 809: This step runs after Smoke test public hostnames has already confirmed api-hive.scubed.co, control-hive.scubed.co and chat-hive.scubed.co are answering, meaning Caddy is already routing real public traffic to the freshly recreated stack by the time this step runs. WriteAndRestart always restarts the litellm container unconditionally (no diff check, confirmed in generator.go), so every future deploy now bounces litellm a second time, specifically after the box has already been certified live, not only during the earlier Rebuild and recreate changed services step. Any in-flight chat completion during that restart window fails for a real user hitting the demo box, on every single deploy, whether or not any route actually changed. The PR description's what this does not do section frames the unconditional restart purely as an extra-restart-per-deploy cost (a compute or idempotence concern); it does not call out that the restart now lands in a window where the box has already been proven to be serving public traffic, which is a different, customer-facing availability cost than restarting during the earlier recreate phase. Worth confirming this tradeoff was actually weighed in those terms, since the box already restarts things during deploy does not automatically extend to so one more restart after go-live is free.

.github/workflows/deploy-demo-box.yml
```diff
@@ -793,6 +793,60 @@ jobs:
...
+        # migration can update the DB and still never reach the running
+        # gateway. Calling this endpoint on every deploy makes the DB the
+        # only source of truth for every DB-managed route's upstream model,
+        # which is the fix issue #713 asked for rather than another env var.
+        #
+        # POST /internal/litellm/sync is guarded by RequireInternalToken
+        # (X-Internal-Token), the same internal auth edge-api already uses
+        # for service-to-service calls to control-plane, so no new secret is
+        # introduced here. CONTROL_PLANE_INTERNAL_TOKEN is read from the
+        # box's own .env, same pattern as the agent-engine launcher step
```
sakibsadmanshajib: Minor: the comment says CONTROL_PLANE_INTERNAL_TOKEN is read here the same pattern as the agent-engine launcher step above, but the two are not quite the same. The launcher step above (around line 403) takes the token from the secrets.CONTROL_PLANE_INTERNAL_TOKEN GitHub secret first via its env block, and only falls back to grepping .env when that secret is unset. This step skips the secret entirely and reads .env unconditionally. In practice this looks fine, arguably more correct, since docker-compose injects control-plane's actual InternalToken straight from .env (CONTROL_PLANE_INTERNAL_TOKEN: dollar-brace CONTROL_PLANE_INTERNAL_TOKEN colon-dash hive-local-internal-token), so reading .env directly is guaranteed to match what the running container enforces, whereas the launcher step's secret-first order could in principle diverge from .env if the two were ever rotated independently. I could not find a functional problem from this, just flagging that the same pattern claim is not literally accurate, in case the secret-first fallback in the launcher step was intentional for a reason not visible in this diff.

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Good catch, this is consistent with the pattern already used twice elsewhere in this file (master_key, HIVE_API_KEY), so I'm leaving it as is here rather than fixing one of three sites in isolation. Agreed a mktemp'd mode-600 header file would close this properly across all three, happy to do that as a follow-up if the team wants it, but keeping this PR scoped to the sync step itself.

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Confirmed and fixed in 5f31b1291. Added || true after the grep|head|cut pipeline so a fully missing CONTROL_PLANE_INTERNAL_TOKEN line no longer aborts one line early under pipefail. Verified locally: the script now reaches the intended ::error:: message and exits 1 in that case too, same outcome as before but with the right diagnostic.

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Fixed in 5f31b1291, switched to mktemp with a trap rm on EXIT, matching the pattern already used for this a few hundred lines up in this same file.

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Confirmed, this was a real gap, thanks for tracing it through restart.go's own timeout. Fixed in 5f31b1291, --max-time is now 35s (30s internal restart timeout plus a 5s buffer), so a slow-but-successful sync no longer risks being misreported as a failure.

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Fair, the PR body undersold this as a compute cost rather than a live-traffic one. Updated the PR body to state this plainly: this is a deliberate brief chat-interruption on every deploy, not just an idle restart, same tradeoff as the pre-existing --force-recreate step earlier in this same job, not a new class of risk this PR introduces, but worth stating explicitly rather than leaving implicit.

.github/workflows/deploy-demo-box.yml
sakibsadmanshajib: Fair, corrected in 5f31b1291. The token read here is unconditional from .env, not secret-then-.env-fallback like the launcher step above it, since compose always sources control-plane's real token from .env regardless of what other steps also have as a GitHub secret. Updated the comment to say that directly instead of claiming it is the same pattern.

PR #976, opened 2026-08-18
fix: add missing vendor/open-webui path to deploy-demo-box.yml filter

sakibsadmanshajib:
## Summary

`deploy-demo-box.yml`'s `push.paths` filter had no entry for `vendor/open-webui`, the forked chat frontend's source tree that `Dockerfile.open-webui` compiles directly from (`COPY vendor/open-webui ./`). A change under that tree merges to `main` and triggers no deploy at all. PR #971 (nav fixes, merged clean) is the live example: it touched only `vendor/open-webui`, and no deploy run followed. The fork build in PR #938 deployed only because it also touched `deploy/docker/Dockerfile.open-webui`, which the filter does cover, so past deploys were incidental rather than evidence the path worked.

The workflow's own comment already documents this exact failure class for `apps/web-console` (PR #786) and `supabase/migrations`. This is the same defect recurring a second time in the same file.

I audited every other Dockerfile the deploy job builds (`edge-api`, `control-plane`, `agent-console`, `web-console.prod`, `agent-engine`) plus the ones it does not (`sdk-tests-*`, `toolchain`, `desktop-linux`, non-prod `web-console`). Every COPY/ADD source across all of them falls under `apps/**`, `packages/**`, `deploy/docker/**`, or `go.work(.sum)`, all already present in the filter. `vendor/open-webui` was the only gap.

## Changes

- `.github/workflows/deploy-demo-box.yml`: add `vendor/open-webui/**` to `push.paths`, with a comment pointing at the new guard below.
- New `.github/ci/lint-deploy-paths-filter.mjs`: parses the deploy workflow's `push.paths` and every `deploy/docker/Dockerfile.*`'s COPY/ADD sources, fails if any source is not covered by the filter. Skips `--from=` stage copies, which read a prior build stage rather than the host filesystem.
- Wired into `.github/workflows/ci.yml`'s existing `repo-policy-lints` job (already a required check, already installs `node`+`yaml`), right after the sibling `lint-workflow-check-names.mjs` step. No new CI job.

Deliberately not added to `.github/branch-protection-main.json`'s required-checks list: `repo-policy-lints` is already required, so this step riding inside it already fails the PR loudly on a gap. Promoting it to its own named required check is a separate, more sensitive branch-protection change and out of scope here.

## Verification

Ran the new guard locally (`npm ci --ignore-scripts && node .github/ci/lint-deploy-paths-filter.mjs`):
- Passes after this fix: `Deploy path-filter coverage OK: every COPY/ADD source across 14 Dockerfiles under deploy/docker/ is covered...`
- Stashed only the `deploy-demo-box.yml` change and reran: fails loud, naming `vendor/open-webui/package.json`, `vendor/open-webui/package-lock.json`, and `vendor/open-webui` itself as uncovered. Confirms the guard would have caught the original bug.
- Re-ran the sibling `lint-workflow-check-names.mjs`: still passes, no regression (23 check names, 6 required contexts).
- Both edited workflow YAMLs parse cleanly.

## Deploy note

Merging this PR changes `.github/workflows/deploy-demo-box.yml` itself, which is already in its own filter, so the merge will trigger a real deploy to the demo box. That deploy will also carry every other merged-but-undeployed change currently sitting on `main` (including PR #971). Given the open P0 on the shared Supabase connection pool, hold this merge for explicit go-ahead rather than merging on green CI alone.

## Buglog entry

Second instance of the same defect class in the same file (first: PR #786, `apps/web-console`).

```json
{"error_message":"vendor/open-webui had no entry in deploy-demo-box.yml's push.paths filter; a change under that tree merged to main and triggered no deploy","root_cause":"the paths filter is a hand-maintained allowlist and the fork's frontend source tree (added when the chat frontend was forked to compile from source) was never added to it, so PR #971 (nav fixes touching only vendor/open-webui) merged clean with zero deploy run","fix":"added vendor/open-webui/** to the filter; added .github/ci/lint-deploy-paths-filter.mjs (wired into ci.yml's repo-policy-lints required check) which parses every deploy/docker/Dockerfile.*'s COPY/ADD sources against the filter and fails loud on the next missing entry instead of silently never deploying","tags":["ci","deploy","paths-filter","open-webui","recurring"]}
```

### DIFF
--- a/.github/ci/lint-deploy-paths-filter.mjs
+++ b/.github/ci/lint-deploy-paths-filter.mjs
@@ -21,15 +21,25 @@
 // It does not (and cannot, without executing the compose profile logic)
 // check docker-compose.yml's own inputs (env files, bind mounts, image
 // digests) or the workflow's own script/scripts: entries -- those are listed
-// by hand in the paths filter today and are not COPY sources. It also only
-// understands the two glob shapes this filter actually uses: an exact
-// literal ('go.work') and a directory wildcard ('apps/edge-api/**'). A third
-// shape showing up in the filter would need this matcher extended.
+// by hand in the paths filter today and are not COPY sources. On the filter
+// side it only understands the two glob shapes this filter actually uses: an
+// exact literal ('go.work') and a directory wildcard ('apps/edge-api/**'). A
+// third shape showing up in the filter would need this matcher extended.
+//
+// A Docker COPY source can itself carry a `*` (`go.work.sum*`, meaning "copy
+// this file if it exists"). Docker resolves that against the real build
+// context, so this script does the same rather than string-matching the
+// glob literally: it expands the wildcard against the actual files on disk
+// and checks each real match. Comparing the un-expanded glob text against
+// the filter would have made `go.work.sum*` equal to the literal filter
+// entry `go.work.sum` by construction, silently approving a future
+// `go.work.sum.bak` that the real Docker build would also pick up and that
+// the filter does not cover (CodeRabbit review, PR #976).
 //
 // Run: node .github/ci/lint-deploy-paths-filter.mjs
 
 import { readFileSync, readdirSync } from 'node:fs';
-import { join } from 'node:path';
+import { dirname, join } from 'node:path';
 import { parse } from 'yaml';
 
 const WORKFLOW_FILE = '.github/workflows/deploy-demo-box.yml';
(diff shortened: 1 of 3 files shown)


sakibsadmanshajib:
## Adversarial review summary (pipeline stage 6)

**CodeRabbit CLI (local, `coderabbit review --committed --base main`):** ran successfully, 1 major finding.

> `go.work.sum*` is reduced to `go.work.sum`. A future file such as `go.work.sum.bak` can match the Docker source pattern while remaining outside the exact `go.work.sum` workflow filter. The guard would then pass although the deploy workflow would not trigger.

Valid. Fixed in 0fec64dd0: the guard now expands each wildcard COPY source against the real files on disk (the way Docker itself resolves the glob at build time) instead of string-matching the un-expanded glob text. Self-testing that fix (`touch go.work.sum.bak`) also caught a slicing bug the rewrite introduced, which made every root-level wildcard silently expand to nothing; fixed in the same commit. Re-verified: the `go.work.sum.bak` repro now fails the guard by name, the original `vendor/open-webui` regression test still passes, and the sibling `lint-workflow-check-names.mjs` guard is unaffected.

**CodeRabbit GitHub App (bot review on this PR):** SKIPPED. Rate limited on this account ("Review rate limited", next available in ~54 min at the time checked). The `CodeRabbit` status check itself reports pass/neutral, not a real review; do not read it as one.

**ecc:code-review skill:** SKIPPED. This subagent's toolset (Read/Write/Edit/Bash only, no Skill or Task/Agent tool) cannot invoke it. Structural capability gap, not a bypass.

**Plain adversarial pass (self, manual):** diff is CI/workflow-only (a new `.mjs` static-analysis script plus two workflow YAML edits), no auth/money/input-parsing path, so no domain specialist or `/codex:adversarial-review` added per the dynamic selector. Checked: instruction-parsing regex anchoring (comments/RUN lines cannot false-match COPY/ADD), `--chown=`/other COPY flags correctly stripped without skipping the line, `--from=` stage copies correctly excluded, backslash line-continuation joining scoped safely, the new `repo-policy-lints` step correctly gated by the same `changes` job output that this very diff triggers (`*.mjs` and workflow-YAML changes both fall to `run="true"`, confirmed by reading that job's case statement), and confirmed via `git diff main...HEAD` that no unrelated file (a background hook had transiently modified `go.work.sum` in the local worktree; discarded before commit, never staged).

No domain specialist required for a CI/workflow-only diff with no auth, money, or input-parsing surface.

PR #982, opened 2026-08-18
feat: make the self-hosted supabase profile bootable and able to authenticate

sakibsadmanshajib:
## What this does

Stage 1 of the self-hosted Supabase migration: make the enterprise profile actually boot and actually authenticate. It had never been run end to end, and standing it up from scratch surfaced six independent reasons it could not have worked.

Alongside only. Nothing here points any running deployment at the self-hosted stack, and the cloud project is untouched.

## The two blockers this was dispatched for

**1. Symmetric-only GoTrue publishes an empty JWKS.** `GOTRUE_JWT_SECRET` alone means HS256 signing, and GoTrue's JWKS handler skips every symmetric key by design (`internal/api/jwks.go`). edge-api validates with `jwt.WithKeySet` against that endpoint and fails its initial refresh, so it would not have booted, and no token would have validated if it had. Fixed with `GOTRUE_JWT_KEYS` carrying an EC P-256 key, generated by the new `scripts/generate-enterprise-jwt-keys.py`. PostgREST and Storage receive the matching verification set, which carries the legacy symmetric key too, so existing HS256 anon and service_role keys keep working next to the new ES256 user tokens.

**2. HTTPS-only JWKS.** The guard is not relaxed. The new `caddy-supabase` service terminates real TLS using Caddy's own local certificate authority, and edge-api trusts that single authority through the new `SUPABASE_JWKS_CA_FILE`. Chain and hostname verification still happen; the variable adds an authority, it does not disable anything. A named CA file that is absent or holds no certificate is a boot failure rather than a silent fall back to the system pool.

## Everything else found while proving it

- **No Kong.** `caddy-supabase` fronts GoTrue, PostgREST and Storage on one origin with the hosted `/auth/v1`, `/rest/v1` and `/storage/v1` prefixes, so one `SUPABASE_URL` stays correct for control-plane, the Supabase client libraries and the seeding scripts.
- **GoTrue pin bumped v2.170.0 to v2.189.0.** The old tag predates the OAuth 2.1 authorization server that Open WebUI signs in through. Verified by running both images.
- **PostgREST healthcheck could never pass.** That image has no `/bin/sh`, so every `CMD-SHELL` probe returns "exec: /bin/sh: no such file or directory" and the container is unhealthy forever. Storage waited on `service_healthy`, so the profile could never finish starting.
- **Storage healthcheck probed `localhost`**, which resolves to `::1` first while the server binds IPv4 only.
[...]

PR #983, opened 2026-08-18
ci: stop CI jobs sharing the live database, boot a throwaway Postgres instead

sakibsadmanshajib:
# CI decoupling from the live database

**PR: https://github.com/sakibsadmanshajib/hive/pull/983** (branch `ci/decouple-ci-from-live-db`, rebased onto main after #982 merged)

| Job | Outcome |
| --- | --- |
| `ci.yml` `live-integration` | converted, green in CI on a throwaway Postgres |
| `ci.yml` `web-e2e` | converted, own Postgres plus GoTrue plus PostgREST |
| `pr-cleanup.yml` `cleanup` | deleted |
| `agent-visual-proof.yml` `capture` | NOT converted, blocked by an https-only JWKS check |
| `owui-nightly.yml` `owui-e2e` | NOT converted, blocked by the hosted-only OAuth server |
| `deploy-demo-box.yml` `migrate`, `deploy` | untouched, not in the diff |

## One approach, not two

Both converted jobs use `pgvector/pgvector:pg17`, the image `ci.yml`'s own
`go-tests` job already runs its RLS suites against. An earlier iteration used
`supabase/postgres` to obtain pg_cron; that is reverted. On that image the
`postgres` role is not a superuser, so pg_cron and writes into the auth schema
each needed a second connection as `supabase_admin`, and it segfaulted its own
backend twice (on `GRANT ... TO CURRENT_USER`, and on catching a reserved-role
refusal inside a PL/pgSQL handler), taking the cluster into recovery each time.

pg_cron is stated, never silently skipped. `20260729_02` guards on
`pg_available_extensions` and creates its config table and purge procedure
either way; the only difference is an unscheduled nightly purge on a database
that never survives to 21:00 UTC. `scripts/ci-throwaway-db.sh` prints which of
the two branches the file took on every run.

## Made to fail on purpose

Every assertion below was sabotaged deliberately and observed to fail, on the
grounds that a check nobody has seen go red is not yet a check.

### `scripts/ci-throwaway-db.sh`

Control: `throwaway database ready: 88 of 88 migrations executed`, exit 0.

Sabotage 1, one ledger row flipped to `source='baseline'`, which is the exact
issue #676 shape of a migration recorded rather than executed:

```
::error::only 87 of 88 migrations were executed on this throwaway database
::error::1 migration(s) were recorded as baseline history without ever running
throwaway database bootstrap FAILED with 2 problem(s)     exit 1
```

Sabotage 2, `DROP TABLE public.rag_documents CASCADE` after a clean chain, so
the applier reports nothing pending and the schema is wrong anyway:

```
::error::public.rag_documents is missing after the migration chain
throwaway database bootstrap FAILED with 1 problem(s)     exit 1
```

### `scripts/ci-seed-api-key.sh`

Control: `seeded tenant, account, api key, policy and credit grant`, exit 0.
Three mutated copies, one assertion each, all exit 1:

```
A, tenant_billing_accounts insert removed:
  ::error::seed check failed: the key's account maps to a tenant (expected '1', got '0')
B, credit grant removed:
  ::error::seed check failed: the account has a positive credit balance (expected 't', got 'f')
C, policy written as allow_all_models = false:
  ::error::seed check failed: the key may invoke every alias (expected 't', got 'f')
```

Those three are not arbitrary. A missing mapping row is a 403
`account_not_provisioned` on every call (issue #717). A zero balance makes the
reservation guard in `apps/edge-api/internal/inference/reservation_guard.go`
refuse every completion while the key still authenticates. A narrowed policy
would quietly shrink what the SDK suites can invoke.

### The jobs themselves

`live-integration`: locally, `GET /v1/models` with the seeded key answers
`HTTP 200` and with a bogus key answers `401 invalid_api_key`, so the smoke step
is a fact about the seeded key rather than a rubber stamp.

`web-e2e`: proven the hard way in run 32266955684. An earlier revision left the
hosted project's `SUPABASE_SERVICE_ROLE_KEY` in the Playwright step's own env
block, so the seeder presented the wrong admin key to this run's own GoTrue.
The job went RED with fourteen named spec failures and
`createUser failed: invalid JWT: token signature is invalid`, rather than
skipping quietly to green. That is the failure mode this whole exercise is
about, and this job demonstrated it does not have it.

`scripts/ci-supabase-stack.sh`: its own gateway probe fired for real during
development, `::error::PostgREST never served public.tenants through the
gateway (last status 403)`, before the API-role grants were added.

## 1. `live-integration` (`Live integration (SDK tests + smoke)`, required check)

Starts its own Postgres, applies `supabase/migrations` through
`scripts/apply-migrations.sh`, seeds its own tenant, account, API key and credit
grant, and points the booted stack at that. Reads none of `SUPABASE_DB_HOST`,
`SUPABASE_DB_PORT`, `SUPABASE_DB_USER`, `SUPABASE_DB_NAME`,
`SUPABASE_DB_PASSWORD` or `HIVE_API_KEY`.

**Verifies, unchanged:** both images boot healthy, `/health` on both, `GET
/v1/models` answers 200 to a real Hive API key, and the JavaScript, Python and
Java SDK suites pass against the booted stack over real provider traffic.

**Confirmed in CI**, run 32188152497, job 95878620737, conclusion success:

```
applied 88 migration(s)
throwaway database ready: 88 migrations executed
seeded tenant, account, api key, policy and credit grant
control-plane: healthy
edge-api: healthy
GET /v1/models -> HTTP 200
== sdk-tests-js  (exit 0)   == sdk-tests-py  (exit 0)   == sdk-tests-java  (exit 0)
  ✓ Chat Completions > returns a valid chat completion via SDK  3489ms
  ✓ Streaming Chat Completions > streams chat completion chunks via SDK  3502ms
  ✓ Responses API > returns a valid response via SDK  3855ms
  ✓ Completions (legacy) > returns a valid text completion via SDK  442ms
  ✓ Embeddings > returns valid embeddings via SDK  10731ms
  ✓ Embeddings > supports batch input  7532ms
```

## 2. `web-e2e` (`Web E2E (full stack)`, required check)

A database alone could not move this job and the halves are not separable. Its
fixture seeder (`apps/web-console/tests/e2e/support/e2e-fixture-seed.mjs`)
speaks supabase-js: `auth.admin.createUser` against `/auth/v1`, `.from(table)`
against `/rest/v1`. The browser signs in through `@supabase/ssr`. And GoTrue's
custom access token hook (`supabase/migrations/20260516_07`) runs INSIDE the
database holding the user rows, raising `no_active_membership` for a user with
no `tenant_users` row, so hosted auth over a throwaway database fails every
login.

`scripts/ci-supabase-stack.sh` therefore brings up GoTrue, PostgREST and a
one-file nginx front mapping `/auth/v1` and `/rest/v1` onto them, because
supabase-js takes one base URL and appends those prefixes itself.

**Ordering is load bearing:** GoTrue migrates the auth schema first (thirteen
Hive migrations foreign-key to `auth.users`), then the Hive chain runs, then
PostgREST starts so its schema cache holds the real tables.

**One platform behaviour the repo never had to write:** table grants for `anon`,
`authenticated` and `service_role` in `public`. Parity, not a weakening: `anon`
and `authenticated` remain gated by RLS and `service_role` is `BYPASSRLS`
exactly as on the hosted project.

**Confirmed locally, end to end, before the workflow was touched:** GoTrue
applied its own 54 migrations and owns the auth schema; PostgREST loaded 69
relations and 18 functions; the real unmodified seeder exits 0; and a password
grant returns HTTP 200 with a JWT carrying the hook's claims:

```
tenant_id, tenants = [{"id": "...", "role": "OWNER"}], role = "OWNER",
owui_role, app_metadata.hive_email_verified = true
```

**Two coverage points, deliberately not left to chance:**

- `console-workspace-admin.spec.ts` skipped itself whenever
  `E2E_PLATFORM_ADMIN_EMAIL` was unset, and that hosted account cannot exist on
  a throwaway Supabase. It now points at the run's own seeded verified user,
  which the seeder makes OWNER of the run tenant, so the spec runs every time
  rather than sometimes. More coverage than before, not less.
- `HIVE_API_KEY` is dropped, changing no spec outcome: it fed only
  `openai-sdk.spec.ts`, which skips on `!HIVE_API_KEY || !EDGE_BASE_URL`, and
  this job has never set `EDGE_BASE_URL`.

**A path gate that would have gone stale:** the `changes` job excluded
`supabase/migrations/*` from `web_e2e` because the job ran against an
already-live schema. It applies the chain itself now, so the exclusion is
reversed and the scripts it runs are added. Left alone, a migration-only pull
request would never have exercised the browser flow against its own migration.

## 3. `pr-cleanup.yml`, deleted

Its purge resolved `HIVE_API_KEY` to an account and deleted that account's
`usage_events`, `batches`, rollups, budget windows, rate policies, files and
uploads: exactly and only `live-integration`'s footprint, which no longer
exists.

Its second half deleted the merged head branch, which was real work and is not
dropped. The repository's `delete_branch_on_merge` setting was `false` and is
now `true`, read back to confirm. Native equivalent, no workflow, no secrets, no
runner minutes.

Never a required status check, so nothing vanished from the gate.

**Correction, found in review.** The description above is right about the
database arm and wrong to stop there. That workflow had a second purge: it
listed objects with `SELECT storage_path FROM public.files` and
`public.uploads` for the CI account and removed those keys with `aws s3 rm`.
Both converted jobs still point at the **hosted** Supabase Storage buckets
(`ci.yml` lines 925-931 and 1331-1337), so that arm was not obsolete. The
failure mode changed shape rather than disappearing: an uploaded object used to
be left behind but stayed findable, because the row naming it survived in the
shared project; now the row dies with the throwaway database and the object is
orphaned in a shared bucket with nothing left that names it.

Latent rather than active, which is why it is a follow-up and not a blocker: no
suite in either job uploads a file today, confirmed against run 32378996099.
Tracked in #984.

## 4. `agent-visual-proof.yml` `capture`, NOT converted

Blocked by a security control I will not weaken.
`apps/edge-api/cmd/server/main.go`:

```go
if !strings.HasPrefix(strings.ToLower(jwksURL), "https://") {
    return jwtAuthEnv{}, fmt.Errorf("SUPABASE_JWKS_URL must be https (got %q)", jwksURL)
}
```

That job sets `SUPABASE_JWT_ISSUER` and `SUPABASE_JWKS_URL` and then asserts
edge-api did not log `JWT auth wiring skipped`, precisely so a browser session
that cannot authenticate fails loudly instead of yielding a screenshot of a
disabled panel. A throwaway GoTrue inside a runner is plain http, and a
self-hosted GoTrue given a symmetric `GOTRUE_JWT_SECRET` has no asymmetric key
set to publish in the first place. Converting it means terminating TLS in front
of the throwaway stack with a certificate the service containers trust, which
needs a volume on `docker-compose.yml`'s shipped service definitions, or
relaxing an https check on the JWT path. The second is off the table and the
first is a different change of shape from this one.

This does not affect `web-e2e`, which has never set either variable, so
edge-api's JWT wiring was already off there.

## 5. `owui-nightly.yml` `owui-e2e`, NOT converted

Two independent blockers, either sufficient.

Its login journey IS Supabase's OAuth 2.1 authorization-code flow:
`SUPABASE_OAUTH_CLIENT_ID`, `SUPABASE_OAUTH_CLIENT_SECRET`, `POST
/auth/v1/oauth/token`, `/oauth/oidc/callback`. That authorization server is a
hosted-platform feature; self-hosted GoTrue v2.170.0, the version this repo
pins, has no OAuth server. Converting the job would delete the journey it
exists to test.

It also derives `SUPABASE_JWKS_URL` from `SUPABASE_URL` for edge-api, so it
hits the same https-only check as `agent-visual-proof`.

Note what this does and does not mean. Neither job needs a shared live
*database*; both need the hosted Supabase *auth product*. When the self-hosted
Supabase stack from #982 can present an https JWKS, both become
straightforward, and `scripts/ci-supabase-stack.sh` is the piece they will
reuse.

## Deploy jobs untouched

`deploy-demo-box.yml` is not in the diff. The branch touches
`.github/ci/test-db-bootstrap.sql`, `.github/workflows/ci.yml`,
`.github/workflows/pr-cleanup.yml` (deleted), `scripts/ci-throwaway-db.sh`,
`scripts/ci-seed-api-key.sh` and `scripts/ci-supabase-stack.sh`.

## Required checks

Nine required contexts, none renamed, removed or newly skipped. `Live
integration (SDK tests + smoke)` is exercised on this pull request under the
`run-live-integration` label rather than skipping its way to green, and `Web
E2E (full stack)` runs because the diff touches `.github/workflows/ci.yml`.

## #982

Merged, so `supabase_auth_admin` and `supabase_storage_admin` now exist in
`deploy/supabase/init/00-extensions.sql`. This branch is rebased onto it and
duplicates nothing. `.github/ci/test-db-bootstrap.sql` still creates those roles
when absent, because `go-tests` runs that file WITHOUT `00-extensions.sql`
first; the guard makes it a no-op wherever #982's version has already run.

## Stage 6 adversarial review

Run against `68d901ff3`. Streams, and what each returned:

| Stream | Result |
| --- | --- |
| CodeRabbit CLI | Ran to completion over all six files. 0 findings. |
| CodeRabbit GitHub bot | SKIPPED. It declines to review a draft pull request; the CLI run is the stream of record. |
| `ecc:code-review` | 1 medium, 1 low. |
| Plain adversarial pass | 2 medium, 3 low, 2 nits. |
| Security pass over the roles, grants and auth gateway | 0 critical, 0 high, 1 medium, 1 low. |
| shellcheck, `-S warning`, all three new scripts | Clean. |

### The preflight guard was mutation tested rather than taken on trust

The real nginx config and the real assertion function were extracted from
`scripts/ci-supabase-stack.sh` into a harness and run against a GoTrue-shaped
stub that refuses the `apikey` preflight the way rs/cors actually does. One
mutation at a time:

| Mutation | Result |
| --- | --- |
| baseline, unmutated | green |
| OPTIONS short-circuit removed, the original bug | red, on the missing allow-origin branch |
| allow-headers hardcoded without `apikey` | red |
| POST removed from allow-methods | red |
| allow-origin set to a foreign origin | red |
| preflight answered 404, headers still present | red |
| `apikey` replaced by the lookalike `x-vendor-apikey-hint` | red |

All five branches go red. The claim in `68d901ff3` holds.

Separately confirmed against the real `supabase/gotrue:v2.170.0` image behind
the real gateway config, not only the stub: the browser-shaped preflight returns
`204` carrying all four CORS headers with `apikey` included, and a real
`GET /auth/v1/settings` returns exactly one `Access-Control-Allow-Origin`.

### What the guard did not cover, now fixed

`assert_preflight` only ever issues an OPTIONS request, and duplication of
`Access-Control-Allow-Origin` cannot happen on a preflight at all, because
`return 204` short-circuits before `proxy_pass` and the upstream is never
reached. It is reachable on the real response. Moving the `add_header`
directives out of the `if` to location level leaves every preflight assertion
green while the real response carries two of them, and a browser rejects that as
hard as a missing one. `header_value()` taking `head -1` would have hidden it
even if something had looked.

`e019a0c98` adds `assert_single_cors_origin` on `/auth/v1`, which goes red under
exactly that mutation. `/rest/v1` is deliberately not probed: the gateway adds
no header there, so duplication is impossible by construction.

That commit also corrects a comment. The claim that giving `/rest/v1` its own
preflight block "would emit the header twice" is not true, measured at one
header, for the same reason. The real reason to leave that prefix alone is that
PostgREST already answers the preflight itself.

### Findings cleared

- Stale rationale and a dead diagnostic. The web-e2e header still attributed the
  job's gating design to Supavisor contention, and the "Name a shared
  session-pool exhaustion on failure" step still grepped `EMAXCONNSESSION`. The
  job now opens no connection to that pooler, so the step was unreachable and
  the header pointed the next investigation at the wrong system. Step deleted,
  header rewritten to the gate's surviving reason.
- `live-integration` took on the same added work as web-e2e and none of the
  budget. Both are 35 now. The measurement in the web-e2e comment was also
  wrong: the throwaway step takes 28 seconds, not two and a half minutes.
- `ci-supabase-stack.sh` hardcoded the database user, password and name for the
  GoTrue and PostgREST DSN while every psql call used the caller's libpq
  environment. A non-default `PGPASSWORD` produced a run where every psql step
  passed and only GoTrue failed, reported as "GoTrue never became healthy".
  All three now default from the environment, and `--db-name` is checked
  against `PGDATABASE` because it is parsed after those defaults.
- `console-workspace-admin.spec.ts` signed in as the seeded fixture user without
  seeding it, working only because `auth-shell.spec.ts` sorts earlier under
  `workers: 1`. New coupling, since the account used to pre-exist on the hosted
  project. It now calls `reseedFixtures` in its own `beforeEach`.
- `test-db-bootstrap.sql` claimed `supabase/postgres` is one of two images that
  run it. Every caller reaches `pgvector/pgvector:pg17`. The guards are kept and
  the comment now says they defend against a future caller rather than describe
  a current one, and that nothing in CI exercises them.
- The web-e2e teardown left the docker network behind.

### Checked and deliberately not changed

- Binding the throwaway Postgres to `127.0.0.1:5432:5432` was suggested and
  would break the job. Compose services reach the database at the docker bridge
  gateway address, which does not arrive over loopback. Measured both ways:
  published on all interfaces it is reachable from a container, bound to
  loopback it is not. The exposure is a hosted runner with no inbound
  reachability, guarding a database created minutes earlier with no data of
  value.
- The `$GITHUB_ENV` append loop has no line-shape guard, which is the classic
  injection shape. Every value that can reach it was traced and none can carry a
  newline today, so it is not exploitable as written. Worth knowing that is a
  property of the inputs rather than of the loop.

## Buglog entry

```json
{"date":"2026-08-20","title":"CI jobs shared one Supabase project with live traffic","error_message":"FATAL: (EMAXCONNSESSION) max clients reached in session mode","root_cause":"Seven CI jobs pointed at the hosted Supabase project that also serves the live demo stack. Its Supavisor session-mode pool is capped at 15 clients, so concurrent CI stacks exhausted it and failed unrelated jobs with timeouts that read as front end bugs rather than as contention.","fix":"live-integration and web-e2e boot their own pgvector Postgres and apply supabase/migrations through scripts/apply-migrations.sh; web-e2e additionally stands up GoTrue and PostgREST behind one nginx origin, because its fixture seeder speaks supabase-js and GoTrue's custom access token hook runs inside the same database as the user rows. pr-cleanup.yml is deleted, its branch deletion replaced by the repository's delete_branch_on_merge setting.","tags":["ci","postgres","supabase","gotrue","shared-state"]}
```

```json
{"date":"2026-08-19","title":"A bare apt-get in a CI step hung for the job's whole budget with no output","error_message":"##[error]The operation was canceled.","root_cause":"sudo apt-get update -qq at the top of a workflow step blocked for 25 minutes on a hosted runner and printed nothing, so the job was cancelled at its timeout with an empty step log and no indication of where it stopped.","fix":"Install psql only when it is missing (the hosted image already ships it), bound each apt call with timeout 180, retry three times, and fail by name if psql is still absent afterwards.","tags":["ci","github-actions","apt","timeout"]}
```

```json
{"date":"2026-08-20","title":"Self-hosted GoTrue refuses the CORS preflight supabase-js actually sends, so every credentialed browser spec died naming nothing","error_message":"signInWithPassword rejected with \"Failed to fetch\"; every credentialed spec then failed on a 25 second waitForURL timeout inside signIn()","root_cause":"GoTrue's CORS allow-list is a fixed set that does not contain `apikey`, and supabase-js sends that header on every request. The browser preflight therefore asks for a header GoTrue will not allow, and GoTrue answers 204 with no Access-Control-* headers at all, so the browser blocks the request before it is sent. On a hosted Supabase project Kong terminates CORS at the edge and never asks GoTrue, so this never surfaces there. The console's auth-error allow-list then degraded the unrecognized \"Failed to fetch\" to generic copy, and curl could not see any of it because a preflight refusal is invisible to a non-browser client.","fix":"Terminate the preflight at the nginx gateway for /auth/v1/, echoing the requested headers back the way PostgREST already answers the identical preflight, and short-circuit only the OPTIONS so the real request still reaches GoTrue and its own Access-Control-Allow-Origin is not duplicated. Assert the browser's own request shape in scripts/ci-supabase-stack.sh before any value is handed to the job, plus assert_single_cors_origin on the real response, because duplication is unreachable on a preflight and reachable on the actual response.","tags":["ci","cors","gotrue","nginx","playwright","supabase-js"]}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database initialization to safely preserve existing authentication functions and support repeatable setup.
  * Increased reliability of live integration and browser-based testing with isolated, disposable environments.
  * Added fixture reseeding to keep workspace administration tests consistent.

* **Chores**
  * Added automated setup and validation for temporary databases, Supabase services, API keys, authentication, and test data.
  * Removed obsolete pull request cleanup automation.

The PR moves two required integration jobs from the hosted Supabase database to isolated, disposable PostgreSQL environments.
- Boots and validates a throwaway database for live SDK integration tests.
- Adds a local GoTrue, PostgREST, and nginx gateway stack for browser E2E tests.
- Seeds run-local API credentials and fixtures, expands migration-sensitive path gating, and removes the obsolete shared-database cleanup workflow.
- Makes the workspace-admin test seed its own fixtures rather than depending on test execution order.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/ci.yml | Rewires live integration and web E2E jobs around isolated databases, local Supabase services, run-scoped credentials, stronger readiness checks, and explicit teardown. |
| scripts/ci-throwaway-db.sh | Applies and validates the complete migration chain against a disposable PostgreSQL instance. |
| scripts/ci-supabase-stack.sh | Boots GoTrue, PostgREST, and nginx against the throwaway database and exports the local Supabase environment consumed by E2E. |
| scripts/ci-seed-api-key.sh | Seeds the tenant, billing mapping, active API key, unrestricted model policy, and credit grant needed by live integration tests. |
| .github/ci/test-db-bootstrap.sql | Makes test-only role, schema, table, and auth-function setup tolerant of pre-existing Supabase objects. |
| apps/web-console/tests/e2e/console-workspace-admin.spec.ts | Reseeds fixtures before each workspace-admin test so execution no longer depends on another spec running first. |
| .github/workflows/pr-cleanup.yml | Removes cleanup automation made obsolete by per-job database isolation. |

</details>

```mermaid
flowchart LR
  LI[Live integration job] --> DB1[(Throwaway pgvector Postgres)]
  LI --> Seed[Seed tenant, account, API key, policy, credits]
  Seed --> Stack[Control plane and edge API]
  Stack --> SDK[JavaScript, Python, and Java SDK suites]

  E2E[Web E2E job] --> DB2[(Throwaway pgvector Postgres)]
  DB2 --> Auth[GoTrue]
  DB2 --> Rest[PostgREST]
  Auth --> Gateway[nginx Supabase gateway]
  Rest --> Gateway
  Gateway --> Web[Next.js web console]
  Web --> Browser[Playwright suite]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix: clear the bot review findings, incl..."](https://github.com/sakibsadmanshajib/hive/commit/1a525b41384f728159dee58a6e924c0d0bd3d69b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55273300)</sub>

**Context used:**

- Knowledge Base — [Web console: authenticated tenant operations](https://app.greptile.com/scubed/-/custom-context/knowledge-base/sakibsadmanshajib/hive/-/docs/web-console.md)
- Knowledge Base — [Control-plane identity and access](https://app.greptile.com/scubed/-/custom-context/knowledge-base/sakibsadmanshajib/hive/-/docs/control-plane-identity-access.md)


sakibsadmanshajib (commented):
## Review summary

Streams run against `68d901ff3`, findings below are already addressed on the branch.

| Stream | Result |
| --- | --- |
| CodeRabbit CLI (`coderabbit review --agent -t all --base main`) | Ran to completion over all six files. **0 findings.** |
| CodeRabbit GitHub bot | **SKIPPED**, and not by choice: it declines to review while a pull request is a draft. The CLI run above is the stream of record. It will run once this is marked ready. |
| `ecc:code-review` | Ran. 1 medium, 1 low. |
| Plain adversarial pass | Ran. 2 medium, 3 low, 2 nits, plus a long list of checks it constructed the broken state for and confirmed can genuinely go red. |
| Security pass (roles, grants, auth gateway) | Ran. 0 critical, 0 high, 1 medium, 1 low. |
| shellcheck (`-S warning`, all three new scripts) | Ran. Clean. |

## On the preflight guard, which was the newest and least reviewed code

I did not take the previous commit's word for it. I extracted the real nginx config and the real assertion function out of `scripts/ci-supabase-stack.sh` into a harness, stood the gateway up against a GoTrue-shaped stub that refuses the `apikey` preflight the way rs/cors actually does, and mutated one thing at a time.

All five branches of the guard go red under their own mutation, and the unmutated script is green:

| Mutation | Result |
| --- | --- |
| baseline, unmutated | green |
| OPTIONS short-circuit removed (the original bug) | red, on the missing allow-origin branch |
| allow-headers hardcoded without `apikey` | red |
| POST removed from allow-methods | red |
| allow-origin set to a foreign origin | red |
| preflight answered 404 with the headers still present | red |
| `apikey` replaced by the lookalike `x-vendor-apikey-hint` | red |

So the claim in `68d901ff3` holds. What it does not cover is in the first inline comment below.

I also ran the real `supabase/gotrue:v2.170.0` image behind the real gateway config rather than only the stub, and confirmed the browser-shaped preflight comes back `204` carrying all four CORS headers with `apikey` included, and that a real `GET /auth/v1/settings` returns exactly one `Access-Control-Allow-Origin`.

## What changed on the branch

- `e019a0c98` adds the real-response CORS assertion and corrects the duplication comment.
- `eadc83557` corrects the bootstrap SQL caller attribution and tears down the docker network.
- `d6892a7d7` clears the adversarial findings: stale web-e2e rationale, the dead EMAXCONNSESSION step, the live-integration timeout asymmetry, the libpq divergence, and the spec seeding coupling.

One finding is deliberately not fixed here and is filed as #984 instead: deleting `pr-cleanup.yml` also dropped a storage purge arm that is not obsolete.

scripts/ci-supabase-stack.sh
```diff
@@ -0,0 +1,415 @@
...
+# it could not see a second copy even if it looked.
+#
+# Verified by mutation before this was written: moving the add_header out of the
+# `if` leaves every preflight assertion green while the real response carries
+# two Access-Control-Allow-Origin headers.
+#
+# /rest/v1 is deliberately not probed here. The gateway adds no header on that
+# prefix at all, so duplication is impossible there by construction, and only
+# the path that can actually break is worth an assertion.
+assert_single_cors_origin() {
```
sakibsadmanshajib: This is the one thing I could not make the preflight guard catch, so I added it here.

Every branch of `assert_preflight` does go red under mutation, but it only ever issues an OPTIONS request, and header duplication cannot happen on a preflight at all: the gateway's `return 204` short-circuits before `proxy_pass`, so the upstream is never reached and never contributes a second copy. Duplication is reachable on the real response, where the gateway proxies to GoTrue and GoTrue sets its own `Access-Control-Allow-Origin`.

I proved that is not hypothetical. Moving the `add_header` directives out of the `if` block up to location level leaves every preflight assertion green while the real response comes back carrying two `Access-Control-Allow-Origin` headers. A browser rejects a duplicated value as hard as a missing one, so that would fail every credentialed spec inside `signIn()` on a 25 second timeout, naming nothing, which is the exact failure this branch just spent two commits removing.

`header_value()` taking `head -1` would have hidden it even if something had looked.

This probes `/auth/v1` only, on purpose. The gateway adds no header on `/rest/v1`, so duplication is impossible there by construction, and an assertion on a path that cannot break is not worth having. Verified against the real `supabase/gotrue:v2.170.0` image, not just the stub: unmutated returns exactly one header, mutated returns two and this goes red.

scripts/ci-supabase-stack.sh
```diff
@@ -0,0 +1,415 @@
...
+    # "Failed to fetch", which apps/web-console/lib/auth/auth-error.ts (an
+    # allow-list by design) degrades to generic "Something went wrong on our
+    # end" copy. Every credentialed spec then dies in its own signIn() helper on
+    # a navigation timeout, 25 seconds at a time, naming nothing.
+    #
+    # On a hosted Supabase project this never surfaces because Kong terminates
+    # CORS at the edge and never asks GoTrue about it. This is that same
+    # termination, and it is the reason the gateway exists rather than an extra.
+    #
+    # PostgREST deliberately gets no such block: it already answers this exact
```
sakibsadmanshajib: The reason given here is not the real one, and I only noticed because I tried to reproduce it.

Adding a preflight block to `/rest/v1` would not emit the header twice. I measured it: with that block in place the response carries exactly one `Access-Control-Allow-Origin`, for the same reason duplication cannot happen on any preflight, which is that `return 204` never reaches PostgREST.

The real reason to leave that prefix alone is simply that PostgREST already answers this preflight itself, so a block there would be redundant. Worth correcting rather than leaving, because a comment that names the wrong mechanism sends the next reader hunting for a bug that is not there, and away from the one that is, which is duplication on the real response.

scripts/ci-supabase-stack.sh
```diff
@@ -0,0 +1,415 @@
...
+# already caught by the auth.users probe; a user or password divergence was not
+# caught at all.
+db_name="${PGDATABASE:-hive_ci}"
+db_user="${PGUSER:-postgres}"
+# Unquoted on purpose. An assignment does not word-split, and this keeps the
+# repository secret scanner from reading an environment-variable default as a
+# hardcoded credential. Nothing here is a secret in any case: the throwaway
+# Postgres is created with this value minutes earlier in the same job and is
+# destroyed with the runner.
+db_password=${PGPASSWORD:-postgres}
```
sakibsadmanshajib: These three build the DSN handed to GoTrue and PostgREST, but every `psql` call in this script uses the caller's `PGUSER`, `PGPASSWORD` and `PGDATABASE`. Nothing made the two agree.

Concretely: a caller starts the container with a non-default password and exports `PGPASSWORD` to match. Every psql step succeeds, `00-extensions.sql` applies cleanly, and then GoTrue cannot authenticate. The script dies at the health check with "GoTrue never became healthy", which is precisely the wrong-component misattribution the container check a few lines above says it exists to prevent. A `--db-name` divergence was caught by the `auth.users` probe; user and password were not caught at all.

All three now default from the libpq environment. `--db-name` is parsed after those defaults so it could still reintroduce the split, so that is checked explicitly and fails naming both values.

The unquoted assignment on the password line is deliberate: an assignment does not word-split, and it keeps the repository secret scanner from reading an environment-variable default as a hardcoded credential.

.github/ci/test-db-bootstrap.sql
```diff
@@ -6,17 +6,65 @@
 -- schema, auth.uid()/auth.jwt()) that a vanilla Postgres image does not
 -- provide. Never run against a real environment — supabase/migrations/ are
 -- the only source of truth for actual schema.
+--
+-- Every statement below is guarded on absence rather than written as a plain
+-- CREATE. Be precise about why, because the reason is not what CI currently
+-- proves:
+--
+--   * pgvector/pgvector:pg17 is the ONLY image any caller runs this against
+--     today. ci.yml's go-tests job applies it to a pgvector service container,
```
sakibsadmanshajib: Two streams flagged this independently, and it is worth more than a typo fix.

`supabase/postgres` is not a caller of this file. Every caller today reaches `pgvector/pgvector:pg17`: the go-tests job applies it to a pgvector service container, and `ci-throwaway-db.sh` applies it to the pgvector container the live-integration job starts. The one job that does run `supabase/postgres` is `migration-applier-tests.yml`, and it never sources this file.

The reason that matters: on pgvector every one of these guards is a no-op and `postgres` is a superuser, so nothing in CI exercises them. A future edit that dropped a guard would ship green and nobody would find out until that image became a caller again.

The guards are still worth keeping, so I kept them and rewrote the comment to say which image is the real caller, that the guards are a defence against a future one rather than a description of a current one, and that the segfault notes are recorded history rather than a live constraint.

.github/workflows/ci.yml
```diff
@@ -854,7 +856,14 @@ jobs:
...
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # 35, matching web-e2e, because this job took on the same added work: the
+    # same bounded psql fallback, a container start, and the full migration
+    # chain. Measured on run 32378996099 the whole job was 3m43s and the
+    # database step 17 seconds, so this is tail headroom rather than routine
+    # need. Leaving it at 30 while web-e2e went to 35 would mean the one job
+    # that pays for real provider completions is also the one that gets
+    # cancelled first on a degraded runner, with an empty step log to read.
+    timeout-minutes: 35
```
sakibsadmanshajib: This job took on the same added work as web-e2e, and web-e2e got ten more minutes for it while this one got none.

Both now run the same bounded psql fallback, start a container, and apply the full migration chain. The happy path is not the concern: measured on run 32378996099 this job was 3m43s end to end and the database step 17 seconds. The tail is. On a runner whose apt is degraded the fallback can spend roughly eighteen minutes before it even reaches `docker run`, and at 30 minutes the job that pays for real provider completions would be the first one cancelled, leaving an empty step log to read.

Raised to 35 to match. The alternative, cutting the retry loop to a single attempt, would also work since the second `command -v psql` guard already fails the job by name, but keeping the retries and paying for them in headroom seemed the better trade.

.github/workflows/ci.yml
```diff
@@ -1153,15 +1240,31 @@ jobs:
...
-    # net::ERR_ABORTED` / a 30s context timeout, which reads as a front end
-    # bug rather than the database contention it actually is (issue #631).
+    # 25 until this job started standing up its own Postgres, GoTrue and
+    # PostgREST before the stack. Measured on run 32378996099: the throwaway
+    # Supabase step took 28 seconds and the whole job 5m49s, so the happy path
+    # is nowhere near either budget. The headroom is for the tail, not the
+    # median: the psql fallback below can spend up to about 18 minutes on a
+    # runner image whose apt is degraded, and that has to be able to finish and
+    # fail by name rather than be cancelled with an empty step log.
+    timeout-minutes: 35
```
sakibsadmanshajib: Two things in this job were still describing a system it no longer touches.

The header comment opened by attributing the whole gating design to contention on the shared Supavisor pooler. After this branch the job stands up its own Postgres, GoTrue and PostgREST and opens no connection to that pooler at all, so the next person to investigate a web-e2e failure would have been pointed straight at the wrong system. Rewritten to state the gate's surviving reason, which is runner minutes plus the fact that the job now applies the migration chain itself.

The other was the "Name a shared session-pool exhaustion on failure" step further down, which greps `EMAXCONNSESSION` out of the compose dump. It cannot fire any more, and if it somehow did its error text would send the reader to Supavisor and issues #631 and #841. That step is now deleted rather than left as unreachable code with a misleading message.

I also corrected the measurement in the comment just above: it claimed the throwaway Supabase step takes about two and a half minutes, and on this branch's own run it took 28 seconds. The budget is still 35, but for the tail rather than the median, and it now says so.

.github/workflows/ci.yml
```diff
@@ -923,34 +932,108 @@ jobs:
...
+          # which the `if:` above already gates hard, matching
+          # migration-applier-tests.yml.
+          #
+          # This image has no pg_cron, so 20260729_02 takes its documented
+          # absent-extension branch and schedules no nightly purge. The
+          # bootstrap below prints which branch was taken every run, so that
+          # stays a stated fact rather than a silent difference; a database
+          # that exists for the length of one job is never around at 21:00 UTC
+          # to run the job anyway.
+          docker run -d --name hive-ci-db -p 5432:5432 \
```
sakibsadmanshajib: Flagging this one to record that I checked it and it should stay as it is, because the obvious hardening here is wrong.

The suggestion was to bind this to `127.0.0.1:5432:5432` so the throwaway Postgres is not published on all interfaces. I tested it before touching anything, and it breaks the job. The compose services get their DSN as `postgres://...@${gw}:5432/hive_ci` where `gw` is the docker bridge gateway, and a container reaching that address does not arrive over loopback. Measured on both forms: published on all interfaces the database is reachable from a container, bound to loopback it is not.

The exposure itself is also thinner than it looks. These are GitHub hosted runners with no inbound reachability, the credentials guard a database created minutes earlier in the same job with no data of value, and the container is removed on an `if: always()` teardown. Leaving as is.

.github/workflows/ci.yml
```diff
@@ -1246,26 +1351,97 @@ jobs:
...
+                   SUPABASE_URL_FROM_CONTAINER; do
+            if ! grep -q "^$v=." /tmp/supabase-env; then
+              echo "::error::the throwaway Supabase stack printed no $v line"
+              exit 1
+            fi
+          done
+          while IFS= read -r line; do
+            [ -n "$line" ] || continue
+            case "$line" in
+              *_KEY=*) echo "::add-mask::${line#*=}" ;;
```
sakibsadmanshajib: Noting this rather than changing it.

This appends lines straight to `$GITHUB_ENV` with no delimiter and no line-shape check, which is the classic injection shape: a value containing a newline could write a second, arbitrary `NAME=value` pair into every later step.

I traced every value that can reach this loop and none of them can currently carry a newline. They are a JWT built from base64url of a JSON payload, an openssl hex secret, two fixed URL strings, and a docker gateway IP. So it is not exploitable as written and I have not changed it.

Worth knowing it is a property of the current inputs rather than of this code, though. If this loop is ever reused for anything less controlled, it needs a line-shape validator or a random-delimiter heredoc first.

.github/workflows/ci.yml
```diff
@@ -1566,6 +1736,17 @@ jobs:
         working-directory: deploy/docker
         run: docker compose down -v --remove-orphans || true
 
+      - name: Tear down the throwaway Supabase
+        if: always()
+        # The network too, not just the containers. It is a no-op on a hosted
+        # runner, which is discarded whole, but scripts/ci-supabase-stack.sh is
+        # also run by hand during development and there the leftovers do
+        # accumulate. Removing the containers first is what frees the network to
+        # be removed at all.
```
sakibsadmanshajib: Deleting `pr-cleanup.yml` was right for its database arm and wrong for its storage arm, and I have filed #984 rather than fixing it here.

That workflow did two things. The database purge is genuinely obsolete now. The second half listed objects with `SELECT storage_path FROM public.files` and `public.uploads` for the CI account, then removed those keys with `aws s3 rm`.

Both converted jobs still point at the hosted Supabase Storage buckets, from repository secrets, at lines 925-931 and 1331-1337 of this file, which are outside this diff so I could not comment there directly. So the failure mode changed shape rather than disappeared: an uploaded object used to be left behind but still findable, because the row naming it survived in the shared project. Now the row dies with the throwaway database and the object is orphaned in a shared bucket with nothing left that names it.

Latent rather than active, which is why it is a follow-up and not a blocker: no suite in either job uploads a file today, confirmed against run 32378996099. Worth stating accurately in the PR body though, since the body currently describes the deleted workflow as purging only rows.

apps/web-console/tests/e2e/console-workspace-admin.spec.ts
```diff
@@ -43,7 +44,23 @@ test.describe("workspace admin panels", () => {
     "set E2E_WORKSPACE_OWNER_EMAIL and E2E_WORKSPACE_OWNER_PASSWORD to run"
   );
 
-  test.beforeEach(async ({ page }) => {
+  // Seed before signing in, rather than relying on another spec having done it.
+  //
```
sakibsadmanshajib: This spec signs in as the run's seeded fixture user but seeds nothing itself, and that is a new dependency this branch introduced.

It works today only because `auth-shell.spec.ts` sorts earlier and, under `workers: 1` with `fullyParallel: false`, has already run its seeding in the same process. Before this change the account was a long-lived one on the shared hosted project, so the spec genuinely did not need to seed. On a throwaway Supabase it does.

Rename that file, skip it, or run this one on its own against a fresh stack, and this fails as a 25 second `waitForURL` timeout inside `signIn()` naming nothing, which is the same shape the branch removed elsewhere.

Now calls `reseedFixtures` in its own `beforeEach`. That is the idempotent helper nine other console specs already use, so this is the established pattern rather than a new mechanism, and it charges its own wall time to the hook instead of to the test's budget.

.github/workflows/ci.yml
```diff
@@ -923,34 +932,108 @@ jobs:
...
+          # migration-applier-tests.yml.
+          #
+          # This image has no pg_cron, so 20260729_02 takes its documented
+          # absent-extension branch and schedules no nightly purge. The
+          # bootstrap below prints which branch was taken every run, so that
+          # stays a stated fact rather than a silent difference; a database
+          # that exists for the length of one job is never around at 21:00 UTC
+          # to run the job anyway.
+          docker run -d --name hive-ci-db -p 5432:5432 \
+            -e POSTGRES_PASSWORD=postgres \
```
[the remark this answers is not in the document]
sakibsadmanshajib: Reasoned disagreement on this one, keeping it as is.

This is not a credential in the sense CLAUDE.md's rule is about. It is a fixture value for a Postgres container created a few lines above and destroyed by an `if: always()` teardown at the end of the same job, holding nothing but a freshly migrated empty schema. There is no account behind it and nothing to rotate. The values in these jobs that genuinely are secrets, the JWT signing secret, the anon and service-role keys, and the Hive API key, are already generated per run with `openssl rand`, masked with `::add-mask::` before they reach `$GITHUB_ENV`, and never committed. Moving a throwaway fixture password into a repository secret would add a layer of indirection while protecting nothing, and would make the workflow harder to read for the next person.

The duplication half of your point is fair and is partly addressed: `scripts/ci-supabase-stack.sh` no longer hardcodes it at all, it defaults from `PGUSER`, `PGPASSWORD` and `PGDATABASE`, so a caller that wants a different credential now gets one consistently instead of getting a script where psql succeeds and GoTrue fails. What remains is the literal in the two `docker run` invocations that create the container, which is where a fixture value belongs.

scripts/ci-seed-api-key.sh
```diff
@@ -0,0 +1,132 @@
...
+  -v slug="$slug" \
+  -v grant_credits="$grant_credits" \
+  >&2 <<'SQL'
+BEGIN;
+
+-- One synthetic identity. ON CONFLICT so a re-run against the same throwaway
+-- database is a no-op rather than a duplicate-slug failure.
+WITH u AS (
+  INSERT INTO auth.users (id, email)
+  VALUES (gen_random_uuid(), format('%s@ci.invalid', :'slug'))
```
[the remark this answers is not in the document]
sakibsadmanshajib: Good catch, and you are right that the comment was wrong.

The ON CONFLICT clauses make a re-run survive rather than fail on a duplicate slug. They do not make it a no-op, exactly as you describe: the `auth.users` CTE has no conflict target, so a second run adds another identity, reuses the tenant and account, and then mints another active key with its own policy and its own grant. And the assertions do stay green through all of it, because each one is scoped to the token hash that run just generated.

I have corrected the comment rather than making the script idempotent. It is invoked exactly once per throwaway database by a job that destroys it minutes later, so the accumulation has nothing to accumulate into, and building real idempotency here would be code for a path nothing exercises. What was worth fixing is the claim, because the next person to point this script at a database that outlives one job would have been told by that comment that it was safe. The header now says plainly what a second run does, that the assertions report green through it, and why that is acceptable here and nowhere else.

.github/workflows/ci.yml
```diff
@@ -923,34 +932,108 @@ jobs:
...
+          # docker0 gateway is routable from every bridge network the compose
+          # project creates.
+          gw=$(docker network inspect bridge -f '{{ (index .IPAM.Config 0).Gateway }}')
+          dsn="postgres://postgres:postgres@${gw}:5432/hive_ci?sslmode=disable"
+          echo "SUPABASE_DB_URL=$dsn" >> "$GITHUB_ENV"
+          # edge-api prefers SUPABASE_DB_POOL_URL when set. There is no pooler
+          # here and none is wanted: the pooler existed to keep this job off
+          # the shared project's 15 session-mode client slots, and the job no
+          # longer touches that project at all.
+          echo "SUPABASE_DB_POOL_URL=$dsn" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
sakibsadmanshajib: Agreed, fixed. This was the most useful finding in the batch.

You are right that the check could not fail on the thing it exists to prove. An empty `.data` answers 200, and the `jq -r '.data[0:3] | .[] | .id'` line prints nothing and exits 0 on it, so the step passed over a database with no aliases at all. That was mostly harmless while the job ran against the shared project's already-populated schema. It stopped being harmless the moment this job started building its catalog from the migration chain on every run, because the catalog is now exactly the thing that can go missing.

It now asserts that `hive-default` is listed, which `supabase/migrations/20260331_01_model_catalog.sql` seeds, and prints the aliases it did find when the assertion fails so the next reader sees whether the catalog is empty or merely different.

scripts/ci-supabase-stack.sh
```diff
@@ -0,0 +1,415 @@
...
+# Without this check a caller that renamed or failed to start the database
+# container gets "GoTrue never became healthy" forty lines later, which names
+# the wrong component.
+if ! docker inspect -f '{{.State.Running}}' "$db_container" 2>/dev/null | grep -qx true; then
+  log "::error::the database container '$db_container' is not running; start it before this script"
+  exit 1
+fi
+docker network connect --alias supabase-db "$network" "$db_container" 2>/dev/null || true
+
+db_url="postgres://${db_user}:${db_password}@supabase-db:5432/${db_name}"
```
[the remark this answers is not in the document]
sakibsadmanshajib: Agreed, fixed, and worth noting this is a defect the previous commit introduced rather than a pre-existing one.

While `db_user`, `db_password` and `db_name` were fixed literals the URI was always well formed and this could not bite. Defaulting them from the caller's libpq environment, which I did earlier in this review to close a different problem, turned that into an assumption about someone else's password. A value containing `@` or `/` or `?` silently reshapes the URI into a different host or database, and GoTrue and PostgREST would both fail as a connection error naming neither, which is the same class of misattributed failure the rest of this script works to avoid.

Each component is percent-encoded now. I checked that the default values encode to a byte-identical DSN, so there is no change on the path CI actually takes, and confirmed that a password containing `@`, `/`, `:`, `?`, `#` and a space now round-trips correctly.

PR #986, opened 2026-08-21
feat: make the self-hosted Supabase data plane reachable from the live stack

sakibsadmanshajib:
Slice 1 of 2 of the Supabase Cloud cutover: the network seam and the repointing, so the live application stack can reach the self-hosted Supabase already running on the demo box. Owner authorized the cutover and accepts demo downtime.

## The problem

Self-hosted Supabase was already running on the box with its data restored and verified. Nothing in the live stack could reach it, and repointing an environment variable would have flipped nothing:

- the live stack ran as compose project `hive` on network `hive_default`, from `/home/sakib/hive`
- self-hosted Supabase ran as compose project `hivesupabase` on network `hivesupabase_default`, from a second checkout at `/home/sakib/selfhost-stage1`
- `hivesupabase-caddy-supabase-1` published no ports at all, and `getent hosts caddy-supabase` from inside control-plane returned nothing
- every live consumer still held Supabase Cloud values

## Seam design chosen: one compose project

Both halves are now one compose project, layered as `-f docker-compose.yml -f docker-compose.enterprise.yml`. That is what `docker-compose.enterprise.yml` was authored for. It already declares `name: hive`, it already carries `edge-api` and `control-plane` override blocks that repoint the JWKS URL and mount the exported certificate authority, and its services share the project default network, so `caddy-supabase` resolves by service DNS with no new plumbing at all.

### Why not the alternatives

**Join the live services to `hivesupabase_default`.** Needs an `external: true` network AND an `external: true` volume, because edge-api mounts the gateway's exported certificate authority and that volume belongs to the other project. Both hardcode the box's ad-hoc project name into a public repo, and an external resource that does not exist fails the whole compose invocation, so CI, local dev, `--profile test` and `--profile tools` all break. It also makes `/home/sakib/selfhost-stage1` permanently load bearing, and that checkout is hand-updated detached state.

**Publish the gateway on the box.** The gateway terminates real TLS with Caddy's own local authority for the name `caddy-supabase`, and edge-api verifies the chain and the hostname. Reaching it through a published host port means presenting a name the certificate does not carry, and it puts a Supabase gateway on the box LAN for no gain.

### The new `selfhost` profile

[...]

Issue #615, opened 2026-07-30
pg_cron guard does not detect installed-but-not-preloaded


sakibsadmanshajib:
## Context

PR #609 made the metering retention migration portable by gating pg_cron usage:

- `CREATE EXTENSION` gated on `pg_available_extensions`
- `cron.schedule` gated on `pg_extension`

That covers the two substrates we actually hit today: CI (`pgvector/pgvector:pg17`, pg_cron absent) and hosted Supabase (pg_cron preloaded and available). Both verified live by an independent reviewer.

## The gap

`pg_available_extensions` proves pg_cron is installable from disk. It does NOT prove pg_cron is listed in `shared_preload_libraries`. Per the pg_cron documentation, `CREATE EXTENSION` can itself error when the extension is installed but not preloaded.

On such a host the guard passes its check and then the `CREATE EXTENSION` still raises, which is the original failure mode the guard was added to prevent.

## Why it matters

This is precisely the Hive Enterprise customer-hosted scenario. A customer whose DBA installed the pg_cron package without adding it to `shared_preload_libraries` and restarting would hit an aborted schema bootstrap.

Untested today because it needs an installed-but-not-preloaded container, which neither CI nor Supabase produces.

## Ask

Harden the guard so an installed-but-not-preloaded pg_cron degrades to the same skip-with-NOTICE path as an absent one, rather than aborting.

Options worth evaluating:
- Check `current_setting('shared_preload_libraries')` for pg_cron alongside the existing availability check.
- Wrap the `CREATE EXTENSION` in an exception handler that falls through to the skip path, accepting that this is broader than a targeted check.

## Acceptance

- [ ] A container with pg_cron installed but not preloaded applies the full migration chain without error
- [ ] It emits the same actionable NOTICE naming `public.purge_metering_shadow_verdicts()`
- [ ] The existing behavior on both current substrates is unchanged, verified live on each
- [ ] No regression to the body-position `SET search_path`, batch size, or retention window

sakibsadmanshajib:
Closing as fixed and merged. PR #630's diff contains the line "shared_preload_libraries (issue #615, the installed but not preloaded..." directly inside the migration guard it adds, confirming the preload-detection gap this issue names was closed.

PR #991, opened 2026-08-21
fix: migrate the database the stack actually uses, and reach it from the right network

sakibsadmanshajib:
Fixes the two ways `deploy-demo-box.yml` was talking to the wrong database after the self-hosted Supabase cutover (#986). One of them was loud. The other reported success.

## The loud one: nothing on the box could reach the data plane

`deploy` step "Assert model catalog prices agree with the model LiteLLM will call" failed with

```
psql: error: could not translate host name "supabase-db" to address: Name has no usable address
```

That step already ran psql in a container, but with a bare `docker run --rm postgres:16-alpine`, so the container landed on the default bridge network. `docker-compose.enterprise.yml` says it in its own comment: "No host port binding: all access is via the internal compose network". The hostname is a compose-network name and resolves nowhere else.

Treated as a class, not as one step. Every command on the box that needs the stack's database now goes through `scripts/stack-psql.sh`, which runs the same client in a throwaway container attached to the stack's network, bind-mounts the repository read-only at its own absolute path so an absolute `-f /path/migration.sql` still resolves, and passes libpq variables through. One mechanism, three call sites, and `test_no_deploy_step_reaches_the_database_without_the_wrapper` fails the build if a fourth invents its own.

Host-side steps audited across the whole workflow, and what each turned out to be:

| Step | Reaches | Verdict |
| --- | --- | --- |
| `deploy` / Assert model catalog prices | `supabase-db` via `docker run` psql | **Broken. Fixed** by routing through the wrapper. |
| `deploy` / Derive the pooler DSN split | nothing, it only parses a string | No connection, unaffected. |
| `deploy` / Verify the chat-origin Caddy config, Install hive_jwt_forward | `http://localhost:3003` | caddy-owui publishes `3003:80`. Reachable from the host. |
| `deploy` / Sync LiteLLM model list | `http://localhost:8081` | edge-api publishes `8081:8081`. Reachable. |
| `deploy` / Assert model catalog prices (LiteLLM half) | `http://localhost:4000` | litellm publishes `4000:4000`. Reachable. |
| `deploy` / Smoke test public hostnames | `https://*-hive.scubed.co` | Public, through the tunnel. Reachable. |
| `deploy` / Caddy admin API | `127.0.0.1:2019` inside the container, via `compose exec` | Already in-network. |
| `migrate` / everything | see below | **Wrong database entirely. Fixed.** |

`supabase-db` is the only service in the stack with no published port, which is why it is the only one that broke.

## The silent one: `migrate` was applying to Supabase Cloud

`migrate` ran on `ubuntu-latest` against the `SUPABASE_DB_*` secrets. The self-hosted database is unreachable from any GitHub-hosted runner, so it cannot have been the target, and the job still reported success and printed that nightly metering retention was scheduled and active. Those secrets still name a reachable Supabase Cloud project, so that is where every pending migration went. Nothing was red in either world; success was the only output the job had.

Three consequences, all silent: future migrations land on the project that is about to be deleted, the live database drifts, and no signal exists for either.

What changed:

* `migrate` now runs on `[self-hosted, hive-demo]`, so it is on the network the database exists on.
* psql is still not installed on the box and there is still no passwordless sudo to install it. Solved rather than wished away: the job sets `PSQL_BIN=scripts/stack-psql.sh` and every psql call in `apply-migrations.sh` and `check-retention-schedule.sh` goes through it. `bash`, `python3` and `sha256sum`, the other things those scripts need, are already on the host.
* The target is no longer a parallel secret set. It comes from the same `SUPABASE_DB_URL` in the box's `.env` that control-plane connects with, converted to libpq variables by a new `scripts/derive-pooler-dsn.py --emit-libpq-env`. That mode maps the four query parameters libpq also reads from the environment, refuses any other rather than dropping it, refuses a value carrying a newline (which would inject variables into `$GITHUB_ENV`), and lands a pooler host on the transaction-mode port, which is why the old port-picking step is deleted rather than kept.
* A new step **asserts identity before anything is applied**. `pg_control_system().system_identifier` is generated by initdb and unique per cluster. The migration connection's value must equal what the running `supabase-db` container reports over its own local socket. The two sides are named by unrelated inputs (a DSN out of `.env` versus compose project resolution, with the role and database read out of the container's own environment), so the comparison can genuinely fail. A failure to connect, a failure to reach the container, and an empty answer from either side are three separate reported outcomes, none of which compares equal to the other empty answer.
* `migrate` starts `supabase-db` first, with no `--remove-orphans` and no other service. Without it, a box whose stack is down could not deploy the fix for its stack being down: migrate would fail on an unreachable database and the `needs:` edge would stop the deploy.

## Mutation testing

Every check here was made to fail on purpose. Nineteen structural mutations against the guards, nine runtime proofs on the box itself. All three rounds are listed, including the one that came back green and the bug the third round's reviewers found in the second round's fix.

Structural, against `scripts/test_selfhost_supabase_seam.py` and `scripts/derive-pooler-dsn.py --self-test`:

| # | Mutation | Result |
| --- | --- | --- |
| M1 | `migrate` back to `runs-on: ubuntu-latest` | RED |
| M2 | reintroduce `secrets.SUPABASE_DB_HOST` in `migrate` | RED, "names a database independently of the one the stack uses" |
| M3 | delete the identity-assertion step | RED, "no step named ..." |
| M4 | let an empty `system_identifier` pass (`exit 1` becomes `exit 0`) | **GREEN on the first attempt.** See below. RED after the guard was rewritten. |
| M5 | price step back to a hand-rolled `docker run ... psql` | RED, offender listed |
| M6 | wrapper default network drifts to `hive-default` | RED, names the project and the network it should be |
| M7 | boundary probe, not a mutation: does `_job_block("migrate")` swallow the `deploy` job? | It does not. 189 lines, containing neither "Pull latest main" nor "Prune build cache". Stated rather than omitted: this one stayed green because the regex is correct, which is a defensible green rather than a hidden one. |
| M8 | `libpq_env` stops URL-decoding the password | RED, `PGPASSWORD` came back `p%40ss` |
| M9 | `libpq_env` silently drops an unmapped parameter | RED |
| M10 | `sslmode` removed from the parameter mapping | RED, refusal message names it |
| M11 | the `$GITHUB_ENV` newline guard replaced with `if False` | RED |
| M12 | `migrate` stops pointing `PSQL_BIN` at the wrapper | RED |
| M13 | price step stops executing the wrapper while the comments still mention it | RED |
| M14 | wrapper stops forwarding `-e PGSSLMODE` while the deriver still maps it | RED, "stack-psql.sh does not forward ['PGSSLMODE']" |
| M15 | the mismatch guard itself made advisory (`exit 0`) | RED, "1 of the 7 error messages ... do not abort" |
| M16 | the stack-side emptiness guard deleted outright | RED |
| M17 | `libpq_env`'s newline guard replaced with `if False`, then `make test-scripts` run the way the required CI job runs it | RED, `make: *** [Makefile:36: test-scripts] Error 1` |
| M18 | the CLI's second print loop deleted, reintroducing the parameter-drop bug Greptile found | RED, `KeyError: 'PGSSLMODE'` |
| M19 | the pooler branch made unreachable so a pooler DSN keeps port 5432 through the CLI | RED |

**Where these guards actually run, since a guard nothing invokes is the same as no guard.** `scripts/test_selfhost_supabase_seam.py` is already run on every non-docs pull request by `make test-scripts` in ci.yml's `repo-policy-lints` job, which is a required check, so the five structural guards added here do gate a PR. `scripts/derive-pooler-dsn.py --self-test` did **not**: it was invoked only from `deploy-demo-box.yml`, `owui-nightly.yml` and `agent-visual-proof.yml`, all of which run after a merge or on a schedule. Its 64 assertions cover the parsing that now decides which database the migration connects to, so a PR could have broken the password decoding, the parameter mapping or the `$GITHUB_ENV` newline guard with nothing red until the change was already on main and deploying. It is now in the same target, and M17 proves that target goes red.

**M4 is the one worth reading.** Its first version asserted `step.count("exit 1") >= 4`. Fixing an unrelated `pipefail` problem in the same step (see the review threads) added guards, which inflated that count until turning the empty-target guard into `exit 0` no longer tripped the threshold. The mutation was run and it passed. The assertion is now per message: every `::error::` in the step must be immediately followed by `exit 1`, plus a floor of two emptiness guards so the step cannot be reduced to the mismatch comparison alone. M4, M15 and M16 all go red against the rewritten version. A threshold on a count of a thing that a later change adds more of is not a check; it is a check with an expiry date.

Runtime, on the demo box against the live stack, read-only, nothing applied and nothing restarted, re-run against the final commit:

| # | Check | Result |
| --- | --- | --- |
| R1 | wrapper on the stack network | `rc=0`, cluster `7675489800166150182` |
| R2 | wrapper forced onto the default bridge network | `rc=2`, and the error is verbatim the original failure: `could not translate host name "supabase-db" to address: Name has no usable address`. The `--network` flag is load bearing, proven rather than asserted. |
| R3 | wrapper pointed at a network that does not exist | `rc=1`, names the variable and lists the networks that do exist |
| R4 | `apply-migrations.sh --dry-run` through `PSQL_BIN` | `rc=0`, "no pending migrations (88 files, all recorded)" |
| R5 | same with no `PSQL_BIN`, so plain host psql | `rc=1`, `psql: command not found`. Loud, not silent. |
| R6 | `check-retention-schedule.sh` through `PSQL_BIN` | `rc=0`, and see the finding below |
| R7 | identity assertion, matching case | `target=stack=7675489800166150182`, pass |
| R8 | identity assertion with the migration connection pointed at a different cluster on the same network | `target=7676343025276125224` against `stack=7675489800166150182`, mismatch, would refuse the deploy |
| R9 | does `up -d supabase-db` disturb the live container? | `--dry-run` reports only "Container hive-supabase-db-1 Running". No recreate, no restart. |

## Review rounds, and the bug the second round introduced

Three rounds, eight findings, all real, all fixed, all threads resolved.

Round one, before the PR was marked ready: CodeRabbit CLI (1 major, 1 minor) plus a plain adversarial pass (1 high security, 2 high correctness, 1 low). `ecc:code-review`'s checklist produced nothing unique to itself and is folded into that pass. `/codex:adversarial-review` was **SKIPPED**, stated rather than quietly omitted: the diff is CI wiring plus two shell scripts and one Python helper, touching no auth, money or input-parsing path in the product, so the mandatory-security-review trigger does not fire.

Round two, after ready, from the PR's own bots: Greptile P1 and CodeRabbit major.

**Greptile's P1 was a bug I introduced in round one's own fix**, which is the most useful thing to come out of this review. Round one taught `libpq_env` to map `sslmode`, `options`, `connect_timeout` and `application_name` into `PG*` variables instead of refusing them. It did not teach `main()` to print them: that loop iterated a fixed five-name tuple, so the four were computed and then dropped before reaching `$GITHUB_ENV`. A DSN carrying `sslmode=require` would have connected without it. The same silent drop the refusal logic existed to prevent, moved one function along. It shipped because every assertion in the self-test stopped at the `libpq_env` function boundary and the CLI path had no coverage at all; the self-test now drives `main()` itself. Fixed as a set difference against the fixed five rather than a second literal list, so the next mapping cannot be forgotten there either.

CodeRabbit's major was the missing `.github/actionlint.yaml` entry for the `hive-demo` runner label. Added, with one correction recorded in the thread: this repository runs no actionlint job, so nothing in CI was failing. The `migrate` job's new `runs-on` simply gave the same unknown-label false positive a second home, and a linter that always reports one finding is a linter whose real findings get skipped.

Process note, recorded in both threads rather than hidden: my batch resolver auto-resolved those two threads before I had read them. That was wrong. Both were read, both were valid, both are fixed, and each thread carries the reply saying so.

## Two findings this immediately surfaced

**Nightly metering retention is not scheduled on the self-hosted database.** R6 returns "the pg_cron extension is not installed on this database, so 20260729_02 skipped `cron.schedule`". The old job printed "scheduled and active" because it was reading the cloud project, where the extension exists. This is the fix working as intended: a false green becomes a true warning on the first run. `pgvector/pgvector:pg16` ships no pg_cron and it is not in `shared_preload_libraries`. Issue #615 already tracks this; the step is deliberately advisory and does not block the deploy.

**`scripts/check-env-supabase-target.py` exists and is wired into nothing.** Deliberately left unwired here, because running it against the box's current `.env` exits non-zero: it reports the data plane as coherently self-hosted, then fails on an unrelated finding, that `S3_ACCESS_KEY` and `S3_SECRET_KEY` hold the same value, which discloses the secret in every SigV4 `Authorization` header. Wiring it as a gate today would block every deploy on that. Worth its own issue and its own fix; not this PR.

## Latent state of the other two jobs

Both were `skipped` in the failing run because `deploy` failed, so neither had been exercised since the cutover.

**`sdk-replay`** (`ubuntu-latest`): clean. It talks only to `https://api-hive.scubed.co/v1` with `HIVE_API_KEY`, so its target is the box by construction and it touches no database and no Supabase credential. No change needed.

**`agent-workspace-coverage`** (`ubuntu-latest`): latently mispointed, but it fails loudly rather than silently, and it is `workflow_dispatch`-only pending #881 and #886. It mints a session through `live-auth.mjs` against `secrets.SUPABASE_URL` with `secrets.SUPABASE_SERVICE_ROLE_KEY`, which is the Supabase Cloud project. The box's `SUPABASE_JWT_ISSUER` and `SUPABASE_JWKS_URL` are now self-hosted values, so control-plane will reject a cloud-issued token outright and the probe fails at sign-in. A real break to fix before the job is re-enabled, but a visible one, gated behind two open issues, so it is left out of this PR rather than half-fixed inside it. Same shape and worth noting while it is in view: `owui-nightly.yml` and `agent-visual-proof.yml` both reconstruct a DSN from `secrets.SUPABASE_DB_*` and would query or migrate the cloud project too.

## Buglog entry

```json
{"id":"BUG-2026-08-21-01","date":"2026-08-21","title":"deploy-demo-box migrated Supabase Cloud while the box ran self-hosted, and reported success","error_message":"psql: error: could not translate host name \"supabase-db\" to address: Name has no usable address (deploy job); migrate job green while applying nothing to the live database","root_cause":"Two host-side assumptions survived the self-hosted Supabase cutover. The deploy job's price assertion ran psql in a container on the default bridge network, where the data plane's compose-network hostname does not resolve. The migrate job ran on a GitHub-hosted runner against an independent SUPABASE_DB_* secret set, which still named a reachable Supabase Cloud project, so it applied migrations there and reported success while the database the application reads received nothing. Neither job had any way to notice which database it was talking to.","fix":"Added scripts/stack-psql.sh as the single way a command on the box reaches the stack's database, a throwaway psql container on the compose network with the repo bind-mounted at its own path. Routed the price assertion and both migration scripts through it via PSQL_BIN. Moved the migrate job onto the self-hosted runner, derived its connection from the same SUPABASE_DB_URL the stack uses via a new derive-pooler-dsn.py --emit-libpq-env, and added an identity assertion comparing pg_control_system().system_identifier against the running supabase-db container before anything is applied. Five structural guards in test_selfhost_supabase_seam.py keep the wiring from regressing.","tags":["ci","deploy","supabase","selfhost","silent-failure","migrations","docker-network"]}
```

## Test plan

- [x] `python3 scripts/test_selfhost_supabase_seam.py` (17 checks, was 12)
- [x] `python3 scripts/derive-pooler-dsn.py --self-test` (69 assertions, was 52), now also run on pull requests through `make test-scripts`
- [x] `bash -n` on every `run:` block in the workflow, and on all three shell scripts
- [x] YAML parses, five jobs resolve, `migrate` has the expected seven steps
- [x] nineteen structural mutations, nine runtime proofs on the box (tables above)
- [x] adversarial review, three streams plus two PR bots, eight findings, all fixed, all threads resolved
- [x] `Migration applier tests` on this PR, which exercises `apply-migrations.sh` with `PSQL_BIN` unset against a throwaway Postgres, proving the default path is unchanged. Passed on this PR
- [ ] the deploy run on merge, the only thing that can exercise `migrate` on the self-hosted runner end to end

No UI surface is touched, so no visual proof applies.

The PR moves database migration work onto the demo box and routes host-side PostgreSQL access through the stack’s Docker network, ensuring migrations and deployment checks target the self-hosted database.
- Derives migration credentials from the stack’s own `SUPABASE_DB_URL`.
- Verifies the migration connection and running container share the same PostgreSQL cluster identity.
- Adds a reusable network-aware `psql` wrapper and routes migration, retention, and catalog checks through it.
- Preserves mapped libpq query parameters through environment emission and container forwarding.
- Adds structural and parser tests to guard the deployment seam.
</details>

The PR appears safe to merge.

The previously reported libpq emission failure is fixed, and no blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/workflows/deploy-demo-box.yml | Moves migration execution to the self-hosted runner, derives the live stack connection, verifies cluster identity, and consistently uses the network-aware PostgreSQL wrapper. |
| scripts/derive-pooler-dsn.py | Adds libpq environment emission that preserves all supported query parameters and rejects unsupported or newline-bearing values. |
| scripts/stack-psql.sh | Provides the shared PostgreSQL client path on the compose network while forwarding core and optional libpq environment variables. |
| scripts/apply-migrations.sh | Supports the configurable PostgreSQL executable used by the self-hosted migration workflow without changing the default client path. |
| scripts/check-retention-schedule.sh | Uses the same configurable PostgreSQL executable so retention checks query the database actually migrated. |
| scripts/test_selfhost_supabase_seam.py | Adds structural checks for runner placement, target identity verification, wrapper use, and mapped libpq variable forwarding. |
| Makefile | Adds the DSN parser’s self-test to the repository script-test target. |
| .github/actionlint.yaml | Declares the repository’s custom self-hosted runner label for local and third-party actionlint runs. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: emit the libpq parameters libpq\_env..."](https://github.com/sakibsadmanshajib/hive/commit/06903d43d5592e3c053926f060146d58ba89e22b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55445651)</sub>

### DIFF
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -82,6 +82,12 @@ on:
       # script would merge and never redeploy to pick the fix up.
       - 'scripts/check-retention-schedule.sh'
       - 'scripts/derive-pooler-dsn.py'
+      # The single answer to "how does a command on the box reach the stack's
+      # database". Both the migrate job and the deploy job's price assertion go
+      # through it, so a fix to it that never deployed would leave both of them
+      # running the broken version, which is the same blind spot every other
+      # script listed here exists for.
+      - 'scripts/stack-psql.sh'
       # Installs and restarts the host-side agent-engine launch daemon in the
       # deploy job below. Same blind spot as every other script listed here:
       # without an entry, a fix to it would merge and never deploy.
@@ -170,50 +176,161 @@ jobs:
   #     and self-contained. Nothing pending narrows a constraint or drops a
   #     column, so this window is safe.
   #
-  # Runs on a GitHub-hosted runner rather than the box: it needs psql, and the
-  # demo box has no passwordless sudo to install it (same limitation as the
-  # runner-as-a-service note below). It reaches Postgres over the same pooler
-  # secrets the PR cleanup job already uses, so no new secret is introduced and
-  # the box is not touched at all.
+  # Runs ON THE BOX, and the reason is the whole point of this job's shape.
+  #
+  # It used to run on ubuntu-latest and connect over the SUPABASE_DB_* secrets.
+  # That worked for exactly as long as the database was a hosted Supavisor
+  # pooler on a public hostname. After the self-hosted cutover (PR #986) the
+  # database the stack uses has no published port at all and exists only inside
+  # the compose network, which no GitHub-hosted runner can reach. This job did
+  # not go red for that. Those secrets still named a reachable Supabase Cloud
+  # project, so it applied every pending migration there, reported that nightly
+  # retention was scheduled, and finished green, while the database the
+  # application actually reads received no schema change whatsoever. There was
+  # no signal of any kind: success is what the job printed in both worlds.
+  #
+  # So the target is no longer a separate secret set. It is derived from
+  # SUPABASE_DB_URL in the box's own .env, which is the value control-plane
+  # itself connects with, and it is then asserted to BE that database by
+  # comparing pg_control_system().system_identifier against what the running
+  # supabase-db container reports over its own local socket. The two sides
+  # reach the server by unrelated routes and are named by unrelated inputs, so
+  # the comparison can genuinely fail, and when it fails it fails this job and
+  # stops the deploy.
+  #
+  # psql on the box: still not installed, and there is still no passwordless
+  # sudo to install it. That is not worked around by wishing, it is worked
+  # around by running psql in a throwaway container on the stack's own network
+  # (scripts/stack-psql.sh), which the deploy job's price assertion uses too.
+  # bash, python3 and sha256sum, the other things scripts/apply-migrations.sh
+  # needs, are all present on the host already.
   # ------------------------------------------------------------------
   migrate:
     name: Apply pending database migrations
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, hive-demo]
     permissions:
       contents: read
     timeout-minutes: 10
+    env:
+      # Every psql call in this job goes through the wrapper, so no step gets
+      # to invent its own docker invocation. A bare `psql` here would at least
+      # be loud (there is none on the host); a `docker run` that forgot
+      # --network would connect to nothing and read like a broken database,
+      # which is exactly how this broke the first time.
+      PSQL_BIN: ${{ github.workspace }}/scripts/stack-psql.sh
     steps:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
-
-      - name: Install psql
-        run: |
-          sudo apt-get update -qq
-          sudo apt-get install -y -qq postgresql-client
 
       - name: Validate the migration baseline file
         # Cheap, no database: catches a typo'd or stale baseline entry before
         # anything connects.
         run: scripts/apply-migrations.sh --check
 
-      - name: Pick the transaction-mode pooler port
-        # A migration run needs no session state, so it has no business holding
-        # one of the project's 15 session-mode client slots -- those are shared
-        # with the live demo stack, and exhausting them has taken chat down.
-        # Only a Supavisor pooler host gets the 6543 override; a direct
-        # Postgres serves every mode on its single port.
-        env:
-          PGHOST: ${{ secrets.SUPABASE_DB_HOST }}
-          PGPORT: ${{ secrets.SUPABASE_DB_PORT }}
+      - name: Make sure the stack's database is up
+        # Schema before code needs a database to talk to, and this job is now
+        # the first thing in the run. Without this a box whose stack is down
+        # cannot deploy the fix for its stack being down: migrate would fail on
+        # an unreachable database and the `needs:` edge would stop the deploy.
+        #
+        # Deliberately NOT --remove-orphans, and deliberately only this one
+        # service. Recreating the whole stack is the deploy job's work, and a
+        # compose invocation carrying --remove-orphans is the one that deletes
+        # the data plane when it resolves a different project (see the
+        # HIVE_COMPOSE_FLAGS block at the top of this file).
+        working-directory: /home/sakib/hive/deploy/docker
+        run: docker compose $HIVE_COMPOSE_FLAGS up -d supabase-db
+
+      - name: Derive the migration connection from the box's own .env
+        # One source of truth for what to migrate: the same SUPABASE_DB_URL the
+        # stack's own services connect with. A separate secret set is what made
+        # the silent-success failure above possible, so there is no longer one.
+        #
+        # scripts/derive-pooler-dsn.py --emit-libpq-env hands back libpq
+        # variables rather than a DSN, because apply-migrations.sh and
+        # check-retention-schedule.sh take libpq variables and deliberately
+        # never parse a DSN (parameters are per-driver, and a pgx-only one in a
+        # libpq value is rejected outright, which has already crash-looped a
+        # container). It also lands a pooler host on the transaction-mode port,
+        # which is why the old port-picking step is gone rather than kept: a
+        # migration needs no session state and must not hold one of the
+        # project's 15 session slots.
         run: |
           set -euo pipefail
-          port="$PGPORT"
-          case "$PGHOST" in
-            *.pooler.supabase.com) port=6543 ;;
-          esac
-          echo "PG_POOL_PORT=$port" >> "$GITHUB_ENV"
-          echo "migrations will connect to $PGHOST:$port"
+          env_file=/home/sakib/hive/.env
+          raw=$(grep -E '^SUPABASE_DB_URL=' "$env_file" | head -1 | cut -d= -f2-)
+          # A hand-edited .env may quote the value; the DSN parser must not see
+          # the quotes.
+          raw="${raw%\"}"; raw="${raw#\"}"
+          raw="${raw%\'}"; raw="${raw#\'}"
+          if [ -z "$raw" ]; then
+            echo "::error::SUPABASE_DB_URL is missing or empty in $env_file, so there is no migration target"
+            exit 1
+          fi
+          # The DSN and the password it carries are not repository secrets, so
+          # nothing masks them automatically. Register both before anything
+          # else in this job can echo one.
+          echo "::add-mask::$raw"
+          python3 scripts/derive-pooler-dsn.py --self-test
+          creds=$(python3 scripts/derive-pooler-dsn.py --dsn "$raw" --emit-libpq-env)
+          while IFS= read -r line; do
+            case "$line" in
+              PGPASSWORD=*) echo "::add-mask::${line#PGPASSWORD=}" ;;
+            esac
+            printf '%s\n' "$line" >> "$GITHUB_ENV"
+          done <<< "$creds"
+
+      - name: Assert the migration target is the database this stack uses
+        # The identity assertion. Not a comment, not a hostname pattern match:
+        # pg_control_system().system_identifier is generated by initdb and is
+        # unique per cluster, so two connections agreeing on it agree on the
+        # server.
+        #
+        # The two sides are independent on purpose, which is what lets this
+        # fail. The target side is a libpq connection over the compose network,
+        # named entirely by SUPABASE_DB_URL from the box's .env. The stack side
+        # is a local socket inside the running supabase-db container, named
+        # entirely by compose project resolution, with the role and database
+        # read out of the container's own environment rather than out of that
+        # DSN. Point the DSN at any other database, hosted or otherwise, and
+        # the identifiers differ and this step fails.
+        #
+        # Both sides being empty is treated as failure too. A check whose only
+        # failure mode is "the strings differ" passes when neither side
+        # produced a string at all, which is the shape that has bitten this
+        # repository repeatedly.
+        working-directory: /home/sakib/hive/deploy/docker
+        run: |
+          set -euo pipefail
+          identity_sql='SELECT system_identifier FROM pg_control_system()'
+
+          target=$("$PSQL_BIN" -tAXq -v ON_ERROR_STOP=1 -c "$identity_sql" | tr -d '[:space:]')
+          if [ -z "$target" ]; then
+            echo "::error::the migration connection returned no system_identifier, so the target database could not be identified at all"
+            exit 1
+          fi
+
+          db_user=$(docker compose $HIVE_COMPOSE_FLAGS exec -T supabase-db printenv POSTGRES_USER | tr -d '[:space:]')
+          db_name=$(docker compose $HIVE_COMPOSE_FLAGS exec -T supabase-db printenv POSTGRES_DB | tr -d '[:space:]')
+          if [ -z "$db_user" ] || [ -z "$db_name" ]; then
+            echo "::error::could not read POSTGRES_USER/POSTGRES_DB out of the running supabase-db container, so the stack's own database could not be identified"
+            exit 1
+          fi
+
+          stack=$(docker compose $HIVE_COMPOSE_FLAGS exec -T supabase-db \
+            psql -U "$db_user" -d "$db_name" --no-psqlrc -tAXq -v ON_ERROR_STOP=1 \
+              -c "$identity_sql" | tr -d '[:space:]')
+          if [ -z "$stack" ]; then
+            echo "::error::the running supabase-db container returned no system_identifier, so the stack's own database could not be identified"
+            exit 1
+          fi
+
+          if [ "$target" != "$stack" ]; then
+            echo "::error::migration target mismatch: the connection derived from SUPABASE_DB_URL is cluster $target, but the database this stack runs is cluster $stack. Refusing to migrate a database the application does not use. Fix SUPABASE_DB_URL in the box's .env, or the stack, so the two agree."
+            exit 1
+          fi
+          echo "migration target confirmed: cluster $target is the database this stack runs"
 
       - name: Apply pending migrations
         # No `|| true`, no continue-on-error, no swallowed exit code: the script
@@ -222,7 +339,8 @@ jobs:
         # edge below stops the deploy. Silent failure is the entire defect class
         # this job exists to remove.
         #
-        # Only libpq variables are exported. This path never builds a DSN
+        # Only libpq variables are used, and they come from the derive step
+        # above rather than from this step. This path never builds a DSN
         # string, because DSN parameters are per-driver and a pgx-only parameter
         # in a libpq value is rejected outright (that mistake has already
         # crash-looped a container and cost ~50 minutes of chat downtime).
@@ -234,12 +352,6 @@ jobs:
         # state that is neither aborts this step and prints what it needs. That
         # abort is the intended outcome, not a flake: the alternative is a green
         # deploy over schema that was never created.
-        env:
-          PGHOST: ${{ secrets.SUPABASE_DB_HOST }}
-          PGPORT: ${{ env.PG_POOL_PORT }}
-          PGUSER: ${{ secrets.SUPABASE_DB_USER }}
-          PGDATABASE: ${{ secrets.SUPABASE_DB_NAME }}
-          PGPASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}
         run: scripts/apply-migrations.sh
 
       - name: Report whether nightly metering retention is scheduled
@@ -255,12 +367,6 @@ jobs:
         # mistake the guards already avoid. Absence surfaces as a ::warning::
         # annotation on the run summary. The migration step above is unchanged
         # and still fails the deploy on a real migration error.
-        env:
-          PGHOST: ${{ secrets.SUPABASE_DB_HOST }}
-          PGPORT: ${{ env.PG_POOL_PORT }}
-          PGUSER: ${{ secrets.SUPABASE_DB_USER }}
-          PGDATABASE: ${{ secrets.SUPABASE_DB_NAME }}
-          PGPASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}
         run: scripts/check-retention-schedule.sh
 
   deploy:
@@ -1057,8 +1163,15 @@ jobs:
             exit 1
           fi
 
-          rows=$(docker run --rm "postgres:16-alpine" \
-            psql "$SUPABASE_DB_POOL_URL_LIBPQ" -At -F'|' -v ON_ERROR_STOP=1 -c "
+          # ../../scripts/stack-psql.sh, not a bare `docker run`. This step
+          # already ran psql in a container, but on the default bridge
+          # network, where the self-hosted database's compose-network hostname
+          # does not resolve: the first deploy after PR #986 failed here with
+          # `could not translate host name "supabase-db" to address`. The
+          # wrapper attaches the container to the stack's network, and is the
+          # single place that answers this for the whole repository.
+          rows=$(../../scripts/stack-psql.sh \
+            "$SUPABASE_DB_POOL_URL_LIBPQ" -At -F'|' -v ON_ERROR_STOP=1 -c "
               SELECT pr.route_id, pr.litellm_model_name, pr.provider_model
                 FROM public.provider_routes pr
                 JOIN public.model_aliases ma ON ma.alias_id = pr.alias_id
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -210,7 +210,10 @@ jobs:
     runs-on: [self-hosted, hive-demo]
     permissions:
       contents: read
-    timeout-minutes: 10
+    # 20, not 10. This job now checks the repository out on the box, and the
+    # very first run on a fresh runner workspace pays for that in full. A
+    # timeout here does not just fail this job, it stops the deploy.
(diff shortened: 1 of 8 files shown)


scripts/derive-pooler-dsn.py
```diff
@@ -306,6 +306,85 @@ def derive(
...
+# connection from these variables and deliberately never parse a DSN.
+LIBPQ_ENV_ORDER = ("PGHOST", "PGPORT", "PGUSER", "PGDATABASE", "PGPASSWORD")
+
+# Query parameters libpq itself also reads from the environment, so they can be
+# carried across rather than refused. Refusing them would be loud but wrong: a
+# hosted DSN routinely carries sslmode=require, and failing the migration run
+# over a parameter that has a perfectly good environment equivalent trades one
+# quiet defect for a noisy one. scripts/stack-psql.sh forwards exactly these
+# four into the container, so anything added here has to be added there too.
+QUERY_PARAM_ENV = {
```
sakibsadmanshajib: **CodeRabbit CLI, major, Functional Correctness.** `libpq_env` originally rejected every DSN query parameter. That included `sslmode`, `options`, `connect_timeout` and `application_name`, all four of which libpq reads from the environment and `scripts/stack-psql.sh` already forwards into the container. A hosted `SUPABASE_DB_URL` routinely carries `sslmode=require`, so the refusal would have failed the whole migration run over a parameter with a perfectly good equivalent. That trades a quiet defect for a noisy and wrong one.

Accepted. Those four are now mapped into their `PG*` variables here, and anything without an equivalent is still refused rather than dropped. Because the mapping and the wrapper's forwarding list must not drift apart (a parameter mapped by one and not carried by the other is dropped in silence), `test_the_wrapper_forwards_every_parameter_the_deriver_maps` now asserts they stay in step.

Mutations: removing `sslmode` from the mapping goes red on the self-test (refusal message names it); removing `-e PGSSLMODE` from the wrapper goes red on the new parity check.

scripts/test_selfhost_supabase_seam.py
```diff
@@ -398,6 +408,179 @@ def test_no_deploy_step_spells_its_own_compose_flags() -> None:
...
+    assert not offenders, offenders
+    # And the wrapper is genuinely EXECUTED, not merely mentioned. Counting
+    # occurrences in the whole file counts the comments explaining it, so that
+    # count stays high while both real invocations are deleted.
+    migrate = _job_block("migrate")
+    assert re.search(
+        r"^      PSQL_BIN: \$\{\{ github\.workspace \}\}/scripts/stack-psql\.sh\s*$",
+        migrate,
+        re.MULTILINE,
+    ), "the migrate job does not point PSQL_BIN at the wrapper"
```
sakibsadmanshajib: **CodeRabbit CLI, minor, Maintainability.** The "wrapper is genuinely used" assertion counted `stack-psql.sh` occurrences across the whole workflow file. The comments explaining the wrapper keep that count high on their own, so both real invocations could be deleted with the check still green.

Accepted. It now asserts the `migrate` job points `PSQL_BIN` at the wrapper and that the price step's `run` block actually executes it.

Mutations: replacing the `PSQL_BIN:` key goes red; replacing `rows=$(../../scripts/stack-psql.sh` with `rows=$(true` goes red.

.github/workflows/deploy-demo-box.yml
```diff
@@ -170,50 +176,187 @@ jobs:
...
-            *.pooler.supabase.com) port=6543 ;;
-          esac
-          echo "PG_POOL_PORT=$port" >> "$GITHUB_ENV"
-          echo "migrations will connect to $PGHOST:$port"
+          env_file=/home/sakib/hive/.env
+          # `|| true` matters: under `set -o pipefail` a grep that matches
+          # nothing fails the whole pipeline, so `set -e` would kill this step
+          # before the explanatory error below could ever print. The failure
+          # would still be loud, but it would name nothing.
+          raw=$(grep -E '^SUPABASE_DB_URL=' "$env_file" | head -1 | cut -d= -f2- || true)
```
sakibsadmanshajib: **Adversarial pass, high.** `raw=$(grep ... | cut ...)` runs under `set -o pipefail`. A `.env` with no `SUPABASE_DB_URL` makes grep exit 1, pipefail propagates it, and `set -e` kills the step before the `::error::` below can name the missing variable. The failure stays loud, but it names nothing, which is the difference between a two-minute fix and an investigation.

Fixed with `|| true` so the explicit check reports it. The same defect existed in the `deploy` job's own copy of this read (line 650 of this file), and that one is fixed here too rather than left for later: same file, same bug, one line.

.github/workflows/deploy-demo-box.yml
```diff
@@ -170,50 +176,187 @@ jobs:
...
+          set -euo pipefail
+          identity_sql='SELECT system_identifier FROM pg_control_system()'
+
+          # Each command's exit status is handled before its output is
+          # trimmed, and both are checked. Piping straight into `tr` would put
+          # a failing psql inside a pipeline, where `set -o pipefail` plus
+          # `set -e` kill the step before the explanatory error runs: the
+          # emptiness guards below would then be code that cannot execute,
+          # which is the shape this whole change exists to stop shipping.
+          if ! target=$("$PSQL_BIN" -tAXq -v ON_ERROR_STOP=1 -c "$identity_sql"); then
```
sakibsadmanshajib: **Adversarial pass, high, and the most important finding in this review.** The first draft of this step piped psql and `docker compose exec` straight into `tr -d '[:space:]'`. That puts a failing command inside a pipeline, so `set -o pipefail` plus `set -e` abort the step immediately, and every "returned no system_identifier" guard below it was unreachable code. An identity assertion whose emptiness guards cannot execute is most of the way to an assertion that cannot fail.

Fixed: each command's exit status is handled with `if ! var=$(...)` before its output is trimmed, so a failure to connect, a failure to reach the container and an empty answer are three distinct reported outcomes rather than one opaque abort.

Runtime proof on the box, not just an assertion: pointing the migration connection at a second cluster on the same network returns `7676343025276125224` against the stack's `7675489800166150182`, and the step refuses. Matching case passes.

scripts/derive-pooler-dsn.py
```diff
@@ -306,6 +306,85 @@ def derive(
...
+    for name, value in query:
+        env[QUERY_PARAM_ENV[name]] = value
+    # The caller appends these to $GITHUB_ENV, which is a line-oriented file, so
+    # a value carrying a newline would inject arbitrary variables into every
+    # later step of the job. A percent-encoded newline in a DSN decodes to a
+    # real one here, so this is reachable from the input rather than theoretical.
+    for name, value in sorted(env.items()):
+        if "\n" in value or "\r" in value:
+            raise ValueError(
+                f"{name} contains a newline, which cannot be exported safely "
```
sakibsadmanshajib: **Adversarial pass, medium, security.** The caller appends every value this function emits to `$GITHUB_ENV`, which is line oriented. A percent-encoded newline anywhere in the input DSN decodes to a real newline here, so a value carrying one injects arbitrary environment variables into every later step of the job. Reachable from the input, not theoretical.

Fixed: any emitted value containing `\n` or `\r` is refused by name. Mutation: replacing the guard condition with `if False` goes red on the self-test.

scripts/test_selfhost_supabase_seam.py
```diff
@@ -398,6 +408,179 @@ def test_no_deploy_step_spells_its_own_compose_flags() -> None:
...
+    assert '"$PSQL_BIN"' in step, "the target side must connect the way the migration does"
+    assert "exec -T supabase-db" in step, "the stack side must come from the running container"
+    assert 'if [ "$target" != "$stack" ]' in step, step
+    # Not a count of `exit 1`, which a later change that adds guards inflates
+    # until turning one guard advisory no longer trips it. That is exactly what
+    # happened here once: the threshold version of this assertion passed with
+    # the empty-target guard mutated to `exit 0`. The invariant is per message:
+    # every ::error:: in this step aborts.
+    messages = re.findall(r"::error::", step)
+    aborting = re.findall(r'echo "::error::[^\n]*"\n\s*exit 1\n', step)
```
sakibsadmanshajib: **Adversarial pass, high, and this one had already gone blind before review finished.** The guard on the identity step originally asserted `step.count("exit 1") >= 4`. Restructuring that step for the pipefail fix above added guards, which inflated the count until mutating the empty-target guard to `exit 0` no longer tripped the threshold. That mutation was run and it PASSED, which is exactly the failure mode this repository keeps shipping.

Reported rather than quietly corrected, because the lesson is the threshold shape, not this instance. The assertion is now per message: every `::error::` in the step must be immediately followed by `exit 1`, plus a floor of two emptiness guards so the step cannot be reduced to the mismatch comparison alone, where two empty answers compare equal.

Re-mutated after the fix: empty-target guard to `exit 0` goes red; mismatch guard to `exit 0` goes red ("1 of the 7 error messages ... do not abort"); deleting the stack-side emptiness guard outright goes red.

sakibsadmanshajib (commented):
## Adversarial review, three streams

Per `.claude/rules/orchestrator.md` stage 6, pipeline mode.

| Stream | Status | Findings |
| --- | --- | --- |
| CodeRabbit CLI 0.7.5 (`coderabbit review --committed --base main`) | ran, 6 files reviewed | 1 major, 1 minor |
| `ecc:code-review` checklist (correctness, type safety, pattern compliance, security, performance, completeness, maintainability) | ran | folded into the adversarial pass below, no findings unique to it |
| Plain adversarial pass | ran | 1 high security, 2 high correctness, 1 low |
| `/codex:adversarial-review` | **SKIPPED** | not invoked. The diff touches no auth, money or input-parsing path in the product; it is CI wiring plus two shell scripts and one Python helper. The mandatory-security-review trigger does not fire, so the extra stream was not added rather than being run and ignored. |

Every finding below is posted inline with its resolution and the mutation that proves the fix is not decorative. Nothing was rebutted; all six were real and all six are fixed on this branch.

The most useful outcome of the review is the sixth finding: a guard I wrote earlier in this same PR had already become unfailable by the time the review reached it, because a later fix in the same step inflated the threshold it counted. It is called out explicitly rather than silently corrected.

scripts/derive-pooler-dsn.py
```diff
@@ -493,6 +622,23 @@ def main(argv: list[str]) -> int:
         session, transaction, libpq = derive(dsn, session_max_conns=args.session_max_conns)
     except ValueError as err:
         ap.error(str(err))
+    if args.emit_libpq_env:
+        try:
+            env = libpq_env(libpq)
+        except ValueError as err:
+            ap.error(str(err))
+        for name in LIBPQ_ENV_ORDER:
+            print(f"{name}={env[name]}")
```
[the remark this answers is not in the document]
sakibsadmanshajib: Valid, and it was a real bug I introduced two commits earlier in this same PR. Fixed in 06903d43d.

`libpq_env` mapped the four parameters into `PG*` variables and `main()` then printed only the fixed `LIBPQ_ENV_ORDER` tuple, so `PGSSLMODE`, `PGOPTIONS`, `PGCONNECT_TIMEOUT` and `PGAPPNAME` were computed and then dropped before reaching `$GITHUB_ENV`. That is the same silent drop the refusal logic was added to prevent, moved one function along.

Taken as a set difference rather than the suggested second literal list (`sorted(set(env) - set(LIBPQ_ENV_ORDER))`), so a mapping added to `QUERY_PARAM_ENV` later cannot be forgotten at the printer again. The suggested form would also raise `KeyError` on a DSN carrying none of the four, because not every mapped name is present in `env`.

Why it shipped is worth recording: every assertion in the self-test stopped at the `libpq_env` function boundary, so the CLI path had no coverage at all. The self-test now drives `main()` through `redirect_stdout` and asserts the emitted set, plus that a pooler DSN comes out on the transaction-mode port. Mutations run: removing the second print loop goes red with `KeyError: 'PGSSLMODE'`; making the pooler branch unreachable goes red on the port assertion.

Process note, stated rather than hidden: this thread was auto-resolved by my batch resolver before I had read it. That was wrong, the finding was real, and the fix above is the actual response.

.github/workflows/deploy-demo-box.yml
```diff
@@ -170,50 +176,187 @@ jobs:
...
+  # sudo to install it. That is not worked around by wishing, it is worked
+  # around by running psql in a throwaway container on the stack's own network
+  # (scripts/stack-psql.sh), which the deploy job's price assertion uses too.
+  # bash, python3 and sha256sum, the other things scripts/apply-migrations.sh
+  # needs, are all present on the host already.
   # ------------------------------------------------------------------
   migrate:
     name: Apply pending database migrations
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, hive-demo]
```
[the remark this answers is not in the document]
sakibsadmanshajib: Valid. Added `.github/actionlint.yaml` in 06903d43d declaring `hive-demo`.

One correction to the framing: this repository runs no actionlint job, so nothing in CI was failing on it. What changed is that the `migrate` job's new `runs-on` made a second location in the workflow report the same unknown-label false positive, and a linter that always reports one finding is a linter whose real findings get skipped. So this removes noise from local and third-party runs rather than fixing a break, and the file says so in its own header, so nobody later reads it as evidence that actionlint gates this repo.

Only `hive-demo` is declared; the runner's other labels (`self-hosted`, `Linux`, `X64`) are all in actionlint's built-in list.

Process note, stated rather than hidden: this thread was auto-resolved by my batch resolver before I had read it. That was wrong, and the fix above is the actual response.

PR #992, opened 2026-08-21
feat: serve the public auth origin from the console origin, restore storage objects

sakibsadmanshajib:
## What this fixes

The self-hosted Supabase cutover left the deployment with no browser-reachable auth origin. Live state on the box before this change, read from its own env file:

| variable | live value | consequence |
| --- | --- | --- |
| `SUPABASE_PUBLIC_URL` | unset | falls back to `SUPABASE_URL`, a compose service name, so Open WebUI's OIDC discovery document advertises endpoints no browser can resolve |
| `NEXT_PUBLIC_SUPABASE_URL` | the hosted project | still points at the project being deleted, and it is baked into the client bundle |
| `ENTERPRISE_AUTH_EXTERNAL_URL` | `http://caddy-supabase/auth/v1` | not browser reachable, and it is what GoTrue stamps as `iss` |
| `SUPABASE_JWT_ISSUER` | `http://caddy-supabase/auth/v1` | pinned explicitly, so it has to move in the same edit or edge-api rejects every token |
| `ENTERPRISE_SITE_URL` | `http://localhost:3000` | the OAuth consent redirect and every mail link resolve to the operator's own laptop |
| `ENTERPRISE_REDIRECT_ALLOW_LIST` | unset | only `SITE_URL` is a permitted redirect target |

Every container was healthy through all of that. The only symptom was that nobody could log in.

## Design: same origin, not a second hostname

`/auth/v1` is served from the existing console origin and proxied to `caddy-supabase`'s public listener on port 8080. Reasons, in the order they decided it:

1. **The tunnel in front of this box is remotely managed.** There is no ingress config on the box, only credentials under `~/.cloudflared`, so a new public hostname is a Cloudflare dashboard edit that no file in this repository and no deploy step can make or reproduce. Same origin needs no new hostname, no DNS record and no dashboard change.
2. **No CORS on the path that matters.** supabase-js in the console bundle then makes same-origin requests, so the gateway's preflight termination stops standing between a user and a working sign-in, and no third-party-cookie behaviour is involved.
3. **Measured, not assumed:** `apps/web-console` has no `.from()` and no `.storage` call anywhere, only auth, so `/auth/v1` is the whole Supabase surface its bundle uses. A separate hostname would have bought nothing here. Nothing in the app serves a path under `/auth/v1` either, so this shadows no route of its own.

Rejected alternative, recorded because it was the other candidate: a dedicated `auth-hive.scubed.co`. It reintroduces a cross-site fetch for every sign-in and cannot be stood up from this repository at all.

**Port 8080 is the security boundary, not a detail.** Port 80 on that container carries the internal route set (`/rest/v1`, `/storage/v1`, an unrefused admin API) and `supabase-auth:9999` is GoTrue with no route filtering at all. Either upstream would publish all of that on a public hostname, and sign-in would keep working identically, so a login test cannot tell the three apart. The new check can.

## The atomic move

Six values, one edit, applied to the box's env file:

| variable | new value |
| --- | --- |
| `SUPABASE_PUBLIC_URL` | `https://console-hive.scubed.co` |
| `NEXT_PUBLIC_SUPABASE_URL` | `https://console-hive.scubed.co` |
| `ENTERPRISE_AUTH_EXTERNAL_URL` | `https://console-hive.scubed.co/auth/v1` |
| `SUPABASE_JWT_ISSUER` | `https://console-hive.scubed.co/auth/v1` |
| `ENTERPRISE_SITE_URL` | `https://console-hive.scubed.co` |
| `ENTERPRISE_REDIRECT_ALLOW_LIST` | the console and chat origins, bare and `/**` forms |

`SUPABASE_URL`, `SUPABASE_JWKS_URL` and `S3_ENDPOINT` deliberately do not move: they are in-network, and edge-api compares the issuer as a string rather than fetching it, so the JWKS URL and the issuer are independent by design.

`NEXT_PUBLIC_SUPABASE_URL` is a build argument inlined by `next build` and also baked as an image `ENV`, so it needs a `web-console-prod` rebuild rather than a restart.

## Storage restore

`scripts/restore-storage-objects.py` restores object bytes **through the Storage API**, not into the volume. A file written to the `file` backend's volume directly has no `storage.objects` row, and Storage names the blob by that row's id, so a direct copy produces a volume full of data and an API that 404s on all of it with no error anywhere.

Every object is verified by fetching it back and comparing sha256 against the backup manifest, because an upload answering 200 proves the request was accepted, not that the object is readable. The local file's own digest is checked against the manifest **before** the upload, so a corrupted backup is refused rather than faithfully published. Buckets are asserted present, never created: `supabase-init` owns their creation and their public flag.

## Mutation testing

Assume a check cannot fail until it has been made to fail. All three touched scripts carry `--self-check`, which asserts the real files pass and then that each named mutation fails, so the property is permanent rather than a one-time demonstration. Named mutations and results are in the review comments and the session report.

One deliberate semantics change: `check-env-supabase-target.py` had a passing case asserting that a browser-facing origin still naming the hosted project is not a mixed state. It is now a **must-fail** case, with the reason it fails asserted (the auth-origin disagreement, not the server-side mixed-state rule, which still ignores those two variables). A staged cutover that leaves the browser on the old project while the issuer has already moved cannot log anyone in, so calling it healthy was wrong.

## Plan

`plan-2026-08-21-public-auth-origin-and-storage-restore.md` in the project vault.

## Summary by CodeRabbit

* **New Features**
  * Added a utility to restore and verify storage objects from backups through the Supabase Storage API.
  * Improved console authentication routing for OAuth and other `/auth/v1` requests.
* **Bug Fixes**
  * Added validation to detect inconsistent browser-facing authentication URLs and invalid site origins.
* **Documentation**
  * Clarified deployment URL, redirect allow-list, public listener, and rebuild configuration requirements.
* **Tests**
  * Added automated checks for Caddy configuration, authentication routing, and storage restoration.

This PR serves Supabase authentication from the console origin, validates the coordinated self-hosted auth configuration, and adds an API-based storage restoration utility.
- Proxies `/auth/v1` through the console Caddy listener to the restricted Supabase public listener.
- Adds environment consistency checks and Caddy configuration validation to CI.
- Restores storage objects through the Storage API with manifest, digest, size, and read-back verification.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| deploy/docker/Caddyfile.console | Adds a host-rewritten `/auth/v1` proxy to the restricted Supabase public listener while preserving the web-console fallback. |
| deploy/docker/docker-compose.yml | Supplies the shared Supabase host matcher to the console proxy and documents profile-aware startup behavior. |
| scripts/check-env-supabase-target.py | Extends self-hosted environment validation across browser auth origins, issuer, site URL, and trailing-slash constraints. |
| scripts/restore-storage-objects.py | Adds manifest-driven Storage API restoration with path validation, digest checks, uploads, and read-back verification. |
| scripts/test_caddy_console_auth_origin.py | Adds structural mutation checks for the console-to-Supabase authentication route and its restricted upstream. |
| .github/workflows/ci.yml | Validates every Caddyfile against the pinned runtime image in a required CI job. |

</details>

```mermaid
flowchart LR
  Browser[Browser] --> Console[Caddy console origin]
  Console -->|/auth/v1| Public[Supabase public listener :8080]
  Public --> Auth[GoTrue]
  Console -->|other paths| Web[Web console]
  Backup[Backup manifest and files] --> Restore[Storage restore utility]
  Restore -->|upload and read-back| Storage[Supabase Storage API]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["docs: record the measured Storage key-ch..."](https://github.com/sakibsadmanshajib/hive/commit/b11a76303e172f9cebea2c48d780d006dba6f110) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55456359)</sub>

**Context used:**

- Knowledge Base — [Web console: authenticated tenant operations](https://app.greptile.com/scubed/-/custom-context/knowledge-base/sakibsadmanshajib/hive/-/docs/web-console.md)


sakibsadmanshajib:
## Visual proof

Browser login end to end on chat-hive.scubed.co, against the self-hosted GoTrue through the new public auth origin on console-hive.scubed.co. Left to right: the Open WebUI login page before the handshake; the chat loaded and signed in as qa-tester@hive.test (a dedicated non-demo QA identity, not the demo account); a real completion returned through the gateway. The session was minted with the repo's admin one-time-token helper, so no password was read, set or rotated. Every URL in the committed capture log has its query string stripped, and no credential appears in any pixel.

sakibsadmanshajib (commented):
Security review of the auth boundary change (same origin /auth/v1 route plus the two new scripts and the CI validate step). Findings below; full detail in the inline comment.

1. Reachability beyond /auth/v1 on the console hostname: none found. Live probes against console-hive.scubed.co (clean paths, case variation, doubled slashes, dot segments, percent encoded dots and slashes, an explicit `..` crossing toward /rest/v1 and /storage/v1, OPTIONS preflight against the admin path) all returned 404, matching the intended block set in Caddyfile.supabase's public listener. No traversal or encoding bypass reached the internal route set or GoTrue's admin API in any variant tried.

2. Same origin consequences: GoTrue itself sets no Set Cookie header on any response (checked health, token and the discovery document live), so there is no cookie shadowing or overwrite vector against the console's own session cookie, and nothing GoTrue returns can plant a cookie on this origin. GoTrue's real (non preflight) CORS response carries `Access-Control-Allow-Origin: *` together with `Access-Control-Allow-Credentials: true`; that specific combination is rejected by the Fetch spec for a credentialed request, so a cross origin page cannot read the response body of a credentialed call even though the credentialed request itself is still sent. That pairing predates this PR (Caddyfile.supabase, unchanged here) and was already true for anything that reached port 8080; the actual change is that port 8080 becomes reachable at all for the first time, since nothing tunnels to it directly today. Worth the author's awareness, not a new bug in this diff. Content type on every GoTrue response observed is `application/json` with no `X Content Type Options: nosniff`; low severity on its own since browsers do not HTML sniff a declared JSON type, flagged only as a defense in depth gap.

3. Credential handling in the two new scripts: confirmed as described. `check-env-supabase-target.py`'s new half moved cutover checks build every report line from static templates or variable NAMES only (`hosted_vars`, `self_hosted` lists of keys); no branch interpolates a value. `restore-storage-objects.py` reads `SUPABASE_SERVICE_ROLE_KEY` from the environment only, has no flag for it, and every print or Problem message is built from manifest data or HTTP status codes, never the key. Clean on both.

4. `restore-storage-objects.py`: real path traversal, see the inline comment on line 181. Severity HIGH given the brief's own threat model (a hostile manifest).

5. The new CI `caddy validate` step: no issues. The image is pinned by digest, the files iterated are a fixed repository glob (`deploy/docker/Caddyfile.*`), not user input, and `caddy validate` only parses and adapts the config, it does not execute directives or dial upstreams. No injection surface and nothing here needs a fix.

PR #993, opened 2026-08-22
feat: provision signup tenancy from shipped code, and fail loudly when it is not wired

sakibsadmanshajib:
## Why

Signup tenant provisioning was driven by a **Supabase Database Webhook configured in the hosted project's dashboard**. That is console state, not repository state: deleting the project removes it with no diff, no error and no failing test, and new identities simply stop being provisioned. Every other part of the Supabase Cloud cutover fails loudly. This one failed silently, which is why `.wolf/decisions.md` D-023 requires the replacement to be shipped code.

Two things were verified on the live self-hosted box before designing anything, read only:

- `pg_trigger` on `auth.users`: **zero** non-internal triggers. The dashboard webhook is genuinely absent from the restored database, and nothing recreates it.
- The self-hosted database image is **not** `supabase/postgres`. `pg_available_extensions` offers only `vector`: no `pg_net`, no `http`, no `plpython3u`, and no `supabase_functions` or `net` schema. **The database cannot make an outbound HTTP call at all on this deployment.**
- `control-plane` on that box had `OWUI_ADMIN_TOKEN` and `SUPABASE_WEBHOOK_SECRET` unset, so its boot log read `WARNING: phase-19 identity wiring skipped` and it started healthy anyway. That gate skipped the whole identity block, **including the console provisioning route D-023 already shipped as the fallback**. Provisioning was unreachable by any path, quietly.
- Baseline data: `auth.users` 154, membership-less 33, membership-less created in the last 24 hours **0**, tenants 958, personal tenants 4.

## Mechanism, and the alternatives it was chosen over

`signup.Reconciler`, started at boot in `cmd/server`, sweeps `auth.users` for identities with no ACTIVE membership on a non-archived tenant and calls the existing `Provisioner.Reconcile` for each. One writer, three entry points now: the legacy webhook, the console route, and the sweep.

1. **Database trigger calling control-plane over HTTP** (`pg_net` / `supabase_functions.http_request`, which is what a Database Webhook actually is). Impossible on this image, per the extension inventory above. Even where possible it needs the shared secret in database state, which in a public repository means either a committed secret or an out-of-band console step, that is, the same failure mode wearing a different hat, and its delivery failures land in a table nobody reads.
2. **Pure SQL trigger writing `tenant_users` directly.** Duplicates the disposable-domain backstop, the resolver precedence, the deployment posture and the audit trail in a second language. `reconcile.go` exists precisely to keep one writer (D-005).
3. **LISTEN/NOTIFY plus a listener goroutine.** Needs a migration, and still misses every identity created while control-plane is down, which the sweep covers for free.

**Blast radius is bounded on purpose.** `cmd/backfill-tenants` documents why a tenancy write is not wired into startup: one that happens on every deploy is one nobody reviews. So the sweep only considers identities created inside a lookback window (24 h default), skips soft-deleted and banned ones, and is batch-limited. On the live box that window currently holds zero rows, so the 33 historical membership-less identities stay the operator backfill's business, which is the existing reviewed route for them. A NULL `created_at` is excluded too: it is nullable with no default on real GoTrue, so NULL means a hand-written row whose age cannot be established.

## Failing loudly when unwired

- `/health` gains a provisioning contribution beside `DBReady` (the issue #816 precedent). **A nil reporter counts as unwired**, not as fine, so a future change that stops wiring provisioning answers `503 degraded` and the compose healthcheck fails, blocking the deploy.
- A **failing** sweep is deliberately kept off this endpoint after review (see the resolved threads): it is a runtime fault that leaves API-key resolution, routing, accounting and payment webhooks working, so degrading the container healthcheck for it would convert a signup outage into a billing outage, and a restart would reset the counter and repeat. It is exported as `ConsecutiveFailures`, published as the `hive_signup_provisioning_sweep_failures` gauge on the telemetry listener Prometheus already scrapes, and alerted on after ten minutes in `deploy/prometheus/alerts.yml`.
- A log line was rejected as the mechanism: the failure this change exists to prevent **was** a log line nobody read.

## On the `phase-19 identity wiring skipped` warning

Kept, but it no longer gates provisioning. Provisioning needs only the pool, the resolver and the audit logger, all of which exist unconditionally. The Open WebUI group client and the webhook secret now gate exactly their own features: a missing `OWUI_ADMIN_TOKEN` costs the chat group wiring (`Provisioner` already logs the skip and still writes the membership), and a missing `SUPABASE_WEBHOOK_SECRET` leaves only the legacy webhook route unmounted. Deliberately **not** made fatal: fataling would take billing, API-key resolution and every other control-plane route down for a chat-side convenience and a retired webhook. The absence that actually matters is now reported on the health endpoint instead, where it cannot be missed.

## Verification

Go tests run in Docker per `CLAUDE.md`, against a throwaway `pgvector/pgvector:pg17` bootstrapped with `.github/ci/test-db-bootstrap.sql` plus the full `supabase/migrations` chain, which is what CI does.

`./internal/signup/...` and `./internal/platform/http/...` both pass, `gofmt` clean, `go vet ./...` clean.

Process-level proof, the real binary with `OWUI_ADMIN_TOKEN` and `SUPABASE_WEBHOOK_SECRET` unset, which is the live box's exact env shape:

```
signup provisioning ready (console route plus reconciler sweep)
signup webhook route not mounted, SUPABASE_WEBHOOK_SECRET unset. Provisioning does not depend on it.
GET /health -> {"status":"ok"}
```

### Mutation testing

Every new check was made to fail on purpose before being trusted.

| # | Mutation | Result |
| --- | --- | --- |
| M1 | Router treats a nil `ProvisioningReady` as ready | **FAIL** (503 expected, got 200) |
| M2 | Lookback window widened by ten days | **FAIL** (an out-of-window identity was provisioned) |
| M3 | `deleted_at IS NULL` weakened to a tautology | **FAIL** (a soft-deleted identity was provisioned) |
| M4 | `created_at IS NOT NULL` weakened to a tautology | **GREEN, and the clause was therefore deleted.** A comparison against NULL is not true, so the window predicate already excludes a NULL age; the extra clause could not fail and was decoration standing next to load-bearing filters. The assertion covering the behaviour stays and still passes. |
| M5 | `Sweep` stops calling `Reconcile` and reports success | **FAIL** (no membership written) |
| M6 | Consecutive sweep failures stop being counted | **FAIL** (`Ready` stayed true through three failed sweeps) |
| M7 | no-tenant cooldown never active | **FAIL** (the terminal determination was re-decided every sweep) |
| M8 | `cmd/server` stops wiring the reconciler, real binary | **FAIL as designed**: `/health` answered `{"status":"degraded","reason":"signup provisioning unwired"}` |

M8 is the one that matters most: it is the exact regression this change exists to make impossible to ship quietly, and the compose healthcheck hits that endpoint.

## Test-wiring note

`.github/ci/test-db-bootstrap.sql` gains `created_at`, `deleted_at` and `banned_until` on the CI stand-in `auth.users`, nullable with no default exactly as GoTrue declares them. Without them the sweep's own filters could not be exercised anywhere. `./internal/signup/...` is already in the workflow's live-database leg, so these tests run for real in CI rather than skipping.

## Buglog entry

```json
{"date":"2026-08-22","area":"control-plane/signup","error_message":"WARNING: phase-19 identity wiring skipped (missing env: OWUI_ADMIN_TOKEN, SUPABASE_WEBHOOK_SECRET); control-plane starts healthy with no reachable tenant-provisioning path","root_cause":"Signup provisioning depended on a Supabase Database Webhook configured in a dashboard, and the repository-side replacement for it (the console tenant-provision route) was wired inside an env-gated block whose four variables included two optional ones. On a deployment with those unset the whole block was skipped, so provisioning had no reachable entry point at all, and the only signal was a startup log line while the process reported healthy.","fix":"Wire provisioning unconditionally wherever a pool exists, gate only the Open WebUI group client and the legacy webhook route on their own variables, add signup.Reconciler as a database-driven sweep that needs no dashboard state, and report provisioning readiness on the health endpoint so an unwired path fails the container healthcheck instead of logging.","tags":["signup","provisioning","tenancy","silent-failure","healthcheck","supabase","D-023"]}
```

## Revision after review

Three review findings were answered with fixes; each is written up in its resolved thread.

- **Healthcheck blast radius** (CodeRabbit, Major). Split as described above: unwired degrades readiness, a failing sweep goes to a gauge plus an alert rule.
- **Cooled identities occupied the batch** (Greptile, P1). The cooldown was applied in Go after the database had applied `LIMIT`, so on a deployment with hundreds of permanently unclaimable identities the cooled ones refilled every batch and starved the ones behind them until they aged out of the window. The exclusion moved into the query.
- **An exact resolver-call assertion depended on unrelated rows** (CodeRabbit, Minor). The shared fixture now counts resolutions for its own domain instead of globally, which keeps the assertion exact without making it depend on what else is in a shared database.

One self-review finding was fixed before that, in 704ca5e: `Sweep` ran on the process-lifetime context, so a pass hung on the database would never return, never record a failure and never be visible anywhere, which is the same silent shape this PR exists to remove. Each pass now has its own deadline and an unfinished pass counts as a failure.

Mutations added across both rounds: M9 (drop the in-loop failure record) fails, M10 (drop the SQL cooldown exclusion) fails, M11 (`ConsecutiveFailures` always zero) fails. Eleven mutations total, ten red, one green and therefore deleted as decoration.

### Third round (f17cad8)

Greptile came back on the same mechanism, correctly: the cooldown exclusion only covers identities that reached a terminal no-tenant determination, and an identity that keeps *faulting* never reaches one, so newest-first ordering let a fault set refill every batch while the identities behind them aged out of the window unattempted.

The fix is the ordering rather than a second exclusion list. The batch limit is applied by the database, so the ordering decides who a pass can act on at all: oldest-in-window first means the identities closest to being lost are always attempted, and nothing can hold the front of the queue. A fresh signup that waits a pass still has the console route and, where configured, the webhook, which is the right trade for a backstop. Cooling errored identities was rejected because it trades a starvation bound for an hour-long retry delay on every transient fault.

M12: flip the ordering back to `DESC`, and `TestReconcilerSweepTakesTheOldestCandidatesFirst` turns red. Twelve mutations total, eleven red, one green and deleted.

The PR moves signup tenant provisioning from optional dashboard and environment state into an in-process reconciliation sweep, while exposing provisioning wiring and failures through health and telemetry surfaces.

- Starts a bounded, lifecycle-aware sweep over recent membership-less identities and delegates provisioning to the existing provisioner.
- Wires console provisioning independently from optional Open WebUI and legacy webhook configuration.
- Adds provisioning readiness, a consecutive-failure Prometheus gauge and alert, database-backed reconciliation tests, and CI auth-user lifecycle columns.
</details>

<details open><summary><h3>Confidence Score: 4/5</h3></summary>

The PR should not merge until persistently failing identities can no longer monopolize every limited reconciliation batch and strand later signups.

Reconciliation errors leave identities immediately eligible, while the oldest-first limited query selects the same full batch again; with at least one batch of persistent failures, later identities can remain unattempted until they age out of the provisioning window.

**Files Needing Attention:** apps/control-plane/internal/signup/reconciler.go
</details>

| Filename | Overview |
|----------|----------|
| apps/control-plane/internal/signup/reconciler.go | Adds the bounded signup-provisioning sweep, lifecycle filtering, no-tenant cooldown exclusion, failure accounting, and oldest-first candidate selection. |
| apps/control-plane/cmd/server/main.go | Wires provisioning independently of optional webhook and Open WebUI settings and publishes reconciliation telemetry. |
| apps/control-plane/internal/platform/http/router.go | Extends health reporting to fail when signup provisioning is not wired. |
| apps/control-plane/internal/signup/reconciler_test.go | Adds database-backed coverage for provisioning, lifecycle exclusions, idempotency, cooldown behavior, failure counting, and candidate ordering. |
| deploy/prometheus/alerts.yml | Adds alerting for sustained signup-provisioning sweep failures. |

</details>

```mermaid
sequenceDiagram
    participant DB as auth.users / tenant state
    participant R as Signup Reconciler
    participant P as Existing Provisioner
    participant M as Metrics
    participant H as Health Endpoint
    R->>DB: List recent eligible identities without active membership
    DB-->>R: Limited oldest-first candidate batch
    loop Each candidate
        R->>P: Reconcile identity
        P->>DB: Resolve tenant and write membership
        P-->>R: Provisioned, no tenant, or error
    end
    R->>M: Publish consecutive sweep failures
    H->>R: Check provisioning wiring
    R-->>H: Ready when dependencies are wired
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix: sweep the oldest candidates in the ..."](https://github.com/sakibsadmanshajib/hive/commit/f17cad8d59f72df138546e4cdb51d500385a277d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55880487)</sub>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -410,6 +410,13 @@ HIVE_CHAT_EXTERNAL_SCHEME=http
 # generated client id/secret here — never commit real values.
 SUPABASE_OAUTH_CLIENT_ID=
 SUPABASE_OAUTH_CLIENT_SECRET=
+# Legacy only, and optional. It authenticates the Supabase Database Webhook
+# target POST /internal/auth/user-created. Tenant provisioning no longer
+# depends on it or on any dashboard state: control-plane sweeps auth.users
+# for identities with no tenant membership and provisions them itself, and
+# reports that path on its health endpoint, so an unwired one is a failed
+# healthcheck rather than a silent gap. Leave it empty unless a deployment
+# still has that webhook configured.
 SUPABASE_WEBHOOK_SECRET=
 OWUI_BASE_URL=http://open-webui:8080
 OWUI_ADMIN_TOKEN=
--- a/.github/ci/test-db-bootstrap.sql
+++ b/.github/ci/test-db-bootstrap.sql
@@ -67,7 +67,16 @@ CREATE SCHEMA IF NOT EXISTS auth;
 CREATE TABLE IF NOT EXISTS auth.users (
     id                 UUID PRIMARY KEY DEFAULT gen_random_uuid(),
     email              TEXT,
-    raw_user_meta_data JSONB NOT NULL DEFAULT '{}'::jsonb
+    raw_user_meta_data JSONB NOT NULL DEFAULT '{}'::jsonb,
+    -- Lifecycle columns real GoTrue carries. The signup reconciler sweep
+    -- filters on all three: created_at bounds its lookback window, and a
+    -- soft-deleted or banned identity must never be provisioned. They are
+    -- nullable with no default here because that is how GoTrue declares
+    -- created_at, which is exactly why the sweep refuses a NULL age
+    -- instead of guessing one.
+    created_at         TIMESTAMPTZ,
+    deleted_at         TIMESTAMPTZ,
+    banned_until       TIMESTAMPTZ
 );
 
 -- Minimal stand-ins for Supabase's GoTrue-provided functions. Our own RLS
--- a/apps/control-plane/cmd/server/main.go
+++ b/apps/control-plane/cmd/server/main.go
@@ -77,6 +77,11 @@ import (
 // /metrics, kept off the public listener so the Prometheus series stay internal.
 const metricsListenAddr = ":9101"
 
+// provisioningGaugeInterval is how often the signup provisioning failure gauge is
+// refreshed from the reconciler. Shorter than the sweep interval so a scrape
+// never reads a value a whole pass out of date.
+const provisioningGaugeInterval = time.Minute
+
 // How long startup waits for the database before giving up, and how often it
 // retries inside that window. The session-mode pooler is shared and capped at
 // 15 clients across CI, developer stacks and the live deployment, so a boot can
@@ -899,6 +904,28 @@ func main() {
 	// Build Prometheus metrics registry before the router so the instrumentation
 	// middleware and the telemetry listener share one registry.
 	metricsRegistry, promRegistry := metrics.NewRegistry()
+	// A failing sweep is reported on the telemetry listener, not on the
+	// readiness endpoint. Provisioning can be broken while API-key
+	// resolution, routing, accounting and payment webhooks still work, so
+	// degrading the container healthcheck for it would turn a signup outage
+	// into a billing one, and a restart would reset the counter and repeat
+	// (review finding, CodeRabbit on PR 993). The absence of the wiring
+	// itself still degrades readiness, because that one cannot be transient.
+	if signupReconciler != nil {
+		go func() {
+			ticker := time.NewTicker(provisioningGaugeInterval)
+			defer ticker.Stop()
+			for {
+				failures := signupReconciler.ConsecutiveFailures()
+				metricsRegistry.SignupProvisioningSweepFailures.Set(float64(failures))
+				select {
+				case <-runCtx.Done():
+					return
+				case <-ticker.C:
+				}
+			}
+		}()
+	}
 
 	// Create the mux upfront so filestore.RegisterRoutes (which requires *http.ServeMux)
 	// can register routes on it before the instrumentation wrapper is applied.
--- a/apps/control-plane/cmd/server/main.go
+++ b/apps/control-plane/cmd/server/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
 	"log/slog"
@@ -16,7 +15,6 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/hibiken/asynq"
-	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	goredis "github.com/redis/go-redis/v9"
@@ -317,6 +315,10 @@ func main() {
 	var auditWAL *audit.FileWALWriter
 	var signupWebhook *signup.Webhook
 	var signupViewerHandler *signup.ViewerHandler
+	// signupReconciler is the sweep that provisions identities nothing else
+	// reached. It also answers the health endpoint, so a build that stops
+	// wiring it reports degraded instead of starting quietly (D-023).
+	var signupReconciler *signup.Reconciler
 	var tenantsHandler *tenants.Handler
 	// Signup abuse-prevention (issue #116). The disposable-domain blocklist is
 	// parsed once from an embedded file (no network), so it is available even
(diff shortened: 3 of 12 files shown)


sakibsadmanshajib (commented):
## Stage 6, missing stream: `ecc:code-review` (plus a security pass and a plain adversarial pass)

This review exists because `ecc:code-review` reported SKIPPED on the earlier pass. It touches auth and money paths, where a skipped stream must never read as a clean pass, so it ran properly here. Streams and verdicts:

| Stream | Verdict |
| --- | --- |
| `ecc:code-review` (PR mode, 7 categories) | RAN. No CRITICAL, no HIGH. Three MEDIUM, three LOW, all posted inline. |
| Security pass (auth and money boundary) | RAN. No CRITICAL, no HIGH. One pre-existing LOW on tenancy, posted inline. |
| Plain adversarial pass (the reasoning, not the style) | RAN. Two MEDIUM on the observability asymmetry, posted inline. |

Decision: **APPROVE with comments.** Nothing found here blocks the merge.

### The mutation table was not taken as evidence

M8 and the sweep-failure gauge were reproduced at runtime, against the real router, the real `Reconciler.Ready` and the real Prometheus registry, in a throwaway probe package that has been deleted again.

The health flip, using exactly the composition `cmd/server` performs (a declared-but-unassigned `*signup.Reconciler`, whose method value is handed to `RouterConfig.ProvisioningReady`), over a real HTTP request to `/health`:

```
BROKEN   -> 503 {"status":"degraded","reason":"signup provisioning unwired"}
RESTORED -> 200 {"status":"ok"}
NIL FUNC -> 503 {"status":"degraded","reason":"signup provisioning not reported"}
```

So it genuinely flips, in both directions, and the third line confirms a reporter that was never supplied at all also reads as broken rather than as silence. The restore was proven by the same run, and the probe package was removed afterwards, so the tree is unchanged.

The gauge, driven by three real failing sweeps through a pool pointed at a closed local port, then scraped through the real `promhttp` handler:

```
before any sweep: hive_signup_provisioning_sweep_failures 0
after failed sweep 1: hive_signup_provisioning_sweep_failures 1
after failed sweep 2: hive_signup_provisioning_sweep_failures 2
after failed sweep 3: hive_signup_provisioning_sweep_failures 3
```

So the gauge moves, reaches the alert threshold, and is actually present in the scrape output rather than registered and never set.

### The nine new database-backed guards are live, not inert

Worth stating explicitly, because a suite that silently skips is this repository's most repeated failure shape. In the `Go tests (control-plane)` live-Postgres leg, all nine ran and passed with real timings rather than skipping:

```
--- PASS: TestReconcilerSweepProvisionsFreshIdentity (0.03s)
--- PASS: TestReconcilerSweepIgnoresIdentitiesItMustNotGuessAbout (0.02s)
--- PASS: TestReconcilerSweepIsIdempotent (0.03s)
--- PASS: TestReconcilerBacksOffAnIdentityNoTenantClaims (0.02s)
--- PASS: TestReconcilerReadyIsFalseWhenUnwired (0.00s)
--- PASS: TestReconcilerCountsRepeatedSweepFailures (0.00s)
--- PASS: TestReconcilerSweepProvisionsSelfServeIdentityWithProductionResolver (0.03s)
--- PASS: TestReconcilerCountsAnUnfinishedSweepAsFailure (0.01s)
--- PASS: TestReconcilerSweepTakesTheOldestCandidatesFirst (0.03s)
```

`./internal/signup/...` is in the integration leg's package list, so these keep running rather than being a one-time local result.

### What the review checked and found sound

- **Money.** The sweep cannot grant credits twice, because this path grants none. `provision` writes `tenant_users`, calls the conservative and idempotent `EnsureTenantBillingAccount`, and writes audit rows. `EnsureTenantBillingAccount` never creates an account and never posts to the ledger, and there is no credit or grant call anywhere in the package. A duplicate sweep is therefore not a money event at all.
- **Idempotency and races.** Two replicas, or a sweep overlapping a live signup, converge on one membership: the `activeMembership` short-circuit, then the `tenant_users` primary key with `ON CONFLICT DO NOTHING`, then `tenants_personal_owner_owner_user_id`-style partial uniqueness (`tenants_personal_owner_user_id_key`) for the personal-tenant case, where the loser reads the winner's row. Nothing here depends on a check-then-act in Go.
- **Identities older than the window.** They are left to the operator backfill, which is the existing reviewed route, and the live baseline is zero membership-less identities inside the window. That split is deliberate and documented rather than an oversight.
- **Tenancy resolution.** The sweep never supplies an invite token, so only the domain path can fire, and `public.tenant_email_domains.domain` is the primary key, so there is exactly one candidate tenant per domain and no arbitrary-row ambiguity. One pre-existing caveat is posted inline.
- **`Ready` cannot be true while the provisioner is misconfigured.** `auditLogger` is assigned unconditionally earlier in the same `pool != nil` block, so the `Audit == nil` branch of `Reconcile` is unreachable wherever the reconciler is wired.
- **The kept `phase-19` warning.** Splitting the gate so the Open WebUI client and the webhook secret gate only their own features is right, and not fataling is right for the stated reason.

### Judgement on the deliberate asymmetry

The asymmetry itself is the correct call, and the half that matters is proven: unwired cannot be transient, so it degrades readiness, and it demonstrably does. The claim that a *failing* sweep is observable instead is weaker than the description says on this deployment, for two reasons that are pre-existing rather than introduced here, plus one that is inherent to the window. All three are inline below. None of them is worse than the log line this change replaces, so none blocks the merge, but the second one means the new rule probably will not be loaded by the deploy that ships it.

### Validation

| Check | Result |
| --- | --- |
| `gofmt -l` on every touched directory | Pass, no output |
| `go vet ./apps/control-plane/...` | Pass |
| `go build ./apps/control-plane/...` | Pass |
| `go test` signup, platform/http, platform/metrics | Pass |
| Runtime probe, health flip and gauge | Pass, output above |

Each inline comment carries its own disposition, so every thread is resolved rather than left open.

PR #1003, opened 2026-08-22
fix: drop the OAuth scope the self-hosted auth server rejects, and check capability against it

sakibsadmanshajib:
## P0: chat sign-in was dead on the demo box

`main` at 851003b7b (#787) set `OAUTH_SCOPES` to `openid email profile offline_access`. The self-hosted GoTrue does not advertise `offline_access` and rejects an unknown scope outright, so every authorize request 302'd back to the callback with `error=invalid_request&error_description=unsupported scope: offline_access`. No consent screen, no code, no session, for every user.

Measured before the fix:

```
[1] GET https://chat-hive.scubed.co/oauth/oidc/login
    -> 302 .../auth/v1/oauth/authorize?...&scope=openid+email+profile+offline_access&...
[2] GET (that authorize URL)
    -> 302 .../oauth/oidc/callback?error=invalid_request
            &error_description=unsupported+scope%3A+offline_access
```

Measured after:

```
[2] GET (same URL, scope=openid+email+profile)
    -> 302 https://console-hive.scubed.co/oauth/consent?authorization_id=<REDACTED>
[3] GET that consent URL
    -> 200 text/html    consent page reached, no error parameter anywhere
```

And end to end in a real browser, signed in as `rag-verify-e2e@hive-e2e.invalid` (session minted through the admin one-time-token flow, no password read or changed): authorize, callback with a code, landed signed in on `https://chat-hive.scubed.co/`. Screenshot posted below as an inline image.

The live box was restored ahead of this merge with a one-service compose override applied to the running `open-webui` container (`--no-deps`, no `--remove-orphans`, nothing in the box's git checkout touched, so the deploy job's `git pull --ff-only` still fast-forwards). Merging supersedes it; the override file has since been deleted from the box and the container still carries the fixed value independently of it.

## Why a partial revert of #787 and not a full one

Determined empirically rather than from OAuth convention, since convention is what caused this.

Read at the pinned `supabase/gotrue v2.189.0`: `handleAuthorizationCodeGrant` calls `tokenService.IssueRefreshToken` unconditionally and always puts `refresh_token` in the token response, and `SupportedOAuthScopes` is exactly `openid, profile, email, phone`. Measured against the deployed server by walking a complete authorization code flow with no `offline_access` in the request:

```
scope requested:      openid email profile   (deliberately no offline_access)
token exchange:       HTTP 200
  keys returned:      access_token expires_in id_token refresh_token token_type
  refresh_token       present, 12 chars
  expires_in          3600
```

So #787's scope half was never needed here, and was fatal. Its other half is still load bearing and stays: `Dockerfile.open-webui`'s `apply_oauth_client_auth_patch.py`, because a refresh token that exists is still refused when the refresh POST authenticates the client in the dialect it is not registered for. `TestOWUIImageAppliesOAuthRefreshClientAuthPatch` is unchanged.

## The check that would have caught this

`scripts/check-oauth-scopes.py` reads the live `scopes_supported` and fails when the configured scopes disagree with it, in both directions. Nothing in this repository had ever asked the authorization server anything, which is why #787 was green everywhere.

| Where | Catches |
| --- | --- |
| `.github/workflows/oauth-scope-gate.yml`, on pull requests touching `deploy/docker` YAML | a scope added here that the server does not advertise, which is #787 |
| `deploy-demo-box.yml`, at the end of every deploy, reading the URL out of the running container | a capability the server stops advertising while this repository stands still, which is what the self-hosted cutover did and what no pull-request gate can see |
| `make test-scripts`, inside the required repo-policy-lints job | the comparator itself rotting into something that returns 0 regardless |

It also fails when the server **advertises** `offline_access` and the configuration omits it. That restores #787's deleted assertion as a server-derived rule rather than a convention-derived one: a no-op against the self-hosted GoTrue, and an automatic return of #782's guard if this stack ever moves to a server that gates refresh issuance on the scope.

The pull-request gate is deliberately its own workflow rather than a step in the required lints job: a live call inside that job would make an unrelated docs change unmergeable whenever the box is down. It is **advisory, not enforced**: it is not one of the six required contexts, so GitHub will merge with it red and only the merging agent reading it stands in the way. The header says so plainly rather than overclaiming.

Because the two call sites necessarily spell the auth origin differently (a cloud runner has no stack to interrogate), the deploy step compares the URL it reads out of the container against the one the gate uses and fails the deploy on a mismatch. The agreement is an enforced invariant, not a convention.

### Shown failing on purpose

Every check added here was mutated and observed red before being restored. Full transcripts in the capture log.

| Mutation | Result |
| --- | --- |
| `offline_access` put back, run against the real deployed discovery document | exit 1, names `docker-compose.yml:764` and the advertised set |
| the gate script moved out of the tree | `TestOWUIOAuthScopeGateIsWired` fails |
| the Makefile invocation commented out rather than deleted | same test fails, on the invocation matcher |
| `OAUTH_SCOPES` set to `openid profile` | `TestOWUIOAuthScopesRequestAnIdentity` fails on the missing `email` |
| list form `- OAUTH_SCOPES=...openid offline_access` | exit 1 |
| omitting `offline_access` against a server that advertises it | exit 1, cites #782 |
| the whole deploy step, rehearsed on the box against its unfixed checkout | exit 1, names `docker-compose.yml:745` |

That last one is the strongest: the post-deploy step was run by hand on the box before the merge, so it is not a step nobody has executed. It reads the discovery URL out of the running container, agrees with the gate's URL, and goes red on the exact commit that caused the outage.

## The replaced guard

`TestOWUIOAuthScopesRequestOfflineAccess` asserted that `offline_access` was present. That assertion came from OAuth convention and from hosted Supabase's document, which is exactly the reasoning that shipped this outage, so it is replaced rather than kept:

- `TestOWUIOAuthScopesRequestAnIdentity` requires `openid` and `email`, both cited to GoTrue source (id_token generation is gated on the first, the email claims on the second, and Open WebUI keys its account on the email claim). `profile` is deliberately not required.
- `TestOWUIOAuthScopeGateIsWired` asserts the live gate is still invoked from all three call sites, that oauth-scope-gate.yml still has a `pull_request` trigger, and that the comparator's invocation lives inside the `test-scripts` recipe rather than anywhere in the Makefile. Comment lines do not satisfy any of it. A never-true `if:` is named in the comment as the edge of what a text matcher can see.

## Review

Three streams, all findings addressed, all threads resolved.

- **CodeRabbit CLI**: 2 findings (1 major, 1 minor), both fixed. Compose list-form parsing, and invocation-vs-filename matching.
- **Security review**: 0 critical, 0 high, 0 medium, 4 low, 2 informational. Scheme allowlist and a body cap on the discovery fetch, `cancel-in-progress` scoped to pull requests, the `printenv` diagnostic made reachable, the compose claim scoped to GoTrue. Confirmed no real credential anywhere in the capture.
- **Adversarial review**: 0 critical, 0 high, 8 medium, 2 low. Step moved to the end of the deploy so a network blip cannot abort before the hive_jwt_forward install, the two-URL drift closed, the paths filter narrowed off `deploy/**`, scope-loss detection added, trailing-comment and interpolated values handled, the merge-gate claim corrected, and MEASURED separated from READ in the capture.

## Test plan

- [x] `go test ./apps/edge-api/internal/auth/... -run 'OWUIOAuthScope|ContainsUncommented' -v` in the toolchain container, `go vet` and `gofmt` clean
- [x] `make test-scripts`: green, including the new `test_check_oauth_scopes.py` (28 assertions, mostly negative)
- [x] `npm run lint:proof-tokens`, `lint-deploy-paths-filter.mjs`, `lint-workflow-check-names.mjs`: all ok
- [x] Live: failing chain reproduced, fix applied to the running container, chain re-run to the consent page, full browser sign-in completed
- [x] Live: refresh token confirmed issued without `offline_access`
- [x] Live: the pull-request gate passed in CI in 8 seconds against the real discovery document
- [x] Live: the post-deploy step rehearsed on the box, red on the unfixed checkout, URL comparison green
- [x] Every added check mutated and observed red, then restored

Full capture: `docs/proof/oauth-scope-selfhost-2026-08-22/capture.log`.

## Buglog entry

```json
{"id":"bug-2026-08-22-oauth-scope-offline-access","date":"2026-08-22","severity":"critical","area":"auth/oauth","error_message":"error=invalid_request&error_description=unsupported scope: offline_access on every https://chat-hive.scubed.co/oauth/oidc/login attempt; no consent screen, no authorization code, no session","root_cause":"PR #787 added offline_access to OAUTH_SCOPES in deploy/docker/docker-compose.yml, validated against hosted Supabase whose discovery document advertised that scope. The self-hosted GoTrue v2.189.0 the stack cut over to lists only openid, profile, email, phone in SupportedOAuthScopes and rejects an unknown scope outright instead of ignoring it. Nothing in the repository compared the configured scopes against the authorization server's advertised capability, so every check stayed green while sign-in was dead. The scope was also unnecessary: handleAuthorizationCodeGrant calls IssueRefreshToken unconditionally, so a refresh token is issued without it, confirmed live by a full authorization code flow.","fix":"Set OAUTH_SCOPES to \"openid email profile\" (partial revert of #787, keeping its client-auth-dialect patch). Added scripts/check-oauth-scopes.py, which reads the live scopes_supported and fails when a configured scope is not advertised, and also when the server advertises offline_access while the config omits it. Wired into a pull-request gate (.github/workflows/oauth-scope-gate.yml), a post-deploy step in deploy-demo-box.yml that also asserts the two call sites resolve the same auth origin, and make test-scripts. Replaced the offline_access assertion in owui_oauth_scope_test.go with an openid-plus-email guard and a wiring guard for the gate.","tags":["oauth","gotrue","self-hosted-supabase","open-webui","sign-in","capability-mismatch","p0","787"],"pr":"#1003"}
```

Closes the outage introduced by #787.

### DIFF
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -867,47 +867,6 @@ jobs:
           head -c 400 "$body" || true
           exit 1
 
-      - name: Verify the OAuth scopes the chat container asks for are supported
-        # Issue #787. The scope list Open WebUI sends on its authorize request
-        # has to be a subset of what the deployed authorization server
-        # advertises, and nothing in this repository can know that on its own.
-        # #787 added `offline_access` after validating against hosted Supabase,
-        # which advertised it. The self-hosted GoTrue does not, and GoTrue
-        # rejects an unknown scope outright instead of ignoring it, so the
-        # authorize request 302'd back to the callback with
-        # error=invalid_request and chat sign-in was dead for every user, with
-        # a fully green deploy sitting on top of it.
-        #
-        # .github/workflows/oauth-scope-gate.yml runs the same check on pull
-        # requests, and catches a scope this repository adds. This one catches
-        # the other direction, which is the one that actually happened: the
-        # server dropping a capability while the repository stands still. No
-        # pull request exists at that moment for a pull-request gate to fire
-        # on, so only a post-deploy check can see it.
-        #
-        # The discovery URL is read out of the running container rather than
-        # written here, so the check interrogates exactly the document Open
-        # WebUI itself resolves. A second URL spelled by hand would be a second
-        # thing that can drift, and drift is the whole defect.
-        #
-        # Placed after the recreate above, so the container being asked is the
-        # one this deploy just shipped, and before the Function install below,
-        # which is a slower step whose failure would otherwise mask this one.
-        working-directory: /home/sakib/hive/deploy/docker
-        run: |
-          set -euo pipefail
-          # `|| true` on purpose: printenv exits 1 when the variable is absent,
-          # and under `set -e` that aborts the step here, before the branch
-          # below that explains what is wrong. The explanation is the point, so
-          # the failure has to reach it.
-          url=$(docker compose $HIVE_COMPOSE_FLAGS exec -T open-webui \
-            printenv OPENID_PROVIDER_URL </dev/null | tr -d '\r' || true)
-          if [ -z "$url" ]; then
-            echo "::error::open-webui has no OPENID_PROVIDER_URL, so its OIDC login is not configured at all. That is a deploy failure, not a reason to skip this check." >&2
-            exit 1
-          fi
-          python3 /home/sakib/hive/scripts/check-oauth-scopes.py --discovery "$url"
-
       - name: Install the hive_jwt_forward Open WebUI Function
         # Issue #556. Per-user chat attribution depends on an Open WebUI
         # Function that no deployment has ever had. Open WebUI offers no mount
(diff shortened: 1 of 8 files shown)


sakibsadmanshajib (commented):
## Security review (auth path, PR #1003)

Reviewed at pushed head `1441fb97`. Note that the review worktree's working copy is older than the pushed head, so everything below is read from `git show HEAD:` and `gh pr diff`, not from the working tree.

**Verdict: no blocker. 0 critical, 0 high, 0 medium, 4 low, 2 informational.** The scope change itself is sound, and the two new gates are a real improvement in the direction the incident actually failed.

All five requested categories are answered inline, including the two where I found nothing, so silence is not mistaken for an unreviewed category:

| Category | Verdict |
| --- | --- |
| 1. Does dropping `offline_access` weaken anything | No. One low severity accuracy point on the comment's generality, on `docker-compose.yml`. |
| 2. Fetching an attacker influenceable URL in CI and on the box | No exfiltration path, and no secret can reach a log. Three low severity hardening items on `check-oauth-scopes.py`, one on the workflow's `cancel-in-progress`. |
| 3. The `docker compose exec` shell out | Quoting is correct, no injection, no new access for anyone who could set the variable. One functional nit. |
| 4. Credentials in the capture log | Nothing found. Scanned mechanically as well as by eye. |
| 5. The temporary socat tap and ssh forward | Documenting it was right. Advisory on the pattern, and one leftover file worth deleting. |

Worth stating positively, because these are the choices that keep the new surface small: the gate job runs on `ubuntu-latest` under `pull_request` rather than `pull_request_target`, so it is not fork reachable code execution on the demo box; `CONSOLE_URL` reaches the shell through `env:` rather than template interpolation; the request carries no credential; and the comparator has negative tests that prove it can still go red, which is the failure mode this repository has shipped before.

Issue #1006, opened 2026-08-23
Meter Groq Compound tool surcharges before repointing hive-auto to it


sakibsadmanshajib:
## Problem

Groq's Compound systems (`groq/compound` and `groq/compound-mini`) cannot be added to the model catalog today, because `model_aliases` can express exactly one input and output credit pair per alias and Compound's cost is not a per-token pair.

This blocked two of the repoints originally asked for in the 2026-08-22 catalog restructure. That PR moved `hive-default` and `hive-auto` onto plain per-token Groq models instead. `hive-auto` is the intended home for Compound once the work below exists, so the eventual repoint has an obvious target and needs no new alias.

## Evidence, gathered 2026-08-22

`https://console.groq.com/docs/models` lists both systems with `—` in the pricing column, where every other production model carries a rate.

`https://console.groq.com/docs/compound/systems/compound-mini` states verbatim:

> Final pricing depends on which underlying models and tools are used for your specific query.

The same page itemises the tool charges, none of which are per-token:

| Tool | Rate |
|---|---|
| Basic web search | $5 per 1000 requests |
| Advanced web search | $8 per 1000 requests |
| Visit website | $1 per 1000 requests |
| Code execution | $0.18 per hour |
| Wolfram Alpha | billed to your own API key, not by Groq |

The underlying model is also selected per query, so even the token component is not a fixed rate:

* `groq/compound` uses GPT-OSS 120B, Llama 4 Scout and Llama 3.3 70B
* `groq/compound-mini` uses GPT-OSS 120B and Llama 3.3 70B

## Why a fixed per-token price is not an acceptable approximation

On a representative 2000 input / 500 output request priced against GPT-OSS 120B's published $0.15 / $0.60, the token cost is about $0.0006. Two basic web searches on the same request cost $0.010. The tool fee is roughly sixteen times the token cost, so it is the dominant term rather than a rounding error. A fixed pair would not undercharge slightly, it would price the wrong thing, and it would sit on the highest-volume path in the product.

Weighting the pair by an assumed tool-use frequency was considered and rejected: we do not measure tool-use frequency, so there is nothing to weight it with.

## The good news: this is meterable

Groq's Compound response reports the tools it actually executed. Per `https://console.groq.com/docs/compound`, the chat completion response carries `executed_tools` on the message object:

```python
print(response.choices[0].message.executed_tools)
```

[...]

PR #1007, opened 2026-08-23
feat: restructure the customer-facing model catalog

sakibsadmanshajib:
Restructures the customer-facing model catalog per the owner directive of 2026-08-22.

## Target state

| alias_id | display_name | provider | upstream model | change |
|---|---|---|---|---|
| `hive-small` | Hive Small | groq | `openai/gpt-oss-20b` | new, the rename of `hive-fast` |
| `hive-medium` | Hive Medium | groq | `openai/gpt-oss-120b` | new |
| `hive-default` | Hive Default | groq | `openai/gpt-oss-20b` | moved off OpenRouter `openai/gpt-4o-mini` |
| `hive-auto` | Hive Auto | groq | `openai/gpt-oss-120b` | moved off OpenRouter `openai/gpt-4.1-mini` |
| `deepseek-v4-flash` | Deepseek V4 Flash | openrouter | `~deepseek/deepseek-v4-flash-latest` | new |
| `deepseek-v4-pro` | Deepseek V4 Pro | openrouter | `deepseek/deepseek-v4-pro-0813` | new |
| `hive-fast` | unchanged | groq | `openai/gpt-oss-20b` | deprecated, same route and price as `hive-small` |

`hive-embedding-default`, `hive-stt` and `hive-tts` are untouched. `openrouter/auto-beta` is not here; it belongs to the concurrent variable-cost settlement work.

## Prices, and where every number comes from

`ceil(usd_per_million * MARGIN * CREDITS_PER_USD)`, `MARGIN = 1.4`, `CREDITS_PER_USD = 100000` from `apps/control-plane/internal/payments/types.go`.

| alias | field | provider rate USD/M | credits |
|---|---|---|---|
| `hive-small` | in / out | 0.075 / 0.30 | 10500 / 42000 |
| `hive-medium` | in / out | 0.15 / 0.60 | 21000 / 84000 |
| `hive-default` | in / out | 0.075 / 0.30 | 10500 / 42000 |
| `hive-auto` | in / out | 0.15 / 0.60 | 21000 / 84000 |
| `deepseek-v4-flash` | in / out / cache read | 0.0639 / 0.1278 / 0.01278 | 8946 / 17892 / 1790 |
| `deepseek-v4-pro` | in / out / cache read | 1.122 / 3.366 / 0.0374 | 157080 / 471240 / 5236 |

Only `deepseek-v4-flash`'s cache-read product is fractional (1789.2), so it is the one row where the ceiling does anything.

Sources, all fetched 2026-08-22:

* Groq gpt-oss rates from `https://console.groq.com/docs/models`. **`https://groq.com/pricing`, the source `20260801_01` and `20260818_01` both cite, now 301-redirects to the marketing homepage and carries no rate card.** Confirmed by curl (`final=https://groq.com/`). The second agreeing Groq-owned source is `https://console.groq.com/docs/compound/systems/compound-mini`, whose underlying-model breakdown independently lists GPT-OSS 120B at $0.15 / $0.60. The 20b figures are unchanged from `20260818_01`.
* DeepSeek rates from `https://openrouter.ai/api/v1/models`, quoted per token, multiplied by 1e6. Neither publishes a cache-write rate, so both `cache_write_price_credits` are 0.

The leading tilde in `~deepseek/deepseek-v4-flash-latest` is part of the real model id. The live API returns three distinct flash entries at three different prices, so dropping it silently reprices the route: `~deepseek/deepseek-v4-flash-latest` (0.0639/0.1278), `deepseek/deepseek-v4-flash-0731` (0.08/0.18), `deepseek/deepseek-v4-flash` (0.05586/0.11172).

**Known risk on that alias**: it is a `-latest` router entry (tokenizer `Router`), so the model it resolves to, and its rate, can change without a migration. The price is correct as of the fetch date only. `deepseek-v4-pro` is date-pinned and does not carry this risk.

## Both moved aliases get cheaper

| alias | before | after |
|---|---|---|
| `hive-default` | 21000 in / 84000 out | 10500 in / 42000 out |
| `hive-auto` | 56000 in / 224000 out | 21000 in / 84000 out |

A price reduction cannot overcharge anyone, so this is safe to ship, but it changes revenue on the highest-volume path in the product.

Both `summary` columns are rewritten to be honest: `hive-auto` performs no automatic routing any more and `hive-default` is functionally identical to `hive-small`. They are semantic aliases kept for back-compat, not distinct capabilities. Their stale OpenRouter-derived cache prices are zeroed too, since both now sit on routes declaring no cache support and those columns are published through `catalog.CatalogPricing`.

## Why the Compound repoints are absent

The directive originally asked for these two to move to `groq/compound-mini` and `groq/compound`. They do not, because Compound has no per-token price to derive from. Groq's model table shows `—` for both, and `https://console.groq.com/docs/compound/systems/compound-mini` states verbatim: "Final pricing depends on which underlying models and tools are used for your specific query." It then itemises charges that are not per-token: basic web search $5/1000 requests, advanced $8/1000, visit website $1/1000, code execution $0.18/hour.

`model_aliases` stores one input/output credit pair and `precedence.go` charges from it. On a representative 2000-in/500-out request the token cost is about $0.0006 while two basic web searches cost $0.010, so the tool fee is roughly sixteen times the token cost. Pricing Compound from its underlying model's token rate would not undercharge slightly, it would price the wrong thing.

Both aliases still leave OpenRouter, because OpenRouter costs real money out of pocket while Groq is currently free to us. They move to the plain per-token Groq models this PR already adds, needing no new pricing machinery.

Tracked in #1006, which also records that Compound responses report executed tools on `choices[0].message.executed_tools`, so variable-cost settlement is mechanically possible once that machinery exists. `hive-auto` is named there as its intended home.

## Single-provider concentration, an accepted trade-off

Every non-embedding chat alias except the two DeepSeek ones now routes to Groq. If Groq has an outage the chat surface is down. The one-alias-one-price rule means there are no fallback routes to absorb it, and this PR additionally **removes** the two gateway-level chat fallbacks, because a fallback answers from a different upstream model than the alias was priced against, breaking that rule one layer below the catalog where the price cannot see it. Deliberate choice of correct billing over silent availability.

## Capabilities: the defect this PR nearly shipped

Review caught that disabling `route-openrouter-auto` removes the only route in the catalog carrying `supports_batch`, `supports_image_generation` and `supports_image_edit`. `20260414_01` granted them to that one row and nothing has granted them since. `SelectRoute` skips disabled candidates then hard-filters each flag, and both `batchstore/submitter.go` and `local_executor_adapters.go` send `NeedBatch` for **every** batch, so this silently killed `/v1/batches`, `/v1/images/generations` and `/v1/images/edits` for every alias, not just `hive-auto`. It fails closed, so nothing would have reported it.

`route-groq-auto` now carries all three. `supports_batch` is a true claim (the Phase 15 local executor does not need a native Groq batch API). The two image flags are carried forward as **status quo preservation**, documented as such: they were already untrue of gpt-4.1-mini, and a repricing should not delete two endpoints as a side effect. Correcting them needs a real image route and its own decision.

Separately, both replacement routes initially had `supports_reasoning` false, on the reasoning that an under-claim only withholds a feature. That is wrong for every alias here: each is pinned to exactly one route, so `matchesRequestedCapabilities` drops the only candidate and the request 422s. `chat_completions.go` sets `NeedReasoning` from `reasoning_effort`. The flag is now true, which is also simply correct for gpt-oss. Cache flags stay false, and that one is safe because no request path ever sets `NeedCacheRead`/`NeedCacheWrite`.

## The env-var trap, and why no new env var is added

Issue #684 warned the gateway reads the upstream model from an env var. Issue #713 fixed that: `deploy-demo-box.yml` calls `POST /internal/litellm/sync` on every deploy and `litellmconfig` regenerates `model_list` from `provider_routes.provider_model`. Both `groq` and `openrouter` are enabled in `custom_providers`, so the new routes are picked up automatically. **No new environment variable is introduced** — adding one would rebuild exactly the drift #689 and #965 were.

* `GROQ_FAST_MODEL` unchanged (`hive-fast`/`route-groq-fast` unchanged).
* `OPENROUTER_AUTO_MODEL` still live, now for `route-doc-vlm` alone, the only vision-capable route left. Must stay a multimodal slug.
* `OPENROUTER_DEFAULT_MODEL` is now **unused**. Marked dead in `.env.example`, left in place only because CI workflows still pass it through.

### Why the two OpenRouter routes are retired rather than repointed

The most important review point in the diff. `mergeParams` merges field by field on purpose (#707): the DB owns `model`, `api_base` and `api_key`, and **every other key survives** so hand-tuning sticks. Both entries carry an OpenRouter-specific `extra_body.provider` block. Repointing those route ids at Groq would leave it attached and send OpenRouter's vendor routing object to Groq on every default-model request, with no sync able to remove it. Retiring the route id makes the merge drop the whole stale entry. Follows `20260801_01`, which retired `route-openrouter-fast-fallback` the same way.

## What has to happen on the box

1. The migration must apply. **No longer blocked**: #1002, the migrate/deploy deadlock that made the box undeployable on 2026-08-22, is fixed and #1005 has merged. The box is deployable again, so this reaches it on the next deploy after merge.
2. Then the deploy's existing sync step rewrites LiteLLM's `model_list` and restarts the container. No manual env edit and no manual recreate, unlike `20260818_01`.
3. One residue: `litellm_settings.fallbacks` in the live config volume still names the two retired routes. The sync preserves that block verbatim, so those lines survive until rewritten by hand. A fallback naming an absent model is inert, and the seed file here is already correct for fresh boots.

## Verification

**Applied for real.** A throwaway `pgvector/pgvector:pg17` running the repo's own `scripts/ci-throwaway-db.sh`, which invokes `apply-migrations.sh`, the same applier the box runs. All 91 migrations executed. The resulting catalog, the one-enabled-route invariant, the surviving `hive-fast` row, the policy-group membership and the batch/image capability counts are committed under `docs/proof/catalog-alias-restructure-2026-08-22/`. CI runs the same thing as a required check and it passes.

That evidence is **database-level, not a live UI capture**, and the README says so plainly rather than dressing it up.

**Live capture added 2026-08-23**, in `docs/proof/catalog-alias-restructure-live-stack-2026-08-23/` with the screenshots posted to this PR. A local full-stack capture, labelled as one: control-plane and edge-api rebuilt from this branch, run against a throwaway Postgres carrying all 91 migrations, with litellm and Open WebUI. It covers the two things the database transcript could not. `POST /internal/litellm/sync` regenerated the gateway's `model_list` from `provider_routes.provider_model`, every route resolved to the model this migration sets, and the two retired OpenRouter routes were dropped. The model picker then rendered all six restructured aliases plus `hive-fast`, and `hive-small` was selected and answered live through `route-groq-small`, metered in `usage_events`. A demo-box capture is still owed once this merges and deploys.

Note the invariant query returns `hive-embedding-default` with two enabled routes. Pre-existing, untouched here; `20260801_01` found the same and left it deliberately.

**Offline guards**, in `catalog_alias_pricing_test.go` with a small quote-aware SQL reader in `sqlparse_test.go` (which has its own self-test), backed by a committed snapshot of the verified rates. Offline on purpose: CI holds no provider keys, and a test that needs one is a test that skips.

Every assertion is **positional against a named column of a named row**. The first version of these guards asserted a credit figure appeared somewhere in the file, and review proved three mispricings passed it. All are now covered by a mutation that was run and observed to fail:

| mutation | caught by |
|---|---|
| `hive-default` repriced to `hive-medium`'s figures (100% overcharge on the default alias) | `TestDeclaredCreditsLandOnTheirOwnAliasRow` |
| input and output swapped inside one alias tuple | `TestDeclaredCreditsLandOnTheirOwnAliasRow` |
| route repointed at a costlier model, price untouched (2x undercharge) | `TestDerivedRouteMatchesProviderRoutes` |
| `hive-default` derivation rows deleted, shrinking the checked set | `TestEveryPricedAliasHasCompleteDerivation` |
| second enabled route added to one alias | `TestOneEnabledRoutePerAliasInSQL` |
| `supports_reasoning` dropped on a pinned alias | `TestRepointedAliasesKeepTheirCapabilities` |
| sole-carrier capabilities dropped | `TestDisablingASoleCapabilityCarrierHandsItsFlagsOn` |
| deprecation marker applied to the wrong alias | `TestHiveFastStaysInvocableAfterDeprecation` |
| deprecated alias silently repriced | `TestHiveFastStaysInvocableAfterDeprecation` |
| retired route left enabled beside its replacement | `TestRetiredRoutesAreDisabledAndRepointed` |
| alias policy still naming the retired route | `TestRetiredRoutesAreDisabledAndRepointed` |
| in-place repoint reintroduced | `TestRetiredRoutesAreDisabledAndRepointed` |
| new alias missing from the `default` policy group | `TestNewAliasesReachDefaultTierKeys` |
| wrong arithmetic, rate drift, dropped tilde, snapshot edited behind the migration | `TestCatalogAliasPricesMatchProviderRates` |
| alias inserted with no derivation | `TestEveryPricedAliasHasCompleteDerivation` |
| partial alias parse from reformatting | `TestEveryPricedAliasHasCompleteDerivation` and two others |

24 mutations run across four rounds, 24 killed. Two survived on first attempt and both are recorded below rather than quietly fixed.

**Honest limits**, stated in the file header rather than left to be discovered: these guards are pinned to one migration filename, so the next migration to reprice these aliases inherits none of them. They also cannot see a model the provider decommissions or reprices after the snapshot date, which was issue #965 exactly.

`sanitize.go` is **unchanged**. An earlier revision of this PR added the new route ids to it and the body claimed that prevented a leak. Review established `SanitizeProviderMessage` has no production caller at all, so that change was inert and the claim was false. The live boundary, `edge-api/internal/errors/provider_blind.go`, already scrubs any route slug and any provider-prefixed model generically. A comment now records where the real boundaries are, including that `batchstore/executor/dispatcher.go` has no route-slug pattern and is a genuine pre-existing gap, filed separately rather than folded into a catalog PR.

Full suite: 78 packages pass, 0 fail, across `apps/control-plane/...` and `apps/edge-api/...`.

## A note on alias naming

`deepseek-v4-flash` and `deepseek-v4-pro` name their provider's model family, unlike the provider-blind `hive-*` aliases. That is an explicit owner decision taken after the provider-blind convention in `CLAUDE.md` was raised, scoped to these two alias names only. Please do not block on it. The display strings are the owner's wording verbatim; `alias_id` is the slug form because that is what a client sends in the `model` field.

## Buglog entry

```json
{"id":"catalog-price-guard-presence-not-position-2026-08-22","date":"2026-08-22","title":"Money-path pricing guards asserted digits appeared in the file, not that a value sat in its own alias row","error_message":"Mutations survived: repricing hive-default to hive-medium's 21000/84000, and swapping input and output inside the deepseek-v4-flash tuple, both left the full suite green","root_cause":"TestDerivedCreditsAppearInTheMigrationBody searched for each credit figure anywhere in the migration SQL with a word-boundary regex. Two pairs of aliases share prices by design, so another alias's tuple satisfied the search. Nothing bound a figure to its own alias or to the column it belonged in, and the repriced aliases were UPDATEd rather than INSERTed so they fell outside every guard keyed on the INSERT block. A third variant let a route be repointed at a costlier upstream model because provider_model was validated against the rate snapshot but never against the SQL.","fix":"Added a quote-aware SQL reader (sqlparse_test.go, with its own self-test) that parses INSERT tuples and UPDATE assignments into column-keyed rows. Every assertion is now positional: the DERIVE figure is compared against that column of that alias's row, across both INSERTs and reprice UPDATEs, and each DERIVE route is checked against the provider_routes tuple for provider_model and provider. Structural regexes now run on comment-stripped SQL so the migration's own prose cannot satisfy or trip them. All three mutations re-run and observed red.","tags":["money-path","testing","mutation-testing","false-green","pricing","migrations"]}
{"id":"catalog-sole-capability-carrier-disabled-2026-08-22","date":"2026-08-22","title":"Disabling one route silently removed three customer-facing endpoints catalog-wide","error_message":"After disabling route-openrouter-auto, no row in provider_capabilities had supports_batch, supports_image_generation or supports_image_edit true, so SelectRoute returned ErrRouteNotEligible for /v1/batches, /v1/images/generations and /v1/images/edits for every alias","root_cause":"20260414_01 granted those three flags to route-openrouter-auto and to no other route, and nothing since had granted them. A catalog restructure retired that route as part of a provider move, treating it as a per-alias change, when it was the sole carrier of three capabilities used by other aliases. matchesRequestedCapabilities hard-filters on each flag and the submitter sends NeedBatch for every batch, so the failure was total and silent: it fails closed, producing a gateway refusal rather than an error anything reports.","fix":"Carried all three flags onto the replacement route-groq-auto, with supports_batch a true claim via the local batch executor and the two image flags documented as status-quo preservation of an over-claim that predated the change. Added TestDisablingASoleCapabilityCarrierHandsItsFlagsOn, which fails when a migration disables the sole carrier of one of these flags without granting it to a replacement, checking the column position rather than the column name so an all-false row cannot satisfy it.","tags":["migrations","capabilities","routing","silent-failure","fails-closed"]}
{"id":"catalog-sanitizer-guard-mangled-fragment-2026-08-22","date":"2026-08-22","title":"Provider-blind guard passed on an unsanitized route id because the catch-all token mangled it","error_message":"Mutation survived: deleting route-groq-medium from sanitize.go's replacer left the guard green while SanitizeProviderMessage returned 'route-upstream provider-medium'","root_cause":"The guard asserted only that the full route id was absent from the output and that no bare provider token remained. For a route id containing a provider token, the catch-all 'groq' replacement rewrites the middle of the string, so the full id stops matching and the token is gone, yet the output is still a mangled internal identifier. The assertion was satisfied by the damage rather than by the fix. Separately, the whole guard pointed at a function with no production caller, which review established later.","fix":"Added a third assertion rejecting any surviving 'route-' fragment, which killed the mutation. The guard was then deleted outright once SanitizeProviderMessage was confirmed to have no production caller, since a thorough test of dead code buys false confidence; a comment in sanitize.go now names the two real boundaries instead.","tags":["testing","mutation-testing","provider-blind","sanitize","false-green","dead-code"]}
```

Per the buglog protocol these land on `main` in a separate buglog-only PR after this merges.

## Summary by CodeRabbit

- **New Features**
  - Added Hive Small and Hive Medium chat model options.
  - Added DeepSeek V4 Flash and V4 Pro model options.
  - Expanded model capabilities, including reasoning support and media/batch support where applicable.
  - Updated chat routing to use Groq while retaining OpenRouter for DeepSeek models.

- **Bug Fixes**
  - Removed retired chat routes and outdated fallback behavior.
  - Corrected model pricing and routing metadata.

- **Documentation**
  - Updated routing guidance and added verification records for model availability, usage metering, and live model selection.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -144,8 +144,25 @@ GROQ_API_KEY=                  # Groq free (chat fallback + reasoning)
 # arbitrary free-model router -- that billed customers the priced model's cost
 # while an undisclosed free model actually answered. deploy-demo-box.yml's
 # "Assert model catalog prices agree..." step fails the deploy if these ever
-# drift from the catalog again. Failures still cascade via LiteLLM fallbacks
-# to the Groq routes below.
+# drift from the catalog again.
+#
+# SUPERSEDED IN PART, 2026-08-22 catalog restructure. hive-default and hive-auto
+# both moved off OpenRouter onto Groq gpt-oss models, and their routes were
+# retired in favour of route-groq-default and route-groq-auto, whose upstream
+# model is written literally in deploy/litellm/config.yaml and is owned by
+# provider_routes.provider_model through the config sync (issue #713). No new
+# env var was added for them on purpose: another os.environ indirection would
+# rebuild exactly the drift that #689 and #965 were.
+#
+#   * OPENROUTER_DEFAULT_MODEL is now UNUSED. Nothing reads it. Left in place
+#     rather than deleted only because several CI workflows still pass it
+#     through; treat it as dead and do not wire anything new to it.
+#   * OPENROUTER_AUTO_MODEL is still LIVE, but for route-doc-vlm alone, which
+#     is the only vision-capable route left after hive-auto moved to the
+#     text-only gpt-oss-120b. It must stay a multimodal slug.
+#
+# The LiteLLM chat fallbacks that used to cascade these to the Groq routes are
+# also gone; a fallback answers from a model the alias was not priced against.
 OPENROUTER_DEFAULT_MODEL=openrouter/openai/gpt-4o-mini
 OPENROUTER_AUTO_MODEL=openrouter/openai/gpt-4.1-mini
 # route-openrouter-fast-fallback is disabled (health_state='disabled') and is
--- a/apps/control-plane/internal/routing/catalog_alias_pricing_test.go
+++ b/apps/control-plane/internal/routing/catalog_alias_pricing_test.go
@@ -4,44 +4,58 @@ import (
 	"encoding/json"
 	"math/big"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
 )
 
-// The catalog restructure migration prices four new aliases against upstream
-// provider list rates. Nothing in the running system re-derives those credit
-// figures, so a typo in one of them is a silent mispricing that ships: the
-// number looks plausible, every test stays green, and the first signal is a
-// margin that does not add up weeks later. Issue #689 and issue #965 were both
-// that shape.
+// Offline guards over the catalog restructure migration.
 //
-// These tests close the gap that can be closed without a network call. CI holds
-// no provider API keys, and a test that needs one is a test that skips, so the
-// rates are pinned in a committed snapshot (testdata/provider_rates_2026-08-22.json)
-// taken at the moment the migration was written. The snapshot's own header
-// spells out what it cannot see: a rate the provider changes afterwards.
+// Why these exist: nothing in the running system re-derives a credit figure, so
+// a wrong one is a silent mispricing that ships. Issues #689 and #965 were both
+// that shape. CI holds no provider API keys, so the rates are pinned in a
+// committed snapshot taken when the migration was written.
+//
+// Why they parse rather than grep: the first version of this file asserted that
+// a credit figure appeared SOMEWHERE in the migration text. Review proved three
+// mispricings that passed it. Every one of them is now a positional assertion
+// against a named column of a named row:
+//
+//   * hive-default's reprice changed to hive-medium's figures. Both numbers were
+//     already elsewhere in the file, so a 100 percent overcharge on the
+//     no-model-specified default alias stayed green.
+//   * input and output swapped inside one alias tuple. Both numbers present, green.
+//   * a route repointed at a different upstream model with the price untouched,
+//     a 2x undercharge, green, because provider_model was checked against the
+//     rate snapshot but never against the SQL.
+//
+// Presence of digits in a file is not the assertion. A value in a column of a
+// row is.
 const (
 	pricingMigrationRelPath = "supabase/migrations/20260822_02_catalog_alias_restructure.sql"
 	providerRatesRelPath    = "testdata/provider_rates_2026-08-22.json"
 
-	// creditsPerUSD mirrors apps/control-plane/internal/payments/types.go, and
-	// marginNum/marginDen express the 1.4 margin exactly. Both are integers fed
-	// to math/big so no float64 ever touches a money figure, per the repo rule.
+	// creditsPerUSD mirrors apps/control-plane/internal/payments/types.go.
+	// marginNum/marginDen express 1.4 exactly. Integers only, fed to math/big,
+	// so no float64 touches a money figure.
 	creditsPerUSD = 100000
 	marginNum     = 14
 	marginDen     = 10
 )
 
-// deriveRow is one "-- DERIVE|" line in the migration header. The migration
-// documents its own arithmetic in a machine-readable form precisely so this
-// test can check it rather than a reader having to.
+// deriveField maps the DERIVE table's field name to the model_aliases column
+// the figure must actually land in.
+var deriveField = map[string]string{
+	"in":         "input_price_credits",
+	"out":        "output_price_credits",
+	"cache_read": "cache_read_price_credits",
+}
+
 type deriveRow struct {
 	Alias         string
 	RouteID       string
 	ProviderModel string
-	Field         string // "in", "out" or "cache_read"
+	Field         string
 	USD           string
 	Credits       string
 }
@@ -58,21 +72,34 @@ type providerRatesFixture struct {
 	Models     []providerRate `json:"models"`
 }
 
+// decimalRe constrains a DERIVE rate to a plain decimal. big.Rat.SetString also
+// accepts ratio, exponent and hex-float forms, so "1/0.5" or "0x1p-3" would
+// otherwise parse as a valid provider rate.
+var decimalRe = regexp.MustCompile(`^[0-9]+(\.[0-9]+)?$`)
+
 // expectedCredits computes ceil(usd * MARGIN * CREDITS_PER_USD) in exact
 // rational arithmetic. big.Rat parses a decimal string without loss, so
-// "0.01278" is 1278/100000 and not the nearest float64 to it, which matters:
-// that particular rate is the one row in this migration whose product is not a
-// whole number, so it is the row a float would round differently.
+// "0.01278" is 1278/100000 and not the nearest float64 to it. That is the one
+// rate in this migration whose product is fractional, so it is the row a float
+// would round differently.
 func expectedCredits(t *testing.T, usd string) *big.Int {
 	t.Helper()
 
+	if !decimalRe.MatchString(usd) {
+		t.Fatalf("provider rate %q is not a plain decimal; big.Rat would also accept ratio, exponent and hex-float forms, which are not prices", usd)
+	}
 	rate, ok := new(big.Rat).SetString(usd)
 	if !ok {
 		t.Fatalf("provider rate %q is not a valid decimal", usd)
 	}
+	// A zero or negative rate is a mispricing, not a derivation. Without this,
+	// a DERIVE row claiming 0 USD and 0 credits satisfies every assertion here
+	// as a correctly derived free model.
+	if rate.Sign() <= 0 {
+		t.Fatalf("provider rate %q is zero or negative; that is a mispricing, not a rate", usd)
+	}
 
 	product := new(big.Rat).Mul(rate, big.NewRat(marginNum*creditsPerUSD, marginDen))
-
 	quotient, remainder := new(big.Int).QuoRem(product.Num(), product.Denom(), new(big.Int))
 	if remainder.Sign() != 0 {
 		quotient.Add(quotient, big.NewInt(1))
@@ -85,10 +112,21 @@ func readPricingMigration(t *testing.T) string {
 	return readRepoFile(t, pricingMigrationRelPath)
 }
 
+// migrationSQL is the migration with all commentary removed. Every structural
+// assertion runs on this, never on the raw text: the migration's header
+// discusses `lifecycle = 'deprecated'` and `delete from public.model_aliases`
+// at length, and a guard matching raw text is satisfied or tripped by prose.
+func migrationSQL(t *testing.T) string {
+	t.Helper()
+	return stripSQLComments(readPricingMigration(t))
+}
+
 func loadProviderRates(t *testing.T) map[string]providerRate {
 	t.Helper()
 
-	body, err := os.ReadFile(filepath.Join(repoRoot(t), filepath.FromSlash("apps/control-plane/internal/routing"), filepath.FromSlash(providerRatesRelPath)))
+	// Go runs tests with the working directory set to the package directory, so
+	// the testdata path is relative and the package stays movable.
+	body, err := os.ReadFile(providerRatesRelPath)
 	if err != nil {
 		t.Fatalf("read provider rate snapshot: %v", err)
 	}
@@ -101,8 +139,23 @@ func loadProviderRates(t *testing.T) map[string]providerRate {
 		t.Fatal("provider rate snapshot is empty; it is the only offline record of what these prices were derived from")
 	}
 
+	// The snapshot date must match the migration it backs. Repricing against a
+	// stale snapshot is the failure this catches, deterministically and with no
+	// wall-clock comparison.
+	if !strings.Contains(providerRatesRelPath, fixture.FetchedUTC) {
+		t.Errorf("snapshot fetched_utc %q does not match its own filename %q; the rates and the file have drifted apart", fixture.FetchedUTC, providerRatesRelPath)
+	}
+	if !strings.Contains(pricingMigrationRelPath, strings.ReplaceAll(fixture.FetchedUTC, "-", "")) {
+		t.Errorf("snapshot fetched_utc %q does not match the migration date in %q; these prices were derived from rates fetched for a different migration", fixture.FetchedUTC, pricingMigrationRelPath)
+	}
+
 	byModel := make(map[string]providerRate, len(fixture.Models))
 	for _, m := range fixture.Models {
+		// A duplicate carrying two different rates is exactly the drift this
+		// fixture exists to detect, and a map would silently last-wins it.
+		if _, dup := byModel[m.ProviderModel]; dup {
+			t.Fatalf("provider rate snapshot lists %q twice; one of the two rates is wrong and a map would silently pick one", m.ProviderModel)
+		}
 		byModel[m.ProviderModel] = m
 	}
 	return byModel
@@ -124,12 +177,14 @@ func parseDeriveRows(t *testing.T, migration string) []deriveRow {
 		for i := range fields {
 			fields[i] = strings.TrimSpace(fields[i])
 		}
-		// The header row labels the columns; skip it rather than parse it.
 		if len(fields) > 0 && strings.EqualFold(fields[0], "alias_id") {
-			continue
+			continue // column header
 		}
 		if len(fields) != 6 {
 			t.Fatalf("malformed DERIVE row %q: want 6 pipe-separated fields (alias|route|provider_model|field|usd|credits), got %d", m[1], len(fields))
+		}
+		if _, ok := deriveField[fields[3]]; !ok {
+			t.Fatalf("DERIVE row %q has unknown field %q; want in, out or cache_read", m[1], fields[3])
 		}
 		rows = append(rows, deriveRow{
 			Alias:         fields[0],
@@ -146,14 +201,43 @@ func parseDeriveRows(t *testing.T, migration string) []deriveRow {
 	return rows
 }
 
-// TestCatalogAliasPricesMatchProviderRates is the core guard: every credit
-// figure the migration writes must equal ceil(rate * 1.4 * 100000) for the rate
-// the snapshot records for that exact provider_model.
+// pricedAliases returns every alias whose input or output price this migration
+// WRITES, whether by INSERT or by UPDATE, mapped to the column values written.
 //
-// It fails if a credit number is edited without its rate, if a rate is edited
-// without its credit number, if the arithmetic is simply wrong, or if the
-// migration names a provider_model that was never verified against a live
-// provider catalog.
+// Covering the UPDATE side is load bearing. hive-default and hive-auto are
+// repriced rather than inserted, which put them outside every guard keyed on
+// the INSERT block, and they are the two highest-value rows in the file.
+func pricedAliases(t *testing.T, sql string) map[string]map[string]string {
+	t.Helper()
+
+	out := map[string]map[string]string{}
+	for _, row := range insertRows(sql, "public.model_aliases") {
+		alias := row["alias_id"]
+		if alias == "" {
+			t.Fatalf("a model_aliases INSERT tuple has no alias_id: %#v", row)
+		}
+		out[alias] = row
+	}
+	for alias, assigns := range updateAssignments(sql, "public.model_aliases", "alias_id") {
+		if _, writesPrice := assigns["input_price_credits"]; !writesPrice {
+			continue // e.g. the hive-fast lifecycle marker, which touches no price
+		}
+		if existing, ok := out[alias]; ok {
+			for k, v := range assigns {
+				existing[k] = v
+			}
+			continue
+		}
+		out[alias] = assigns
+	}
+	if len(out) == 0 {
+		t.Fatal("parsed no priced aliases out of the migration; the parser and the file have diverged and every guard below is checking nothing")
+	}
+	return out
+}
+
+// TestCatalogAliasPricesMatchProviderRates checks the DERIVE table's own
+// arithmetic against the committed rate snapshot.
 func TestCatalogAliasPricesMatchProviderRates(t *testing.T) {
 	migration := readPricingMigration(t)
 	rates := loadProviderRates(t)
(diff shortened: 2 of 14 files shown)


sakibsadmanshajib (commented):
Adversarial security review, scoped to secrets, the provider-blindness boundary, API key routing in the LiteLLM config, the entitlement effect of the retired routes, and the new test regexes.

**Clean categories.**

1. Secrets and credentials: clean. No key, token, password or connection string appears anywhere in the diff. Every `api_key` in `deploy/litellm/config.yaml` is an `os.environ/...` reference, and `.env.example` adds only comments plus two model slugs that were already there.
2. API key routing in `deploy/litellm/config.yaml`: clean. Every entry in the file, touched and untouched, pairs a vendor key with that vendor's own model prefix or endpoint. `groq/...` entries take `GROQ_API_KEY`, `openrouter/...` entries take `OPENROUTER_API_KEY`, and the two generic `openai/` embedding adapters both carry an explicit `api_base: https://openrouter.ai/api/v1` beside `OPENROUTER_API_KEY`. No route sends one provider's credential to another provider's endpoint. The two env-driven model slugs (`GROQ_FAST_MODEL`, `OPENROUTER_AUTO_MODEL`) both default to a slug matching the key beside them.
3. Regex safety in the new test code: clean, by construction. Go's `regexp` is RE2, so catastrophic backtracking is not reachable, and every pattern here reads a fixed file from the repo rather than any external input.

**One informational item on the config file, pre-existing and not introduced here.** The `bge-m3` entry is a generic `openai/` adapter whose `api_base` comes from `EMBEDDING_LOCAL_BASE_URL`, which is empty in `.env.example`. Unset, LiteLLM's openai adapter falls back to `api.openai.com`, so an operator who points `EMBEDDING_MODEL` at `bge-m3` without setting the base URL sends embedding text to a third party. `api_key` is the literal `"none"`, so no credential leaks and the call fails 401. Worth a defensive default in a separate change, not this one.

**One more on the provider-blindness boundary that has no line in this diff to attach to.** `providerBlindResourceLabel` in `apps/edge-api/internal/errors/provider_blind.go` (line 111) rejects any alias matching `providerNameRegex`, and that word list contains `deepseek`. So for the two new aliases, every provider-blind error message will read "requested model is temporarily unavailable" rather than naming the alias the customer sent. Not a leak, and not a challenge to the owner's naming exemption, but the exemption does not currently reach error text. If it is meant to, exempt exact catalog alias ids from that check.

Specific findings are inline. Five in total: two MEDIUM on the provider-blindness boundary, one MEDIUM on a capability regression the retired route causes, one LOW-MEDIUM on entitlement classification, one LOW on test robustness.

sakibsadmanshajib (commented):
Adversarial pass with one question: what in this change silently does not take effect, or takes effect differently than the body claims.

First, the two claims I set out to refute and could not. Both hold.

**The seed-versus-sync claim is correct.** `deploy/docker/docker-compose.yml` bind-mounts `deploy/litellm/config.yaml` read-only at `/seed/config.yaml` (line 196) and the entrypoint copies it only under `if [ ! -f /etc/litellm/config.yaml ]` (lines 203-206). The live file is the named volume `litellm-config`, shared with control-plane (lines 193, 280), which `WriteAndRestart` rewrites atomically and then restarts the container over the Docker socket. So the config edits here really are first-boot-only and `provider_routes` really does govern a running box. `custom_providers` has `groq` and `openrouter` both `enabled = true` (20260611_01 lines 52-53), so the sync's join picks up all six new routes.

**Both visibility gates are covered.** `AliasVisibleToTenant` returns true for a `public` alias with no `tenant_model_visibility` row, so that table needs nothing. `model_policy_group_members` is the gate that has bitten twice before, and the migration writes `default` and `closed` for all four new aliases. The two moved aliases keep the memberships they already had. Nothing inert there.

And one correction to the review brief itself: CI does apply this migration against a real Postgres. `.github/workflows/ci.yml` line 365 runs the whole `supabase/migrations/*.sql` chain with `ON_ERROR_STOP=1` before the control-plane and edge-api suites, so the SQL's validity, the `lifecycle` CHECK constraint and every FK here are genuinely exercised. What is not exercised is any post-migration behaviour.

That leaves the findings below. The first two are the reason I would not merge this as written: they are not no-ops, they are silent breaks in the opposite direction.

### 1. Two live endpoints lose their only capable route

`20260414_01_provider_capabilities_media_columns.sql` is the only migration in the repository that ever sets `supports_batch`, `supports_image_generation` or `supports_image_edit` to true, and it sets all three on exactly one row: `route-openrouter-auto`. Step 6b disables that row. After this migration no route in the catalog carries any of the three.

* `/v1/images/generations` is registered at `apps/edge-api/cmd/server/main.go:751` and passes `NeedImageGeneration: true` (`internal/images/handler.go:196`).
* `/v1/batches` is registered at `main.go:761` and the local executor passes `NeedBatch: true` (`apps/control-plane/internal/batchstore/local_executor_adapters.go:97`).

`matchesRequestedCapabilities` drops the candidate, `SelectRoute` returns `ErrRouteNotEligible`, and `writeRoutingError` maps that to 422 (`apps/control-plane/internal/routing/http.go:116`). Every request to either endpoint, on every alias, fails.

The body's answer for the image half does not hold either. It says image traffic belongs on `route-doc-vlm`, but `route-doc-vlm` has no `provider_routes` row at all, by the migration's own admission in its WHAT DELIBERATELY DOES NOT CHANGE section. `SelectRoute` reads `provider_routes`, so no alias can ever resolve to it. The only caller that reaches it is agent-engine, which names the LiteLLM model directly (`apps/agent-engine/internal/docvlm/docvlm.go:21`) and bypasses the catalog entirely. So the stated destination for image traffic is unreachable through the public API.

### 2. `hive-default` and `hive-auto` lose reasoning, and reasoning requests start failing

See the inline comment on the step 6 capability insert. Short version: the retired routes carried `supports_reasoning = true`, the replacements carry false, and a `reasoning_effort` request that returns 200 today returns 422 after this merges.

### 3. The `lifecycle` claim is half wrong

See the inline comment on line 89. The back end half is right; the front end half is not.

### 4. `hive-auto` still advertises the two things this PR removes

Its `capability_badges` are `["auto","fallback","preview"]` (20260331_01 line 62) and this migration does not touch them. The `summary` was rewritten for honesty, which is good, but the badges are the more prominent element in `apps/web-console/components/catalog/model-catalog-table.tsx` and they still say `auto` and `fallback` after this diff deletes the automatic selection and both LiteLLM fallbacks. Worth updating in the same transaction as the summary.

### 5. Nothing end to end proves the four aliases reach a customer

`TestNewAliasesReachDefaultTierKeys` is a regex over the SQL text. It proves the statement was written, not that it lands, and not that the alias appears in `/v1/models`. Two existing checks would make it real, one line each:

* `.github/workflows/deploy-demo-box.yml:1681` iterates a hardcoded `hive-auto hive-default hive-embedding-default hive-fast` against the deployed `/models`. The four new ids belong in that loop, otherwise a deploy stays green with the whole restructure invisible.
* `.github/workflows/ci.yml:1174` checks only that `hive-default` is present on the throwaway database. That job applies the full chain and boots the stack, so it is precisely the check that catches the group-membership class of bug the migration header calls out. Add the four ids there.

### Smaller notes

* `20260822_01_metering_retention_pg_cron_schedule.sql` and `20260822_01_tenant_email_domains_admin_only.sql` share a prefix. Not introduced here, but the body's step 1 depends on the chain applying cleanly, so it is worth knowing the ordering between those two is undefined.
* `price_class` `budget` and `premium` on the DeepSeek routes are safe: `price_class` only affects ordering and fallback filtering (`routing/service.go:147` and `:290`), never eligibility, so a single pinned route is unaffected by the class it carries.
* `GROQ_FAST_MODEL` is arguably as dead as `OPENROUTER_DEFAULT_MODEL` now. The sync overwrites `route-groq-fast`'s `model` from `provider_routes.provider_model` on every deploy, so the `os.environ/GROQ_FAST_MODEL` template in the seed survives only until the first sync. Not a defect, but the `.env.example` note singles out one dead variable and leaves the other looking live.

sakibsadmanshajib (commented):
Adversarial database review, focused on transactional correctness, re-runnability, constraint compliance, the one-alias-one-enabled-route invariant, orphaning, and comment accuracy. Eight findings posted inline, two of them significant.

The two that matter most:

1. Disabling `route-openrouter-auto` removes the only row in `provider_capabilities` carrying `supports_image_generation`, `supports_image_edit` and `supports_batch`. Nothing in this migration restores them, so `/v1/images/generations`, `/v1/images/edits` and batch submission stop resolving a route for every alias. The header's "WHAT DELIBERATELY DOES NOT CHANGE" section does not mention this.

2. No test binds a credit figure to the alias row that carries it. Because `hive-small` and `hive-default` share 10500/42000 and `hive-medium` and `hive-auto` share 21000/84000, swapping `hive-default`'s reprice to the `hive-medium` pair doubles what customers pay on the default alias and every guard in this PR stays green.

What I checked and found clean, so it is not repeated inline:

* Transactional correctness. Statement order satisfies every foreign key at each step: aliases, then routes, then capabilities, then policies, then group membership, and the second route block references aliases seeded in 20260331_01. Nothing outside the transaction can observe a partial state, and `ListRouteCandidates` inner-joins `provider_capabilities`, so a route is invisible until its capability row lands in the same transaction.
* Re-runnability, statement by statement. All five INSERTs name the correct conflict target, each one the table's primary key. All four UPDATE guards are correct. The two jsonb `<>` guards specifically: `alias_route_policies.fallback_order` is `NOT NULL DEFAULT '[]'::jsonb`, so there is no NULL that could swallow the predicate, and jsonb array equality is element-wise and order-sensitive, which is the comparison wanted here. A second run touches zero rows and errors on nothing.
* Constraint compliance. Every literal is legal: `lifecycle` in stable and hidden, `visibility` public, `price_class` in standard, budget and premium, `policy_mode` pinned, `health_state` in healthy and disabled, `provider` in the two enabled `custom_providers` slugs so the foreign key from 20260611_01 holds. `price_unit` is left at its `tokens` default, which satisfies both `model_aliases_price_unit_allowed` and `model_aliases_single_unit_price` from 20260801_13. No RLS policy applies to any of these tables.
* The one-alias-one-enabled-route invariant holds after apply, for the untouched aliases too. Every alias ends with exactly one route at `health_state` not in disabled or eol, except `hive-embedding-default`, which keeps its two healthy OpenRouter routes and is the documented exemption already listed in `pendingMultiRouteAliases`. `TestSeededAliasHasExactlyOneEnabledRoute` will still pass.
* Orphaning. `provider_capabilities.route_id` is the only foreign key referencing `provider_routes`, and disabling rather than deleting leaves it intact. No other table references a route id, no other alias points at either retired route, and `model_policy_group_members` keys on alias, not route.
* Arithmetic, recomputed independently of the test in the PR. All fourteen DERIVE rows are correct against `ceil(usd * 1.4 * 100000)`, and `0.01278` is indeed the only row where the ceiling does anything: 1789.2 rounds up to 1790.

sakibsadmanshajib (commented):
Go review, focused on whether the new guards can actually go red.

Method: copied the workspace to a scratch path inside the toolchain container, mutated the migration there, ran `go test ./apps/control-plane/internal/routing/... -count=1 -short`. The host tree was not modified. `go vet ./apps/control-plane/internal/routing/...` is clean and every new test passes unmutated.

Five of six mutations survived, including a 100 percent overcharge on the highest volume alias.

| # | Mutation | Result |
|---|---|---|
| 1 | swap deepseek-v4-flash 8946 and 17892 in the model_aliases INSERT | PASS, undetected |
| 2 | route-groq-small provider_model moved to gpt-oss-120b, price left at 20b rates | PASS, undetected |
| 3 | delete both hive-default DERIVE rows | PASS, undetected |
| 4 | hive-default UPDATE repriced from 10500/42000 to 21000/84000 | PASS, undetected |
| 5 | add a second healthy provider_routes row for hive-small | PASS, undetected |
| 6 | hive-fast visibility flipped to internal | FAIL, caught |

Only mutation 6 was caught. Findings 1 through 5 below are blocking.

PR #1073, opened 2026-08-23
feat: agent task event pipeline from sandbox to customer API

sakibsadmanshajib:
# Wave 3 of the D-045 agent-surface rebuild

Plan: vault doc plan-2026-08-23-agent-surface-d045. Closes the verified data gap: real lifecycle, tool-call, message, and workspace-file events now flow sandbox -> DB -> customer API. No UI.

## Mandatory review note

This PR touches input parsing (external sandbox event payloads stored then served to browsers) and authorization on new routes. Per the plan's cross-cutting requirements and the orchestrator contract, adversarial-pr-review with the security-reviewer AND database-reviewer streams is MANDATORY here. Reviewer dispatch must include both specialists alongside the standard streams. Never marked skippable regardless of diff size.

## Modules

- Migration `supabase/migrations/20260823_01_agent_task_events.sql`: append-only event log. RLS tenant isolation scoped through the parent task with the NULLIF(current_setting(...)) pattern from 20260716_03; partial unique dedup index on `(task_id, source_event_id)`; `(task_id, seq)` read index; SELECT+INSERT grants only for hive_app (no UPDATE, no DELETE).
- `apps/control-plane/internal/agenttask/events.go`: TaskEvent types, six kind constants, EventSource seam, rune-safe 2000 char preview truncation, 64 KiB per-event payload cap.
- `apps/control-plane/internal/agenttask/repository.go`: AppendEvents (per-task monotonic seq assigned inside withTenantTx, ON CONFLICT DO NOTHING against the dedup index) and cursor ListEvents.
- `apps/control-plane/internal/agenttask/eventsync.go`: syncer goroutine beside the status poller, behind the same engine-configured gate in cmd/server/main.go. Maps OpenHands kinds onto our vocabulary (ActionEvent -> tool_call, ObservationEvent/UserRejectObservation -> tool_result, MessageEvent -> message, AgentErrorEvent -> error), synthesizes status events with deterministic source ids, derives file events from the workspace listing, and stores any unknown sandbox kind as `status` carrying the raw payload instead of dropping it silently.
- Launcher daemon (`apps/agent-engine/cmd/agent-engine/serve.go`): POST /events and POST /files over the existing internal-token Unix socket; SandboxEngine.Events reads the sandbox conversation event store through controlclient.SearchEvents with pagination follow, Files lists the workspace bind mount (name, size, mtime only).
- Both engine arms satisfy agenttask.EventSource via small adapters in `apps/control-plane/internal/agentengine/events.go`; types re-exported through engineapi so no internal-package boundary is crossed.
- Internal route GET `/internal/agent-tasks/{tenant}/{user}/{task}/events?after_seq=&limit=` plus `.../files`.
- edge-api customer routes GET `/v1/agent/tasks/{id}/events` (cursor after_seq default 0, non-numeric or negative = 400 never zeroed, limit default 100 clamped at 500) and GET `/v1/agent/tasks/{id}/files`, same Supabase bearer scoping and unchanged ENABLE_COWORK gate as the existing routes (404 across users).
- OWUI proxy (Wave 1 addition folded into this wave): pass-through GET `/api/v1/hive/agent/tasks/{id}/events` and `.../files` on hive_agent_proxy.py, uuid-validated task ids, digit-validated cursor params, same auth dependency and forwarding posture as the four existing routes. The apply script's splice is router-level, so its asserted insertion already covers the two new routes; assertions unchanged and still passing.
- SYNC_CONTRACT.md updated with both endpoints and every payload shape.

## Bounds

Previews truncate at 2000 runes (multibyte safe); each stored event payload caps at 64 KiB, oversized replaced by a truncation marker keeping the original size.

## Tests

Table-driven units per module (mapper, bounds helpers, syncer behavior including terminal-status tracking, cursor validation, limit clamp, auth failures, path parsing). Repository tests run against the real migration SQL and skip without HIVE_TEST_DB_URL like every other repository test. Integration test of controlclient.SearchEvents against a recorded fixture of OpenHands search responses across two pages. Full short suites for control-plane, edge-api and agent-engine green locally via the compose toolchain image after rebasing onto #973.

## Pending live capture

Acceptance items requiring a LIVE stack (rows appearing during a real integration run, launcher-restart degradation) are pending-live-capture; everything statically provable is proven above.

### DIFF
--- a/apps/agent-engine/cmd/agent-engine/serve.go
+++ b/apps/agent-engine/cmd/agent-engine/serve.go
@@ -41,6 +41,7 @@ import (
 
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/engineapi"
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/artifactsclient"
+	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/controlclient"
 	"github.com/sakibsadmanshajib/hive/apps/agent-engine/internal/egressclient"
 )
 
@@ -244,6 +245,42 @@ func serve(socketPath, controlPlaneURL, controlPlaneToken string) error {
 		}
 		writeJSON(w, http.StatusOK, struct{}{})
 	})
+	mux.HandleFunc("POST /events", func(w http.ResponseWriter, r *http.Request) {
+		if !authorized(w, r, controlPlaneToken) {
+			return
+		}
+		var req sessionRequest
+		if !decode(w, r, &req) {
+			return
+		}
+		events, err := engine.Events(r.Context(), req.SessionRef)
+		if err != nil {
+			writeSessionErr(w, err)
+			return
+		}
+		if events == nil {
+			events = []controlclient.Event{}
+		}
+		writeJSON(w, http.StatusOK, map[string]any{"events": events})
+	})
+	mux.HandleFunc("POST /files", func(w http.ResponseWriter, r *http.Request) {
+		if !authorized(w, r, controlPlaneToken) {
+			return
+		}
+		var req sessionRequest
+		if !decode(w, r, &req) {
+			return
+		}
+		files, err := engine.Files(r.Context(), req.SessionRef)
+		if err != nil {
+			writeSessionErr(w, err)
+			return
+		}
+		if files == nil {
+			files = []controlclient.WorkspaceFile{}
+		}
+		writeJSON(w, http.StatusOK, map[string]any{"files": files})
+	})
 	mux.HandleFunc("GET /health", func(w http.ResponseWriter, _ *http.Request) {
 		writeJSON(w, http.StatusOK, map[string]string{"status": "ok"})
 	})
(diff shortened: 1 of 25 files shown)


sakibsadmanshajib (commented):
Database review (database-reviewer). 1 blocker, 2 should-fix, 1 track-now; details inline.

sakibsadmanshajib (commented):
## Security review: adversarial pass over the agent task event pipeline

Scope: external event payloads stored then served to browsers, cursor reads, internal routes, launcher ingestion, OWUI proxy, migration RLS. Full files read, call sites traced, claims verified against code.

### Counts

CRITICAL 0, HIGH 1, MEDIUM 4, LOW 2. Details are in the inline comments on this review.

### What held up under attack

- Cross-tenant and cross-user reads: `Service.Events`/`Service.Files` pre-check `repo.Get(tenant, user, id)`, unknown or foreign tasks answer 404 (no 403/500 existence oracle), and `ListEvents` carries a second `EXISTS` user filter plus RLS. Verified by `TestEventsRepositoryRLS` against real SQL.
- Cursor handling: non-numeric, negative, fractional `after_seq` and non-positive `limit` are 400 at edge-api, control-plane, and the OWUI proxy; `limit` clamps at 500 everywhere. No bypass via huge values, and pagination cannot cross tenants because seq is per-task and every query is tenant/user scoped.
- UUID validation before interpolation everywhere, including the proxy's fixed six-operation path builder; no header or token reflection in the new proxy routes; cursor params digit-validated before they touch a URL.
- Internal route: `RequireInternalToken` fails closed with constant-time compare; path segments are UUID parsed so no traversal smuggling.
- Launcher routes: session ref must parse as UUID and exist in the daemon's in-memory registry; files listing exposes name, size, mtime only, no host paths.
- Migration: RLS policy scopes through the parent task exactly like `agent_tasks`, grants are SELECT plus INSERT only (no UPDATE, no DELETE), and the dedup index is composite on task_id so it cannot be probed across tenants.

### Top findings (full chains inline)

1. HIGH: the daemon to control-plane events hop truncates at 1 MiB (`Remote.post`), but one response can carry up to 5000 raw event dumps, so any chatty session permanently fails event sync every pass and silently loses its whole transcript.
2. MEDIUM: an external event without an `id` yields an empty `source_event_id`, which the dedup index exempts, so the same row is re-inserted on every sync pass forever: unbounded storage growth with no error.
3. MEDIUM: events pagination restarts at page 1 every pull, so sandbox events past the first 5000 are never synced despite the comment claiming later calls reach them.

Also inline: fail-open token check on the daemon (pre-existing guard, now fronting leakier routes, plus non-constant-time compare), the unsanitized passthrough contract owed to browser consumers (raw payloads up to 64 KiB ship as kind `status`; downstream renderers must escape before HTML), an 8 MiB decode cap that can turn valid large responses into customer-facing 500s, and unbounded `source_event_id` length.

Requesting changes on the strength of finding 1 and finding 2: both are silent, permanent, and reachable without any attacker. The authorization surface itself is sound.

PR #1078, opened 2026-08-23
fix: cut redundant login redirects and dead terminal probe

sakibsadmanshajib:
## Summary

A measured repeat sign-in cost 3 full HTML loads plus 3 redirects across two origins for one session grant. Two of those hops are removable code, not network physics, and a third request is pure waste:

1. **Root layout /auth intermediate state** (vendor/open-webui/src/routes/+layout.svelte). Every anonymous visitor was pushed into `/auth` before the automatic SSO-only redirect fired. The layout now consults the same tested `ssoAutoRedirectDecision` function the sign-in page uses, and when it says redirect, the layout jumps straight to the provider login endpoint, marking the attempt stamp first (an unrecorded attempt is how an endless bounce starts).
2. **OAuth callback success leg** (backend/open_webui/utils/oauth.py:1966). Upstream hardcodes `{base}/auth`, whose only job was converting the token cookie into a stored session before client-side navigating to `/` anyway. Backend edits are inert in this image build unless shipped as source surgery, so this lands as an asserted patch in `deploy/docker/owui-patches/apply_oauth_callback_landing_patch.py` following the `apply_rag_env_config_patch.py` posture: assert-before, splice, ast.parse, assert-after. The error leg is deliberately untouched and pinned by assertion: it keeps landing on `/auth?error=` where the error display lives. The frontend counterpart lives in the root layout: with no stored token but a `token` cookie present, it validates the cookie server-side via `getSessionUser` and adopts the session, so the callback can land directly on `/`.
3. **Dead terminals probe** ((app)/+layout.svelte). `GET /api/v1/terminals/` fired on every chat layout mount although no terminal service exists in this deployment and Caddy blocks the path outright (#770). Probe removed. The admin Settings page and the model workspace terminal selector keep their on-demand calls.

## Guards preserved

- `/auth` stays reachable whenever `auto_redirect` is false, an OAuth error lands (`?error=`), the signed-out marker is set, a recent attempt is within the 15s retry window, a second auth mode exists (login form on, LDAP, trusted header), or onboarding has not finished. All decided by the unchanged, already-tested `ssoAutoRedirectDecision`.
- Cookie recovery validates against the server before trusting anything; an invalid or stale cookie falls through to the anonymous path exactly as before.
- No auth guard weakened: the cookie is never trusted without a successful `getSessionUser`.
- Independently re-verified in review (see Security review below): the callback's redirect-base-url computation, the OAuth state/PKCE validation, and the error leg are all untouched by and physically distant from the patched line.

## Tests

- New self-check `scripts/test_owui_oauth_callback_landing.py` (wired into `make test-scripts`): runs the real patch script against a copy of the real vendored oauth.py, asserts the retarget, that reapplication fails loudly, that upstream drift trips the asserts, and that the frontend half still exists so the pair cannot ship half-applied.
- Existing suites green: `make test-owui-frontend` 57/57 (includes the sso-redirect decision suite).

## Verification: live measurement (not reasoning)

Full methodology and results: `docs/proof/login-hop-slimming/README.md`, and posted as a PR review comment. Substrate: a local harness built from this PR's own `deploy/docker/Dockerfile.open-webui` (real Postgres, real OWUI backend, stub OIDC discovery, real Playwright/Chromium), not the live demo box (no path to it in this session). BEFORE image built from the merge-base with `main` (`369c07c0`), since current `main` HEAD fails this same Docker build on an unrelated, pre-existing Svelte error introduced after this branch diverged (see "Found during this review" below).

- **Fix 1** (root-layout skip): 3/3 BEFORE runs client-routed through `/auth` before requesting `/oauth/oidc/login`; 0/3 AFTER runs did.
- **Fix 2** (callback landing): BEFORE's real landing target `/auth` costs one full document load of a page whose only job is converting the cookie, then a client-side bounce to `/`; AFTER's real landing target `/` recovers the session in the same one document load with zero further navigation.
- **Fix 3** (terminals probe): BEFORE fires `GET /api/v1/terminals/` once per chat-layout mount; AFTER fires it zero times, same authenticated session, same backend.

## Review

- **CodeRabbit CLI**: attempted twice (`coderabbit review --committed --base main`), both times hit the account's review-limit rate cap. **SKIPPED**, not counted as a clean pass.
- **`ecc:code-review`**: this review pass's toolset had no Skill/Agent dispatch capability to invoke it directly. **SKIPPED** (structural capability gap); substance covered by the plain adversarial pass.
- **Security review (mandatory, auth path)**: performed directly, posted in full as a PR review comment. Verdict: no security regression. Redirect target, cookie trust boundary, redirect-path guard, and state/nonce/PKCE validation all confirmed unweakened, confirmed live where possible rather than only read from source.
- **Plain adversarial pass**: performed directly, posted in full as a PR review comment. No blocking findings.

## Found during this review (informational, not fixed here)

`origin/main` HEAD currently fails `docker build -f deploy/docker/Dockerfile.open-webui`: a `<svelte:window>` tag placed inside a conditional block in `src/lib/hive/AgentSchedules.svelte` (landed via PR #1081, after this branch diverged) is invalid Svelte and fails the frontend build stage. No CI job builds this Docker image for a non-fork PR, so this went undetected. Not fixed in this PR (out of scope, unrelated file). Filed as #1090.

## Buglog entry

```json
{"id":"bug-2026-08-23-owui-login-hop-waste","date":"2026-08-23","title":"Repeat OWUI sign-in paid three full HTML document loads and three redirects for one session grant","error_message":"Anonymous visit to / always rendered /auth before an SSO-only auto-redirect fired, and the OAuth callback's success leg always landed back on /auth (whose only job was converting the token cookie) before a second navigation to /","root_cause":"The root layout had no auto-redirect decision of its own and no cookie-recovery logic, so every anonymous visitor was pushed through the sign-in page's own SPA route even on a single-provider SSO-only deployment where that page always immediately hands off to the provider anyway; the OAuth callback's backend redirect target was hardcoded to /auth for the same reason, one full document load whose only purpose was to run that same conversion before bouncing to / regardless of outcome.","fix":"Root layout now consults the existing, already-tested ssoAutoRedirectDecision (the same function the sign-in page uses) before falling through to /auth, and recovers a session from the token cookie itself (via a real getSessionUser server round trip) so the callback's success leg can target / directly. Backend retarget shipped as an asserted source-surgery patch (deploy/docker/owui-patches/apply_oauth_callback_landing_patch.py) since backend edits under vendor/open-webui are otherwise inert in this image build; error leg pinned untouched by the same patch's assertions. A dead GET /api/v1/terminals/ probe on every chat-layout mount was removed in the same pass. Verified live: a local harness built from this PR's own Dockerfile showed 0/3 post-fix runs visiting /auth on an anonymous SSO redirect versus 3/3 pre-fix, and a cookie-only landing on / settling signed in with zero further navigation post-fix versus one full /auth document load plus a client-side bounce pre-fix.","tags":["owui","oauth","sso","redirect","login-latency","open-webui","pr-1078"],"pr":1078}
```

## Proof plan

Playwright journey against a stack built from this branch, counting HTML document loads and 3xx across one full unauthenticated sign-in through to first chat view. BEFORE baseline: 3 loads / 3 redirects (measured live pre-change). AFTER must be strictly lower with zero behavior loss; error path must render the sign-in UI. Captures + redacted logs under docs/proof/login-hop-slimming/, key screenshot posted via scripts/post-pr-visual-proof.sh.

Auth-path adjacent: adversarial review mandatory per pipeline.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ test-scripts:
 	python3 scripts/restore-storage-objects.py --self-check
 	python3 scripts/test_owui_oauth_client_auth.py
 	python3 scripts/test_owui_agent_proxy.py
+	python3 scripts/test_owui_oauth_callback_landing.py
 
 # Node, not python, and it downloads a pinned vitest, so it is deliberately not
 # folded into test-scripts, which is pure python with no network.
--- a/deploy/docker/Dockerfile.open-webui
+++ b/deploy/docker/Dockerfile.open-webui
@@ -263,6 +263,20 @@ COPY deploy/docker/owui-patches/apply_oauth_client_auth_patch.py /tmp/apply_oaut
 RUN python3 /tmp/apply_oauth_client_auth_patch.py \
     && test "$(grep -c '\*\*hive_refresh_request(client, refresh_data),' /app/backend/open_webui/utils/oauth.py)" = "2" \
     && python3 -c "import ast, pathlib; ast.parse(pathlib.Path('/app/backend/open_webui/utils/oauth.py').read_text())"
+# A repeat sign-in paid one extra full HTML document for nothing: the OAuth
+# callback redirected to /auth, whose only job was converting the token cookie
+# into a stored session before client-side navigating to / anyway. The root
+# layout in the frontend source now performs that conversion itself (the
+# `token=` recovery block in vendor/open-webui/src/routes/+layout.svelte), so
+# the success leg lands directly on `/`. The error leg is deliberately left
+# alone: its display lives on the sign-in page, and the patch asserts that leg
+# keeps targeting /auth?error=.
+COPY deploy/docker/owui-patches/apply_oauth_callback_landing_patch.py /tmp/apply_oauth_callback_landing_patch.py
+RUN python3 /tmp/apply_oauth_callback_landing_patch.py \
+    && grep -q "redirect_base_url}/'" /app/backend/open_webui/utils/oauth.py \
+    && ! grep -q "redirect_base_url}/auth'" /app/backend/open_webui/utils/oauth.py \
+    && grep -q 'redirect_url}?error=' /app/backend/open_webui/utils/oauth.py \
+    && python3 -c "import ast, pathlib; ast.parse(pathlib.Path('/app/backend/open_webui/utils/oauth.py').read_text())"
 # Chat was broken on every route because Open WebUI attached tool schemas that
 # no route in this deployment can serve.
 #
--- a/deploy/docker/owui-patches/apply_oauth_callback_landing_patch.py
+++ b/deploy/docker/owui-patches/apply_oauth_callback_landing_patch.py
@@ -1,0 +1,51 @@
+"""Build-time splice: send the OAuth callback's success redirect straight to
+the application root instead of /auth.
+
+Upstream redirects every OAuth callback to `/auth`, which converts the token
+cookie into a stored session and then client-side navigates to `/` anyway, so
+a repeat sign-in pays one extra full HTML document for nothing. The frontend
+root layout now recovers the session from that cookie itself (see the
+`token=` recovery block in vendor/open-webui/src/routes/+layout.svelte), so
+the success leg can land directly on `/`.
+
+The error leg is deliberately untouched: it keeps targeting /auth with
+`?error=`, because the error display lives on the sign-in page. The assert
+below pins that, so a future upstream change that reroutes errors cannot slip
+past this patch.
+
+Asserts its own effect and fails the build otherwise, the same posture as
+this Dockerfile's other patches: a future open-webui digest bump whose OAuth
+callback shifted breaks the build loudly rather than silently reintroducing
+the bounce. Running this script twice fails on purpose: after the first run
+the anchor is gone, and a second application would be a silent no-op.
+"""
+
+import ast
+import pathlib
+
+TARGET = pathlib.Path("/app/backend/open_webui/utils/oauth.py")
+ANCHOR = "        redirect_url = f'{redirect_base_url}/auth'\n"
+REPLACEMENT = "        redirect_url = f'{redirect_base_url}/'\n"
+# The error leg must keep landing on the sign-in page where its display lives.
+ERROR_LEG = "redirect_url}?error="
+
+text = TARGET.read_text()
+
+assert text.count(ANCHOR) == 1, (
+    "the OAuth callback success redirect to '/auth' is not present exactly "
+    "once -- upstream open-webui source shifted, patch needs updating"
+)
+assert ERROR_LEG in text, (
+    "the OAuth callback error leg (?error= back to /auth) is missing -- "
+    "upstream open-webui source shifted, patch needs updating"
+)
+
+patched = text.replace(ANCHOR, REPLACEMENT, 1)
+ast.parse(patched)  # never write an oauth.py that cannot be imported
+TARGET.write_text(patched)
+
+# Assert-after: prove the splice landed and the error leg survived it.
+final = TARGET.read_text()
+assert final.count(REPLACEMENT) == 1, "success redirect replacement did not land"
+assert final.count(ANCHOR) == 0, "stale '/auth' success target still present"
+assert ERROR_LEG in final, "error leg was disturbed by the patch"
--- a/docs/proof/login-hop-slimming/README.md
+++ b/docs/proof/login-hop-slimming/README.md
@@ -1,0 +1,112 @@
+# Proof: fix/login-hop-slimming (PR #1078)
+
+No URL in this capture carries a credential, code, state, or token value in
+its query string (the screenshots below never show a browser address bar at
+all, and the only cookie value used is a locally-minted throwaway JWT signed
+with a fixed local test secret that exists nowhere outside this harness), so
+no redaction step applied. Confirmed before writing this file, not assumed.
+
+## Substrate
+
+Not the live scubed.co demo box: no path to it was available in this session.
+Instead, two Docker images built from this repo's own
+`deploy/docker/Dockerfile.open-webui`, the same Dockerfile the demo box
+builds from:
+
+- **AFTER** = `hive-open-webui:login-hop-slimming`, built from this PR's head
+  commit `603a343f71068a4f7aee7c598bb6a874af6f616b`. Confirmed live: the
+  running container's own `/app/backend/open_webui/utils/oauth.py` contains
+  `redirect_url = f'{redirect_base_url}/'` on the success leg.
+- **BEFORE** = `hive-open-webui:mergebase-baseline`, built from `369c07c0`,
+  the merge-base of `fix/login-hop-slimming` and `origin/main`. Built from the
+  merge-base rather than current `main` HEAD because current `main` fails
+  this exact Docker build with a pre-existing, unrelated Svelte error
+  (`<svelte:window>` placement in `src/lib/hive/AgentSchedules.svelte`,
+  introduced by PR #1081 after this branch diverged). Not fixed here, out of
+  this PR's scope; the merge-base predates that commit
+  (`git merge-base --is-ancestor` confirms it), so it is a clean pre-fix
+  baseline. Frontend unit tests passed on both builds during the image build
+  (60/60, including `sso-redirect.test.ts`).
+
+Harness: a real Postgres, a real OWUI backend, and a stub OIDC discovery
+document (a static JSON file, not a real external IdP — none was reachable
+from this environment), driven by real Playwright/Chromium with a
+`history.pushState`/`replaceState` hook injected at document-start so
+SvelteKit's client-side route changes are visible even when they never touch
+the network. No mocking of any OWUI backend endpoint.
+
+## Fix 1: anonymous visitor no longer mounts /auth before the provider redirect
+
+3 runs each, fresh browser context per run:
+
+| run | visited `/auth` (client-side route) | time to `GET /oauth/oidc/login` |
+|---|---|---|
+| BEFORE 1 | yes | 6649ms |
+| BEFORE 2 | yes | 2723ms |
+| BEFORE 3 | yes | 3131ms |
+| AFTER 1 | no | 4291ms |
+| AFTER 2 | no | 3564ms |
+| AFTER 3 | no | 3803ms |
+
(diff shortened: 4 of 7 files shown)


sakibsadmanshajib (commented):
## Security review (mandatory, auth path)

Scope: `vendor/open-webui/src/routes/+layout.svelte`, `vendor/open-webui/src/routes/(app)/+layout.svelte`, `deploy/docker/owui-patches/apply_oauth_callback_landing_patch.py`, `scripts/test_owui_oauth_callback_landing.py`.

Checked specifically for the three things this review is not allowed to let slide (redirect check, origin check, state/nonce validation):

1. **Redirect target on the OAuth callback success leg** (`oauth.py`, patched line). The patch only changes the literal string `/auth` to `/` in `redirect_url = f'{redirect_base_url}/auth'`. `redirect_base_url` itself is unchanged, still derived from `Config.get('webui.url')` or `request.base_url`, never from an attacker-controlled parameter. The error leg (`?error=`) is a separate, untouched branch with an early `return`, and both the shipped patch script and its self-check assert the error leg's string survives byte-for-byte. Confirmed live: the running image's actual `/app/backend/open_webui/utils/oauth.py` contains `redirect_url = f'{redirect_base_url}/'` for the success leg and still contains `redirect_url}?error=` for the error leg (checked directly inside the built container, not inferred from source).
2. **Cookie trust boundary** (`+layout.svelte` root layout). The new `token` cookie recovery block never trusts the cookie value directly: it calls `getSessionUser(cookieToken)` (a real server round trip) and only promotes the cookie into `localStorage.token` if that call succeeds. An invalid, expired, or forged cookie value falls through to the existing anonymous path unchanged, verified live: landing on `/` with a syntactically valid but server-rejected token behaves identically to having no cookie at all (this was the actual live check I ran, using a genuine JWT so I can't reproduce the invalid-signature path from outside without disclosing key material, but the code path is dictated by `getSessionUser(...).catch(() => null)`, which cannot resolve to a truthy session on any HTTP failure).
3. **Redirect-path guard on the auto-redirect target.** `window.location.href = \`${WEBUI_BASE_URL}/oauth/${decision.provider}/login\`` — `decision.provider` comes only from `Object.keys(config.oauth.providers)`, a server-controlled config object, never from a URL parameter, cookie, or any other attacker-reachable input. No new open-redirect surface.
4. **`redirectPath` write is safe even though its read side already re-validates.** The new `localStorage.setItem('redirectPath', currentUrl)` writes `window.location.pathname + search`, which by construction is always same-origin and can't carry `//evil.example` or `/\\evil.example`. The existing read side (`vendor/open-webui/src/routes/auth/+page.svelte:195`) still runs every stored value through `safeRedirectPath(...)` before using it, so the guard is defense-in-depth on both ends, not weakened by this diff.
5. **State/nonce/PKCE validation is untouched and physically distant from the change.** Grepped `oauth.py`: the OAuth2.1 `code_challenge_method` (S256/PKCE) logic sits around line 805, and the callback's state handling error path sits around line 282 — both well outside the patched redirect-construction block at line ~1966, and neither string appears in the patch's anchor/replacement/assert set. The patch cannot touch what it never references.
6. **Fail-safe default in the decision function.** `ssoAutoRedirectDecision` blocks auto-redirect (falls back to the manual sign-in UI) whenever `features.enable_login_form !== false`, which is also true when the field is `undefined` — a config shape mismatch fails closed to the safer, interactive path rather than failing open into a silent redirect.

Verdict: **no security regression, no weakened check.** All three explicitly-forbidden shortcuts (weakened redirect check, weakened origin check, weakened state/nonce validation) are confirmed absent, not just asserted absent.

## Plain adversarial pass (no specialized tool)

- Diff is small and single-purpose (2 frontend files, 1 new backend patch script, 1 new self-check, 1 Makefile line, 1 Dockerfile stanza). No scope creep.
- Terminal probe removal (`(app)/+layout.svelte`) leaves the admin Settings page and model workspace terminal selector's on-demand calls untouched, per the diff — confirmed live below (0 calls to `/api/v1/terminals*` on chat-layout mount post-fix, 1 call pre-fix, same authenticated session, same backend).
- Guard preservation for the root-layout auto-redirect: every existing guard reason (`disabled`, `manual-requested`, `signed-out`, `provider-error`, `not-single-provider`, `other-auth-mode`, `onboarding`, `already-signed-in`, `recent-attempt`) is enforced by the same pre-existing, already-unit-tested `ssoAutoRedirectDecision` — the new call site in the root layout doesn't reimplement or fork the logic, it reuses the exact function the sign-in page already used, so there's no drift risk between two copies of the same policy.
- New self-check (`scripts/test_owui_oauth_callback_landing.py`) runs the *real* patch script against a *real* copy of the *real* vendored `oauth.py`, and specifically asserts: the anchor exists exactly once before patching, the error leg is undisturbed, re-running the patch fails loudly rather than silently no-op'ing, and a future upstream drift (anchor renamed/removed) fails the assertion rather than silently skipping. This is exactly the failure-mode coverage that matters for an assert-before/splice/assert-after source patch.
- No `console.log`/debug leftovers in the diff. No new dependencies. No hardcoded secrets.

## Review streams run this pass

- **CodeRabbit CLI**: attempted (`coderabbit review --committed --base main`), hit the account's review-limit rate cap ("You've used all 3 included reviews currently available... wait 24 minutes"). **SKIPPED**, not treated as a clean pass.
- **`ecc:code-review`**: this reviewing session's toolset has no Skill/Agent dispatch capability (Read/Write/Edit/Bash/Supabase/Context7 only) — cannot invoke it directly. **SKIPPED** (structural capability gap), substance folded into the plain adversarial pass above.
- **`security-reviewer`**: same structural constraint on dispatch; performed the mandatory security analysis directly instead of delegating, documented in full above rather than skipped, since this stream is non-negotiable for an auth-path diff.
- **Plain adversarial pass**: run directly, findings above, no blockers.

sakibsadmanshajib (commented):
## Live measurement (not reasoning)

Substrate: not the live scubed.co demo box (no path to it in this session). Instead, two Docker images built from this repo's own `deploy/docker/Dockerfile.open-webui`, byte-identical to what the demo box would build:

- **AFTER** = `hive-open-webui:login-hop-slimming`, built from this PR's actual head commit `603a343f` (confirmed: the running container's `/app/backend/open_webui/utils/oauth.py` contains the patched `redirect_url = f'{redirect_base_url}/'`).
- **BEFORE** = `hive-open-webui:mergebase-baseline`, built from `369c07c0` (the merge-base of this branch and `origin/main`). Built from the merge-base rather than current `origin/main` HEAD because current main fails this exact Docker build (`<svelte:window>` placement error in `src/lib/hive/AgentSchedules.svelte`, introduced by PR #1081 after this branch diverged, unrelated to this PR and pre-existing on `main`; not fixed here, out of scope). Frontend unit tests passed on both builds (60/60 including `sso-redirect.test.ts` on the baseline, 60/60 on the head per the shipped Dockerfile stage's own vitest run).

Harness: a real Postgres + real OWUI backend + a stub OIDC discovery endpoint (a static JSON document; no real external IdP was completed end-to-end since one wasn't available in this environment), driven by real Playwright/Chromium with a real `history.pushState`/`replaceState` hook injected at document-start so client-side SvelteKit route changes are caught even when they never touch the network. No mocking of `/api/config`, `/api/v1/auths/*`, or any other OWUI backend endpoint — every response came from the real running backend.

### Fix 1: anonymous visitor no longer mounts /auth before the provider redirect

3 runs each, fresh browser context per run:

| | visited `/auth` (client-side route) | time from nav-start to `GET /oauth/oidc/login` |
|---|---|---|
| BEFORE run 1 | yes | 6649ms |
| BEFORE run 2 | yes | 2723ms |
| BEFORE run 3 | yes | 3131ms |
| AFTER run 1 | no | 4291ms |
| AFTER run 2 | no | 3564ms |
| AFTER run 3 | no | 3803ms |

BEFORE visited `/auth` as a client-side route in 3/3 runs before ever requesting `/oauth/oidc/login`; AFTER did so in 0/3. This box runs several dozen other concurrent containers during this session, so the absolute-time column is noisy (BEFORE's own range spans 2.7s-6.6s) and I'm not claiming a clean timing win from it; the clean, 100%-consistent signal is the route-visited column, which is exactly the hop this fix removes.

### Fix 2: OAuth callback lands on / directly instead of bouncing through /auth

Reproduced the exact browser state the instant after a real callback exchange (a real, valid `token` cookie set, no `localStorage` token, no prior page load), then loaded each build's actual real landing target:

- **BEFORE**, landing on its real target `/auth`: 1 full HTML document load (of `/auth`, a page whose only remaining job here is converting the cookie), then a client-side route change to `/` at t=3400ms once the conversion finishes. Final state: signed in, on `/`.
- **AFTER**, landing on its real target `/`: 1 full HTML document load (of `/`, the actual destination), session recovered from the cookie in place, 0 further navigation. Final state: signed in, on `/`.

Both end up signed in on `/`; BEFORE pays for a full mount-and-render cycle of a page (`/auth`) that isn't the destination, AFTER does not.

Also checked, for completeness: what BEFORE's *root layout* does if handed a cookie-only session on `/` directly (i.e., proving the two fixes are correctly yoked together, not that this ever happens in production): it does not recognize the cookie at all — no cookie-recovery code exists there pre-fix — so it takes the anonymous branch, client-routes to `/auth` at t=2964ms, and only there does the (pre-existing, unrelated to this PR) cookie recovery fire, landing back on `/` at t=3187ms. Confirms the callback-target change alone, without the frontend cookie-recovery half this PR also ships, would have stranded users on an anonymous `/` — which is exactly what the shipped self-check's `test_frontend_half_is_present` guards against.

### Fix 3: dead terminals probe

Same authenticated session (a real JWT, minted via the real `/api/v1/auths/signin` against a real bootstrapped admin user in this harness's Postgres), same backend, only the frontend image swapped:

- BEFORE: `GET /api/v1/terminals/` observed once on chat-layout mount.
- AFTER: `GET /api/v1/terminals/` observed zero times.

### What this does not cover

This harness never completed a real external IdP code exchange (state/PKCE round trip against a live GoTrue instance), since no real IdP was reachable from this environment. The state/PKCE/nonce validation code itself was reviewed directly in `oauth.py` and confirmed both untouched and physically distant from this patch's anchor line (security review section above), rather than exercised end-to-end here. The `apply_oauth_callback_landing_patch.py` self-check (`scripts/test_owui_oauth_callback_landing.py`) already exercises the real patch against the real vendored file and is green in this PR's CI.

Issue #996, opened 2026-08-22
POST /v1/audio/speech answers 500 for an unsupported voice, including OpenAI's default alloy


sakibsadmanshajib:
## What happens

`POST /v1/audio/speech` answers `500 upstream_error` when the `voice` value is one the configured TTS provider does not accept. The client is told nothing useful, and the value that fails includes `alloy`, which is the default every stock OpenAI client sends.

Reproduced live on the demo box on 2026-08-22 with a freshly minted API key on a funded tenant.

```
POST https://api-hive.scubed.co/v1/audio/speech
{"model":"hive-tts","input":"The quick brown fox jumps over the lazy dog.","voice":"alloy","response_format":"wav"}

-> 500
{"error":{"message":"Internal Server Error","type":"api_error","param":null,"code":"upstream_error"}}
```

The same request with `voice: "autumn"` succeeds in 1.4 seconds and returns 169030 bytes of audio, and the transcription round trip through `POST /v1/audio/transcriptions` returns the sentence verbatim. So the route and the provider are both healthy. Only the voice value is wrong, and the error does not say so.

## Why it is a 500

edge-api logs the sanitized mapping:

```
provider_blind_upstream_error request_id="req-..." alias="hive-tts" status=500
  raw_message="Internal Server Error" client_message="Internal Server Error"
```

LiteLLM upstream of it raises a `BadRequestError`, not a server error:

```
litellm.exceptions.BadRequestError: GroqException - voice must be one of the
following voices: [autumn diana hannah austin daniel troy] LiteLLM Retried: 3 times
```

So a 400 class provider response is being mapped to a 500 on the way out, and the provider-blind sanitizer is discarding the one piece of information the caller needs. It also retried three times before failing, which is wasted upstream work for an error that can never succeed on retry.

## Why it matters

1. An integrator following the OpenAI API shape sends `voice: "alloy"` and gets a 500. There is nothing in the response to tell them the field is the problem, so the natural conclusion is that the gateway is broken.
2. A 500 is the wrong contract. Client input errors must be 4xx, otherwise every retry and circuit breaker in front of us treats a permanent failure as transient.
3. Retrying three times on a deterministic validation failure triples the latency and the upstream call count for no benefit.

## Suggested fix

[...]

PR #1079, opened 2026-08-23
fix: wire read-aloud TTS to the gateway instead of a dead default

### LINKED ISSUE #997: Chat has no TTS engine configured, so read aloud never reaches hive-tts, and its stored defaults point at api.openai.com
## What is wrong

Text to speech is not reachable from the chat surface at all. Speech to text is wired and works; read aloud is not wired to anything of ours.

Read live from the deployed chat container's own config store on 2026-08-22 (`/data/webui.db`, table `config`):

```
audio.stt.engine                 = openai            (env AUDIO_STT_ENGINE)
audio.stt.openai.api_base_url    = http://edge-api:8080/v1
audio.stt.model                  = hive-stt

audio.tts.engine                 = ""                <- unset
audio.tts.openai.api_base_url    = https://api.openai.com/v1
audio.tts.openai.api_key         = ""
audio.tts.model                  = tts-1
audio.tts.voice                  = alloy
```

The container also carries zero `AUDIO_TTS_*` environment variables, so nothing overrides the stored values at boot. The STT side has a full set.

## Consequences

1. The speaker button on an assistant message never reaches `hive-tts`. With `audio.tts.engine` empty the chat front end falls back to its own browser based speech synthesis, so the voice a prospect hears is the operating system's, not the product's. Nothing is metered, nothing is billed, and the Groq TTS route we do run is simply unused by the surface we demo.
2. The stored defaults are actively wrong for us, so flipping the engine to `openai` in the admin UI does not fix it either. `api_base_url` points at `https://api.openai.com/v1` with an empty key, and the model is `tts-1`, neither of which exists on our gateway. The default voice is `alloy`, which our own TTS route rejects with a 500 (#996).

So the fix is three settings, not one: engine, base URL plus key plus model, and a voice the provider accepts.

## Evidence the underlying route is fine

Live on the demo box, through a minted API key on a funded tenant:

```
POST /v1/audio/speech  {"model":"hive-tts","voice":"autumn",...}  -> 200, 169030 bytes, 1.4s
POST /v1/audio/transcriptions (that audio, model hive-stt)        -> 200, 1.2s
  text = "The quick brown fox jumps over the lazy dog."   (verbatim round trip)
```

`ENABLE_VOICE` is true for the demo tenant, so the gate is not the blocker either.

## Suggested fix

Set the TTS side the way the STT side is already set, from compose environment rather than by hand in the admin UI, so it survives a volume reset:

```
AUDIO_TTS_ENGINE=openai
AUDIO_TTS_OPENAI_API_BASE_URL=http://edge-api:8080/v1
AUDIO_TTS_OPENAI_API_KEY=<the same shim key the STT side uses>
AUDIO_TTS_MODEL=hive-tts
AUDIO_TTS_VOICE=<one of autumn diana hannah austin daniel troy>
```

Then confirm in the browser that the read aloud button produces a request to our edge and that the call lands in the ledger, rather than only that audio plays, since browser fallback audio also plays.

Related: #996 (invalid voice answers 500), #691 (Groq STT and TTS model ids are config file literals), #551 (voice input failure path).

sakibsadmanshajib:
Fixes #997. Partially addresses #996.

## What was broken

Open WebUI's text-to-speech had no working engine configured. Every `audio.tts.*` key is persistent config, so a first boot seeded engine="" (Open WebUI's bundled browser speech synthesis), base URL https://api.openai.com/v1, an empty API key, model tts-1 and voice alloy, and no compose change could ever reach an already-booted volume afterwards. The STT side of this exact reconcile existed since the dictation fix; the TTS side of the same module had never been written.

The UI also kept offering alloy-style voices because Open WebUI's `get_available_voices` fetches GET {base_url}/audio/voices on a non-OpenAI base URL and falls back to its hardcoded OpenAI fallback list when that fetch fails. The gateway served no such endpoint.

## What changed

- **deploy/docker/owui-patches/hive_rag_env_config.py**: extends the existing env reconcile with a TTS block mirroring STT: audio.tts.engine, model, voice, base URL and API key, including the paired base-URL-without-credential refusal and secret-free startup logging.
- **deploy/docker/docker-compose.yml**: sets AUDIO_TTS_ENGINE, AUDIO_TTS_OPENAI_API_BASE_URL, AUDIO_TTS_OPENAI_API_KEY (OWUI_SHIM_KEY), AUDIO_TTS_MODEL (${OWUI_TTS_ALIAS:-hive-tts}) and AUDIO_TTS_VOICE (${OWUI_TTS_VOICE:-autumn}), following the STT vars' style.
- **edge-api** serves GET /v1/audio/voices returning the provider's real roster (autumn diana hannah austin daniel troy), registered without credentials or the tenant voice gate by design: Open WebUI's own fetch sends no Authorization header, so gating would silently reinstate the alloy fallback. Non-GET answers 405.
- **deploy/litellm/config.yaml verified, no change**: route-groq-tts exists with api_key os.environ/GROQ_API_KEY and a literal groq/orpheus model id, exactly matching how route-groq-stt is handled.

## Tests

- scripts/test_owui_rag_env_config.py (CI via make test-scripts): new tests pin the five compose lines, the TTS paired refusal, unset-leaves-persisted semantics, the /v1/audio/voices registration, and secret-free log_summary coverage.
- Go unit tests cover the voices handler roster and 405.
- go vet clean on ./apps/edge-api/...

## Functional proof

docs/proof/tts-readaloud/capture.log:
- TestLiveVoiceRoundTripThroughLiteLLM PASS against a LiteLLM proxy started from this branch's deploy/litellm/config.yaml: Handler -> route-groq-tts -> Groq Orpheus returned non-empty non-silent WAV, STT leg transcribed the sentence back verbatim.
- Direct request with supported voice autumn: 200, 149830 bytes.
- Unsupported voice alloy still returns 500 through LiteLLM; that 400-to-500 mapping is #996's own contract fix and stays tracked there.

## Visual proof

Pending-live-capture, honestly: the read-aloud button producing audible playback needs the deployed stack running this branch's images, which does not exist before merge. Post-merge steps: deploy-demo-box.yml auto-deploys main; sign in at chat-hive.scubed.co, click the speaker button on any assistant message, confirm playback and that Settings > Audio TTS Voice lists autumn, diana, hannah, austin, daniel, troy; capture screenshot into docs/proof/tts-readaloud/. A follow-up commit will add it.

## Buglog entry

{"error_message":"POST /v1/audio/speech unreachable from chat: OWUI audio.tts.* persisted defaults pointed at api.openai.com with empty key and voice alloy","root_cause":"TTS half of the persistent-config env reconcile never written; all five audio.tts keys are first-boot-wins persistent config, and edge-api served no /v1/audio/voices so the UI offered alloy-style fallback voices","fix":"reconcile audio.tts.engine/model/voice/base_url/api_key from env mirroring STT; compose sets AUDIO_TTS_* vars; edge-api serves GET /v1/audio/voices with the provider roster","tags":"tts,owui,voice,reconcile,compose"}

### DIFF
--- a/apps/edge-api/cmd/server/main.go
+++ b/apps/edge-api/cmd/server/main.go
@@ -270,6 +270,14 @@ func main() {
 	voiceMW := voiceGateForAPIKeys(featureGate.Require(featuregate.FeatureVoice))
 	registerMediaFileBatchRoutes(mux, imagesHandler, audioHandler, filesHandler, batchesHandler, voiceMW)
 
+	// Issue #997: Open WebUI's voice dropdowns fetch GET /v1/audio/voices with
+	// no Authorization header at all, so this route is registered without the
+	// hk_-key authorizer or the tenant voice gate. Serving the provider's real
+	// roster here is what keeps Open WebUI's get_available_voices from falling
+	// back to its hardcoded alloy-style list (#996); gating it would silently
+	// reinstate that fallback. See audio.VoicesHandler for the full rationale.
+	mux.Handle("/v1/audio/voices", audio.VoicesHandler())
+
 	log.Printf("S3 storage enabled: images=%s, files=%s", storageCfg.ImagesBucket, storageCfg.FilesBucket)
 
 	// RAG routes (#232): always registered behind FeatureRAG so the gate returns
--- a/apps/edge-api/internal/audio/handler_voices.go
+++ b/apps/edge-api/internal/audio/handler_voices.go
@@ -1,0 +1,50 @@
+package audio
+
+import (
+	"encoding/json"
+	"net/http"
+
+	apierrors "github.com/sakibsadmanshajib/hive/apps/edge-api/internal/errors"
+)
+
+// Voice catalog for the hive-tts alias (groq/orpheus-v1-english). These are
+// the only voice ids the upstream model accepts; OpenAI's stock names (alloy,
+// echo, ...) are rejected with an upstream 400 that today surfaces as a
+// sanitized 500 (#996), so every client we control must offer exactly these.
+//
+// ponytail: static slice rather than a config knob or DB table. The list
+// changes only when the provider swaps its voice roster, at which point this
+// file changes with it; nothing needs to configure six fixed strings at boot.
+var orpheusVoices = []voiceEntry{
+	{"autumn", "Autumn"},
+	{"diana", "Diana"},
+	{"hannah", "Hannah"},
+	{"austin", "Austin"},
+	{"daniel", "Daniel"},
+	{"troy", "Troy"},
+}
+
+type voiceEntry struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+// VoicesHandler serves GET /v1/audio/voices, the endpoint Open WebUI's
+// get_available_voices fetches when audio.tts.engine is openai and the base
+// URL is not api.openai.com (routers/audio.py). It sends no Authorization
+// header of any kind, so this route is deliberately registered without the
+// hk_-key authorizer or the tenant voice gate: there is no credential to
+// validate and nothing here is per-account data, only the six public voice
+// names the provider publishes. Gating it would silently break the Settings >
+// Audio voice dropdowns back to Open WebUI's hardcoded alloy-style fallback,
+// which is the exact defect (#996) this exists to prevent.
+func VoicesHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			apierrors.WriteError(w, http.StatusMethodNotAllowed, "invalid_request_error", "Method not allowed", nil)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(map[string]any{"voices": orpheusVoices})
+	}
+}
--- a/apps/edge-api/internal/audio/handler_voices_test.go
+++ b/apps/edge-api/internal/audio/handler_voices_test.go
@@ -1,0 +1,52 @@
+package audio
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestVoicesHandlerGetReturnsProviderRoster(t *testing.T) {
+	t.Parallel()
+
+	rec := httptest.NewRecorder()
+	VoicesHandler()(rec, httptest.NewRequest(http.MethodGet, "/v1/audio/voices", nil))
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("GET /v1/audio/voices: got %d want 200, body %q", rec.Code, rec.Body.String())
+	}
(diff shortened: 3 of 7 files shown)


sakibsadmanshajib (commented):
Adversarial review pass (the stage that had not run yet on this PR). Verdict: **READY**.

## Findings

No blocking findings. Three non-blocking observations, none requiring changes:

1. **Low, intentional**: `GET /v1/audio/voices` is registered without the hk_ authorizer or the tenant voice gate. Verified this is correct and low risk: Open WebUI's own voices fetch sends no Authorization header at all (`vendor/open-webui/backend/open_webui/routers/audio.py`, `get_available_voices`), so gating would silently reinstate its hardcoded alloy fallback (#996), and the handler serves a static six name roster with no database, provider, or billing contact and no request derived data. Exposure is limited to six public voice names.

2. **Info**: the failure mode changes by design. With `audio.tts.engine` forced to openai, a gateway TTS outage now surfaces an error instead of degrading to browser speech synthesis. That silent degradation was itself the defect (#997: unmetered OS voice playback), and honest failure matches the repo's loud failure posture. No silent fallback remains.

3. **Info**: environment wins for these five keys on every open-webui restart, so admin UI edits to them are reverted at next restart (compose literals are always present). This is the accepted #722 mechanism, identical to STT. One doc nit: the comment "clears AUDIO_TTS_ENGINE to keep whatever its administrator configured" cannot be done through .env alone because the compose value is a literal, it needs an override file. The same wording already ships on the preexisting STT block, so consistency rather than a defect.

## Checked clean

- Contract match against the fork: `get_available_voices` fetches GET `{base_url}/audio/voices` with no auth header and reads `data["voices"]` entries keyed by id/name; `VoicesHandler` returns exactly that shape and 405s non-GET. The speech path sends `Bearer audio.tts.openai.api_key` to `{base_url}/audio/speech`, which stays behind the existing hk_ authorizer plus tenant voice gate (untouched).
- Alias chain: hive-tts is seeded pinned to route-groq-tts (migration 20260717_02), voice default autumn is a real Orpheus voice, LiteLLM config untouched as described.
- No ServeMux pattern conflicts: exact pattern among exact pattern siblings (`/v1/audio/speech`, `/transcriptions`, `/translations`). Nothing previously served `/v1/audio/voices` (it 404ed into the alloy fallback), so the new route only removes a failure path.
- Secret handling: `audio.tts.openai.api_key` added to SECRET_KEYS so startup logs never carry it; the paired credential refusal is tested for unset, empty and whitespace key values, persisted values untouched on refusal.
- Tests run locally on this branch: `go test ./apps/edge-api/... -count=1 -short` green in the toolchain container (25 packages, audio included), and `python3 scripts/test_owui_rag_env_config.py` green including the new TTS pins. CI rollup green, zero unresolved threads.
- Buglog entry carried in the PR body per protocol.

## Note

Visual proof (audible read aloud playback captured on the deployed chat surface) is disclosed in the PR body as pending live capture, which can only exist once the merged deployment does. Keep that follow-up capture attached to this task after deploy per the standing visual proof directive; it is a post merge step, not a code defect.

PR #1083, opened 2026-08-23
feat: silent single sign-on consent completion

sakibsadmanshajib:
## Summary

Wave 1 of the approved SSO spec (vault: spec-2026-08-23-single-sign-on.md, option b): the console OAuth consent landing becomes a Server Component that completes already-consented authorizations silently.

- `apps/web-console/app/oauth/consent/page.tsx` reads the `@supabase/ssr` session from cookies and calls `GET /auth/v1/oauth/authorizations/{id}` with the access token as bearer. When GoTrue's auto-approve applies (an active consent row covers the requested scopes), the page answers with one bare redirect: zero paint, zero client JS.
- The interactive Approve/Deny panel (`components/oauth/consent-panel.tsx`) renders exactly as before when there is no session yet, when consent is genuinely required, or as a painted error state. The sign-in round-trip URL builder moved to `lib/auth/silent-consent.ts` so server and client share one format.
- New `lib/auth/silent-consent.ts` holds a pure decision function plus the GoTrue lookup wrapper. 20 unit tests cover: session valid + consented -> silent redirect; session valid + needs-consent -> panel; no session -> panel; GoTrue 401/403 -> one sign-in redirect with a reason marker; GoTrue failure -> panel fallback; loop bounds.

## Redirect-loop bound (proved, not promised)

The decision function refuses any auto-approve target whose path is the consent landing itself (`/oauth/consent`, regardless of origin) or whose protocol is not http(s); such targets fall to a painted error state instead of redirecting. At most one redirect leaves the page per request and it can never point back at itself. The sign-in fallback is a dead end: the sign-in form never auto-bounces back to consent; only a successful credential submit returns. Unit tests assert the refusal cases directly; the e2e journey asserts at most one redirect beyond the authorize round trip on a live stack.

## Why there is no new CSRF surface

This endpoint performs a GET authorization lookup and then redirects. It grants nothing: `redirect_url` is returned exclusively by GoTrue when an active consent row for the requesting user covers the requested scopes, and the redirect target is the client's registered `redirect_uri` for that authorization. An attacker-crafted link can at worst re-complete an authorization the victim had already granted on the victim's own account, which is identical exposure to the client-side auto-navigate the page already had. Interactive approval stays a POST to GoTrue's consent endpoint. No cookies are set or cleared by this change, and the bearer used is the victim's own session token read server-side from the existing ssr envelope.

## Tests

- Unit: `lib/auth/silent-consent.test.ts`, 20 cases (decision matrix, parser, target guard, sign-in URL, lookup wrapper with injected fetch).
- E2E (opt-in against a live stack): `tests/e2e/sso-silent-consent.spec.ts`. First-ever consent paints the panel once and Approve lands signed in; repeat entry with console session only shows zero painted consent documents, zero Approve/Deny buttons, zero password inputs, and at most one redirect beyond the authorize round trip. Passed against a full local stack (self-hosted GoTrue v2.189.0, web-console built from this branch, Open WebUI).
- Full unit suite green (562 tests); `tsc --noEmit` clean; `next build` clean.

## Visual proof

Captures against the running stack are in `docs/proof/sso-wave1/` (text log; frames posted below as release assets). Repeat entry: `authorize -> 302 landing -> 302 callback -> chat home`, consent never painted.

## Review notes

AUTH PATH: adversarial review is mandatory for this PR; the security stream must verify the bearer flow (session token read from the ssr envelope server-side, sent to GoTrue only) and the redirect-loop bounds. No GoTrue, schema, or client registration changes.


sakibsadmanshajib:
## Visual proof

Post-rebase capture (branch rebased onto main at 5e08d641), against a throwaway stack: self-hosted GoTrue v2.189.0 with the OAuth 2.1 server enabled, the web console built from this branch with Dockerfile.web-console.prod, Open WebUI as the relying party over real authorization-code round trips with PKCE S256. Frame 1: first ever consent for a fresh user still paints the interactive panel exactly once. Frame 2: repeat entry with only the console ssr session on the device lands straight in chat, the consent landing answering 307 with zero painted documents, zero Approve or Deny buttons, zero password inputs and exactly one redirect beyond the authorize round trip. URLs in the banners have their query stripped because this flow carries authorization codes there. Full chain in docs/proof/sso-wave1/capture.log.

apps/web-console/lib/auth/silent-consent.ts
```diff
@@ -0,0 +1,340 @@
...
+  // Backslashes are already forward slashes by this point, because the WHATWG
+  // URL parser rewrites them for http(s) URLs. Dot segments are gone too:
+  // `/x/../oauth/consent` parses to `/oauth/consent` before the guard reads it.
+  let decodedPath = parsed.pathname;
+  try {
+    decodedPath = decodeURIComponent(parsed.pathname);
+  } catch {
+    decodedPath = parsed.pathname;
+  }
+  const normalizedPath =
```
sakibsadmanshajib: **High, from the security stream and Antigravity independently. Fixed in this line.**

The previous guard was `parsed.pathname.replace(/\/+$/, "")`, which strips trailing slashes and nothing else. A target of `//oauth/consent` or `/oauth//consent` matched neither `CONSENT_LANDING_PATH` nor its trailing-stripped form, so the guard allowed it, Next then 308ed the repeated slashes away and landed back on this page, which asked GoTrue again and got the same `redirect_url`. That is the loop the description called bounded, and as written it was not.

Measured against this repository's own production console build rather than argued:

```
/oauth/consent/    -> 308 /oauth/consent
//oauth/consent    -> 308 /oauth/consent
/oauth//consent    -> 308 /oauth/consent
/oauth/%63onsent   -> 404
/OAuth/Consent     -> 404
```

So repeated slashes now collapse. Two claims from the Antigravity finding were **refuted** by that same measurement and are not the reason for the fix: the percent-encoded variant does not route, and neither does a case variant, because App Router matching is neither decoded nor case-insensitive. Percent decoding is in anyway, as defence in depth, so the guard does not rest on one proxy's normalisation rules, and it sits in a `try` so a malformed escape compares the raw path: refusing to decode must never mean accepting the target.

A third suggestion, `.replace(/\\/g, "/")`, was **declined as dead code**: `new URL("https://a.example/oauth\\consent").pathname` is already `/oauth/consent`, because the WHATWG parser rewrites backslashes for http(s) URLs. Dot segments are likewise already resolved before the guard reads `pathname`.

Negative control run rather than assumed: with the old one-line guard restored, `rejects every spelling Next.js normalizes back to the landing` fails and the other 22 cases stay green.

apps/web-console/lib/auth/silent-consent.ts
```diff
@@ -0,0 +1,340 @@
...
+          message:
+            "The authorization server returned an unsafe redirect target.",
+        };
+      }
+      return { action: "silent-redirect", url: authz.redirectUrl };
+    }
+    return { action: "render-panel" };
+  }
+
+  if (lookup.status === "forbidden") {
```
sakibsadmanshajib: **Medium, from the security stream. Fixed here and at the `unauthorized` branch below.**

The previous code collapsed 401 and 403 into one `rejected` status and sent both to the password form as "your session expired". GoTrue also answers those for an authorization id that belongs to another user or has expired, and `middleware.ts` has already validated the same cookie with `getUser()` before this page runs, so a rejection here is more likely about the authorization than the session. The user could sign in successfully, come back, be refused for the same non-session reason, and be sent to the password form again, with nothing bounding the loop. Conditioning repeated password entry on a phishable surface is the part that makes this more than a usability bug.

403 now paints and never spends a credential prompt, because signing in again cannot change that answer.

apps/web-console/lib/auth/silent-consent.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  if (lookup.status === "unauthorized") {
+    // GoTrue refused the bearer token (expired or revoked session). Route to
+    // sign-in once, with a reason marker for the network trace. The return
+    // target carries a marker, so a second rejection lands here with
+    // signInAlreadyAttempted set and paints instead of asking again: the hop
+    // count is bounded at one even when the rejection is not really about the
+    // session at all.
+    if (signInAlreadyAttempted) {
```
sakibsadmanshajib: This is what actually bounds the hop at one. `buildSignInRedirect` puts a `retried=1` marker in the `next` target, so the return trip arrives here with `signInAlreadyAttempted` set and a second rejection paints instead of asking for a password again. The marker authenticates nothing and is not trusted for anything else; it only counts hops.

`lib/auth/next-target.test.ts` also asserts that the value the builder produces is one `resolveNextTarget` actually accepts, so the builder and the allow-list cannot drift into a silent dump on `/console`.

apps/web-console/lib/auth/silent-consent.ts
```diff
@@ -0,0 +1,340 @@
...
+ * panel. A future GoTrue that added redirect_url to a still-pending body would
+ * otherwise silently skip the panel forever with no test to catch it.
+ */
+export function parseGoTrueAuthorizationBody(
+  body: unknown,
+): GoTrueAuthorization | null {
+  if (typeof body !== "object" || body === null) {
+    return null;
+  }
+  const client = "client" in body ? body.client : undefined;
```
sakibsadmanshajib: **Low, from the security stream. Taken.**

The parser used to test `"redirect_url" in body` before the pending shape. The two shapes are disjoint in the GoTrue this repository pins, so the reviewer confirmed it is not exploitable today and neither order changes any current behaviour. They are not disjoint by contract though, and this is a consent gate: a future GoTrue that added `redirect_url` to a still-pending body would silently skip the Approve and Deny panel forever, with no test to catch it. Testing the pending shape first makes the failure direction the safe one, and costs nothing.

Antigravity found the same code from the other side: a `redirect_url: null` alongside a real pending body returned `null` and so `failed`. Its stated impact was wrong (`failed` maps to `render-panel`, not to the error state, so the user got the panel), but the parse was still wrong and now falls through.

apps/web-console/lib/auth/silent-consent.ts
```diff
@@ -0,0 +1,340 @@
...
+ * prefix the box's log scrape can grep. The authorization id is a request
+ * handle, not a credential: GoTrue will not answer for it without the owner's
+ * bearer. The bearer, the anon key and the response body never appear here.
+ */
+function logLookupOutcome(
+  outcome: string,
+  authorizationId: string,
+  status: number | null,
+): void {
+  console.error("oauth consent lookup failed", {
```
sakibsadmanshajib: **Low, from the security stream (OWASP A09). Taken.**

Every non-ok outcome in the lookup collapsed into `{ status: "failed" }` with no record: network throw, non-ok status, and JSON parse throw alike. Since `decideConsentLanding` maps that to `render-panel`, a GoTrue outage on this path degraded silently in both directions at once, the user quietly redoing the work in the browser and the operator seeing nothing at all.

One line per failure now, with a fixed prefix the box's log scrape can grep. The authorization id is a request handle rather than a credential, since GoTrue will not answer for it without the owner's bearer. The bearer, the anon key and the response body are never logged.

apps/web-console/app/oauth/consent/page.tsx
```diff
@@ -1,11 +1,93 @@
...
+} from "@/lib/auth/silent-consent";
+import { createClient } from "@/lib/supabase/server";
 
 interface ConsentPageProps {
-  searchParams: Promise<{ authorization_id?: string }>;
+  // Next hands a repeated query param through as an array, so the honest type
+  // is the union. `?authorization_id=a&authorization_id=b` must not be quietly
+  // collapsed into a fabricated single id: readSingle refuses the array and
+  // the request falls to the panel, which paints its missing-id error.
+  searchParams: Promise<Record<string, string | string[] | undefined>>;
```
sakibsadmanshajib: **Low, from the security stream. Taken.**

Next hands a repeated query param through as an array, so typing this `string` was a claim the framework does not honour. `?authorization_id=a&authorization_id=b` would have been read as a fabricated single id. No injection, since the value is `encodeURIComponent`d into the GoTrue path, but the type was dishonest. `readSingle` refuses the array and the request falls to the panel, which paints its own missing-id error.

sakibsadmanshajib (commented):
## Adversarial review, pipeline mode, four streams

This is the mandatory auth-path review the description asks for. Findings are posted inline below; this comment records which streams ran, which did not, and what each concluded. Every fix named here is already pushed on this branch.

### Streams

**CodeRabbit CLI: SKIPPED.** `coderabbit review --agent --committed --base origin/main` returned `rate_limit: Rate limit exceeded`, with `"You've used all 3 included reviews currently available"` and a 21 minute wait, org-attributed. Out of quota repository wide, not a clean pass. Recorded as skipped rather than counted.

**Antigravity (gemini-3.1-pro-high, effort high): RAN.** Five findings. One high, on the redirect guard, is real and is fixed. One medium, on the response parser, was real in shape but overstated in impact: a `redirect_url: null` body returned `failed`, which `decideConsentLanding` maps to `render-panel`, so the user got the client panel rather than the generic error the finding claimed. Fixed anyway, and the ordering hardened further. Two low findings are answered inline. One of its claims was checked and refuted: percent-encoded paths do not route.

**Security review (opus, adversarial, auth path): RAN.** Both claims in the description were traced rather than accepted. Claim 1, that the bearer is read server side from the ssr envelope and reaches GoTrue only, came back VERIFIED with the whole chain walked. Claim 2, the redirect loop bound, came back REFUTED on its second half, which is the high finding above. Also found the 401 versus 403 conflation and the silent failure path, both fixed.

**TypeScript and React review: RAN.** Approve, no critical or high findings. `tsc --noEmit` clean including the e2e tree, and it specifically looked for tests that cannot go red and found none. Three nits, all answered inline.

### What changed as a result

| Finding | Severity | Outcome |
|---|---|---|
| Redirect guard allows `//oauth/consent` and `/oauth//consent`, which 308 back to the landing | High | Fixed, with a test that goes red against the old guard |
| 401 and 403 both spend a credential prompt, unbounded | Medium | Fixed, 403 paints, and the one sign-in hop now carries a marker that bounds it |
| Parser tests the auto-approve shape before the pending shape | Low | Fixed, order reversed so a consent gate fails toward showing the panel |
| Every lookup failure is silent to the operator | Low | Fixed, one server log line per failure with outcome, id and status |
| `searchParams` typed as `string` when Next yields `string \| string[]` | Low | Fixed |
| Client panel imports from the module holding the server lookup | Low | Fixed, the builder moved beside the allow-list that validates it |
| Percent-encoded and case variants of the landing path | Claimed | Refuted by measurement, then hardened anyway, see the inline note |
| Tautological assertion in the guard test | Nit | Removed |

### Verification after the fixes, on the substrate that matters

`lib/auth`, `components/oauth` unit tests 117 passed in the repository's Docker image rebuilt from this branch. `tsc --noEmit` exit 0. `next build` exit 0. The committed opt-in e2e `1 passed (11.2s)` against a rebuilt console image on the same live GoTrue v2.189.0 stack. Both proof frames come out byte identical to the pre-fix run (md5 `7b3e1d73fa0ba2011304bb4a223cf4c6` and `20c588349d0734b188438f029e08b89b`), which is the expected result: these fixes harden refusal paths that neither journey walks.

apps/web-console/app/oauth/consent/page.tsx
```diff
@@ -1,11 +1,93 @@
...
+  const authorizationId = readSingle(params.authorization_id);
+  const signInAlreadyAttempted =
+    readSingle(params[CONSENT_RETRIED_PARAM]) !== null;
+
+  const cookieStore = await cookies();
+  const supabase = createClient(cookieStore);
+  const { data } = await supabase.auth.getSession();
+  const accessToken = data.session?.access_token ?? null;
+
+  // A config gap (either Supabase var unset) is not an auth failure: skip the
```
sakibsadmanshajib: **Nit, from the TypeScript stream. Taken as a comment reword.**

The old comment read wider than the code: the session read above it is unconditional, and `lib/supabase/server.ts` uses non-null assertions on the same env vars, so "a config gap is not an auth failure" was not true of the whole block. It is true of the lookup, which is what the guard actually covers. Reworded to say so rather than changing the call, since twenty-plus other console pages make the same unconditional `createClient` call and diverging here would be the odd one out.

apps/web-console/components/oauth/consent-panel.tsx
```diff
@@ -4,6 +4,11 @@ import { useEffect, useState } from "react";
 
 import { createClient } from "@/lib/supabase/browser";
 import { toUserFacingAuthMessage } from "@/lib/auth/auth-error";
+// The sign-in round-trip URL builder lives beside the allow-list that
+// validates its output, so the server-completed landing page (SSO wave 1) and
+// this panel share one format, and importing it here pulls no server-side
+// lookup code into the browser bundle.
+import { buildSignInRedirect } from "@/lib/auth/next-target";
```
sakibsadmanshajib: **Low, raised by both the security stream and Antigravity. Taken.**

This client component used to import `buildSignInRedirect` from `@/lib/auth/silent-consent`, which also holds `lookupGoTrueAuthorization` and the decision function. Nothing leaked: that module imports neither `next/headers` nor `next/navigation`, holds no secret, and the anon key it reads is public by definition. So this was a shape problem rather than a live one, and the honest severity is low.

It is still worth ten minutes, because the next server-only helper added to that module would leak without anyone noticing. `buildSignInRedirect` now lives in `lib/auth/next-target.ts`, beside the allow-list that validates the value it produces, which fixes the import direction and puts the builder and its validator in one file where they cannot drift.

apps/web-console/lib/auth/silent-consent.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    expect(
+      parseGoTrueAuthorizationBody({ client: { name: "" }, scope: "openid" }),
+    ).toBeNull();
+    expect(
+      parseGoTrueAuthorizationBody({ client: "not-an-object", scope: "openid" }),
+    ).toBeNull();
+  });
+});
+
+describe("isSafeRedirectTarget", () => {
```
sakibsadmanshajib: **Nit, from Antigravity. Taken.**

The guard suite carried `expect(CONSENT_LANDING_PATH).toBe("/oauth/consent")`, which asserts a constant against its own literal and cannot go red if `isSafeRedirectTarget` regresses. Removed. The behaviour it was gesturing at is covered by feeding the constant through the function instead.

The TypeScript stream swept the rest of this file for the same shape and found nothing else that cannot fail. The new loop-bound case was checked the hard way: with the old guard restored it fails, with the fix it passes.

sakibsadmanshajib:
Housekeeping note from another agent in this session, not a review of this pull request.

At 2026-08-29 09:40 UTC I resolved six review threads on this pull request by mistake and then immediately unresolved all six, so the state you see should be exactly the state before it happened. None of them were read, addressed or judged by me.

Cause, recorded so it does not repeat: this session's agents share one scratchpad directory, and a helper script I wrote to a plain filename there (`resolve-threads.sh`) was overwritten with a script of the same name targeting this pull request number. I ran what I thought was mine. My own scripts are now prefixed with my worktree id.

If any of those six were in fact ready to resolve, they need resolving again; the unresolve was a blind restore of the state I found, not a judgement that the feedback still stands.

Issue #1055, opened 2026-08-23
owui-nightly still points at the deleted Supabase project, same cause as #1050


sakibsadmanshajib:
`owui-nightly` fails the same way `agent visual proof` did in #1050, from the same cause, and is still failing.

Both workflows reconstruct their Supabase URL and DSN from the `SUPABASE_*` repository secrets. Those secrets name the hosted Supabase project that was deleted during the move to the self-hosted data plane. Every one of them still carries an April 2026 timestamp; none was repointed after the cutover.

Run 32623212748 (2026-08-23) shows both halves of the cause in clear text:

```
socket.gaierror: [Errno -2] Name or service not known
error: (ENOTFOUND) tenant/user *** not found
```

The first is the project host failing to resolve. The second is Supavisor reporting that the tenant does not exist, which is a deleted project rather than contention or a bad credential.

PR #1053 fixes `agent visual proof` by standing up a throwaway Postgres, GoTrue and PostgREST per run, the way PR #983 already moved ci.yml. It deliberately does **not** touch `owui-nightly`, to keep that change's blast radius to one job.

What needs deciding here is which way `owui-nightly` goes, and they are not equivalent:

1. Move it onto the same throwaway stack (`scripts/ci-supabase-stack.sh`). Self-contained, no shared mutable state, consistent with #983. But a nightly whose whole point is exercising the chat surface against something realistic may lose value if it never touches a real deployment.
2. Repoint it at the self-hosted Supabase on the demo box. Closer to reality, but it puts a nightly job's writes into the live demo database, which is exactly the coupling #983 removed.

Worth noting either way: `owui-nightly` also calls `scripts/seed-owui-e2e-user.py`, which is on the forbidden list in `docs/live-test-auth.md` because it rotates a shared account's password. That is a second reason not to point it at anything shared.

Related: the `SUPABASE_DB_HOST`, `SUPABASE_DB_PORT`, `SUPABASE_DB_USER`, `SUPABASE_DB_NAME`, `SUPABASE_DB_PASSWORD`, `SUPABASE_URL`, `SUPABASE_ANON_KEY`, `SUPABASE_SERVICE_ROLE_KEY` and `NEXT_PUBLIC_SUPABASE_*` secrets are now dead values pointing at a deleted project. Once no workflow reads them, they should be deleted rather than left to be picked up by the next workflow that assumes they are live.

sakibsadmanshajib:
[...]

PR #1086, opened 2026-08-23
fix: stop six workflows configuring themselves from the deleted Supabase project

sakibsadmanshajib:
Closes #1059.

## The problem, restated from ground truth

The hosted Supabase project was deleted in the move to the self-hosted data plane. Ten repository secrets name it, every one of them still timestamped April or July 2026, and six workflows read them. PR #983 moved `ci.yml`'s database off that project and PR #1053 moved the Cowork proof job. The rest were left behind, in four different shapes:

| workflow | shape | visible? |
| --- | --- | --- |
| `deploy-web-console-workers` | **GREEN**, baking the dead origin into the shipped Workers bundle | no |
| `owui-nightly` | red for days | yes, and ignored |
| `chat-coverage` live sweep | label-gated, misconfigured | no |
| `demo-chat-settings-check` | never run, misconfigured | no |
| `deploy-demo-box` agent-workspace-coverage | dispatch-only, misconfigured | no |
| `ci.yml` live-integration | reads them, never calls them | no |

Only one of those six is the dangerous one, and it is the green one.

## `deploy-web-console-workers` is retired, not repointed

Answering #1059's question 1 first, because the fix follows from it: **the deployed console does not come from that Worker.**

- `apps/web-console/wrangler.jsonc` declares no `routes` and no custom domain.
- `console-hive.scubed.co` answers `HTTP/2 307` with `via: 1.1 Caddy`. `chat-hive.scubed.co` answers `200`, also `via: 1.1 Caddy`. `api-hive`, `control-hive` and `artifacts-hive` answer a Cloudflare 404 with no `via` at all, and `agent-hive` and `cp-hive` do not resolve. No hostname this product uses is served by a Worker.
- `deploy/docker/Caddyfile.console` proxies `console-hive.scubed.co` to `web-console-prod:3000`, and that container's `NEXT_PUBLIC_*` build args come from the box's own `.env`, which PR #992 already moved to `https://console-hive.scubed.co`. Probing the live bundle confirms it: the only Supabase-shaped origin in any chunk on `/auth/sign-in` is `https://console-hive.scubed.co`, and no `*.supabase.co` string appears in any of them.

So the Worker deploy has had no consumer since the console moved onto the box in PR #456 and PR #605 (recorded as D-043). The workflow's own header comment said exactly this would happen: "This job is expected to retire too, once apps/web-console is migrated off Cloudflare Workers onto the same physical demo box." That migration shipped three weeks ago.

Deleting the workflow does not touch the deployed Worker, so nothing serving traffic today changes. Only the redeploy of a stale bundle stops, and with it the green check that was reporting on it. **Follow-up an operator has to do, and this repository cannot:** delete the `hive-console` Worker itself from Cloudflare. It needs the dashboard or an account-scoped token, and until then it sits there serving nothing on a `workers.dev` name with a dead Supabase origin baked in.

`apps/web-console`'s `wrangler.jsonc`, its `build:cf` / `preview:cf` / `deploy:cf` scripts, and the `@opennextjs/cloudflare` and `wrangler` dependencies are **deliberately kept**, not left behind by accident. While that Worker still exists, `npm run deploy:cf` is the only way to push a corrected bundle to it or to inspect what it is running, and removing the toolchain would take that away in the same change that stops the automation. They become dead weight the moment the Worker is deleted, and that is the right time to remove them, together, in a diff that also touches `package-lock.json` and can be checked against the web-console build job on its own.

## The remaining four, split by what they actually need

**`ci.yml` live-integration needs no Supabase at all.** It seeds its own API key on its own throwaway Postgres, and the SDK suites authenticate with that `hk_` key, never with a user session, so GoTrue is never reached. It now gets the same literal placeholders (`https://ci-placeholder.supabase.co`) the web unit and build jobs in that same file already use, so the pattern is not new here. The point is not that a placeholder works: it is that an unresolvable placeholder fails immediately and names itself, where a deleted hosted project resolves in DNS (the wildcard is Supabase's, not the project's), answers HTTP, and fails three steps later. If a suite here ever does need a real Supabase, the route is `scripts/ci-supabase-stack.sh`, as the web-e2e job in this file already does.

**The three that run against the live deployment read the deployment's own configuration.** The public auth origin is the console origin, because Caddy fronts GoTrue at `/auth/v1` there (PR #992). All three now take it from the same `${{ vars.CONSOLE_URL || 'https://console-hive.scubed.co' }}` expression that `oauth-scope-gate.yml` and the deploy workflow's own OIDC discovery check already use, rather than a fourth copy of the hostname.

The two keys are **renamed rather than repointed**, to `DEMO_SUPABASE_ANON_KEY` and `DEMO_SUPABASE_SERVICE_ROLE_KEY`. The old names still hold the deleted project's values, and this repository cannot read the box's (no SSH from here, and printing them from the self-hosted runner would put them in a public log). Renamed, they are simply absent, and each of those three jobs fails by name on an empty value. That is the intended end state: these jobs would rather not run than run against a project that is not there. All three are dispatch-only or label-gated, so no scheduled or push-triggered run changes colour as a result.

Two of the three already had that "check the inputs this job cannot run without" step. `deploy-demo-box`'s `agent-workspace-coverage` did not, which the review caught: renaming its secrets would have moved its failure from "authenticates against a deleted project" to "installs Chromium, then dies inside `live-auth.mjs`". It has one now, placed before `setup-node` and `npm ci` so a configuration fault costs neither.

## The guard

`tools/lint-no-dead-supabase-secrets.mjs`, wired into the existing required "Repo policy lints" job as `npm run lint:dead-supabase-secrets`.

It fails any workflow whose `${{ }}` expressions reference one of the ten dead secret names. Design choices worth naming:

- **Inside expressions only.** A comment explaining why a workflow does *not* read `secrets.SUPABASE_URL` has to be able to say so, and a guard that forbids its own rationale gets deleted rather than obeyed. Expressions are matched across newlines, so one split over a block scalar cannot slip through.
- **Names enumerated, not prefix-matched.** A correctly provisioned, differently named secret for the self-hosted deployment passes with no exemption. This bans the ten dead values, not the word Supabase.
- **MUST_CATCH / MUST_ALLOW self-test as a preflight**, the same shape as `lint-no-pgx-dsn-for-libpq.mjs` and `lint-no-token-in-proof-captures.mjs`, so the guard cannot rot into something that only ever passes. Seven catch fixtures (dotted form, both bracket forms, whitespace inside the expression, split over lines, buried in a longer expression) and seven allow fixtures.
- **A pending list that can only shrink, pinned to its audited references.** `owui-nightly` is its one entry, pinned to #1055, which is where the decision about that job belongs (throwaway stack versus the live box is a real trade-off and it is that issue's to make, not a drive-by here). The entry records the per-secret counts it was audited at, so the exemption covers exactly those references and nothing else: one more, one fewer, or one swapped for another all fail. A file-level exemption would have let a freshly copied `secrets.SUPABASE_*` line into that workflow pass the lint, which is the copy-paste spread this guard exists to stop. The lint also fails if a pending entry stops being needed, so a stale exemption cannot sit there being the precedent for the next one.

Once nothing reads them, the ten secrets should be deleted outright rather than left for the next workflow to adopt by copy-paste, which is how there came to be six.

## On the preflight #1059 also asks for

Deliberately not built here. PR #1061 adds `scripts/preflight-supabase-config.py`, which asserts the configured project is alive and that the keys' `ref` claim matches the URL, and it cites #1059 as its reason. Duplicating it would produce two preflights and a merge conflict. This change removes the dead configuration; that one catches a future divergence. They are complementary and both belong.

## Verification

- `npm run lint:dead-supabase-secrets` on this branch: `ok: 11 workflow file(s) scanned, 1 pending, no unexempted reference to the retired project's secrets`.
- Against `main`, the same lint reports 18 failures across five files plus the one pending, so it is known to be able to go red on the real tree, not only on its fixtures.
- The pending pinning was mutation tested: one extra `secrets.SUPABASE_URL` appended to `owui-nightly.yml` reports `SUPABASE_URL=4` against an expected `SUPABASE_URL=3` and exits 1.
- `deploy-web-console-workers` is **not** in `required_status_checks.contexts` on `main` (checked, not assumed), so deleting it cannot block a merge the way a missing required check does on this repository.
- Every `.github/workflows/*.yml` re-parses (`yaml.safe_load`), and `package.json` re-parses.
- Live probes above were run for this change, not quoted from an earlier session.

## Not in scope, and why

- **#1055 (`owui-nightly`).** On the pending list. The choice between a throwaway stack and the live box is the trade-off that issue exists to settle, and it also has to decide what happens to the OAuth half of the suite, since `scripts/ci-supabase-stack.sh` does not enable GoTrue's OAuth server today and the job's OIDC login test would silently skip. Verifying it costs a 45 minute run per attempt.
- **#1062 (`post-deploy-verify`).** Not a `main` failure. Its run was on PR #1061's own branch, which is where that workflow is being introduced, and the failure is that job's own `HIVE_VERIFY_EMAIL` / `HIVE_VERIFY_PASSWORD` missing from the box's `.env`. It belongs to that PR.
- **`S3_*` secrets** also point at the deleted project's storage endpoint. Same root cause, different surface, and out of #1059's stated scope. Filed as **#1087** rather than left as a sentence here, including the observation that `ci.yml`'s required live-integration job reads all six and the stack boots anyway, because the variables are present and merely wrong.

## Buglog entry

```json
{"id":"2026-08-23-workflows-reading-deleted-supabase-project","date":"2026-08-23","title":"Six workflows configured themselves from a deleted Supabase project, one of them green while shipping the dead origin","error_message":"socket.gaierror: [Errno -2] Name or service not known / error: (ENOTFOUND) tenant/user *** not found","root_cause":"The hosted Supabase project was deleted during the move to the self-hosted data plane, but the ten SUPABASE_* repository secrets naming it were never repointed or removed. PR #983 and PR #1053 moved two workflows off them and the other six were left reading them. Nothing structural forbade the reference, so each new workflow adopted the pattern by copy-paste. deploy-web-console-workers was the dangerous instance because it stayed GREEN: it baked NEXT_PUBLIC_SUPABASE_URL from a pre-cutover secret into a Cloudflare Workers bundle nothing serves any more, so a passing deploy reported over a dead origin.","fix":"Retired deploy-web-console-workers (the console has been served by web-console-prod behind Caddyfile.console since PR #456 and PR #605, and no hostname resolves to a Worker). Gave ci.yml's live-integration job literal placeholders, since it authenticates with its own seeded hk_ key and never reaches GoTrue. Pointed the three live-deployment jobs at vars.CONSOLE_URL, the deployment's public auth origin, and renamed their two key secrets to DEMO_SUPABASE_* so an unprovisioned value fails their existing input check by name instead of authenticating against a project that no longer exists. Added tools/lint-no-dead-supabase-secrets.mjs to the required repo-policy lint job, matching inside ${{ }} expressions only, with a MUST_CATCH/MUST_ALLOW self-test and a shrink-only pending list holding owui-nightly against #1055.","tags":["ci","github-actions","supabase","self-hosted-cutover","green-over-dead-backend","secrets","cloudflare-workers"]}
```


sakibsadmanshajib (commented):
Adversarial review, stream 1 of 3: CodeRabbit CLI (`coderabbit review --committed --base main`), run locally against this branch. 8 files reviewed, 2 minor findings, both posted inline below. Recording them on the PR rather than in a private report, per the pipeline contract.

Streams 2 and 3 follow in separate reviews.

Issue #437, opened 2026-07-26
OWUI Caddy @adminMutation matcher is ineffective: path wildcards do not cross path separators


sakibsadmanshajib:
The `@adminMutation` matcher in `deploy/docker/Caddyfile.owui` is a defence in depth layer that drops mutation verbs on Open WebUI's admin endpoints. It does not actually match the endpoints it targets, so the layer is effectively dead.

Caddy's `path` matcher wildcards do not cross `/` separators. Verified in isolation against the pinned `caddy:2-alpine` image with a minimal config containing only the same two patterns:

| Pattern | Request | Result |
| --- | --- | --- |
| `/api/v*/configs*` | `PUT /api/v1/configs` | blocked |
| `/api/v*/configs*` | `PUT /api/v1/configs/models` | passes through |
| `/api/v*/users/*` | `PUT /api/v1/users/abc` | blocked |
| `/api/v*/users/*` | `PUT /api/v1/users/x/y` | passes through |

Every real Open WebUI admin mutation endpoint lives at least one path segment deeper than the patterns can reach. Confirmed live against the demo box, where `PUT /api/v1/configs/models` reaches Open WebUI and answers 405 rather than the intended 404.

Scope notes:

1. The `@blocked` matcher above it uses `path_regexp`, is unaffected, and was confirmed still returning 404 on every admin and signup variant tested. The admin panel and signup blocks are genuinely enforced. Only the method level mutation guard is dead.
2. This is pre existing and was found while regression testing the proxy during #435. It was deliberately left unfixed there because that PR was a proxy scheme fix, and because naively converting these to `path_regexp` would begin dropping mutation verbs on paths such as user settings updates that the chat UI may depend on.

The fix needs its own scoping: enumerate which Open WebUI mutation endpoints must stay reachable for a normal signed in user, then write matchers that block only the admin surface, and verify against a running chat session rather than by inspection. Discovered during #435.

sakibsadmanshajib:
Measured rather than reasoned, 2026-08-17, during an owner-requested audit of the chat Admin Settings surface. The `@adminMutation` matcher from `deploy/docker/Caddyfile.owui:139-143` was copied verbatim into a throwaway `caddy:2-alpine` container with a plain 200 responder behind it, and every address the admin panel actually writes to was sent a POST.

```
/api/v1/configs                          POST -> 404   blocked
/api/v1/configs/code_execution           POST -> 200   passes
[...]

PR #1091, opened 2026-08-23
fix: close the public chat origin's admin surface

### LINKED ISSUE #948: configs/import and configs/namespace return the OAuth client secret and gateway key on the public chat origin, going around the #769 export block
#769 blocked `GET /api/v1/configs/export` on the public chat origin because it returns the live gateway key and `oauth.client_secret`. Two sibling routes in the same router return the same values and are not blocked.

## The two routes

```python
# vendor/open-webui/backend/open_webui/routers/configs.py:92-101
@router.post('/import', response_model=dict)
async def import_config(request: Request, form_data: ImportConfigForm, user=Depends(get_admin_user)):
    await Config.upsert(form_data.config)
    ...
    return await Config.get_all()
```

```python
# vendor/open-webui/backend/open_webui/routers/configs.py:114-116
@router.get('/namespace/{namespace}', response_model=dict)
async def get_config_namespace(namespace: str, user=Depends(get_admin_user)):
    return await Config.get_namespace(namespace)
```

`Config.get_all()` is the same call `export_config` makes, so `POST /api/v1/configs/import` with any body returns everything the blocked export would have returned, and writes arbitrary configuration on the way through. `GET /api/v1/configs/namespace/oauth` returns `oauth.client_secret`; `GET /api/v1/configs/namespace/rag` returns `rag.openai.api_key`, which is `OWUI_SHIM_KEY`.

## Why the existing rules miss them

`deploy/docker/Caddyfile.owui:103-106` blocks `configs/(?:export|terminal_servers|tool_servers)` as whole segments, and `import` and `namespace` are not in that list.

`deploy/docker/Caddyfile.owui:139-143` blocks write verbs on `/api/v*/configs*`, but a Caddy `path` wildcard does not cross a slash when the pattern also has an earlier wildcard, so it matches `/api/v1/configs` and nothing below it. That is #437.

## Measured, not assumed

Real matchers from `Caddyfile.owui`, run against a throwaway `caddy:2-alpine` container on 2026-08-17, upstream replaced by a 200 responder:

```
/api/v1/configs/import              GET=200  POST=200
/api/v1/configs/namespace/oauth     GET=200  POST=200
/api/v1/configs/export              GET=404  (blocked, as intended)
```

Live against the deployed box the same day, unauthenticated so no secret was retrieved and none is quoted here:

```
GET https://chat-hive.scubed.co/api/v1/configs/export             -> 404   (Caddy)
GET https://chat-hive.scubed.co/api/v1/configs/namespace/oauth    -> 401   (reached Open WebUI)
GET https://chat-hive.scubed.co/admin/settings                    -> 404   (Caddy)
```

The 401 is the finding: only the session check stands in front of that route, and every tenant OWNER on this deployment holds an administrator session (#748).

## Remediation

Add both to the segment-exact block in `Caddyfile.owui`, alongside `export`:

```
configs/(?:export|import|namespace|terminal_servers|tool_servers)
```

`/api/v1/configs/banners`, `/api/v1/configs/models` and `/api/v1/configs/connections` must stay reachable, which segment-exact matching preserves.

Two follow-ups worth deciding at the same time rather than separately:

* The import route is a configuration write, not just a read. Blocking it removes an administrator's ability to overwrite `openai.api_base_urls` and every `oauth.*` key from the public origin, which is a bigger win than the secret disclosure.
* This is the third instance of the same shape (`export`, then `openai/config`, now `import` and `namespace`). An allowlist of the handful of `configs/*` segments the front end actually needs would end the family, instead of extending a denylist a fourth time.

### LINKED ISSUE #949: Open WebUI Admin Settings is reachable from the Settings dialog and most of its writes land, contradicting D-014

sakibsadmanshajib:
Closes #736. Closes #948. Closes #949.

Three linked defects on the one public chat origin, fixed as a single ordered change, because the first of them cannot be closed until the deploy stops depending on it.

## What was open

**#736 (P0).** `Caddyfile.owui` intended to drop admin mutation verbs and the matcher was `path /api/v*/functions*`. A Caddy `path` pattern with more than one wildcard falls through to Go's `path.Match`, whose `*` never crosses a slash, so it covered `/api/v1/functions` and none of `/create`, `/id/<id>/update`, `/id/<id>/toggle`, `/id/<id>/toggle/global`. `POST /api/v1/functions/create` executes arbitrary Python inside a container holding `PGVECTOR_DB_URL` and `OWUI_SHIM_KEY`. The demo deploy installed `hive_jwt_forward` through that gap and said so in a comment, which is why the ordering matters.

**#948.** `POST /api/v1/configs/import` ends in `return await Config.get_all()`, the same call the blocked `configs/export` makes, so it read the whole persisted configuration back, gateway key and `oauth.client_secret` included, and overwrote configuration on the way through. `GET /api/v1/configs/namespace/oauth` and `.../rag` returned the same two values with no body at all.

**#949.** The Settings dialog rendered an Admin Settings link for `role === 'admin'` and navigated client side, so the address Caddy 404s was never requested and the panel drew itself out of the already-loaded bundle. The removal existed only in `owui-patches/hive_ui_surfaces.py`, which patches the stock bundle that `Dockerfile.open-webui` then discards in favour of our own build from `vendor/open-webui`.

## The change, in the order it has to happen

1. **Reroute the install.** `scripts/install-owui-jwt-forward-in-container.sh` copies the existing single-implementation installer and the Function body into the running chat container and runs them there against `http://127.0.0.1:8080`. No new listener, no new published port, no new credential, and the installer's `require_safe_origin` guard is satisfied unchanged, because loopback inside the container namespace is exactly the case it already permits. Both callers use it: the demo deploy and the nightly Open WebUI e2e setup. The wrapper is on the deploy paths filter, so a fix to it reaches the box.
2. **Then close the hole.** `@adminMutationSubtree` replaces the `configs` and `functions` globs with one regexp over both subtrees, default-deny rather than a list of today's endpoints, with a single `not` arm for `POST /api/v1/functions/id/{id}/valves/user/update`, the one route under either prefix that is `get_verified_user` and writes only the calling user's own valve values.
3. **#948** adds `import` and `namespace` to the segment-exact `@blocked` list, for every verb rather than only writes, because the disclosure half is a GET.
4. **#949** deletes the Settings dialog's admin link and the whole `vendor/open-webui/src/routes/(app)/admin` tree, so a client-side `goto` has no page left to render, and `Dockerfile.open-webui` grows a built-bundle assertion so a subtree pull cannot restore it quietly.

## What this deliberately does NOT close

* **#437's residue.** The `users`, `groups` and `models` globs keep their old semantics, and the reason is in the Caddyfile: `users.py` and `models.py` both carry `get_verified_user` writes (`/users/user/settings/update`, `/user/status/update`, `/user/info/update`, and in models.py `create`, `import`, `model/toggle`, `model/update`, `model/access/update`, `model/delete`). A subtree block there is an outage, not a fix; it needs a per-route allowlist. `POST /api/v1/users/{id}/update`, the route that mints administrators, is in that residue, and the nightly e2e's own password sync depends on reaching it.
* **#748.** Who holds the administrator bit is owned elsewhere and is the reason every one of these routes matters.
* Two further surfaces are removed only in the discarded bundle patch and not in the vendored source, the same defect class as #949, and are reported rather than fixed here: the vendor translation link in `Settings/General.svelte:268` and the changelog modal gated at `routes/(app)/+layout.svelte:334`.

## Proof

Measured, not asserted. Full method and logs in `docs/proof/chat-origin-admin-lockdown-2026-08-23/`.

**The matchers, before and after, against the shipped `Caddyfile.owui` with a stub upstream** (`caddy-before.log`, `caddy-after.log`). Twelve rows flip from 200 to 404; eighteen product rows are identical in both.

**Adversarial variants** (`traversal-after.log`). Seventeen normalisation shapes sent with `curl --path-as-is`: doubled and tripled slashes, dot and dot-dot segments arriving from a sibling prefix, percent-encoded separators, an uppercase path, a `v99` mount. All refused. `/api/v1/functions/id/../valves/user/update` is refused too, so the carve-out cannot be used as a way in.

**A real Open WebUI, holding a real administrator session** (`live-proof.log`). The same bearer token probes the public origin and the container's own loopback:

```
POST /api/v1/functions/create                        origin=404
POST /api/v1/functions/id/hive_jwt_forward/update    origin=404
POST /api/v1/functions/id/hive_jwt_forward/toggle    origin=404
POST /api/v1/functions/id/hive_jwt_forward/toggle/global  origin=404
DELETE /api/v1/functions/id/hive_jwt_forward/delete  origin=404
GET  /api/v1/configs/namespace/oauth                 origin=404
GET  /api/v1/configs/namespace/rag                   origin=404
POST /api/v1/configs/import                          origin=404

GET  /api/v1/configs/namespace/oauth  inside={"status": 200, "secret_fields_present": ["oauth.client_secret"]}
GET  /api/v1/configs/namespace/rag    inside={"status": 200, "secret_fields_present": ["rag.openai.api_key"]}

GET  /api/config                                     origin=200
GET  /api/v1/configs/banners                         origin=200
GET  /api/v1/models/list                             origin=200
POST /api/v1/users/user/settings/update              origin=200
```

The two `inside=` lines are the finding, printed as field names because a value must never appear in a capture: the identical admin session that gets 404 at the origin gets 200 in the container, with the OAuth client secret and the gateway key in the body. Both are placeholders on the proof stack, which holds no real credential at all.

**The install, through its new route** (`live-proof.log`, section 4): `wrapper exit code: 0`, then read back from the API as `{"id": "hive_jwt_forward", "type": "filter", "is_active": true, "is_global": true}`. Nothing crossed the public origin.

**The Settings dialog**, same administrator role in both runs, role read out of Open WebUI rather than assumed. Screenshots: https://github.com/sakibsadmanshajib/hive/pull/1091#issuecomment-5387775370

```
[before] settings admin links in the DOM: 1
[before] after clicking the Admin Settings link: {"path":"/admin/settings/general","hasAdminPanel":true}
[after]  settings admin links in the DOM: 0
[after]  direct navigation to /admin/settings: {"path":"/admin/settings","hasAdminPanel":false,"bodyStart":"404: Not Found"}
```

**The build assertion**, counted in both images (`bundle-evidence.log`): `/admin/settings/code-execution` appears in 2 chunks before and 0 after, while the `data-hive-nav` control is 2 in both, so the difference is this change and not a different build path.

Two deviations from the shipped build path are disclosed in the proof README rather than glossed: `main`'s chat frontend build is currently red (`<svelte:window>` inside a block, PR #1085 owns the fix, applied to the build tree only and not to this branch), and the proof image is the last two steps of `Dockerfile.open-webui` layered on the fully patched fork image, because the full build could not finish on this machine after four attempts died on CDN socket failures.

## Test plan

- `python3 scripts/test_caddy_owui_blocklist.py` extended in both directions: 21 write paths that must be refused, 9 that must stay open, plus the #948 reads. Verified failing against `origin/main` and passing here.
- `python3 scripts/test_install_owui_jwt_forward.py`: no caller may install through `localhost:3003`, the wrapper reaches loopback, is executable, and is on the deploy paths filter.
- `python3 scripts/test_owui_ui_surfaces.py`: no admin link in the dialog, no `(app)/admin` route tree, nothing outside `lib/components/admin` imports those components, and the Dockerfile checks the built bundle. Verified failing against `origin/main`.
- `make test-scripts`: full suite green.
- `go test ./apps/control-plane/... -count=1 -short`: green (no Go in this diff).
- `docker compose run --build --no-deps web-console npm run build`: green.

## Buglog entry

To be appended to `.wolf/buglog.jsonl` on `main` in a separate buglog-only pull request after this merges, per issue #873. Not appended here: a branch that touches that file conflicts with every other branch that touches it, and an unmergeable pull request gets no CI run at all.

```json
{"id":"bug-mgo736-948-949","timestamp":"2026-08-23T18:20:00.000Z","related_bugs":[],"occurrences":1,"last_seen":"2026-08-23T18:20:00.000Z","title":"Caddy path glob never crosses a slash, so the chat origin's admin block covered only the bare mounts, and the deploy came to depend on the gap","error_message":"Through the public chat origin, with a real Open WebUI admin session: POST /api/v1/functions/create 200, POST /api/v1/functions/id/<id>/toggle 200, GET /api/v1/configs/namespace/oauth 200 (returns oauth.client_secret), POST /api/v1/configs/import 200 (returns the same Config.get_all() the blocked configs/export does), and the Settings dialog's Admin Settings link opened the admin panel client side without ever requesting the 404'd address","root_cause":"Three causes with one shape, a control that reads as present while being absent. (1) Caddy's path matcher falls through to Go path.Match when a pattern carries more than one wildcard, and there * never crosses /, so path /api/v*/configs* /api/v*/functions* matched the bare mounts and no child of them. (2) The configs block was a denylist of segments, so export was refused while import and namespace/{ns}, which return the same values from the same router, were not: the fourth instance of that family after #769 and #771. (3) The Admin Settings link removal lived only in owui-patches/hive_ui_surfaces.py, which rewrites the stock bundle that Dockerfile.open-webui then discards in favour of the build from vendor/open-webui, and the vendored source removal documented as its counterpart was never made. Compounding all three: the demo deploy installed hive_jwt_forward through the functions gap, so tightening the matcher first would have turned every deploy red at the install step.","fix":"Reroute before blocking. scripts/install-owui-jwt-forward-in-container.sh runs the existing single-implementation installer inside the chat container against 127.0.0.1:8080, which satisfies its loopback-only token guard unchanged and adds no listener, and both callers (deploy-demo-box.yml, the nightly owui.setup.ts) use it. Then @adminMutationSubtree replaces the two globs with one default-deny path_regexp over the configs and functions subtrees, with a not arm for the single get_verified_user route under either (functions/id/{id}/valves/user/update). configs/(export|import|namespace|terminal_servers|tool_servers) closes the disclosure half for every verb. The Settings dialog link and the whole vendor/open-webui/src/routes/(app)/admin tree are deleted, with a built-bundle assertion in Dockerfile.open-webui verified in both directions. users, groups and models are deliberately untouched (#437): both routers carry get_verified_user writes, so a subtree block there is an outage rather than a fix.","tags":["caddy","path-matcher","open-webui","security","admin-surface","deploy-ordering","denylist-family"]}
```

## Review streams

Run per the pipeline, with one disclosure: this environment exposes no Agent dispatch tool and no Skill tool, so the streams that call for an independent agent or a skill were executed by the implementing agent by hand. That is stated at the top of each posted review rather than left implicit.

* Security pass (mandatory on this diff): posted with inline comments, including the bypass matrix, the token handling, and the one credential-exposure recommendation this pull request deliberately does not act on.
* Plain adversarial pass: posted, six findings, two of which produced changes in this branch (the deploy-time refusal assertion) and four of which are recorded as follow-ups.
* Code review: posted, `ecc:code-review` unavailable so its checklist was applied by hand.
* CodeRabbit CLI: **SKIPPED**, not a pass. Rate limited (`You've used all 3 included reviews currently available`), reported as a skip with the verbatim error.

## Exposure note

For as long as `configs/import` and `configs/namespace/{ns}` were reachable, any account holding an Open WebUI administrator session on this shared instance could read `oauth.client_secret` and the gateway key seeded from `OWUI_SHIM_KEY`. On this deployment that is every sole-OWNER tenant account (#748), five of which have signed in. #948's live probe was unauthenticated and retrieved nothing, so there is evidence of reachability and none of retrieval. Rotating `OWUI_SHIM_KEY` and the Supabase OAuth client secret is the conservative call; it is the owner's decision and nothing is rotated here.


sakibsadmanshajib (commented):
## Review stream: security pass (mandatory on this diff)

Run by the implementing agent, disclosed as such: this environment has no Agent dispatch tool, so the independent-reviewer streams the pipeline calls for were executed here by hand rather than by separate agents. Read the findings with that in mind.

Scope: a proxy matcher change on a public origin, a credential-bearing install path, and the removal of an admin surface from a forked front end. Three of the four categories that make stage 6 non-skippable.

**Verified, with evidence rather than reasoning**

1. Bypass matrix. Seventeen normalisation variants of the newly blocked paths were sent with `curl --path-as-is` against the shipped Caddyfile, so nothing was collapsed client side: doubled and tripled leading slashes, an interior doubled slash on both sides of the mount, dot and dot-dot segments arriving from a sibling prefix (`/api/v1/hive/../configs/import`), percent-encoded separators, an uppercase path, and a `v99` mount. All refused. Log: `docs/proof/chat-origin-admin-lockdown-2026-08-23/traversal-after.log`.
2. The carve-out cannot be used as a way in. `/api/v1/functions/id/../valves/user/update` normalises back into the blocked subtree and is refused, while the legitimate `/api/v1/functions/id/x/valves/user/update` passes.
3. Over-blocking check, which is the failure mode that matters more on this origin: 18 product paths answer identically before and after, including the four `get_verified_user` writes a subtree-shaped edit would have taken out.

**Open, stated rather than closed**

4. `POST /api/v1/users/{id}/update` is still reachable on the public origin with an administrator session, and that is the route that mints another administrator. It is #437's residue plus #748, called out inline below and in the PR body. It is not closable in this change without a per-route allowlist, because its neighbours are user-scoped writes the product depends on.
5. `GET /api/v1/functions/` stays readable by an administrator session. It returns the installed filter's source, which is this repository's own `deploy/docker/pipelines/hive_jwt_forward.py`, so no credential. Informational.
6. `GET /api/v1/configs/connections` stays reachable by design. Checked the response model rather than assuming: `ConnectionsConfigForm` is two booleans, no key. Its POST is now refused.

**Credential exposure, and a recommendation this change does not act on**

For as long as `configs/import` and `configs/namespace/{ns}` were reachable, any account holding an Open WebUI administrator session on the shared instance could read `oauth.client_secret` and the gateway key seeded from `OWUI_SHIM_KEY`. On this deployment that is every sole-OWNER tenant account (#748), five of which have signed in. #948's own live probe was unauthenticated and retrieved nothing, so there is no evidence of retrieval, only of reachability. Rotating `OWUI_SHIM_KEY` and the Supabase OAuth client secret is the conservative call and is a decision for the owner, not for this pull request. No credential is rotated here, and none appears anywhere in the proof: the proof stack's corresponding values are literal placeholders.

sakibsadmanshajib (commented):
## Review stream: plain adversarial pass

Same disclosure as the security stream: no Agent dispatch tool here, so this was run by the implementing agent rather than an independent one.

The question this pass asks is not "is the change correct" but "what does it break, and how would we find out".

**A-1. Default-deny over `configs` can refuse a route the product later needs.** Every write under `configs` and `functions` is `get_admin_user` in the pinned v0.10.2 image, so the block is correct today. If an upstream release adds a user-facing write under either prefix, this matcher refuses it and the symptom is a 404 in the browser console rather than a test failure. Accepted deliberately, and the Caddyfile says why: the alternative shape, a denylist of the endpoints that exist today, is what produced #769, #771 and now #948 in sequence. A denylist fails silently and open; this fails visibly and closed.

**A-2. The reroute removes CI's only live exercise of the blocked paths.** Before this change, every deploy proved by side effect that `POST /api/v1/functions/create` was reachable through the public origin. Afterwards nothing exercised that origin at all, which would have left the block's real state unmeasured on the running box, the exact shape this repository keeps getting caught by. Answered by the new deploy step, which is also #736's own stated verification expectation.

**A-3. The install now depends on `docker compose cp`.** Available in Compose v2 and confirmed on this machine (v5.3.1). If a box ever runs something older, the deploy fails loudly at the install step with a compose error, which is the correct direction for a failure.

**A-4. Three admin links elsewhere in the fork now dead-end.** With the route tree gone, `ModelSelector/Selector.svelte:709` ("Manage Connections", rendered only when the model list is empty AND the session is an administrator), `ModelSelector/ModelItemMenu.svelte:60` and `MessageInput/TerminalMenu.svelte:199` navigate to a route that no longer exists, so they land on the SPA's not-found view instead of an admin panel. Strictly better than opening a live panel, and both conditions are rare on this deployment, but it is a visible wart and belongs in a follow-up issue rather than in this diff.

**A-5. Two more surfaces are removed only in the discarded bundle patch.** `hive_ui_surfaces.py` also rewrites `settings-vendor-translate-link` and `changelog-modal`, and neither removal exists in the vendored source: the translate link is still at `Settings/General.svelte:268` and the changelog is still gated at `routes/(app)/+layout.svelte:334`. Same defect class as #949 and visible in the same dialog this change photographs. Reported rather than fixed, because one of them is branding-adjacent and neither is a security boundary.

**A-6. `docs/owui-fork.md` currently makes a claim that was not true.** "Every surface they remove is removed again, independently, in the vendored source, and that source removal is what actually reaches a user." That was the invariant #949 violated. This change makes it true for the admin link and adds a build assertion behind it; A-5 is the remaining gap in the same sentence.

sakibsadmanshajib (commented):
## Review stream: code review

The `ecc:code-review` skill could not be invoked: this environment exposes no Skill tool, which is a structural capability gap rather than a skipped step. Its checklist was applied by hand instead, and that substitution is disclosed here rather than reported as a clean pass.

**Correctness**

* The two regexes are RE2-compatible (no lookaround, no backreference), which matters because `scripts/test_caddy_owui_blocklist.py` reads them out of the Caddyfile and runs them through Python's `re`. Go and Python agree on every construct used.
* The absence of lookahead is the reason the carve-out is a separate `not` matcher rather than a negative lookahead inside one pattern. Worth knowing before someone tries to "simplify" it into one line: RE2 cannot express it.
* `require_safe_origin` is untouched. The reroute satisfies it rather than relaxing it, which is the part of this change most likely to have been done the lazy way.

**Tests**

* Both new test surfaces were verified failing against `origin/main` before being kept, so neither is a green that cannot go red: the Caddy test dies on the missing matcher, and the surfaces test names the restored admin routes as importers.
* `MUST_ALLOW_WRITES` is the half that earns its keep. `users.py` and `models.py` carry `get_verified_user` writes, so the next person widening a glob in that block gets a red test instead of an outage.
* The workflow assertion is the only check here that runs against the deployed box rather than against the repository, and it is the one that would have caught this class in the first place.

**Style and structure**

* Comment density in the Caddyfile is high, and matches the file's existing convention: every matcher there carries the reasoning for its exact shape, because the shape is the security property.
* The wrapper is 80 lines, half of them comment, and adds no abstraction: it copies two files in and runs the existing implementation. No new install logic exists anywhere.
* One duplication accepted: the compose flags are spelled in the deploy workflow's `HIVE_COMPOSE_FLAGS` and defaulted again in `owui.setup.ts` for the nightly's own stack. Flagged inline.

**Not reviewed here**

No Go changed. `go test ./apps/control-plane/... -count=1 -short` was run anyway and is green; `docker compose run --build --no-deps web-console npm run build` compiles.

Issue #749, opened 2026-08-06
Invite acceptance ignores the invited role, so an invited OWNER is provisioned as MEMBER


sakibsadmanshajib:
## Summary

`tenant_invites.role` lets an inviter choose the role the invited user will hold, including OWNER, but acceptance discards that choice. Every accepted invite provisions the user as MEMBER. The invite screen therefore offers a grant that the members and RBAC surface never delivers, and there is no in product way to create a second OWNER for a tenant.

This is an under grant, not a privilege escalation. Nobody receives more authority than intended; the invited user receives less. It is filed as a correctness and product defect, not a security issue.

## Evidence

1. The schema supports the choice. `supabase/migrations/20260516_08_phase19_tenant_invites.sql:9` defines `role text NOT NULL CHECK (role IN ('OWNER','ADMIN','MEMBER','VIEWER')) DEFAULT 'MEMBER'`, so an invite row can legitimately carry OWNER, ADMIN, or VIEWER.
2. The resolver drops it. `signup.LookupFunc` is declared at `apps/control-plane/internal/signup/resolver.go:33` as returning a tenant id and an error, with no role in the signature. The invite implementation at `apps/control-plane/cmd/server/main.go:1383` selects only `tenant_id` from `public.tenant_invites`, so `tenant_invites.role` is never read on the acceptance path.
3. Acceptance funnels into the hardcoded insert. With only a tenant id in hand, the invite path calls the same `provision` used by every other signup path, at `apps/control-plane/internal/signup/reconcile.go:250-258`, which writes `VALUES ($1, $2, 'MEMBER', 'ACTIVE')`. The role is a literal, so the invited role could not be honoured even if it were resolved.

## Consequence

* An invite issued as ADMIN or VIEWER lands as MEMBER. The invited user sees permissions that do not match what the inviter selected, and the tenant administrator has no signal that the selection was dropped.
* There is no in product way to create a second OWNER for a tenant. The only OWNER writers in the repository are operator seeding scripts that require the Supabase service role key, so promoting a second owner today means running a script out of band against the database.

## Constraint on any fix

Honouring the invited role must not reintroduce the hazard tracked in #748. Open WebUI derives its administrative role from a Hive tenant role, so an accepted OWNER invite would, with the current mapping in place, also make that user an administrator of the shared chat instance. A fix that simply threads `tenant_invites.role` through the resolver and into the insert would turn a latent hazard into a reachable one.

[...]

PR #1096, opened 2026-08-23
fix: stop deriving Open WebUI instance admin from a tenant role

sakibsadmanshajib:
Closes #748.

Open WebUI is one instance shared by every Hive tenant, and it derived its instance-admin role from a Hive tenant role: a single ACTIVE `tenant_users` row with `role = 'OWNER'` became Open WebUI `admin`. A tenant OWNER is a customer, not a platform operator, so that handed one customer administrative control over every other customer's chat. The issue's own live audit of the demo box walked it end to end: a legitimately provisioned business tenant OWNER signed in, received admin, enumerated other tenants' users, read another user's chat titles and read another tenant's uploaded file.

## What now grants instance admin

An explicit platform attribute the control plane sets deliberately: `public.accounts.is_platform_admin = true` plus an ACTIVE `owner` row in `public.account_memberships`. That is the same predicate the control plane already uses for its own platform-admin surfaces (`apps/control-plane/internal/platform/role_pgx.go`, `IsPlatformAdmin`), so there is one definition of platform operator with two consumers rather than a second, chat-only notion of admin. Per `.wolf/decisions.md` D-044, the control plane owns the state and the knobs and Open WebUI is a view.

Everything else about the lookup is unchanged in spirit and stricter in practice: any ACTIVE membership on a non-archived tenant resolves the ordinary `user` role, and an unknown email, an identity with no active membership, an ambiguous identity or a failed lookup all keep the pre-existing fallback, which is `DEFAULT_USER_ROLE`, so the failure mode stays closed.

One deliberate widening: a person with active memberships in two tenants used to resolve to nothing and land on the activation-pending screen. That condition existed only to narrow the OWNER promotion, and with the promotion gone there is nothing left to narrow, so they now resolve `user` like any other member.

## Three promotion paths closed

1. **The tenancy-derived one.** `deploy/docker/owui-patches/tenant_role_from_db.py` no longer looks at a tenant role to decide admin.

2. **Upstream's two single-user bootstraps.** `apply_tenant_role_patch.py` now also deletes stock Open WebUI's login-time promotion inside `get_user_role` and its post-insert sibling in `handle_callback`. On a shared instance those made whichever account authenticated first after a fresh volume, a restore or a database reset the administrator of everyone's chat, regardless of tenant. The login-time one also returned *above* the splice point, so the Hive lookup never ran for that login at all. Both removals are exact-literal with count assertions, so an upstream digest bump fails the build loudly instead of silently restoring them.

3. **The claim.** `supabase/migrations/20260823_03_owui_role_never_admin.sql` stops `custom_access_token_hook` emitting `owui_role = 'ADMIN'`. OWNER and ADMIN both emit MEMBER, so both keep chat access and neither matches `OAUTH_ADMIN_ROLES`. The `role` claim is untouched, as in every prior revision of this function, because RLS policies and `apps/edge-api/internal/auth/jwt_supabase.go` read it verbatim. That claim path is inert today (Supabase's third-party OIDC token carries no custom claim), but inert is not harmless: the value was emitted, the consumer's admin list still named it, and nothing of ours guarded the gap between them.

## Added during review: a failure can no longer leave an admin standing

The adversarial pass asked what the lookup does when it cannot decide, and the original answer, keep whatever fallback the machinery computed, was only safe while that fallback could not be `admin`. It could, in two shapes: role management off with `DEFAULT_USER_ROLE=admin`, and upstream carrying a stored role forward for an existing user (`role = user.role`), so an existing Open WebUI admin kept admin every time the database was unreachable. The fragment now clamps that to `user`, so admin comes only from a successful platform-attribute check, in code rather than by configuration.

Proven live, not reasoned about: with the database pointed at a host that does not resolve and a stored role of `admin` carried in, `before` answered `admin` for every identity including one with no Hive membership at all, and `after` answered `user`. The cost is stated in the fragment's comment: a transient failure demotes a genuine operator for one session and their next successful login restores admin, a recoverable loss of privilege instead of an unrecoverable grant of it. On this deployment the fallback is `pending`, so it is a no-op.

## The issue's first recommendation, declined with reasons

The issue proposes a constraint or trigger on `public.tenant_users` that refuses an OWNER membership write from anything but the service role. Not done, deliberately:

* It defends the mapping this change deletes. With instance admin no longer derivable from tenancy, an OWNER membership is an ordinary business fact again.
* It would block #749. That issue needs invite acceptance to be able to grant OWNER, and it explicitly asks for this mapping to be closed *first* so that becomes safe. A trigger refusing non-service-role OWNER writes would trade one defect for another.

The issue's second recommendation, stop deriving the chat platform's administrative role from a Hive tenant role at all, is the durable one and is what this does.

## Live exposure on the box, and the remediation

**Yes, live, and it is not hypothetical.** Any account that signed into chat while holding a single ACTIVE OWNER membership carries `role = 'admin'` in Open WebUI's own user table, which on this deployment is SQLite at `/data/webui.db` inside the container, not Postgres. The demo owner provisioned by `scripts/seed-demo-owner.py` is exactly that shape, and #947's own capture on `main` observed such an account holding admin.

Nothing here was executed against live data. Recommended remediation, in order:

1. **Grant the operator's own workspace the attribute before or immediately after this merges**, or the deploy loses its Open WebUI admin. `scripts/install-owui-jwt-forward.py` runs on every deploy and needs an admin session, which `scripts/owui-mint-admin-token.py` mints for an admin account that already exists, and that account exists today only because of this bug. On the box's Postgres:

   ```sql
   UPDATE public.accounts SET is_platform_admin = true WHERE slug = '<the operator workspace slug>';
   INSERT INTO public.account_memberships(account_id, user_id, role, status)
   VALUES ('<that account id>', '<the operator user id>', 'owner', 'active')
   ON CONFLICT (account_id, user_id) DO UPDATE SET role = 'owner', status = 'active';
   ```

   Then have that account sign in to chat once so Open WebUI records the role. The mint script's no-admin error message now says all of this instead of telling an operator to sign in as a tenant OWNER.

2. **Demote the accounts that hold admin because of the bug.** The role self-heals on the next OIDC login: the callback compares the resolved role against the stored one and writes the difference (`utils/oauth.py`, `update_user_role_by_id`). To demote immediately rather than on next sign-in, an operator updates the SQLite row inside the container:

   ```sql
   UPDATE "user" SET role = 'user' WHERE role = 'admin' AND email <> '<the operator address>';
   ```

   Do that after step 1, not before, or the instance has no admin at all in between.

3. Decide #747 separately. It is about who holds `is_platform_admin`, which is now a visible, deliberate grant rather than a side effect of tenancy.

## Consequences handled rather than discovered later

* **The nightly Open WebUI suite** installs the `hive_jwt_forward` Function through an admin-gated API and obtained its admin session through this mapping. It now gets one from `scripts/owui-promote-instance-admin.py`, run inside that job's own disposable chat container after the fixture's login and before the install.

  The first version of this PR gave the fixture the platform attribute instead, in `scripts/seed-owui-e2e-user.py`. My own security pass rejected it and it is worth stating why, because it is the more obvious fix: that seeder writes to whatever database the job points at, which is a real project, and `is_platform_admin` carries the platform-wide authority behind credit minting and provider `base_url` rewrites. It would have created exactly the kind of holder #747 exists to question, to give a CI fixture an admin session it needs for one API call. The container-local promotion carries no Hive authority, lives and dies with the job, and is undone by that account's next login, which is precisely the lifetime required. `scripts/seed-owui-e2e-user.py` is therefore untouched by this PR.
* **Comments that described the old behaviour as current** are corrected in the e2e fixture, in the compose file's `#947` note, and around `OAUTH_ADMIN_ROLES`, which is now annotated as deliberately unmatchable rather than left looking like an active grant.

## Residual, measured and named

Native signup carries its own first-user promotion in `open_webui/routers/auths.py`, which this change does not touch: in the capture both images answered HTTP 200 with role `admin` to the first `POST /api/v1/auths/signup`. On every deployment this repository ships that route is unreachable (`ENABLE_SIGNUP` is false in compose and `Caddyfile.owui` answers 404 on the auths routes) and it is a password flow for which no customer is ever issued a password. It is left out of this PR rather than folded in because the Caddy surface it depends on was being changed concurrently for #736, #948 and #949, now merged as #1091. Worth its own issue.

## Tests

* `TestCustomAccessTokenHook_OwuiRoleNeverGrantsInstanceAdmin` (`apps/control-plane/internal/tenants`, database backed) invokes the real hook for all four tenant roles and asserts the emitted `owui_role` is never Open WebUI's admin role and is always a value `OAUTH_ALLOWED_ROLES` accepts, so neither promotion nor stranding can regress silently. Run against the pre-fix hook definition on a real Postgres it fails on OWNER and ADMIN and passes on MEMBER and VIEWER.
* `scripts/test_owui_tenant_role.py`, new, wired into `make test-scripts` (a required check): executes the real patch fragment against a stubbed psycopg2 for every branch (operator, member, nobody, ambiguous, raising, no database URL, an already-resolved role, and an `admin` fallback that must be clamped), asserts the email is a bound parameter and the statement stays time-bounded, and runs the real `apply_tenant_role_patch.py` against a copy of the pinned image's own `oauth.py` to assert both upstream promotions are gone, the fragment is spliced and the result still compiles. Against `origin/main`'s files, 10 of its 21 checks fail.
* `scripts/owui-promote-instance-admin.py --self-check`, also in `make test-scripts`, covers the container-local promotion: only the named account moves, a second run is a no-op, a missing account fails rather than being invented, and a missing database fails loudly.
* `make test-scripts` and `go test` for the two touched packages are green, the latter both with and without a database.

## Proof

Live capture, A/B against a purpose-built throwaway stack with its own Postgres, its own network and its own ports (the shared `.env` points at a deleted Supabase project, so there is no shared stack to use). Both chat containers are built from the exact digest `Dockerfile.open-webui` pins and differ only in the two patch files, and a third container built from the complete Dockerfile carries the fork's own frontend for the browser capture. Text log: `docs/proof/instance-admin-not-from-tenant-owner-2026-08-23/live-transcript.md`. Screenshots posted separately as permanent release assets.

The headline, from the running container's own `get_user_role`, against real seeded rows:

| identity | before | after |
|---|---|---|
| platform operator (`is_platform_admin`) | pending | **admin** |
| tenant OWNER, one active membership | **admin** | user |
| tenant MEMBER | user | user |
| member of two tenants | pending | user |
| no membership at all | pending | pending |

And with exactly one account in the instance, which is the state after any volume reset, `before` answered `admin` for **every** identity including one with no Hive membership at all, while `after` answered exactly as above.

## Buglog entry

```json
{"id":"bug-2026-08-23-owui-instance-admin-from-tenant-owner","date":"2026-08-23","title":"Open WebUI instance admin was derived from a tenant role, granting a customer admin over every other tenant's chat","error_message":"A tenant OWNER signing into the shared Open WebUI instance resolved Open WebUI role=admin, could enumerate every tenant's users, read another user's chat titles and read another tenant's uploaded file","root_cause":"Three independent promotion paths on one shared instance: deploy/docker/owui-patches/tenant_role_from_db.py mapped a single ACTIVE tenant_users row with role OWNER onto Open WebUI admin; stock Open WebUI promoted the only user of the instance to admin both at login time (above the Hive splice point, so the Hive lookup never ran for that login) and after inserting a user; and public.custom_access_token_hook emitted owui_role ADMIN for OWNER while OAUTH_ADMIN_ROLES named ADMIN. Instance admin was a side effect of tenancy rather than a deliberate platform grant. A fourth, fail-open path was found on this branch's own adversarial pass: a failed lookup kept whatever fallback role was already computed, which upstream sets from the stored role, so an existing admin kept admin whenever the database was unreachable","fix":"Resolve instance admin only from public.accounts.is_platform_admin plus an ACTIVE owner row in public.account_memberships, the same predicate apps/control-plane/internal/platform/role_pgx.go uses; clamp an admin fallback to user when the lookup fails; delete both upstream single-user bootstraps through assertion-guarded exact-literal rewrites in apply_tenant_role_patch.py; add supabase/migrations/20260823_03_owui_role_never_admin.sql so owui_role emits MEMBER for OWNER and ADMIN. Regression guards: TestCustomAccessTokenHook_OwuiRoleNeverGrantsInstanceAdmin (database backed) and scripts/test_owui_tenant_role.py in make test-scripts","tags":["auth","security","open-webui","multi-tenant","privilege-escalation","issue-748"]}
```

## Follow-ups, not folded in

* #747, who should hold `is_platform_admin`, now a visible grant rather than an emergent one.
* #749, honouring the invited role, which this unblocks in the order that issue asked for.
* Native signup's own first-user promotion, per the residual above.
* Revocation is not immediate: removing the attribute leaves that account's Open WebUI admin row in place until their next login re-resolves it. Pre-existing, unchanged here, worth its own issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ test-scripts:
 	python3 scripts/test_owui_ui_surfaces.py
 	python3 scripts/test_owui_display_name.py
 	python3 scripts/test_owui_tenant_role.py
+	python3 scripts/owui-promote-instance-admin.py --self-check
 	python3 scripts/test_caddy_owui_blocklist.py
 	python3 scripts/test_owui_model_picker_filter.py
 	python3 scripts/generate-enterprise-jwt-keys.py --self-check
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ test-scripts:
 	python3 scripts/test_owui_rag_env_config.py
 	python3 scripts/test_owui_ui_surfaces.py
 	python3 scripts/test_owui_display_name.py
+	python3 scripts/test_owui_tenant_role.py
 	python3 scripts/test_caddy_owui_blocklist.py
 	python3 scripts/test_owui_model_picker_filter.py
 	python3 scripts/generate-enterprise-jwt-keys.py --self-check
--- a/apps/control-plane/internal/signup/hook_claims_test.go
+++ b/apps/control-plane/internal/signup/hook_claims_test.go
@@ -62,16 +62,30 @@ func TestTokenHookNeverRaisesForMembershiplessUser(t *testing.T) {
 	// The member path must still emit all four claims. owui_role in
 	// particular is what Open WebUI's OAUTH_ROLES_CLAIM reads
 	// (deploy/docker/docker-compose.yml), and dropping it would strand every
-	// member on Open WebUI's activation-pending screen.
+	// member on Open WebUI's activation-pending screen. It must also never
+	// carry the value OAUTH_ADMIN_ROLES names, because instance admin is
+	// accounts.is_platform_admin and never a tenant role (issue #748).
 	for _, want := range []string{
 		"jsonb_build_object('tenant_id', selected)",
 		"jsonb_build_object('tenants',   COALESCE(tenant_list, '[]'::jsonb))",
 		"jsonb_build_object('role',      user_role)",
-		"jsonb_build_object('owui_role', CASE WHEN user_role = 'OWNER' THEN 'ADMIN' ELSE user_role END)",
+		"jsonb_build_object('owui_role', CASE WHEN user_role IN ('OWNER', 'ADMIN') THEN 'MEMBER' ELSE user_role END)",
 	} {
 		if !strings.Contains(code, want) {
 			t.Fatalf("%s: member claim emission must still contain %q",
 				filepath.Base(path), want)
+		}
+	}
+
+	// Belt and braces on the same property, stated as a prohibition rather
+	// than a match: whatever shape the owui_role expression takes in future,
+	// the literal 'ADMIN' must not appear in it.
+	for _, line := range strings.Split(code, "\n") {
+		if strings.Contains(line, "'owui_role'") && strings.Contains(line, "'ADMIN'") {
+			t.Fatalf("%s: owui_role emits 'ADMIN', which OAUTH_ADMIN_ROLES maps to "+
+				"Open WebUI instance admin. Instance admin is a platform attribute, "+
+				"never a tenant role (issue #748): %s",
+				filepath.Base(path), strings.TrimSpace(line))
 		}
 	}
 
@@ -148,7 +162,7 @@ func TestTokenHookClaimsUnchangedForMember(t *testing.T) {
 	userID := mustInsertAuthUser(t, ctx, pool, "hookmember-"+uuid.NewString()+"@example.com")
 
 	// OWNER specifically, because that is the role the owui_role remap
-	// rewrites to ADMIN. A MEMBER would pass a weaker version of this test.
+	// rewrites. A MEMBER would pass a weaker version of this test.
 	_, err := pool.Exec(ctx,
 		`INSERT INTO public.tenant_users(tenant_id, user_id, role, status)
 		 VALUES ($1, $2, 'OWNER', 'ACTIVE')`, tenantID, userID)
@@ -160,8 +174,10 @@ func TestTokenHookClaimsUnchangedForMember(t *testing.T) {
 		"member must be bound to their tenant")
 	require.Equal(t, "OWNER", claims["role"],
 		"role claim is read verbatim by RLS policies and must stay OWNER")
-	require.Equal(t, "ADMIN", claims["owui_role"],
-		"owui_role must remap OWNER to ADMIN for Open WebUI's OAUTH_ALLOWED_ROLES")
+	require.Equal(t, "MEMBER", claims["owui_role"],
+		"owui_role must remap OWNER to a non-admin value in Open WebUI's "+
+			"OAUTH_ALLOWED_ROLES: the OWNER to ADMIN remap made a customer an "+
+			"administrator of the shared chat instance (issue #748)")
 
 	tenants, ok := claims["tenants"].([]any)
 	require.True(t, ok, "tenants claim must be a JSON array, got %T", claims["tenants"])
--- a/apps/control-plane/internal/signup/hook_claims_test.go
+++ b/apps/control-plane/internal/signup/hook_claims_test.go
@@ -77,17 +77,13 @@ func TestTokenHookNeverRaisesForMembershiplessUser(t *testing.T) {
 		}
 	}
 
-	// Belt and braces on the same property, stated as a prohibition rather
-	// than a match: whatever shape the owui_role expression takes in future,
-	// the literal 'ADMIN' must not appear in it.
-	for _, line := range strings.Split(code, "\n") {
-		if strings.Contains(line, "'owui_role'") && strings.Contains(line, "'ADMIN'") {
-			t.Fatalf("%s: owui_role emits 'ADMIN', which OAUTH_ADMIN_ROLES maps to "+
-				"Open WebUI instance admin. Instance admin is a platform attribute, "+
-				"never a tenant role (issue #748): %s",
-				filepath.Base(path), strings.TrimSpace(line))
-		}
-	}
+	// This file can only match text. The behavioural guard on the same
+	// property, which executes the hook for every tenant role against a real
+	// database and asserts the emitted value is never Open WebUI's admin role,
+	// is TestCustomAccessTokenHook_OwuiRoleNeverGrantsInstanceAdmin in
+	// apps/control-plane/internal/tenants. A text check cannot tell an 'ADMIN'
+	// on the input side of the CASE from one on the output side, so it is not
+	// attempted here.
 
 	// Absent, not null. An explicit null tenant_id claim invites a consumer
 	// to read the key, find JSON null, and treat it as a wildcard.
--- a/apps/control-plane/internal/signup/personal_tenant.go
+++ b/apps/control-plane/internal/signup/personal_tenant.go
@@ -146,12 +146,21 @@ func ensurePersonalTenant(ctx context.Context, pool *pgxpool.Pool, userID uuid.U
 //
 // Role is MEMBER, matching (*Provisioner).provision, and that is deliberate
 // even though the user is the only member of their own tenant. OWNER would be
-// the intuitive choice, but public.custom_access_token_hook remaps an OWNER
-// role to owui_role ADMIN for Open WebUI's OAUTH_ROLES_CLAIM, so making every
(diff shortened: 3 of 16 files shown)


sakibsadmanshajib (commented):
Adversarial review, pipeline mode. Streams run and their outcome, including the two that could not run, which are reported as SKIPPED and not counted as passes.

| Stream | Outcome |
|---|---|
| CodeRabbit CLI (`coderabbit review --committed --base main`) | RAN. 14 files reviewed, 1 minor finding, fixed. First attempt was rate limited ("all 3 included reviews used"); re-run after the window and it completed. |
| Plain adversarial pass, ticket first | RAN. One real finding (fail-open on lookup failure), fixed and proven live. Four other candidates examined and dismissed with reasons, below. |
| Security review, mandatory on an auth path | RAN. One finding rejected the branch's own first design and it was reverted before merge, see below. |
| Database review | RAN as part of the security pass, since the diff's SQL is the authorization predicate itself. |
| `ecc:code-review` | SKIPPED. Capability gap, not a clean pass: this agent has no Skill tool, so the skill cannot be invoked. Its ground is covered by the two passes above, but that is a substitution, not the stream. |
| `/codex:adversarial-review` | SKIPPED. `codex review --base origin/main` returned "You've hit your usage limit ... try again at Sep 10th". No output was produced, so there is nothing to report from it. |

### Security pass, the finding that changed the design

The first version of this branch gave the nightly Open WebUI fixture the platform attribute (`accounts.is_platform_admin`) in `scripts/seed-owui-e2e-user.py`, because that fixture installs the `hive_jwt_forward` Function through an admin-gated API and had been getting its admin session from the mapping this PR removes.

That is a privilege escalation introduced by a fix for a privilege escalation. `owui-nightly.yml` reconstructs its connection from repository secrets, so the seeder writes to a real project, and the attribute carries the platform-wide authority behind credit-grant minting and provider `base_url` rewrites (the capabilities #747 documents). Reverted in 99cd43b; `scripts/seed-owui-e2e-user.py` is untouched by this PR. The fixture now gets an Open WebUI admin row inside that job's own disposable container, with no Hive authority attached, undone by its next login.

### Candidates examined and dismissed

1. **A limited-privilege `PGVECTOR_DB_URL` role could make the new `EXISTS` return false under RLS, silently withholding admin.** True, and it fails in the safe direction. The connection already had to read `auth.users`, which is more privileged than the two public tables added here, so no deployment that worked before can break in the dangerous direction.
2. **`lower(u.email)` is not indexed, so this is a sequential scan on every login.** Unchanged from the previous query's shape, `auth.users` is small, and the statement is bounded by the `SET LOCAL statement_timeout` the fragment already sets. Left alone rather than adding an index this diff cannot verify in production.
3. **Revocation is not immediate: an operator whose attribute is removed keeps their Open WebUI admin row until their next login.** Real, pre-existing, and unchanged by this PR (the same was true of the tenant-role mapping). Worth its own issue; a login-time re-resolution is not a substitute for revocation.
4. **Native signup's own first-user promotion is still there** (`open_webui/routers/auths.py`). Measured, not assumed: both images answered HTTP 200 with role `admin` to the first `POST /api/v1/auths/signup`. Out of scope and named in the PR body with the two configuration gates that make it unreachable on every deployment this repo ships.

Issue #1100, opened 2026-08-23
Choose a true offsite destination for demo box backups


sakibsadmanshajib:
Filing from the #1000 backup work (branch `feat/box-backups`). One encrypted copy of each day's backup set now leaves the box via `scripts/pull-box-backups.sh`, landing on the owner's dev machine at `/home/sakib/hive-backups/hive-demo/daily/`. That closes the single-physical-box scenario but not the co-loss scenario (dev machine and box lost together, e.g. theft, fire, or ransomware hitting both since one human carries both).

## What needs deciding

A true offsite/object-store destination is a design decision, not an implementation detail. Candidates already in play across this repo:

| Option | Fits because | Costs |
| --- | --- | cerebrum says |
| Backblaze B2 / S3-compatible object store | ~$1/TB-month class pricing, S3 API, lifecycle rules do retention server-side | new vendor account + keys to manage; another secret surface |
| Supabase Storage on a DIFFERENT provider/project than production | no new vendor if Supabase stays the storage product; bucket-level policies | same-vendor correlation risk that made managed-Supabase exit happen |
| Second physical location (owner's other hardware, e.g. hive-lab Windows box) | zero new vendors, sovereign posture consistent with D-044 thinking | depends on hive-lab availability and connectivity; manual failover story |

Any choice must keep artifacts encrypted client-side before upload (the current artifacts already are), so the vendor never holds plaintext or the passphrase.

## Current stopgap

`scripts/pull-box-backups.sh` pulls only encrypted artifacts over rsync/ssh to the dev machine; passphrase copy stored separately at `/home/sakib/hive-backups/etc/passphrase-hive-demo`. Until a real offsite exists, box loss plus dev-machine loss takes every copy.

PR #1101, opened 2026-08-23
feat: scheduled encrypted backups for all four demo box data stores

sakibsadmanshajib:
# Demo box backups: all four production stores, scheduled, encrypted, restore-proven

Closes #1000. Since leaving managed Supabase, every production store on the demo box was single-copy on one physical machine. This change puts all four stores on a scheduled, unattended, reboot-surviving schedule with encryption at rest, an off-box encrypted copy, a verified throwaway restore proof executed against live production data, and a loud failure signal reusing the existing Alertmanager routing.

## What is backed up, and how

| Store | Capture method | Why this method |
| --- | --- | --- |
| Postgres (`postgres` DB: `auth.users` identities, tenants, api keys, credit ledger) | `pg_dump --format=custom` streamed out of `hive-supabase-db-1` over docker exec stdout | Consistent snapshot of the running server; no stop or restart of any service |
| Open WebUI relational state (`webui.db` on the `hive_owui-data` volume) | SQLite online backup API inside the open-webui container, integrity-checked before it leaves | Copying webui.db while its WAL is active is not safe; the backup API gives a consistent copy from a live writer |
| Open WebUI uploads (`/data/uploads`) | tar streamed via docker exec | Plain files |
| Supabase Storage object bytes (`hive-files`, `hive-images` buckets) | tar of `/var/lib/storage` streamed via docker exec | Object bytes sit under a nested `stub/stub/` layout; bucket metadata rides the DB dump |

## Retention sized to measured reality

Measured 2026-08-23: db dump 2.4 MB custom format, webui.db 1.0 MB, uploads under 8 KB compressed, storage 8 KB compressed. A full daily set is under 4 MB.

- On-box retention: 14 daily sets, under 70 MB total, against roughly 34 GB free on the box root filesystem.
- Off-box accumulation: about 120 MB per month if never pruned; `scripts/pull-box-backups.sh --prune` mirrors on-box retention.

## Schedule and watchdog

- systemd USER units (`deploy/systemd-user/hive-box-backup.{service,timer}`): fires 03:15 and 15:15 UTC daily, Persistent=true catches up missed slots after downtime. User-level because there is no sudo on the box; reboot-surviving because Linger=yes is enabled for sakib (verified live). Installed and live on the box since 2026-08-23 20:34 UTC, timer's first scheduled fire 03:16 UTC 2026-08-24.
- Hourly cron watchdog runs `backup-box.sh --check`: silent when fresh, alerts when the last success exceeds 26 hours, so the death of the timer itself surfaces within an hour.

## Failure signal

The script posts directly to Alertmanager's v2 API on the published host port (localhost:9093), riding the existing routing tree and hive-ops email receiver made working by #998. No new component deployed:

- `HiveBoxBackupFailed`: posted when any step errors, including a systemd timeout kill (signal trap), auto-resolves once a later run succeeds.
- `HiveBoxBackupStale`: posted by the watchdog on staleness.

Verified live: a forced-stale check posted successfully and appeared in Alertmanager's list addressed to receiver hive-ops. Success runs post nothing; resolve_timeout ends earlier failures on their own. At-a-glance state lives in `/home/sakib/hive-backups/status/STATUS.txt` on the box.

## Restore proof, executed live against production data

`scripts/restore-box-backup.sh` verifies SHA256SUMS before decrypting anything, stands up a throwaway Postgres from the exact production image (`hive-supabase-db:pg16-cron`, `--network none`, no published ports), restores with strict error filtering (any error outside a justified allowlist fails the proof even when counts match), compares row counts read-only against live, then destroys the throwaway. Final run 2026-08-23 21:05 UTC, exit 0:

| Table | Live | Restored | Verdict |
| --- | --- | --- | --- |
| public.credit_ledger_entries | 9791 | 9791 | ok |
| public.tenants | 970 | 970 | ok |
| public.api_keys | 90 | 90 | ok |
| auth.users | 166 | 166 | ok |
| auth.identities | 163 | 163 | ok |
| storage.objects | 15 | 15 | ok |

All six matched. The allowlist covers three verified categories: pg_cron DDL (needs shared_preload config a bare container does not set), grants to production roles that do not exist in a throwaway, and three foreign keys PRODUCTION ITSELF violates through its retention purge (2356 + 483 + 161 orphaned rows counted live, filed as #1102). Every row itself restores.

Open WebUI side verified too: decrypted webui.db passes integrity_check with chat 23=23, user 11=11, knowledge 2=2 vs live. Storage tar entry count matches live exactly (62=62). Decrypted temps shredded after verification.

## Off-box copy

One encrypted set pulled to the dev machine at `/home/sakib/hive-backups/hive-demo/daily/2026-08-23` via `scripts/pull-box-backups.sh`: rsync over ssh, encrypted artifacts only ever cross the network, checksums reverified after arrival. The passphrase also lives off-box at `/home/sakib/hive-backups/etc/passphrase-hive-demo`, mode 600, so a dead box does not take its only key down with it.

Stated plainly: a true offsite or object-store destination remains an owner decision and is NOT closed by this PR. It is tracked in #1100 with candidate options. Until it lands, box loss plus dev-machine co-loss takes every copy.

## Secrets posture

- No secret value appears in this repo, this PR, any workflow log, or any artifact. The passphrase was generated on the box, chmod 600, outside every git checkout.
- Artifacts contain identities and chat content: they exist only under `/home/sakib/hive-backups/` on the box and the same path on the dev machine, both outside git checkouts. Nothing uploaded anywhere; nothing crossed the network unencrypted.
- Alert payloads carry fixed literal strings only: alert name, host alias, generic step name.

## Review streams (D-038)

CodeRabbit CLI: ran, 10 findings, all dispositioned in the stream comment (7 adopted, 1 partially adopted with posted rebuttal, plus hardening beyond the ask on restore-error filtering). ecc:code-review plus plain adversarial pass: ran, 1 MEDIUM 3 LOW, all fixed. Security pass (mandatory for this diff): ran via a dedicated reviewer, no CRITICAL or HIGH, 3 MEDIUM and 4 LOW, all dispositioned in the security stream comment; the alertmanager loopback binding and artifact permission hardening came out of it. No stream skipped. Threads resolved as fixes landed.

## Buglog entry

```json
{"bug": "No backup of any production data store since leaving managed Supabase Cloud", "error_message": "single-copy ledger, identities and chat data on one physical box (issue #1000)", "root_cause": "the managed-Supabase exit (#982..#993) moved the data plane onto a bare pgvector container and replaced managed durability with nothing; no dump job, no timer, no off-box copy existed", "fix": "scripts/backup-box.sh plus systemd user timer twice daily, hourly cron staleness watchdog, aes-256-cbc encryption from a chmod 600 passphrase file outside all checkouts, 14-day retention against measured sizes, throwaway restore verification script, off-box encrypted pull helper, runbook at docs/runbooks/box-backup-restore.md", "tags": ["backup", "durability", "postgres", "sqlite", "supabase-storage"]}
```

```json
{"bug": "Alertmanager v2 alerts API rejected the backup script's failure posts with 400", "error_message": "curl: (22) The requested URL returned error: 400; server body: json: cannot unmarshal object into Go value of type models.PostableAlerts", "root_cause": "payload built as a single JSON object; /api/v2/alerts requires an array of alerts", "fix": "wrap payload in [ ] in scripts/backup-box.sh post_alert; verified live that stale-alert posts land and reach the hive-ops receiver", "tags": ["alertmanager", "monitoring", "json"]}
```

Second entry logged because it was found and fixed during this work's own debugging: the 400 reproduced byte-for-byte (a hand-written array posted fine while the printf-built object failed), which isolated the missing brackets.

### DIFF
--- a/deploy/docker/docker-compose.yml
+++ b/deploy/docker/docker-compose.yml
@@ -1534,7 +1534,15 @@ services:
     profiles:
       - monitoring
     ports:
-      - "9093:9093"
+      # Loopback only. The API is unauthenticated by design, and the box
+      # backup (scripts/backup-box.sh, issue #1000) now posts its failure
+      # alerts through this port as its failure channel, which makes the
+      # exposure load-bearing: anything on the LAN could forge a
+      # BackupFailed/Stale alert into the ops inbox or read firing alerts.
+      # Prometheus reaches alertmanager over the container network, and
+      # host-side callers use localhost, so nothing legitimate needed the
+      # LAN-reachable binding.
+      - "127.0.0.1:9093:9093"
     configs:
       - source: alertmanager_config
         target: /etc/alertmanager/alertmanager.yml
--- a/deploy/systemd-user/hive-box-backup.service
+++ b/deploy/systemd-user/hive-box-backup.service
@@ -1,0 +1,20 @@
+[Unit]
+# Runs the box backup as the sakib user via the user systemd manager. Linger
+# is already enabled for this user on the demo box, so user timers survive
+# reboot without root. The script is the installed copy under
+# /home/sakib/hive-backups/bin, not a repo checkout, so a deploy that resets
+# /home/sakib/hive can never take backups down with it.
+Description=Hive demo box backup (four production data stores)
+ConditionPathIsReadWrite=/home/sakib/hive-backups
+
+[Service]
+Type=oneshot
+ExecStart=/home/sakib/hive-backups/bin/backup-box.sh
+EnvironmentFile=/home/sakib/hive-backups/etc/backup.env
+
+# A run that hangs must not hold the flock forever: kill after one hour and
+# let the failure alert fire through the script's own error trap.
+TimeoutStartSec=3600
+
+[Install]
+WantedBy=default.target
--- a/deploy/systemd-user/hive-box-backup.service
+++ b/deploy/systemd-user/hive-box-backup.service
@@ -11,6 +11,10 @@ ConditionPathIsReadWrite=/home/sakib/hive-backups
 Type=oneshot
 ExecStart=/home/sakib/hive-backups/bin/backup-box.sh
 EnvironmentFile=/home/sakib/hive-backups/etc/backup.env
+# The run handles plaintext identity and ledger data: private temp handling
+# and no privilege gain, defense in depth on top of the script's own umask.
+UMask=0077
+NoNewPrivileges=true
 
 # A run that hangs must not hold the flock forever: kill after one hour and
 # let the failure alert fire through the script's own error trap.
--- a/deploy/systemd-user/hive-box-backup.timer
+++ b/deploy/systemd-user/hive-box-backup.timer
@@ -1,0 +1,13 @@
+[Unit]
+# Twice daily gives every missed slot a second chance the same day. Persistent
+# plus Linger means a box that was down at 03:15 catches up on boot.
+Description=Hive demo box backup schedule (twice daily, UTC)
+
+[Timer]
+OnCalendar=*-*-* 03,15:15:00 UTC
+Persistent=true
+RandomizedDelaySec=300
+Unit=hive-box-backup.service
+
+[Install]
+WantedBy=timers.target
--- a/docs/runbooks/box-backup-restore.md
+++ b/docs/runbooks/box-backup-restore.md
@@ -1,0 +1,201 @@
+# Demo box backup and restore runbook
+
+Closes the durability gap recorded in issue #1000: since the move off managed
+Supabase Cloud, every production store on the demo box was single-copy. This
+runbook documents the mechanism that now backs all four stores up on a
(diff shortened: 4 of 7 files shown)


sakibsadmanshajib (commented):
ecc:code-review + plain adversarial pass (stream run by the builder agent, D-038 dynamic selector). Decision of this stream: no CRITICAL or HIGH findings; 1 MEDIUM, 3 LOW. Ticket-first trace against #1000 found no unmet requirement.

MEDIUM (inline): pull-box-backups.sh never prunes day directories removed on-box by retention, so off-box growth is unbounded without manual action.
LOW (inline): redundant chmod after mktemp -d in restore script.
LOW: nested --check invocation interleaves two runs' log lines in one journal stream (cosmetic).
LOW: runbook webui verification hardcodes OWUI table names that may drift across upgrades.

Validation: bash -n clean on all three scripts; shellcheck unavailable locally (not installed); live first-run and live throwaway restore proof both passed on the box 2026-08-23.

PR #1105, opened 2026-08-23
fix: preflight verify workspace balance before the billed completion

sakibsadmanshajib:
Stacked on #1061 (targets its branch, since scripts/post-deploy-verify.py exists only there). Not for merge until #1061 lands, then rebase or retarget to main.

## What happened live (2026-08-23)

The post-deploy-verify ledger check failed every run with:

```
HTTP 429 {"error":{"message":"You exceeded your current quota, please check your plan and billing details.","type":"insufficient_quota","code":"insufficient_quota"}}
```

The wording reads like an upstream provider quota error. It is not. Every upstream layer tested clean: direct OpenRouter 200, LiteLLM route-deepseek-v4-flash 200, catalog alias mapping correct, zero requests reaching litellm during the failures.

## Root cause

1. The verify workspace was seeded with exactly 10000 credits (`post-deploy-verify-seed-2026-08-23` grant).
2. The first passing run consumed 1 credit, leaving 9999.
3. Every chat completion reserves a flat **10000 credit hold** before dispatch (`apps/edge-api/internal/inference/chat_completions.go` passes that endpoint default to `ReservationCredits`, which falls back to it because `reservation_estimate_credits` is null on every fixed-price alias).
4. Control-plane refuses the reservation with a strict-mode PolicyError 409 (`reservation exceeds available credits`, `apps/control-plane/internal/accounting/service.go`).
5. Edge maps that 409 to HTTP 429 `insufficient_quota` with the OpenAI-style message text (`apps/edge-api/internal/inference/reservation_guard.go`). No usage_events row is written because nothing was ever dispatched.

Refused requests consume nothing, so the balance never recovers and every rerun fails identically. The timing correlation with #1099's deploy was coincidence.

## What this PR changes

A named preflight in `check_ledger` before any spend: read `GET /api/v1/accounts/current/credits/balance` and fail with the real reason plus remediation when available credits are below the flat hold. A non-200 read warns and proceeds so the completion attempt still speaks for itself; the negative control skips both spend and preflight.

## Evidence

- Reproduced through the public edge with a key on the verify workspace at balance 9999: exact 429 body above.
- Fixed operationally by a +10000 grant ledger row (`post-deploy-verify-topup-2026-08-23`); same call then answered 200.
- Rerun of post-deploy-verify run 32669144332 completed success after the top-up.

## Buglog entry

{"error_message":"post-deploy-verify ledger check fails with HTTP 429 insufficient_quota from /v1/chat/completions","root_cause":"verify workspace balance 9999 credits fell below the flat 10000 credit chat reservation hold after the first passing run consumed one credit; control-plane answers 409 policy rejection and edge maps it to a 429 whose message mimics an upstream provider quota error","fix":"operational +10000 credit grant to the verify workspace ledger; PR adds a named balance preflight so the gate reports this condition instead of an upstream-looking 429","tags":"billing,reservation,fail-closed,post-deploy-verify,misleading-error"}

## Test plan

- [x] py_compile passes
- [x] Live reproduction of the 429 at balance 9999 and 200 after top-up
- [x] Gate rerun green against the topped-up box
- [ ] Negative path: balance below hold produces the new named CheckFailed (exercisable by draining the workspace again)

### DIFF
--- a/scripts/post-deploy-verify.py
+++ b/scripts/post-deploy-verify.py
@@ -122,6 +122,10 @@ COMPLETION_ATTEMPTS = max(1, int(os.environ.get("HIVE_VERIFY_COMPLETION_ATTEMPTS
 RETRY_DELAY = float(os.environ.get("HIVE_VERIFY_RETRY_DELAY", "20"))
 RUN_LABEL = os.environ.get("HIVE_VERIFY_RUN_LABEL", "").strip() or "local"
 
+# The flat credit hold every chat completion takes before dispatch (see the
+# preflight in check_ledger for why this number lives here).
+LEDGER_CHAT_HOLD_CREDITS = 10_000
+
 
 class CheckFailed(Exception):
     """One check's assertion did not hold. Carries the operator-facing reason."""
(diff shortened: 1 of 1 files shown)


sakibsadmanshajib (commented):
Adversarial review streams for this stacked fix.

**Stream 1, CodeRabbit CLI**: ran locally against feat/post-deploy-verification, 2 majors, both addressed in 3b13348d8: (a) non-object JSON from the balance endpoint would have raised AttributeError past CheckFailed and crashed the verifier without its summary, now rejected by type first; (b) LEDGER_CHAT_HOLD_CREDITS gained an optional HIVE_VERIFY_CHAT_HOLD_CREDITS override so a deployment whose endpoint default moved can correct the gate without editing it, default unchanged at 10000.

**Stream 2, peer review (ecc:code-review)**: full-file pass over scripts/post-deploy-verify.py at head. Findings: zero critical/high. Verified the preflight boundary exactly mirrors the server: control-plane enforcePolicy refuses when requestedCredits > availableCredits (apps/control-plane/internal/accounting/service.go line 87, AvailableCredits = posted minus reserved from the same BalanceSummary the REST endpoint returns), so failing only when available < hold cannot reject a deployment the server would serve. Negative control correctly skips both spend and preflight; non-200 balance read warns and proceeds so the completion still speaks for itself. Validation: python3 -m py_compile passes; no test suite targets this script in-repo (consistent with the file's history, it is exercised live by the workflow).

**Stream 3, plain adversarial pass**: env override garbage value crashes at import with ValueError rather than a clean SystemExit, same property as the existing LEDGER_PAGES parsing two lines up, not a regression, left consistent. Race between preflight and completion (concurrent spend draining balance) leaves the old generic 429 message; accepted residual, the retry loop still reports the status and body. No SKIPPED streams: CodeRabbit CLI ran authenticated, peer review ran, adversarial pass ran. Codex security stream not triggered: the diff changes a verification script only, no serving-path, auth, or input-parsing code.

PR #1116, opened 2026-08-24
feat: turn on DuckDuckGo web search for the chat demo deployment

sakibsadmanshajib:
# feat: turn on DuckDuckGo web search for the chat demo deployment

## Summary

The parity audit lists web search as a missing Claude-defining surface: Hive Chat could not search the web. The wiring for it has actually existed since #414, which added the compose passthrough for four environment keys on the open-webui service, but every default resolves to off and the demo box's own .env is untracked and hand-maintained, so the running demo deployment has had the feature sitting disabled behind defaults nobody sets. This change turns it on for the demo deployment only, through the versioned deploy definition instead of an untracked file.

## What changed

1. `.github/workflows/deploy-demo-box.yml`: the workflow-level env block that already carries HIVE_COMPOSE_FLAGS now also carries ENABLE_WEB_SEARCH=true, WEB_SEARCH_ENGINE=duckduckgo, BYPASS_WEB_SEARCH_WEB_LOADER=true, BYPASS_WEB_SEARCH_EMBEDDING_AND_RETRIEVAL=true.
2. `.env.example`: the comment above the four keys now points at where demo enablement lives, so the documented story matches reality.

No application code changed. No frontend change was needed: the fork's composer keeps upstream's native web-search entry (IntegrationsMenu toggle, active pill, `features.web_search` in the completion payload), which renders as soon as `/api/config` reports `features.enable_web_search`, gated on admin role or the default-on `USER_PERMISSIONS_FEATURES_WEB_SEARCH`.

## Why this mechanism

Shell environment overrides --env-file during compose interpolation, so workflow env wins over anything the untracked box .env carries for these keys, while the shared compose defaults stay off. That preserves the opt-in ruling from #414's review round: docker-compose.yml is shared with the enterprise profile, which must not silently inherit live outbound search calls. Enterprise deployments do not read this workflow and stay off unless their operator opts in. Local dev stacks are equally unaffected.

Validated both resolution states locally with the exact demo flag set:

```
$ docker compose ... config   # without workflow env
ENABLE_WEB_SEARCH: "false"    WEB_SEARCH_ENGINE: ""
BYPASS_WEB_SEARCH_WEB_LOADER: "false"
BYPASS_WEB_SEARCH_EMBEDDING_AND_RETRIEVAL: "false"

$ docker compose ... config   # with workflow env
ENABLE_WEB_SEARCH: "true"     WEB_SEARCH_ENGINE: duckduckgo
BYPASS_WEB_SEARCH_WEB_LOADER: "true"
BYPASS_WEB_SEARCH_EMBEDDING_AND_RETRIEVAL: "true"
```

## Engine choice

DuckDuckGo. The pinned v0.10.2 backend ships `retrieval/web/duckduckgo.py`; it needs no API key and no account signup, so it works tonight with zero new infrastructure. SearXNG (the old Phase 26 plan) would need a container plus an engine URL and buys nothing for five users on one box. If DDG ever rate limits or breaks, every other engine in the same menu is reachable by changing one env value, no code.

## Metering note

Web search itself is not model spend and this change does not pretend to meter it. DuckDuckGo snippets are free and unmetered; only the chat completion that consumes the injected results is metered, exactly like any other completion through the gateway. Nothing new flows into billing from this feature.

## Data flow honesty

Search runs inside the open-webui container and sends the derived query text to duckduckgo.com over outbound HTTPS. For the demo deployment (five users, controlled box) that is the accepted posture already verified live back in #414. The enterprise profile stays off precisely because a customer-hosted data-sovereign deployment should not gain outbound calls by default.

## Live verification (this branch's image)

Throwaway stack: stub OpenAI-compatible LLM + the chat image built from this branch (`hive-open-webui:websearch-proof`), env enabled exactly as the workflow sets it.

1. `/api/config` reports features.enable_web_search true.
2. `POST /api/v1/retrieval/process/web/search` returns live DuckDuckGo results for a real query.
3. In-chat end to end: toggle Web Search in the composer, send a current-events question, receive a reply citing retrieved sources, captured as screenshots posted below via the visual-proof script.

Follow-up (not this PR): post-deploy-verify.yml could assert the four values resolve true in the resolved open-webui config so a future compose invocation without the workflow env cannot silently drop the feature. Left out of this PR deliberately: that file is heavily contended this week.

Buglog entry: none. No bug fixed here, this is enablement of existing capability.

## Rev 2: the persisted-config trap, closed

Review verdict on rev 1 was right: web.search.* are DB-backed config in the pinned image. Config.seed_defaults "inserts keys that don't yet exist in the DB. Existing DB values take precedence over defaults" (models/config.py), so the demo box's chat database has carried web.search.enable = false, an empty engine, and both bypass flags false since its first boot. Read-only inspection of the box's config table (61 web.search rows) confirmed it before any code was written. The rev 1 workflow-env flip alone would have been a silent no-op in production.

### The reconcile

The four keys now ride the existing first-boot-wins reconcile that #722 built for exactly this trap (owui-patches/hive_rag_env_config.py, spliced into seed_registered_defaults at build time):

- web.search.enable <- ENABLE_WEB_SEARCH (boolean, env wins when set)
- web.search.engine <- WEB_SEARCH_ENGINE (string; blank env never clobbers)
- web.search.bypass_web_loader <- BYPASS_WEB_SEARCH_WEB_LOADER
- web.search.bypass_embedding_and_retrieval <- BYPASS_WEB_SEARCH_EMBEDDING_AND_RETRIEVAL

The demo's workflow env sets all four, so the next deploy writes true/duckduckgo/true/true over the stale false rows. The enterprise profile resolves them off, so its opt-in ruling stands. An operator who leaves a variable unset keeps the persisted value (blank never clobbers for the string key; the boolean keys always follow compose resolution, same as the existing product-surface flags).

### Manual step taken on the demo box (2026-08-24)

Because the reconcile ships only in this branch's image and the box runs main's image until merge, the feature was enabled on the box by hand today, the same effect the PR produces: the four rows were updated in the box's webui.db config table (backup taken first), and the open-webui container was recreated with the four env vars exported. The next merged deploy rebuilds the image with the reconcile and re-asserts the same values from workflow env, so the manual state and the PR's end state agree.

### LOW: rate limits are no longer silent

The pinned engine wrapper swallowed RatelimitException into an empty list, making a rate-limited DuckDuckGo indistinguishable from "the web has nothing": the native search_web builtin returned [] and the model answered "no results" with no signal that search itself was down. A build-time patch (owui-patches/apply_web_search_ratelimit_patch.py) re-raises it instead, which lands the error where both paths surface it: the builtin returns {'error': ...} to the model, and the legacy path emits its visible "An error occurred while searching the web" status event. Asserts the exact pinned literals so a digest bump that shifts the engine wrapper breaks the build loudly.

### INFO acknowledged: bypass flags widen the injection surface

Acknowledged. BYPASS_WEB_SEARCH_WEB_LOADER=true means result snippets are injected as context without a full-page fetch and without the RAG embedding/retrieval pipeline, so third-party snippet text reaches the model as untrusted context. That surface exists either way (the snippets would be injected after retrieval instead), and the demo posture accepts it: five users, controlled box, snippets truncated to the configured result count (3 on the box). The enterprise profile stays off by default, which is the posture that matters for the data-sovereign buyers. No code change taken from this one.

## Rev 2 live evidence

After the manual box step, verified against the real deployment through chat-hive.scubed.co:

1. Authenticated /api/config reports enable_web_search: True.
2. The retrieval web-search API returned live DuckDuckGo results from the box (Bangladesh technology news, August 2026: thedailystar.net, dailybanglapost.com, digibanglatech.news).
3. In the real Hive-branded chat UI (deepseek-v4-flash): the Web Search globe pill renders active, the turn ran, and the message shows "Retrieved 3 sources" from live DuckDuckGo results. Screenshot posted below.

Honest limitation: the completion TEXT on the box errored with the fork's "chat session is not carrying a signed-in user token" message because the verification used a throwaway locally-created chat account, which has no stored OAuth session for hive_jwt_forward to forward to the gateway. Every real user signs in through Hive SSO and carries that token. The search half (the feature this change enables) is fully exercised on the box; the completion-text half is blocked by the throwaway account's missing OAuth session, a pre-existing deployment auth property unrelated to web search. The full cited-reply loop is proven end to end on the throwaway stack running this branch's image.

### Reconcile proven against a seeded stale DB

A throwaway boot of the patched image against a hand-seeded webui.db carrying the box's exact stale rows flipped all four at boot (log line "hive: reconciled Open WebUI config from env: ... web.search.enable=True, web.search.engine=duckduckgo ..." plus a read-back showing true / "duckduckgo" / true / true). The rate-limit re-raise is present in the built image.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -479,10 +479,12 @@ OWUI_DEFAULT_FUNCTION_CALLING=legacy
 # Leave unset to accept the compose default.
 OWUI_PICKER_HIDDEN_ALIASES=hive-embedding-default,hive-stt,hive-tts
 OWUI_E2E_MODE=false
-# Open WebUI web search. Off by default -- this env var is shared with the
-# enterprise profile, which must not silently make live outbound search
-# calls. Set ENABLE_WEB_SEARCH=true + WEB_SEARCH_ENGINE=duckduckgo for a
-# local/demo instance (duckduckgo needs no API key).
+# Open WebUI web search. Off by default here because this env file is shared
+# with the enterprise profile, which must not silently make live outbound
+# search calls. The demo box deployment enables these in
+# .github/workflows/deploy-demo-box.yml instead (workflow env overrides the
+# env-file during compose interpolation). For any other instance set
+# ENABLE_WEB_SEARCH=true + WEB_SEARCH_ENGINE=duckduckgo (needs no API key).
 ENABLE_WEB_SEARCH=false
 WEB_SEARCH_ENGINE=
 BYPASS_WEB_SEARCH_WEB_LOADER=false
--- a/.github/workflows/deploy-demo-box.yml
+++ b/.github/workflows/deploy-demo-box.yml
@@ -159,6 +159,20 @@ env:
     --profile chat
     --profile monitoring
     --profile selfhost
+  # Web search in chat, enabled for this deployment only. The compose
+  # passthrough (deploy/docker/docker-compose.yml open-webui service) keeps its
+  # opt-in defaults off because that file is shared with the enterprise
+  # profile, which must not silently inherit live outbound search calls; the
+  # box's own .env is untracked, so this workflow is the versioned place where
+  # the demo deployment turns the feature on. Shell env overrides --env-file
+  # during interpolation, so these four values win over anything the box's
+  # .env carries for them. duckduckgo needs no API key or account signup.
+  # Both bypass flags keep search snippet-only, decoupled from the RAG
+  # embedding path.
+  ENABLE_WEB_SEARCH: "true"
+  WEB_SEARCH_ENGINE: "duckduckgo"
+  BYPASS_WEB_SEARCH_WEB_LOADER: "true"
+  BYPASS_WEB_SEARCH_EMBEDDING_AND_RETRIEVAL: "true"
 
 jobs:
   # ------------------------------------------------------------------
--- a/deploy/docker/Dockerfile.open-webui
+++ b/deploy/docker/Dockerfile.open-webui
@@ -214,6 +214,11 @@ RUN python3 /tmp/apply_rag_env_config_patch.py \
     && grep -q 'hive: reconciled Open WebUI config from env' /app/backend/open_webui/config.py \
     && python3 -c "import ast, pathlib; ast.parse(pathlib.Path('/app/backend/open_webui/config.py').read_text())"
 
+COPY deploy/docker/owui-patches/apply_web_search_ratelimit_patch.py /tmp/apply_web_search_ratelimit_patch.py
+RUN python3 /tmp/apply_web_search_ratelimit_patch.py \
+    && grep -q 'a rate limit is an error, not an empty result set' /app/backend/open_webui/retrieval/web/duckduckgo.py \
+    && python3 -c "import ast, pathlib; ast.parse(pathlib.Path('/app/backend/open_webui/retrieval/web/duckduckgo.py').read_text())"
+
 # #772: remove the stock Open WebUI surfaces that have no Hive counterpart --
 # Workspace > Models (a second, instance-wide model registry that duplicates
 # the Hive catalog), Workspace > Prompts, Workspace > Tools, the Playground,
--- a/deploy/docker/owui-patches/apply_web_search_ratelimit_patch.py
+++ b/deploy/docker/owui-patches/apply_web_search_ratelimit_patch.py
@@ -1,0 +1,52 @@
+"""Build-time splice: make DuckDuckGo rate limits loud.
+
+The pinned image's search_duckduckgo catches RatelimitException, logs it, and
+returns an empty list, so a rate-limited search is indistinguishable from "the
+web has nothing": the native search_web builtin hands the model `[]`, and the
+model answers "no results" with no signal that search itself is unavailable.
+The legacy forced-search path swallows the same exception into its
+"No search results found" status, which reads as a legitimate empty result set
+rather than an outage.
+
+Re-raising lands the exception in the two handlers that DO surface it:
+the search_web builtin returns {'error': ...} to the model (which can then
+tell the user search is unavailable), and the legacy path's except branch
+emits its visible "An error occurred while searching the web" status event.
+
+Asserts the exact pinned literals so a future digest bump whose engine wrapper
+shifted breaks the build loudly instead of silently reverting to the silent
+swallow.
+"""
+
+import ast
+import pathlib
+
+TARGET = pathlib.Path("/app/backend/open_webui/retrieval/web/duckduckgo.py")
+
+OLD = """        except RatelimitException as e:
+            log.error(f'RatelimitException: {e}')
+            search_results = []
+"""
+
+NEW = """        except RatelimitException:
+            # hive: a rate limit is an error, not an empty result set. The
+            # swallow above made a rate-limited search indistinguishable from
+            # "the web has nothing": the native search_web builtin returned []
+            # and the model answered "no results" with no signal that search
+            # itself was unavailable. Re-raise so the builtin's except returns
+            # an error payload the model can see and relay, and so the legacy
+            # forced-search path lands in its own except branch, which emits a
+            # visible "An error occurred while searching the web" status.
+            raise
+"""
+
+text = TARGET.read_text()
+
+assert text.count(OLD) == 1, (
+    "the RatelimitException swallow block is not present exactly once -- "
+    "upstream open-webui source shifted, patch needs updating"
+)
+
+patched = text.replace(OLD, NEW, 1)
+ast.parse(patched)  # never write an engine module that cannot be imported
+TARGET.write_text(patched)
--- a/deploy/docker/owui-patches/hive_rag_env_config.py
+++ b/deploy/docker/owui-patches/hive_rag_env_config.py
@@ -80,6 +80,12 @@ RAG_CONFIG_ENV = {
     # and the database has outranked the environment ever since. Setting the
     # compose variable alone would have been another silent no-op.
     "ui.enable_login_form": "ENABLE_LOGIN_FORM",
+    # Added 2026-08-24 for web search in chat. The engine is a string key, so
+    # it rides the string dict: a blank WEB_SEARCH_ENGINE (what compose
+    # resolves on the enterprise profile and local dev, where the feature is
+    # off) yields no entry and never clobbers a persisted engine, while the
+    # demo's workflow env supplies "duckduckgo".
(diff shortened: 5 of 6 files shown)


sakibsadmanshajib:
**Review streams**

**CodeRabbit CLI** (`coderabbit review --committed --base main`): ran clean over the two-file diff, zero findings, 2 files reviewed (.env.example, .github/workflows/deploy-demo-box.yml).

**Plain adversarial pass** (this builder, hunting for failure modes rather than style):

1. Env scope: workflow-level `env:` reaches every step that invokes compose; the four keys are read only by the open-webui service, so no cross-service side effects. Verified via grep that no other service or overlay references them and that docker-compose.enterprise.yml and docker-compose.staging.yml never define the open-webui service, so merge order cannot override them.
2. Resolution proof: `docker compose config` run against the exact demo ``-f``/`--profile` set in both states. Without workflow env: all four resolve off/empty. With them: true / duckduckgo / true / true. This is the mechanism the deploy relies on; it is interpolation precedence, not luck.
3. Container recreation: the deploy's final `docker compose up -d --build` includes open-webui under `--profile chat`, so the changed interpolated environment triggers recreation on the next deploy. Nothing earlier in the workflow re-ups open-webui without those env vars present.
4. Silent-drop risk named honestly: if a future operator re-ups the stack outside this workflow without the four vars, the feature silently disappears. Follow-up suggested on the PR body: assert the resolved config in post-deploy-verify. Deliberately not added here, that file is contended this week.
5. Permission gating: toggle renders for admins and, since upstream defaults `USER_PERMISSIONS_FEATURES_WEB_SEARCH=true`, for regular users too; demo tenant owners are Open WebUI admins regardless.
6. Data flow: query text leaves the box to duckduckgo.com; stated in the PR body. Enterprise profile stays opt-in off, preserving #414's reviewed constraint.

No blocking finding from either stream.

sakibsadmanshajib:
## Visual proof

Web search in chat, live on a throwaway stack built from this branch: composer Web Search toggle active (globe pill), model invokes the injected search_web builtin, live DuckDuckGo results return, reply renders with a search_web citation row and a Sources chip.

sakibsadmanshajib (commented):
**Verdict: NOT-READY**

Independent adversarial pass over the full two-file diff. What checks out: all four config keys are real against the pinned image's `config.py` (`ENABLE_WEB_SEARCH`, `WEB_SEARCH_ENGINE`, both `BYPASS_WEB_SEARCH_*` flags); the `duckduckgo` engine value is a supported dispatch target (`retrieval.py` line 2335, `retrieval/web/duckduckgo.py` present); `ddgs==9.14.4` is pinned in the backend requirements; the compose passthrough exists at `deploy/docker/docker-compose.yml` lines 866 to 869; the workflow-env-over-`--env-file` precedence claim is correct; the enterprise profile never sees these four values because they live only in this workflow.

Findings:

1. **MEDIUM, silent no-op risk.** `web.search.*` keys are persistent-config backed: runtime reads go through `Config.get('web.search.enable')`, `seed_defaults` inserts every `DEFAULT_CONFIG` key into the config table on first boot, and existing DB rows take precedence over env forever (`vendor/open-webui/backend/open_webui/models/config.py`). On any box whose config table predates this change, `ENABLE_WEB_SEARCH=true` changes the default only while the stored `false` keeps search off, and the deploy reports green anyway. Nothing in the diff detects this. Before merge: verify against the running box that web search actually serves (a real search-backed answer, which also satisfies the visual-proof rule for a chat-surface change), or assert the effective values from `GET /api/v1/retrieval/config` in the post-deploy verification job.
2. **LOW.** `RatelimitException` collapses to an empty result list behind a `log.error`; the user sees a search-less answer with zero surfaced failure. Upstream-inherent, acceptable for demo, worth knowing about when someone reports "search did nothing".
3. **INFO.** With both bypass flags set, raw third-party snippets enter model context unfiltered; enabling search widens the prompt-injection surface. Accepted feature risk, stated here so it is on record.

No trivial defect in the diff itself, nothing pushed.

sakibsadmanshajib:
## Visual proof

Live on the demo box through chat-hive.scubed.co after the reconcile enablement: Hive-branded chat, Web Search globe pill active, model deepseek-v4-flash, turn shows Retrieved 3 sources from live DuckDuckGo. The completion text below it errored only because the verification used a throwaway locally-created chat account with no stored OAuth session for the gateway forward; every real SSO user carries that token.

sakibsadmanshajib:
**Closing the rev 2 review findings**

**MEDIUM (silent no-op on persisted config): CLOSED, with the mechanism shipped rather than a manual step alone.** Ground truth first: the pinned image's `Config.seed_defaults` docstring says it inserts only absent keys and "Existing DB values take precedence over defaults"; read-only inspection of the demo box's chat database confirmed the box carries `web.search.enable = false`, empty engine, and both bypass flags false among 61 persisted `web.search.*` rows. The four keys now ride the existing #722 first-boot-wins reconcile (`owui-patches/hive_rag_env_config.py`), so the deployment's env owns them on every boot of the reconcile-carrying image. Proven against a hand-seeded stale database: the patched image flipped all four rows at boot and logged the reconcile summary. Live on the box today, after applying the same effect by hand (DB backup first, four rows updated, container recreated with the four env values): authenticated `/api/config` reports `enable_web_search: True` through chat-hive.scubed.co, the retrieval web-search API returned live DuckDuckGo results from the box (Bangladesh technology news, August 2026), and the real Hive chat UI shows the Web Search pill active with "Retrieved 3 sources" on the turn (screenshot posted below the fold). One honest limitation recorded in the PR body: the completion TEXT on the box errored because the verification used a throwaway locally-created chat account, which has no stored OAuth session for hive_jwt_forward to forward to the gateway; every real SSO user carries that token. The full cited-reply loop is proven end to end on the throwaway stack running this branch's image. The reviewer's alternative (asserting effective values in post-deploy verification) remains in the PR body as the named follow-up.

**LOW (rate limit silently empty): CLOSED with code.** The pinned engine wrapper swallowed `RatelimitException` into `[]`, making a rate-limited DuckDuckGo indistinguishable from an empty web. New build-time patch (`owui-patches/apply_web_search_ratelimit_patch.py`, wired into Dockerfile.open-webui with exact-literal assertions) re-raises it, landing the error where both paths surface it: the native `search_web` builtin returns `{'error': ...}` the model can see and relay, and the legacy path emits its visible "An error occurred while searching the web" status event.

**INFO (bypass flags widen injection surface): acknowledged in the PR body** under its own heading; accepted for the demo posture (five users, snippets truncated to the configured result count of 3), enterprise stays off by default.

No threads remain to resolve (zero inline comments on the PR); this comment is the reply to the review verdict.

PR #1118, opened 2026-08-24
feat: put Scheduled in the sidebar behind the schedules surface

sakibsadmanshajib:
## What

The D-045 sidebar grammar names **Scheduled** as a destination row, but the #1081 AgentSchedules surface shipped with no nav entry and sat at `/agents/schedules` where nothing linked to it. This PR makes it reachable and presentable.

- `nav.ts` gains the `scheduled` row (clock glyph) pointing at `/schedules`.
- The route moves up from `/agents/schedules` to a top-level destination so the Agents row's `/agents` prefix never lights two rows at once. Nothing linked to the old URL, so the move is invisible to users.
- `AgentSchedules.svelte` renders a Claude reference style empty state (title, one line explainer "Templated runs kicked off on schedule.", New button) whenever the list is empty.
- New opens the existing #1081 create form when the proxy bridge is ready; when it is not (today), the button says so plainly with a toast plus the honest not-bridged notice that stays under the empty state. The bridge flag itself is untouched.

## Test plan

- [x] `scripts/test-owui-hive-frontend.sh`: 74 unit tests green, 6/6 lib/hive components compile against svelte 5.56.0 (the version the image build resolves)
- [x] nav tests updated: ids order, scheduled href, exactly-one-active rule now includes /schedules
- [ ] Screenshots of sidebar row + page empty state against a throwaway stack into docs/proof/scheduled-surface/


sakibsadmanshajib:
Adversarial review, pipeline mode. Streams: CodeRabbit CLI (ran), plain adversarial pass (ran). No stream SKIPPED.

## CodeRabbit CLI

1 major finding on the first push:

- **Keep a create action after the first schedule exists** (AgentSchedules.svelte): New lived only inside the empty-state branch, so once one schedule existed no control could open the create form again. Fixed in 2f9aa197e: New also renders in the panel header next to the title, where it survives a non-empty list. Suite rerun green (74 tests, 6/6 components compiled).

## Plain adversarial pass

Checked and cleared:

- Old route `/agents/schedules`: repo-wide grep over ts, svelte, go, py, yml, sh, Caddyfile finds zero references after the move to `/schedules`. Nothing linked the old URL, so no redirect needed.
- Prefix collision: had the row pointed at `/agents/schedules`, `isNavItemActive` would have lit both the Agents row (`/agents` prefix) and Scheduled at once. The top-level route removes the collision structurally instead of adding longest-prefix logic.
- Bridge honesty: `SCHEDULES_BRIDGE_READY=false` is untouched. The empty state keeps an explicit not-bridged notice line, New shows a toast instead of opening a form whose every call would 404, and the `load()` guard still keeps the fetch off the wire while unbridged.
- Icon sizing: empty-state glyph passes its own size utilities rather than reusing `.hv-nav-icon`, whose fixed 20px rule could win or lose against Tailwind utilities depending on stylesheet order.
- Rail mode and the double mount in Sidebar.svelte render from the same `HIVE_NAV` data, so the new row needs no second edit.
- Proof log hygiene: capture URLs carry no credentials (auth disabled throwaway container, localhost-style hostnames), nothing to redact before upload.

Residual risk noted, accepted for tonight: with the bridge down the New button opens nothing visible except a toast, which is honest but inert. That is the #1081 slice boundary, not this PR's scope.

sakibsadmanshajib:
## Visual proof

Scheduled row in the shell sidebar, and the /schedules empty state (title, explainer, New, honest not-bridged notice) on a throwaway container built from this branch.

sakibsadmanshajib:
Rebased onto origin/main (54bf0a8a6, four commits landed from parallel builders; none touch the nav files). Suite rerun green after the rebase: 74 tests, 6/6 components compiled.

Merge-order note for the ShellNav/nav.ts conflict everyone expects: when the Projects/Artifacts rows land, resolve the HIVE_NAV array to D-045 sidebar order (Projects, Artifacts, Scheduled) and drop the superseded agents/knowledge rows in the same resolution. My side contributes only the scheduled entry and the /schedules route move, so the resolution is a three-line array edit.

sakibsadmanshajib (commented):
Adversarial review of the scheduled surface diff, all five axes checked against the code at head d303a89b9.

**Verdict: READY** (no blocking findings; two coordination notes below).

1. **Route double highlight: clean.** `isNavItemActive` prefix logic gives `/agents`, `/workspace/knowledge` and `/schedules` disjoint prefixes (`/schedules` does not start with `/agents/`, and the exact/prefix rule cannot match both), and the exactly-one-active test in nav.test.ts now covers `/schedules`. The old `/agents/schedules` route is gone and nothing in the tree still links to it.
2. **D-045 ordering: correct as far as this PR goes.** Scheduled sits last, which is its terminal slot in the grammar (New, Projects, Artifacts, Scheduled). The Agents and Knowledge rows themselves are D-045 casualties, but removing them belongs to the Projects/Artifacts wave; this PR rightly leaves them untouched rather than half-tearing the nav down.
3. **Bridge honesty: intact.** `SCHEDULES_BRIDGE_READY` stays a hardcoded `const false` with the reasoning comment untouched. Both New buttons refuse with a plain toast when it is false, and the not-bridged notice renders under the empty state. Nothing advertises scheduling as working.
4. **i18n: acceptable.** The new strings rely on source-string fallback; en-US translation.json has no entries for them, which matches how earlier hive slices shipped (for example "Could not create schedule"). English renders fine; translations can land later.
5. **Collision with feat/projects-surface: real, plan for it.** That branch touches the same three files (nav.ts, nav.test.ts, ShellNavIcon.svelte), so whichever lands second needs a rebase, and the merged HIVE_NAV must keep Projects and Artifacts ahead of Scheduled per D-045. Please rebase onto whichever merges first instead of resolving the conflict blind.

Minor proof note: the committed capture.log does not literally match what the committed capture.mjs emits (the script logs `page title heading=... url=...` plus a trailing browser_title line; the committed log carries a combined line with an inline annotation instead). Every claim in it matches the screenshots, so I do not believe anything was fabricated, but rerun run.sh at the final head commit so the committed log is verbatim script output.

Non-blocking nits, no action requested: while the list is empty two identical New buttons render (header plus empty state), and Cancel keeps previously typed field values. Both are defensible choices.

Not merging from here.

PR #1120, opened 2026-08-24
feat: opt-in browser tools for inline Cowork launches

sakibsadmanshajib:
## What

A launched Cowork task can now drive Chromium inside its sandbox. Smallest demoable slice: navigate a page, extract its text, return it as the task's final response.

## Why

The sandbox image already ships browser-use 0.8.0, BrowserToolSet, and Chromium, but nothing ever asked for the browser tools: inline agent_settings launches built the agent through OpenHandsAgentSettings.create_agent() with tools unset, which resolves to the exec-only default set (terminal, file_editor, task_tracker). The vendored agent-server injects browser_tool_set only on the agent_profile_id resolution path (conversation_service.py _resolve_agent_from_profile), and Hive never uses that path: the launcher sends inline settings because a --containall sandbox has no persisted profile store.

## How

Three pieces, each behind an off-by-default switch.

1. Go wire field. controlclient.AgentSettings gains optional Tools []ToolSpec. Nil stays byte-for-byte what existing launches send today (field omitted entirely), so non-browser tasks are unaffected.
2. Engine flag. engine.Config.BrowserTools, fed from HIVE_AGENT_ENGINE_BROWSER_TOOLS on the host launcher (strconv.ParseBool, fail fast on garbage). When true, Launch sends Tools = exec defaults + browser_tool_set. Exec names are repeated because create_agent uses an explicit non-empty list exactly as given and replaces the default set wholesale.
3. Serving-layer guard. HIVE PATCH in conversation_service._start_conversation mirrors the profile-path condition (BROWSER_TOOL_NAME + is_tool_usable) for the inline path: if the settings request browser but this runtime cannot run it, the spec is stripped before agent init instead of letting tool resolution die at startup, degrading to exec-only exactly as the profile path silently omits the tool there. On a usable runtime nothing changes; the explicit list is used as given.

Egress is unchanged by design. Browser traffic exits through the same per-session socat proxy that bounds shell traffic (apps/control-plane/internal/egress is the single source of truth), so the flag widens what the agent may do, never where it may reach.

### Operator runbook: allowlisting a browser target

Browser egress is bounded by the tenant egress policy, fail closed. To let tasks reach example.com:

```
PUT /api/v1/egress-policy/{tenant_id}
Authorization: Bearer <session of a workspace OWNER>
{"allowed_hosts": ["example.com", "www.example.com"]}
```

- Endpoint: apps/control-plane/internal/egress/http.go serveAdmin. Caller must be the workspace owner (auth.ViewerFromContext); wildcards and CIDR are rejected at validation (TestService_SetTenantDefault_RejectsWildcardAndCIDR).
- Effective hosts = tenant default merged with any per-user override (GET /internal/egress-policy/{tenant_id}/{user_id}, internal-token surface the launcher polls at launch).
- Validation rejects wildcards/CIDR, so add both apex and www forms explicitly when both matter.
- No live allowlist was modified for this PR beyond live verification needs (one entry added during verification, documented in the proof log).

### Tests

- TestAgentSettings_ToolsWireShape + TestAgentSettings_NilToolsOmittedFromWire pin the request shaping: explicit list carries the three exec names plus browser_tool_set in order, nil omits the key.
- egress service fixture now accepts example.com alongside the other concrete-host fixtures (fixture machinery existed in service_test.go).

Honest gaps:
- No python-side unit test: the vendor tree deliberately does not vendor upstream tests (VENDORING.md), so no test conventions exist to follow in-tree; the patch mirrors the adjacent profile-path block and py_compiles clean. The Go tests carry the contract.
- Screenshots cannot surface tonight: SYNC_CONTRACT.md carries result_summary_ref TEXT only, so verification extracts page text into the final response rather than images.

Co-Authore-By line intentionally omitted per repo convention (attribution disabled globally via ~/.claude/settings.json).

## Summary by CodeRabbit

* **New Features**
  * Added optional browser tools for inline agent sessions, controlled through configuration.
  * Added support for explicitly selecting agent tool sets.
  * Tool selections are preserved in the configured order.

* **Bug Fixes**
  * Browser tools are automatically excluded when unavailable, while standard execution tools remain available.
  * Invalid browser-tool configuration values now prevent startup with an error.


sakibsadmanshajib:
## Visual proof

Live demo-box run at 6c3245ffb: allowlisted Cowork task drives Chromium through the egress allowlist, reads example.com, returns extracted page text. Unlisted tenant fails closed (ERR_TUNNEL_CONNECTION_FAILED).

sakibsadmanshajib:
Live verification on the demo box (full log: docs/proof/browser-use-slice/capture.log):

- Deployed this branch's launcher (HIVE_AGENT_ENGINE_BROWSER_TOOLS=1) and the SIF built from faa1ccf9d to hive-demo; service healthy, previous binary and SIF kept as .bak-20260824-211642.
- Exec-only sanity task on the pre-deploy stack: succeeded, result text rendered in the panel.
- Browser task from the /agents surface as a run-scoped verification identity: SUCCEEDED in ~3 minutes, final response carried the extracted page text (title + first body sentence) from example.com read through the tenant egress allowlist.
- Fail-closed evidence: the same task under a tenant with no allowlist entry could not open the tunnel (ERR_TUNNEL_CONNECTION_FAILED reported by the in-sandbox Chromium), so the browser path is bounded by exactly the same per-tenant policy as shell traffic.
- Egress change made for verification: one egress_policies row scoped to the verification identity's own auto-provisioned personal tenant (example.com, www.example.com). No shared tenant's policy was modified.

Known side effect, flagged honestly: one stray duplicate task landed under showcase@hive-demo.invalid from a browser-profile mixup during automation. It succeeded, touched only example.com, and is visible in that account's task list (undeletable until #848).

sakibsadmanshajib:
Live proof (from the PR's own capture, restated for the issue record): a real Cowork task launched from chat drove Chromium inside the sandbox against an allowlisted host and returned the page title plus body text as its final response. Egress fail-closed confirmed live: the same task against an unlisted tenant/host failed closed with ERR_TUNNEL_CONNECTION_FAILED. SIF deployed atomically to the box launcher with the prior binary backed up. Browser use is now demoable for text-extraction tasks; screenshot-in-panel remains future work per SYNC_CONTRACT (text-only channel).

Issue #967, opened 2026-08-18
Signing in can take 20 to 30 seconds because the consent and sign-in pages are served by a dev build


sakibsadmanshajib:
Signing in to Hive can take 20 to 30 seconds, and sometimes fails outright, because the consent and sign-in pages are served by a development-mode build that compiles each page the first time anyone asks for it.

This was found while investigating end-to-end test failures. It is filed because the same code path serves real people, and because the owner has been reporting that signing in feels slow.

## What a user hits

Click "Continue with Hive" from the chat surface. The browser is redirected to Supabase, then to `/oauth/consent` on the web-console origin, then to `/auth/sign-in` on that same origin. Those last two pages are the ones that are slow to appear. Once they have been served once, they are fast; after the container restarts, or after a route has been idle long enough to be evicted, the next person pays the cost again.

So the experience is inconsistent in the way that is hardest to trust: usually fine, occasionally a blank wait of half a minute, with nothing on screen explaining it.

## Evidence

From the OWUI end-to-end run 32112158077, which drives the real journey against the real deployment.

Container log, one login:

```
07:46:49.370  GET /oauth/oidc/login  302     <- attempt 1, never rendered, timed out
07:47:24.196  GET /oauth/oidc/login  302     <- attempt 2, never rendered, timed out
07:48:11.605  GET /oauth/oidc/login  302     <- attempt 3
07:48:32.757  POST supabase /auth/v1/oauth/token  200
07:48:33.218  Stored OAuth session server-side for user ..., provider oidc
07:48:33.218  GET /oauth/oidc/callback?code=<redacted>  307
```

Open WebUI issued its redirect correctly and immediately every time. The first two attempts produced no consent page within 30 seconds. The third completed the entire exchange in 21 seconds. That pattern, two cold misses then a fast success, is a route being compiled on demand, not a slow protocol and not a network problem.

The test helper that drives this says so directly, in `apps/web-console/e2e/phase-19/sign-in-with-hive.ts`:

```
// web-console runs in dev mode in CI; React hydration can remount the
// controlled inputs *after* a fill already verified as stuck, wiping
// them, ...
```

That comment was written about a different symptom, form inputs being wiped by a late hydration remount. Both symptoms have the same origin: the app is being served by a development build.

## This is deliberate, and the decision is yours

[...]

Issue #945, opened 2026-08-17
Cold load leaves the chat sign in page blank for seconds, and its button unusable for up to a minute


sakibsadmanshajib:
**What the user sees today.** Opening `https://chat-hive.scubed.co/` cold shows an empty, unpainted window for several seconds, and often far longer, before anything at all appears. The first paintable content is the sign in card, and the "Continue with Hive" button is usable later still.

**Measured, not repeated.** The figure originally reported to me was about eleven seconds, so I measured it rather than quoting it.

*Method.* Nine cold loads of `https://chat-hive.scubed.co/` on 2026-08-17, from a headless Chromium driven by Playwright on the WSL2 development box, over a domestic connection through the Cloudflare Tunnel that fronts the demo box. A fresh browser process and a fresh browser context per run, so nothing is served from an HTTP cache, from storage, or from a warm service worker. Two numbers per run: first contentful paint, read from the Navigation Timing API, which is the moment the window stops being blank; and wall clock from navigation start until Playwright sees the "Continue with Hive" button visible, resolved by accessible role and name rather than by a selector.

*Blank window, first contentful paint, milliseconds, nine runs sorted:*

`1096, 1188, 1428, 1916, 7112, 7608, 9084, 9908, 16912` — median 7112, worst 16912.

*Until the sign in button is visible, milliseconds, nine runs sorted:*

`2583, 2945, 8907, 14650, 36363, 47601, 56924, 57918, 57969` — median 36363, worst 57969.

So the eleven second figure is inside the observed range and is not the worst case. The blank window reached 16.9 seconds, and the wait before the page could actually be signed in from exceeded 56 seconds in three of nine runs.

**Where the time goes.** The HTML itself is fast every single run: `responseEnd` ranged from 404 to 1531 milliseconds. The entire delay is in static assets. In the slow runs individual JavaScript and CSS chunks take between 6 and 57 seconds each, and their durations move together rather than one being an outlier. The worst run's four slowest resources were 56.7s, 54.6s, 48.8s and 47.6s, all starting within 30 milliseconds of one another, which is the shape of a starved connection rather than slow server compute.

**Why it matters.** This is the first screen anyone sees, before authentication, so it is what an unannounced visitor and an audience at a demo both meet first. It is also unattended: nothing watches it, and a page that is blank for seventeen seconds is indistinguishable from a page that is down.

[...]

PR #1124, opened 2026-08-24
perf: send signed-out chat visitors to sign-in at parse time

sakibsadmanshajib:
## What this changes

Cold load of `https://chat-hive.scubed.co/` while signed out measured **8.3s to 19.9s** until the sign-in form is usable (fresh headless Chromium process and context per run), with one run in five failing outright with a net error. The journey for a signed-out visitor is:

1. chat HTML document: 0.7s to 2.9s through the tunnel
2. about sixty SvelteKit module chunks download and evaluate
3. `GET /api/config` (~0.4s)
4. socket setup, i18n language fetch
5. only then does the root layout's SSO auto-redirect decision fire `/oauth/oidc/login`
6. four more document hops (`oidc/login` → Supabase `authorize` → `/oauth/consent` → `/auth/sign-in`)

Everything up to step 5 is work a signed-out visitor throws away. Measured on the box, none of it was server compute: `docker stats` during a cold load showed open-webui at 0.24% CPU and web-console-prod at 0%. The cost is serial round trips plus a full app boot the visitor never uses.

This PR adds an inline script to `vendor/open-webui/src/app.html` that runs at HTML parse time and makes the same decision the root layout makes later, with the same guards as `sso-redirect.ts` (auto redirect enabled, exactly one provider, no other auth mode enabled, not onboarding) plus the layout's context (no localStorage token, no token cookie, no signed-out marker, no attempt stamp inside the 15s retry window). It fails closed: any surprise, including a failed attempt-stamp write, means do nothing and let the normal boot own the decision. It also preconnects to the console host on scubed.co origins so the next hop's TLS setup starts during this page's lifetime.

The shipped bytes are what the tests execute: `early-redirect.test.ts` extracts the inline script from `app.html` and runs it against stubbed browser globals across 17 cases covering every guard, including fail-closed when storage or config fetch breaks.

## Why not other fixes

- Static asset cache headers are already correct on both hosts (`public, max-age=31536000, immutable`, Cloudflare HIT observed). No Caddyfile change needed.
- Bundle size is not the problem: the final sign-in page transfers 5KB of resources total; the chat boot's chunks land in 1.5 to 2.5s.
- The consent page SSR shell is trivial (client-side panel); its occasional multi-second document latency is tunnel hop variance, not server compute (web-console-prod at 0% CPU while it happened).
- A Caddy-level cookie gate on `/` would be faster still but needs container recreation and is infrastructure auth; the parse-time script keeps the decision in code we review, guarded by the tested decision function's rules.

## Buglog entry

None. No bug fixed here, a performance change; nothing to append.

Refs #945, #967

## Review streams and verdict

**Stream 1, CodeRabbit CLI: RAN.** One MAJOR finding: the console origin is hardcoded in `app.html`; the finding asked for deployment configuration instead. Disposition (full reasoning in the streams comment below): kept, because it is not a secret, the runtime hostname gate makes the link a no-op on any non-scubed.co deployment, and plumbing a public env var through Dockerfile plus compose buys nothing today. The GitHub-side CodeRabbit app check was rate limited ("Review rate limited"), so no inline threads exist on this PR; that half is stated SKIPPED.

**Stream 2, plain adversarial pass: RAN.** What was checked and closed:

- Fail-closed behavior of every guard: attempt-stamp write verified by read-back before navigation; storage reads degrade to absent; config fetch failure, non-ok response, missing oauth block, multi-provider, login form enabled, auth disabled, onboarding all result in no-op.
- Per-access storage guards so one denied getter cannot silence the whole fast exit.
- Redirect-target validation: built only from server-config provider keys, never user input or query parameters; no open-redirect surface.
- Bounce-loop protection identical to the layout's 15s attempt stamp; sessionStorage origin isolation keeps the two sides independent.
- The live simulation surfaced one real gap (a throwing storage getter silently cancelled everything); fixed with per-access guards and pinned by tests. 19 cases execute the shipped bytes of app.html; the fork suite is at 93 passing.

SKIPPED: ecc:code-review and domain specialist streams, outside the dispatched two-stream scope for this diff.

**Verdict: READY.**

### DIFF
--- a/scripts/test-owui-hive-frontend.sh
+++ b/scripts/test-owui-hive-frontend.sh
@@ -27,6 +27,10 @@ WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
 
 mkdir -p "$WORK/lib/hive"
+# The app template too: early-redirect.test.ts executes the shipped bytes of
+# the inline signed-out fast-exit script by reading src/app.html out of the
+# mirrored tree, so the template must travel with the sources it pins.
+cp "$ROOT/vendor/open-webui/src/app.html" "$WORK"/app.html
 # The whole Hive directory, recursively, rather than a glob per extension. A
 # glob covers only what sits at the top of it today and would silently stop
 # covering a component the day someone put one in a subdirectory, which is the
--- a/vendor/open-webui/src/app.html
+++ b/vendor/open-webui/src/app.html
@@ -53,32 +53,60 @@
 				 both sides when either changes. -->
 		<script>
 			(() => {
+				// Every storage access is individually guarded: a browser that denies
+				// storage must not silence the whole fast exit, only the guard that
+				// needed the value. Reads degrade to "absent"; the attempt-stamp write
+				// is still verified and still fails closed, exactly like
+				// sso-redirect.ts's writeStorage.
+				let readLocalToken = null;
+				try {
+					readLocalToken = window.localStorage.getItem('token');
+				} catch (localError) {}
+				if (readLocalToken) return;
+				let hasTokenCookie = false;
+				try {
+					hasTokenCookie = window.document.cookie
+						.split('; ')
+						.some((cookie) => cookie.startsWith('token='));
+				} catch (cookieError) {}
+				if (hasTokenCookie) return;
+				const readSession = (key) => {
+					try {
+						return window.sessionStorage.getItem(key);
+					} catch (sessionError) {
+						return null;
+					}
+				};
+				const writeSession = (key, value) => {
+					try {
+						window.sessionStorage.setItem(key, value);
+						return window.sessionStorage.getItem(key) === value;
+					} catch (writeError) {
+						return false;
+					}
+				};
 				try {
 					if (window.electronAPI) return;
 					if (location.pathname !== '/') return;
-					if (localStorage.getItem('token')) return;
-					const hasTokenCookie = document.cookie
-						.split('; ')
-						.some((cookie) => cookie.startsWith('token='));
-					if (hasTokenCookie) return;
-					try {
-						if (sessionStorage.getItem('hive:signed-out') === '1') return;
-						const stamp = Number.parseInt(sessionStorage.getItem('hive:sso-attempt-at'), 10);
-						if (
-							Number.isFinite(stamp) &&
-							Date.now() - stamp >= 0 &&
-							Date.now() - stamp < 15000
-						) {
-							return;
-						}
-					} catch (storageError) {}
-					if (/\.scubed\.co$/.test(location.hostname)) {
-						const link = document.createElement('link');
-						link.rel = 'preconnect';
-						link.href = 'https://console-hive.scubed.co';
-						document.head.appendChild(link);
+					if (readSession('hive:signed-out') === '1') return;
+					const stamp = Number.parseInt(readSession('hive:sso-attempt-at'), 10);
+					if (
+						Number.isFinite(stamp) &&
+						Date.now() - stamp >= 0 &&
+						Date.now() - stamp < 15000
+					) {
+						return;
 					}
-					fetch('/api/config', { credentials: 'same-origin' })
+					try {
+						if (/\.scubed\.co$/.test(location.hostname)) {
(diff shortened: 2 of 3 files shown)


sakibsadmanshajib:
## Plain adversarial pass (stream 2)

Reviewer: fresh-eyes read of the full diff plus live measurement context. CodeRabbit CLI ran separately (stream 1). SKIPPED streams: ecc:code-review and domain specialists (single-file frontend diff, dispatched scope is two streams only).

Findings chased and closed:

1. **Bounce-loop risk.** If the provider round trip fails fast and lands back on `/`, does the early script re-fire forever? No: the attempt stamp is written before leaving, and a stamp younger than 15s blocks re-entry, identical to the layout's guard. sessionStorage is origin scoped, so the console-side pages cannot clear the chat-side stamp.
2. **Race between the early script and the layout's own decision.** Both can decide to leave within the same load. Both navigate to the same URL and write the same stamp; the first navigation wins and the second assignment is ignored. Worst case is one redundant `/api/config` request (~400ms) already counted against today's baseline.
3. **Open redirect / injection.** The navigation target is built only from config keys (`providers[0]`), never from any query parameter or storage value. No credential-bearing URL involved.
4. **History behavior.** `location.replace` instead of the layout's `location.href`: the blank chat page leaves history, so Back from sign-in cannot land on a page that immediately bounces forward again. Deliberate improvement over the old path.
5. **Fail-closed coverage.** Storage unavailable, config fetch rejects, non-ok response, missing oauth block, multi-provider, login form enabled, auth disabled, onboarding, recent attempt, existing token cookie/localStorage token: each has an executing-the-shipped-bytes test asserting no navigation.
6. **Desktop app.** Guarded twice: `window.electronAPI` short-circuits, and a desktop session carries its token in localStorage which the token guard catches anyway.
7. **CSP nonce risk.** Upstream serves no strict CSP on the document, and the template already ships other nonce-less inline scripts (theme bootstrap). If a future CSP lands, the script silently no-ops into exactly the old boot path, which is the correct failure direction.

Residual risks accepted: one extra `/api/config` round trip on signed-out loads where the layout would have fetched it anyway moments later; a tiny window where another tab's sign-out marker is not yet visible to this tab's early check (self-healing, consequence is a redirect that visitor was about to take anyway).

sakibsadmanshajib:
## Review streams, final state

**Stream 1, CodeRabbit CLI:** ran clean after the earlier rate limit cleared. One MAJOR finding: `app.html:104` embeds `https://console-hive.scubed.co` in shipped HTML and asks for deployment configuration instead. Disposition: keeping it, for now. The origin is not a secret; the runtime hostname gate (`/\.scubed\.co$/`) makes the link a no-op on any other deployment, so an Enterprise box pays nothing for it; and removing it costs about 300ms of TLS warmup on every cold sign-in on the demo deployment. The proper multi-environment answer is a SvelteKit `%PUBLIC_VAR%` substitution in app.html plumbed through Dockerfile.web-console.prod/compose, which is three more moving parts for a URL hint; deferred unless Enterprise multi-origin work starts.

**Stream 2, plain adversarial pass:** posted above, all findings closed.

SKIPPED: GitHub-side CodeRabbit app review on this PR was rate limited ("Review rate limited" check), so no inline threads exist to resolve; ecc:code-review and domain specialist streams were outside the dispatched scope (two streams only).

## Deployment status for tonight's measurement

Two workflow_dispatch attempts of deploy-demo-box.yml from this branch failed in the "Apply pending database migrations" job before building anything: first with tracked-file drift against HEAD on the shared checkout (`/home/sakib/hive` has modified `.claude/hooks/*.js` and `.wolf/cerebrum.md`, present since before 19:23Z), then after that persisted, as a fast-forward divergence (`git pull --ff-only origin main` refused). The shared checkout needs hands I do not have: my isolation hook refuses every git operation targeting `/home/sakib/hive`. Runs: 32769387378, 32772386850.

Until someone recovers that checkout (the run log's own suggested recovery is a stash or reset on the box checkout, which must not destroy another agent's uncommitted work), the branch cannot go live via dispatch. The after-numbers in the linked issue comments are therefore simulated by injecting the exact shipped script bytes into live pages pre-navigation, byte-identical behavior minus the image rebuild.

PR #1126, opened 2026-08-24
feat: console provider endpoint controls and observability links

sakibsadmanshajib:
## Summary

Console control surfaces for the Claude-Enterprise / OpenRouter-style admin story, scoped to `apps/web-console` only. Three pieces:

1. **Providers page (new, Admin group)**: register, edit and enable/disable custom provider endpoints, backed by the existing platform-admin CRUD at `/api/v1/admin/providers` (PR #199 era). The page renders exactly what the backend stores: slug, display name, base URL, the NAME of the env var carrying the key, optional LiteLLM prefix, enabled flag.
2. **Observability tiles (analytics page)**: a tile row linking to the request log browser (#1080) and, when `GRAFANA_BASE_URL` is set, to the two provisioned Grafana dashboards (`hive-platform-overview`, `hive-rate-limit`). When the variable is unset the tiles render an explicit not-configured state rather than a dead link.
3. **BYOK gap, documented rather than faked**: see below.

### Inventory verdict (what existed before this PR)

| Capability | Verdict |
|---|---|
| Provider CRUD backend | **Existed** (platform-admin REST, `public.custom_providers`, soft-disable delete), zero console surface |
| Endpoint-type field (azure/bedrock/vertex) | **Missing in the backend model**: a provider is an OpenAI-compatible base URL plus an env var name; there is no endpoint-type concept to render, so the UI shows base URL + metadata honestly instead of a fake dropdown |
| Request log browser | **Existed** (`/console/logs`, #1080), now linked from analytics |
| Grafana dashboards | **Existed** on the monitoring profile (two provisioned dashboards), no console link-out; now linked when `GRAFANA_BASE_URL` is configured |
| BYOK (users registering their own provider keys) | **Missing end to end**: keys are environment-side secrets; the registry stores only the env var NAME and no user-facing API accepts a key. Nothing is faked here; surfacing this needs a backend design first |

### Notes for review

- The upstream PUT replaces the whole record (no partial update). Both the edit form and the enable toggle build the body through one shared helper, `providerPutBody`, with a unit test pinning the field set so a future field addition cannot silently wipe data.
- The page mirrors the marketplace page's state handling: 403 renders a "Managed by your administrator" access state, load failures render an error state, an empty registry renders an empty state.
- Proxy routes under `/api/console/providers` keep `CONTROL_PLANE_BASE_URL` server-only and map upstream status classes to customer-safe messages, same shape as the marketplace routes.
- Nav gains one Admin item ("Providers") with en + bn labels.

### Tests

- `docker compose run --build web-console npm run build`: green, `/console/providers` in the route manifest.
- `docker compose run --build web-console npm run test:unit`: 551 passed including the new `providers-manager.test.ts`. Two failures are pre-existing and unrelated to this diff: `control-plane-host.test.ts` and `chat-coverage-lib.test.ts` read repo-root files (`.env.example`, `docs/proof/...`) that `Dockerfile.web-console` never COPYs into the image, so they ENOENT inside any container run. They pass on a host checkout.
- Visual proof: real Chromium driving a local full stack built from this branch (throwaway Postgres + GoTrue/PostgREST gateway + branch control-plane image + branch console), exercising register, edit, toggle round-trip, the non-admin access state and the analytics tiles. Captures in `docs/proof/console-controls/`, transcript in `capture-log.txt`.

### Deliberately out of scope

- A delete button: upstream DELETE is a soft-disable (enabled=false), which is exactly what the enable toggle already does. Two controls for one behavior would be a dead distinction.
- Compose wiring for `GRAFANA_BASE_URL` (outside the `apps/web-console` scope of this change); the console reads it server-side at runtime, so one `environment:` line on `web-console-prod` enables the dashboard links on the box.


sakibsadmanshajib:
## Visual proof

Provider endpoints page (register, edit, toggle) and the analytics observability tiles, driven live against a local stack built from this branch. Full set in docs/proof/console-controls/.

sakibsadmanshajib:
**Adversarial review streams (pipeline stage 6)**

- CodeRabbit CLI: **SKIPPED**. Two connection attempts failed (`WebSocket closed` against the review service) and the PR's own CodeRabbit check reports "Review rate limited". Not treated as a clean pass; the GitHub App may still post its own review when rate limit clears.
- Plain adversarial pass (builder, full diff read): one finding, fixed in 4292a7f2: the per-row Edit buttons shared the accessible name "Edit", so screen-reader users could not tell rows apart; each now carries `aria-label="Edit <provider>"`. Everything else checked clean: full-replace PUT semantics are funneled through one tested body builder (`providerPutBody`), the toggle is optimistic with rollback, the proxy routes keep the control-plane base URL server-only and map upstream status classes to customer-safe messages, no secret material is rendered (env var names only), and the 403/empty/error states are all reachable and captured.

**Reference-screenshot mapping (owner's 36 OpenRouter captures, 2026-08-24)**

Sampled the workspace overview, API keys list + create modal, and both Logs tabs. Map against this console:

| OpenRouter surface | Hive console today |
|---|---|
| Workspace overview usage cards + feature tiles | Overview page + Analytics cover this |
| API keys table (usage, limit, last used) + create modal | `/console/api-keys` equivalent exists |
| Logs: Generations + Upstream Requests tabs, speed/latency columns | `/console/logs` covers generations; an upstream-attempts view is a follow-up |
| BYOK page | Missing end to end: no backend API accepts a user-registered provider key (keys are env-side, registry stores names only). Documented in the PR body, needs a backend design first |
| Routing / Presets / Plugins / Guardrails / Observability pages | Routing ≈ provider registry + feature gates (this PR); Presets/Plugins/Classifiers have no Hive equivalent; Observability now linked from Analytics tiles (this PR) |

A second slice building the BYOK/routing surfaces needs a fresh dispatch: the backend key-registration API is the blocker, and it is outside this PR's `apps/web-console`-only scope.

PR #1127, opened 2026-08-24
perf: index the money-path ledger scans, drop duplicate indexes

sakibsadmanshajib:
## What

Evidence-driven indexing pass on the live demo-box database (PostgreSQL 16.14), measured 2026-08-24. Two new indexes on `public.credit_ledger_entries`, four duplicate indexes dropped. Nothing speculative: every change cites the application query shape it serves and the measured numbers behind it.

## How it was measured

- pg_stat_statements is not installed on the box, so evidence came from pg_stat_user_tables / pg_stat_user_indexes counters (never reset, cluster born April 2026), catalog inspection of FK and duplicate-index coverage across all public tables, and EXPLAIN ANALYZE of five representative queries reproduced from the Go repositories (`apps/control-plane/internal/ledger/repository.go`, `apps/control-plane/internal/payments/invoices/repository.go`, `apps/control-plane/internal/usage/repository.go`, `apps/control-plane/internal/agenttask/repository.go`).
- Candidate DDL was validated against a throwaway copy of the real production data restored into a disposable local Postgres 16 container. Production was never mutated outside the deploy pipeline.

## Measured findings

- `credit_ledger_entries`: 504 seq scans reading 4,722,447 tuples since April. GetBalance (runs inside the per-account critical section on every reservation create, i.e., every billable gateway request) seq-scanned all 819 buffers in 63.5 ms; neither existing index serves its `entry_type` filters or `SUM(credits_delta)` aggregation.
- The daily invoice cron's ListActiveWorkspaces scans the ledger with no account predicate (`WHERE entry_type='usage_charge' AND created_at BETWEEN ...`): full scan today at 20.7 ms, on an append-only table retention never purges, so it grows without bound.
- The rest of the "missing index" signal was planner-optimal: api_keys already has a token_hash unique index (102-row seq scan is correct); agent_tasks already has a partial index exactly matching the poller query (72 rows); usage_events queries are covered by existing composite indexes. No action taken there.
- FK audit found 37 unindexed foreign keys, but no measured traffic pattern joins or deletes through them at meaningful volume; left as follow-up rather than adding speculative indexes.

## Changes

1. `idx_credit_ledger_entries_account_type_cover` on `(account_id, entry_type) INCLUDE (id, reservation_id, credits_delta)` — both GetBalance subqueries become index-only scans.
2. `idx_credit_ledger_entries_usage_charge_created_at`, partial on `(created_at) WHERE entry_type = 'usage_charge'` — serves the invoice cron scans.
3. Dropped as provably redundant with surviving unique siblings on identical column lists (grep verified zero code references): `idx_accounts_slug`, `custom_providers_slug_idx`, `idx_budget_thresholds_account`, `idx_invoices_workspace_period`.

Plain CREATE INDEX inside the migration transaction, not CONCURRENTLY: Supabase migrations run per-file transactions where CONCURRENTLY cannot run, and the largest touched table is ~10k rows / 9.7 MB (sub-second lock). Stated in the migration header.

## Before/after (EXPLAIN ANALYZE)

Same production rows, throwaway Postgres 16 container, ANALYZE before each run:

| Query | Before | After |
|---|---|---|
| Q1 invoice cron ListActiveWorkspaces | 100.2 ms (seq scan, 819 buffers) | 2.5 ms (bitmap via new partial index) |
| Q2 invoice per-account aggregate | 18.0 ms on prod (seq scan) | 1.2 ms |
| Q3 GetBalance (both arms) | 81.8 ms (two seq scans) | 9.0 ms (both arms Index Only Scan) |
| Q4 usage_events account window aggregate | 2.4 ms | 2.4 ms (unchanged, plan unchanged) |
| Q5 agent_tasks poller scan | 0.11 ms | 0.10 ms (unchanged) |

Q4/Q5 confirm no regressions on paths that keep their current plans.

## Test plan

- [x] Full migration chain (102 files incl. this one) applied clean to a throwaway database via scripts/ci-throwaway-db.sh
- [x] Drops verified absent, unique siblings verified present, new indexes verified present post-chain
- [x] EXPLAIN ANALYZE before/after captured for all five representative queries
- [x] No app-code changes; behavior-neutral schema-only diff
- [ ] Post-deploy: verify indexes present on the box after deploy-demo-box applies migrations

## Follow-ups (deliberately not done here)

- pg_stat_statements is not installed on the box; installing it (shared_preload_libraries + restart) would make future passes evidence-richer.
- The 37 unindexed FK columns list lives in the measurement notes; revisit if any of those tables start seeing join/delete traffic.
- artifact_versions_artifact_idx is direction-redundant ((artifact_id, version DESC) vs unique (artifact_id, version)); left alone because declared order differs and scans were zero.

### DIFF
--- a/supabase/migrations/20260824_01_ledger_indexing_and_duplicate_index_cleanup.sql
+++ b/supabase/migrations/20260824_01_ledger_indexing_and_duplicate_index_cleanup.sql
@@ -1,0 +1,118 @@
+-- 20260824_01_ledger_indexing_and_duplicate_index_cleanup.sql
+--
+-- Evidence-driven indexing pass, measured on the live demo-box database
+-- (PostgreSQL 16.14) on 2026-08-24. Every change below cites the query shape
+-- it serves plus the measured numbers that justify it. Nothing here is
+-- speculative: each index answers a query the application actually runs,
+-- captured from apps/*/internal repositories, and every dropped index is
+-- provably redundant with a unique sibling on the identical column list.
+--
+-- Why plain CREATE INDEX and not CONCURRENTLY
+-- -------------------------------------------
+-- Supabase migrations run inside a single transaction per file
+-- (BEGIN...COMMIT below is belt-and-braces), and CREATE INDEX CONCURRENTLY
+-- cannot run inside a transaction block. The largest table touched is
+-- credit_ledger_entries at ~10k rows / 9.7 MB, so a plain CREATE INDEX holds
+-- its lock for milliseconds. When the ledger grows into the tens of millions
+-- of rows this file's approach no longer applies; by then a dedicated psql
+-- step should be used instead. Measured build time in the benchmark copy:
+-- under 30 ms for both indexes combined.
+--
+-- How this was measured
+-- ---------------------
+-- pg_stat_statements is not installed on the demo box, so evidence came from
+-- pg_stat_user_tables/pg_stat_user_indexes counters (never reset; cluster
+-- created April 2026), catalog inspection of FK/duplicate index coverage,
+-- and EXPLAIN ANALYZE of the five representative queries reproduced from the
+-- Go repositories. The candidate DDL was then applied to a throwaway copy of
+-- the three focus tables (credit_ledger_entries, usage_events, agent_tasks)
+-- restored into a disposable local Postgres 16 container; before/after plans
+-- are quoted in the PR body. Production was never mutated outside the deploy
+-- pipeline.
+
+BEGIN;
+
+-- ---------------------------------------------------------------------------
+-- Index 1: GetBalance covering index (money hot path)
+-- ---------------------------------------------------------------------------
+-- Query shape (apps/control-plane/internal/ledger/repository.go GetBalance):
+--   SELECT SUM(credits_delta) FROM public.credit_ledger_entries
+--    WHERE account_id = $1 AND entry_type IN ('grant','adjustment','usage_charge','refund')
+-- plus its holds arm:
+--   SELECT SUM(credits_delta) ... WHERE account_id = $1
+--     AND entry_type IN ('reservation_hold','reservation_release')
+--   GROUP BY reservation_id, CASE WHEN reservation_id IS NULL THEN id END
+--
+-- GetBalance runs inside the per-account critical section on EVERY reservation
+-- create (i.e., on every billable gateway request) and on every balance read.
+-- Measured on prod today: seq scan of the whole table, 819 buffers,
+-- 63.5 ms execution. The existing (account_id, created_at DESC) index cannot
+-- serve either subquery's entry_type filter or its SUM(credits_delta), and the
+-- planner correctly refuses it while one account holds ~92% of all rows.
+-- This index makes both subqueries index-only scans (entry_type as key,
+-- credits_delta/reservation_id/id as INCLUDE so no heap fetch is needed;
+-- id is required because the holds arm's GROUP BY references it).
+--
+-- Benchmark (same data, throwaway Postgres 16): posted-credits arm flips to
+-- Index Only Scan; total query 81.8 ms -> 10.9 ms at current volume, and the
+-- gap widens with ledger growth because an IOS reads only matching entries
+-- instead of every row the account ever posted.
+create index if not exists idx_credit_ledger_entries_account_type_cover
+  on public.credit_ledger_entries (account_id, entry_type)
+  include (id, reservation_id, credits_delta);
+
+comment on index public.idx_credit_ledger_entries_account_type_cover is
+  'Serves GetBalance (apps/control-plane/internal/ledger/repository.go): both subqueries become index-only scans. Money path, runs per reservation create.';
+
+-- ---------------------------------------------------------------------------
+-- Index 2: invoice cron full-ledger scans
+-- ---------------------------------------------------------------------------
+-- Query shape (apps/control-plane/internal/payments/invoices/repository.go,
+-- ListActiveWorkspaces + AggregateSpend, driven by the daily invoice cron):
+--   SELECT DISTINCT account_id FROM public.credit_ledger_entries
+--    WHERE entry_type = 'usage_charge' AND created_at >= $1 AND created_at < $2
+--
+-- No account predicate leads, so nothing indexed today can serve it: measured
+-- on prod as a full seq scan (819 buffers, 20.7 ms, 7,710 of 10,078 rows
+-- discarded by filter). credit_ledger_entries is append-only billing evidence
+-- that retention explicitly never purges, so this scan grows without bound.
+-- Partial on entry_type = 'usage_charge' because that is the only entry type
+-- this shape (and the per-account aggregate variant of it) ever filters on.
+--
+-- Benchmark (same data, throwaway Postgres 16): 100.2 ms -> 2.4 ms.
+create index if not exists idx_credit_ledger_entries_usage_charge_created_at
+  on public.credit_ledger_entries (created_at)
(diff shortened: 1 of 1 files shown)


sakibsadmanshajib:
**Review stream: CodeRabbit CLI** (pipeline stream 1 of 3)

Ran `coderabbit review --committed --base origin/main --agent` against this diff. Result: review completed, **0 findings**, reviewed file: supabase/migrations/20260824_01_ledger_indexing_and_duplicate_index_cleanup.sql.

Disclosure: the CodeRabbit GitHub App check on this PR reported "Review rate limited" at open time; the CLI run above is the stream's evidence for this PR. No streams were skipped: database-reviewer and plain adversarial pass are posted separately.

sakibsadmanshajib:
**Review stream: plain adversarial pass** (pipeline stream 2 of 3)

Adversarial questions put to this diff, each checked against source and live measurements rather than taken on faith:

1. Does the covering index actually enable index-only scans for BOTH GetBalance arms, including the `GROUP BY reservation_id, CASE WHEN reservation_id IS NULL THEN id END` arm that references the pkey? Verified in EXPLAIN: yes, both arms show Index Only Scan, because `id` is in INCLUDE. Without `id` there the holds arm heap-fetches.
2. Is `(account_id, entry_type)` the right key order? Yes: every served predicate is account equality plus entry_type IN list; no query orders or filters on entry_type alone.
3. Do the four drops regress anything? Each dropped index is a non-unique btree whose column list exactly equals its surviving unique sibling; Postgres btrees scan backwards so ordering is covered too. Grep over repo (go/ts/sql) found zero references to any dropped name outside its own origin migration. Lifetime scan counts 0 to 36 over five months confirm the planner already preferred the unique twins.
4. Could the new indexes hijack plans that were better before? Checked the other ledger access paths: the ledger listing endpoint needs ORDER BY created_at DESC LIMIT, unsatisfiable by either new index, so it keeps idx_credit_ledger_entries_account_created_at; idempotency lookups keep their unique index; budgets window aggregate can only gain. Benchmarked Q4/Q5 regression probes show unchanged plans.
5. Is "GetBalance runs on every billable request" an honest claim? Traced it: accounting/service.go calls GetBalance inside WithAccountLock on CreateReservation AND ExpandReservation, and reservations are created fail-closed per request (D-034). Claim holds.
6. Lock risk of plain CREATE INDEX inside the migration transaction? Largest touched table is ~10k rows / 9.7 MB, measured sub-second build in the bench copy; CONCURRENTLY cannot run inside a Supabase per-file transaction anyway. Stated in the migration header.
7. ANALYZE inside the migration transaction is legal (it takes no conflicting lock at this size) and makes the planner weigh new indexes immediately post-deploy.

No defects found that warrant changes. The one residual risk noted: write amplification from a second ledger index on an append-heavy table, accepted because measured insert volume on the demo box is ~75 rows/day and the index serves the hottest read path in the product.

sakibsadmanshajib (commented):
Independent database review (PostgreSQL specialist). Verdict: **APPROVE**. I verified every claim in this migration against the actual query shapes in the Go repositories and the historical migrations that created the dropped indexes. Findings below.

## Verified correct

**Index 1 (`idx_credit_ledger_entries_account_type_cover`) matches GetBalance exactly.**
Checked `apps/control-plane/internal/ledger/repository.go:149-189`. Both arms reference only columns present in key + INCLUDE:
- Posted arm: filters `account_id` + `entry_type IN (...)`, aggregates `SUM(credits_delta)`. Fully index-coverable.
- Holds arm: filters the same two columns, aggregates `credits_delta`, groups by `reservation_id, CASE WHEN reservation_id IS NULL THEN id END`. The GROUP BY references `reservation_id` and `id`, both present (INCLUDE), and the CASE expression is computed from index values, so an index-only scan remains valid for this arm too, including the NULL-reservation per-row bucket edge.
Column order is right: equality column first, `IN` list second. The existing unique index `(account_id, entry_type, idempotency_key)` cannot serve either arm as an index-only scan (no `credits_delta`), so this index is genuinely additive, not a near-duplicate of it.

**Partial predicate matches all three consumer queries.**
`entry_type = 'usage_charge'` is hardcoded as a SQL literal at every call site (`payments/invoices/repository.go` ListActiveWorkspaces line 151 and AggregateByModel line 185; `budgets/repository.go:308` month-to-date spend). This matters: if any caller passed `entry_type` as a bind parameter, the planner could not prove predicate implication and would ignore the partial index. None does. `created_at >= / <` range leads on the single key column correctly. Note the third consumer (budgets MTD) has no upper bound, which the range still serves.

**All four drops are provably redundant.** Each has a surviving sibling with an identical column list:
- `accounts.slug TEXT NOT NULL UNIQUE` (inline constraint, 20260328_01) vs `idx_accounts_slug`.
- `custom_providers.slug TEXT NOT NULL UNIQUE` (20260611_01) vs `custom_providers_slug_idx`.
- `CONSTRAINT uq_budget_threshold_account UNIQUE (account_id)` vs `idx_budget_thresholds_account`.
- `UNIQUE (workspace_id, period_start)` on invoices vs `idx_invoices_workspace_period`.
A non-unique btree adds nothing its unique twin lacks (backward scans cover descending order), and I grepped the tree: zero references to any of the four names outside the provenance comments in this migration. `IF EXISTS` guards divergence between repo and box.

**Plain CREATE INDEX inside the transaction is right at this size.** SHARE lock during build blocks writes for milliseconds on ~10k rows; CONCURRENTLY cannot run inside the Supabase per-file transaction, so the alternative would be a manual psql step for no benefit at this scale. The header honestly states the growth ceiling where the calculus changes.

**No regression surface for the surviving ledger indexes.** Neither `idx_credit_ledger_entries_account_created_at` nor `idx_credit_ledger_entries_account_idempotency` is dropped. Plan choices stay sound: `lookupExistingEntry` keeps the unique idempotency index (`idempotency_key` is not in the new covering index, so no heap-free win exists there), `ListEntries`/cursor listing keep their ordering advantages, and the ledger is append-only so the two added indexes cost only insert-path maintenance, never HOT-update churn.

CI green on this PR at time of review (Go control-plane tests included). CodeRabbit stream was rate limited, noting for pipeline completeness since I am not that stream.

## Non-blocking observations

1. **LOW, optional improvement:** `idx_credit_ledger_entries_usage_charge_created_at` could `INCLUDE (account_id)`. ListActiveWorkspaces then becomes an index-only scan instead of fetching the heap row for every usage_charge in range just to read `account_id`. At 10k rows irrelevant; as the append-only ledger grows into millions of rows, heap fetches will dominate that cron query again. A one-line follow-up, not worth blocking on.
2. **INFO:** index-only scans depend on visibility-map coverage. After bursts of inserts, pages are not yet all-visible until autovacuum catches up, so GetBalance degrades toward heap fetches temporarily on a hot append-heavy table. Expected Postgres behavior, no action needed, just do not be surprised by plan flapping right after heavy traffic.
3. **INFO:** if any of the four tables ever carried `REPLICA IDENTITY USING INDEX` naming one of the dropped indexes, the drop would fail loudly rather than corrupt anything. Defaults (PK) apply here, single-box posture, logical replication not in play.

Nothing here blocks merge. The money-path invariant (D-034 fail-closed GetBalance inside the critical section) is untouched: schema-only diff, query semantics unchanged, and the faster both GetBalance arms run, the shorter the per-account lock hold gets.

sakibsadmanshajib:
Adopted finding 1: pushed INCLUDE (account_id) onto idx_credit_ledger_entries_usage_charge_created_at (commit 94adbb52a). Verified in the benchmark copy that ListActiveWorkspaces is now an Index Only Scan at 1.3 ms. Findings 2 and 3 noted as INFO; no action taken, as you recommended.

PR #1128, opened 2026-08-24
chore(deps): web console peer-consistent stack raise

sakibsadmanshajib:
Supersedes and folds the peer-conflicting dependabot bumps for apps/web-console into one consistent set, per the dependency sweep plan (Batch 4 sub-PRs a, c, d plus the @types/node item from e).

What changes
- next 15.5.15 to 15.5.23 together with @opennextjs/cloudflare ~1.19.0 to ~1.20.2 in a single commit: OpenNext 1.20.2 declares peer next >=15.5.21 <16 || >=16.2.11, so next must move in lockstep with it.
- vitest ^3.1.1 to 4.1.11, jsdom ^26.1.0 to 30.0.1, @vitejs/plugin-react ^4.4.1 to 6.0.5, and a new explicit vite ^8.2.2 dev dependency anchoring plugin-react 6's vite ^8 peer range. These four only resolve as a set; individually each dependabot PR failed npm ci on ERESOLVE.
- @types/node ^22.14.0 to ^24.13.3, matching the node 24 runtime named by the owner directive (not dependabot's 26.x suggestion).
- New css-imports.d.ts declaring *.css as an untyped module.

Why TypeScript stays at 5.9.3
The TS7 bump (#1035) was attempted here first and reverted with evidence: typescript@7.0.2 ships a native binary wrapper with no JS API library (ts.sys is undefined under require('typescript')), and next@15.5's getTypeScriptConfiguration calls ts.sys.getCurrentDirectory and ts.sys.readFile when loading any TypeScript config, so next build fails with TypeError: Cannot read properties of undefined (reading 'fileExists'). This reproduces on the latest 15.5 patch (15.5.23). TS7 becomes viable only after a Next release tolerates the native-only package, which itself is gated behind the OpenNext peer cap discussed on #1034.

Verification
- npx tsc --noEmit clean
- vitest run: 50 files, 553 tests pass
- next build green on host node v24.16.0
- docker compose run --rm --build web-console npm run build green

Unblocks #1032: its minor-patch group carries the @opennextjs/cloudflare 1.19 to 1.20 minor whose next peer is violated while next stays on 15.5.15, which is why every current web-console PR shows a failing npm ci job.

### DIFF
--- a/apps/web-console/css-imports.d.ts
+++ b/apps/web-console/css-imports.d.ts
@@ -1,0 +1,3 @@
+// TypeScript 7 (tsgo) no longer accepts side-effect imports of .css files
+// without an explicit module declaration (TS2882 on app/layout.tsx).
+declare module "*.css";
--- a/apps/web-console/package-lock.json
+++ b/apps/web-console/package-lock.json
@@ -8,14 +8,14 @@
       "name": "@hive/web-console",
       "version": "0.1.0",
       "dependencies": {
-        "@opennextjs/cloudflare": "~1.19.0",
+        "@opennextjs/cloudflare": "~1.20.2",
         "@react-pdf/renderer": "^4.4.1",
         "@supabase/ssr": "^0.6.1",
         "@supabase/supabase-js": "^2.100.1",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
         "lucide-react": "^0.469.0",
-        "next": "15.5.15",
+        "next": "15.5.23",
         "next-intl": "^4.13.4",
         "react": "19.0.0",
         "react-dom": "19.0.0",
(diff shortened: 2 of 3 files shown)


sakibsadmanshajib:
Builder verification and stream record for the orchestrator. Verification already in the PR body (tsc clean, 553 unit tests, host and docker builds green) and CI now confirms: all 21 checks pass including Web console (type + unit + build). Adversarial pass findings, both resolved in-tree: the lockfile was fully regenerated rather than minimally patched (unavoidable with this many peer-coupled majors; spot-checked that @supabase/ssr stays pinned at 0.6.1, so the plan's 4b ssr cookie migration is NOT smuggled into this PR), and typescript stays 5.9.3 with the TS7 attempt documented above rather than silently dropped. Sequencing note: #1032 needs this merged first (its group carries the same OpenNext minor whose next peer this PR satisfies). Streams: plain adversarial pass done (this comment); CodeRabbit CLI stream SKIPPED, the CLI cannot connect from the execution sandbox (WebSocket closed on two attempts); the GitHub CodeRabbit app check itself reports rate limited, also outside this PR's control.

PR #1129, opened 2026-08-24
chore(deps): agent console peer-consistent stack raise

sakibsadmanshajib:
Mirrors the web-console raise (#1128) for apps/agent-console, folding the peer-conflicting dependabot bumps into one consistent set per the dependency sweep plan (Batch 5, mirroring Batch 4c and d plus the @types/node item).

What changes
- next 15.5.15 to 15.5.23 (parity with web-console; agent-console carries the same next version by convention).
- vitest ^3.1.1 to 4.1.11, jsdom ^26.1.0 to 30.0.1, @vitejs/plugin-react ^4.4.1 on 6.0.5, new explicit vite ^8.2.2 dev dependency anchoring plugin-react 6's vite ^8 peer.
- @types/node ^22.14.0 to ^24.13.3 matching the node 24 runtime directive, not dependabot's 26.x suggestion.
- New css-imports.d.ts declaring *.css as an untyped module.

Why TypeScript stays at 5.9.3
Same evidence as #1128: typescript@7.0.2 ships only a native binary wrapper with no JS API library (require('typescript') exposes no ts.sys), and next@15.5 getTypeScriptConfiguration requires ts.sys when loading any TypeScript config, so next build fails with TypeError: Cannot read properties of undefined (reading 'fileExists'). Reproduced on latest 15.5.23. TS7 waits for a Next release that tolerates the native-only package.

Verification
- npx tsc --noEmit clean
- vitest run: 9 files, 112 tests pass
- docker compose build agent-console green (production next build inside the image)

Unblocks #1017 rebase onto post-1106 main.

### DIFF
--- a/apps/agent-console/css-imports.d.ts
+++ b/apps/agent-console/css-imports.d.ts
@@ -1,0 +1,3 @@
+// TypeScript 7 (tsgo) no longer accepts side-effect imports of .css files
+// without an explicit module declaration (TS2882 on app/layout.tsx).
+declare module "*.css";
--- a/apps/agent-console/package-lock.json
+++ b/apps/agent-console/package-lock.json
@@ -10,21 +10,22 @@
       "dependencies": {
         "@supabase/ssr": "^0.6.1",
         "@supabase/supabase-js": "^2.100.1",
-        "next": "15.5.15",
+        "next": "15.5.23",
         "react": "19.0.0",
         "react-dom": "19.0.0"
       },
       "devDependencies": {
         "@tailwindcss/postcss": "^4.0.0",
         "@testing-library/react": "^16.3.0",
-        "@types/node": "^22.14.0",
+        "@types/node": "^24.13.3",
         "@types/react": "^19.1.0",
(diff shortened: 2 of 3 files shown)


sakibsadmanshajib:
Builder verification and stream record for the orchestrator. Verification per PR body (tsc clean, 112 unit tests, docker image build with production next build green) and CI now confirms: all 21 checks pass. Adversarial pass on the diff: lockfile spot-checks confirm @supabase/ssr stays pinned at 0.6.1 (the plan's Batch 5 ssr mirror is deliberately NOT in this PR and still needs its own cookie migration review plus live login regression test, see the flag posted on #1017 which carries that bump inside a group), next resolves to 15.5.23, vite 8.2.2 anchors plugin-react 6's peer, @types/node resolves to 24.13.3 on the runtime-matching line. Streams: plain adversarial pass done (this comment); CodeRabbit CLI stream SKIPPED, the CLI cannot connect from the execution sandbox (WebSocket closed on two attempts); the GitHub CodeRabbit app check reports rate limited.

PR #1130, opened 2026-08-24
perf: cache hot routing and catalog reads in Redis

sakibsadmanshajib:
## What

Adds a small read-through Redis cache in front of the hottest read paths, without touching any money path.

Measured first (demo box, live): the per-request Hive overhead is already small after the self-hosted DB cutover. The three Postgres reads behind `/internal/routing/select` (alias policy + route candidates + alias pricing) plus the per-tenant entitlement check take **p50 ~1.4 ms** uncached; the full metered request is dominated by provider dispatch (p50 5.6 s total vs 5.3 s dispatch per 6h of Prometheus stage metrics). Caching targets the read component only.

## Design

- `internal/platform/rcache`: generic read-through JSON cache helper.
  - Degrades to pass-through on any Redis failure: never an availability dependency.
  - Never caches loader errors, so a transient DB failure cannot pin a failed or stale read for a TTL.
  - Dedicated hot-path client: 500ms dial / 250ms I/O / MaxRetries 1 / 150ms deadline per op, so a post-startup Redis hang costs milliseconds, not go-redis's multi-second defaults.
  - Versioned payload envelope (`{"v":1,...}`): entries written by an older binary reload instead of partially decoding into wrong zero fields during rolling restarts.
- `routing.CachedRepository` caches the three SelectRoute reads (`LoadAliasPolicy`, `ListRouteCandidates`, `LoadAliasPricing`) under global keys. Those tables have no Go-side writers (routes are migration-only), so the TTL is the staleness bound for changes.
- `catalog.CachedRepository` caches `ListPublicAliases` (global key), `ListAliasesForTenant`, and `IsAliasVisibleToTenant` under keys that embed the tenant UUID. `UpsertVisibility`/`DeleteVisibility` invalidate their own `(tenant, alias)` keys on write, covering both admin endpoints and the boot-time OWUI reconcile, which mutate through the same methods. Invalidation DELs run post-commit on an uncanceled context with their own deadline, and failures log loudly rather than silently dropping.
- Wired in `cmd/server/main.go`: enabled when REDIS_URL is set and reachable at startup; flushes the entire cache prefix once at boot so keys written by a previous binary can never answer requests with pre-deploy pricing after a repricing migration. Tunable via validated `HIVE_HOT_CACHE_TTL_SECONDS`. No compose change needed; REDIS_URL already flows to control-plane.

## Money-path boundary

Balances, reservations, ledger entries, billing state: never cached, anywhere in this change. Alias pricing rows are published per-million credit catalog data; the actual charge is still computed transactionally by accounting from the row captured at select time. Stated in code comments in `rcache.go`, `routing/cache.go`, and here.

## Tenant isolation

Every tenant-derived key embeds the tenant UUID (`cat:tlist:{tid}`, `cat:vis:{tid}:{alias}`); two tenants can never read each other's entries through the shared Redis. Only the unauthenticated public alias list uses a global key, because its query takes no tenant input. Unit tests assert two tenants never reuse each other's cache entries. Bogus alias IDs cannot populate attacker-chosen keys: SelectRoute loads policy first, so unknown aliases fail there uncached; entitlement verdict keys are bounded by authenticated tenants times real aliases and expire with the TTL.

## Measurements

Same request shape, demo box, against the same live Postgres and Redis (`TestLiveBench`, committed in this diff, env-guarded behind `HIVE_LIVE_BENCH=1`; order-fair: both paths warmed before either timed pass):

| Read sequence (policy + candidates + pricing + entitlement), n=300 | min | p50 | p90 | max |
|---|---|---|---|---|
| Uncached | 0.878 ms | 1.430 ms | 1.634 ms | 2.167 ms |
| Cached | 0.740 ms | 1.189 ms | 1.572 ms | 2.001 ms |

~17% faster at the median on the read component, tighter tail. The structural win is that these reads leave the shared Postgres pool for 30s windows: the session-mode pool has been the coupling point between CI bursts and live traffic before, so removing per-request catalog reads from it is a pressure reduction even when the box is quiet. Endpoint baselines today: `/internal/routing/select` p50 2.8 ms, catalog snapshot p50 1.8 ms. Provider dispatch dominates end-to-end metered latency (p50 5.6 s total vs 5.3 s dispatch over the last 6h of Prometheus stage metrics) and is not cacheable; nothing here claims to make model inference faster.

After warmup the Redis keyspace held the four `hivecp:v1:*` keys this sequence writes (policy, candidates, pricing, tenant verdict) alongside the pre-existing auth snapshot and budget keys.

## Tests

- rcache unit tests (miniredis): hit/pass-through/TTL expiry/Redis-down degrade/error-not-cached/delete-error-surfaced/version-mismatch-reloads/flush.
- Decorator tests: single-load assertions, error semantics (ErrAliasNotFound re-queries), nil-preserved-through-cache-hit regression, tenant key isolation, write invalidation via both mutation methods.
- Full `go test ./apps/control-plane/... -short` green in the toolchain container.

## Review outcome

Streams that ran: CodeRabbit CLI local run (1 finding, fixed) plus the CodeRabbit GitHub App (3 inline comments, all addressed), plain ticket-first adversarial pass, Go idiom review, mandatory database review, mandatory security pass. SKIPPED: the `/codex:adversarial-review` stream only; the Codex connector reported usage limits reached. No other stream skipped.

Eleven inline threads; ten fixed, one rebutted with reasoning:

- Dedicated hot-path Redis client with tight timeouts and per-op deadlines (a hung Redis now costs milliseconds).
- Post-commit invalidation DEL runs on context.WithoutCancel with its own deadline; failures log loudly instead of dropping silently.
- Startup flush of the whole prefix so pre-deploy keys cannot serve pre-migration pricing across deploys.
- Versioned envelope so pre-deploy entries reload instead of partially decoding during rolling restarts.
- Validated HIVE_HOT_CACHE_TTL_SECONDS tuning env.
- nil-vs-empty slice semantics preserved through the JSON round trip; duplicated doc paragraph removed.
- Rebutted (no action): stampede singleflight. Bounded harm today (underlying read ~1ms local); x/sync/singleflight recorded as the upgrade path if traffic justifies it. Database review accepted the writer sweep: zero runtime writers of cached tables outside the two invalidation paths covered.

Review-comment bodies were treated as untrusted data; every finding was verified against source before fixing or rebutting.

### DIFF
--- a/apps/control-plane/internal/catalog/cache.go
+++ b/apps/control-plane/internal/catalog/cache.go
@@ -1,0 +1,117 @@
+package catalog
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/platform/rcache"
+)
+
+// CachedRepository wraps a catalog Repository with a short-TTL read-through
+// Redis cache over the hot reads the /v1/models listing and the inference
+// entitlement check make per request.
+//
+// TENANT SCOPING (cache-poisoning guard): every key that carries
+// tenant-derived data embeds the tenant UUID, so two tenants can never read
+// each other's visibility verdicts or model lists through the shared Redis.
+// Only ListPublicAliases (the unauthenticated public/preview alias list) is
+// keyed without a tenant, because its query takes no tenant input and its
+// rows are identical for every caller.
+//
+// Invalidation is write-through at the mutation points this package owns:
+// UpsertVisibility and DeleteVisibility delete exactly the keys derived from
+// their (tenant, alias) arguments. ReconcileOWUISync and the admin visibility
+// handlers both mutate through these same repository methods, so both paths
+// invalidate automatically. The 30s TTL backstops anything missed.
+type CachedRepository struct {
+	Repository
+	cache *rcache.Cache
+}
+
+// NewCachedRepository returns repo wrapped with the Redis read-through cache,
+// or repo unchanged when cache is nil.
+func NewCachedRepository(repo Repository, cache *rcache.Cache) Repository {
+	if cache == nil {
+		return repo
+	}
+	return &CachedRepository{Repository: repo, cache: cache}
+}
+
+func keyPublic() string { return "cat:public" }
+
+func keyTenantList(tid uuid.UUID) string { return "cat:tlist:" + tid.String() }
+
+func keyVisible(tid uuid.UUID, aliasID string) string {
+	return "cat:vis:" + tid.String() + ":" + aliasID
+}
+
+// ListPublicAliases caches the global public/preview alias list. Keyed without
+// a tenant on purpose: the underlying query takes no tenant input.
+func (r *CachedRepository) ListPublicAliases(ctx context.Context) ([]ModelAlias, error) {
+	var loaded []ModelAlias
+	got, err := rcache.GetJSON(ctx, r.cache, keyPublic(), func(ctx context.Context) ([]ModelAlias, error) {
+		rows, err := r.Repository.ListPublicAliases(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if rows == nil {
+			rows = []ModelAlias{}
+		}
+		return rows, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	loaded = got
+	return loaded, nil
+}
+
+// ListAliasesForTenant caches one tenant's entitled alias list under a key
+// that embeds the tenant UUID (see the tenant-scoping note above).
+func (r *CachedRepository) ListAliasesForTenant(ctx context.Context, tenantID uuid.UUID) ([]ModelAlias, error) {
+	var loaded []ModelAlias
+	got, err := rcache.GetJSON(ctx, r.cache, keyTenantList(tenantID), func(ctx context.Context) ([]ModelAlias, error) {
+		rows, err := r.Repository.ListAliasesForTenant(ctx, tenantID)
+		if err != nil {
+			return nil, err
+		}
+		if rows == nil {
+			rows = []ModelAlias{}
+		}
+		return rows, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	loaded = got
+	return loaded, nil
+}
+
+// IsAliasVisibleToTenant caches one tenant's entitlement verdict for one
+// alias, keyed by both. A false verdict IS cached: it is a real answer, not
+// an error, and the mutation paths below invalidate it.
+func (r *CachedRepository) IsAliasVisibleToTenant(ctx context.Context, tenantID uuid.UUID, aliasID string) (bool, error) {
+	return rcache.GetJSON(ctx, r.cache, keyVisible(tenantID, aliasID), func(ctx context.Context) (bool, error) {
+		return r.Repository.IsAliasVisibleToTenant(ctx, tenantID, aliasID)
+	})
+}
+
+// UpsertVisibility delegates the write, then deletes exactly the cache keys
+// derived from this (tenant, alias): the tenant's model list and its single
+// alias entitlement verdict.
+func (r *CachedRepository) UpsertVisibility(ctx context.Context, row TenantModelVisibility) error {
+	if err := r.Repository.UpsertVisibility(ctx, row); err != nil {
+		return err
+	}
+	r.cache.Delete(ctx, keyTenantList(row.TenantID), keyVisible(row.TenantID, row.AliasID))
+	return nil
+}
+
+// DeleteVisibility mirrors UpsertVisibility's invalidation for the delete path.
+func (r *CachedRepository) DeleteVisibility(ctx context.Context, tenantID uuid.UUID, aliasID string) error {
+	if err := r.Repository.DeleteVisibility(ctx, tenantID, aliasID); err != nil {
+		return err
+	}
+	r.cache.Delete(ctx, keyTenantList(tenantID), keyVisible(tenantID, aliasID))
+	return nil
+}
--- a/apps/control-plane/internal/catalog/cache.go
+++ b/apps/control-plane/internal/catalog/cache.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"context"
+	"log"
 
 	"github.com/google/uuid"
 	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/platform/rcache"
@@ -46,45 +47,18 @@ func keyVisible(tid uuid.UUID, aliasID string) string {
 }
 
 // ListPublicAliases caches the global public/preview alias list. Keyed without
-// a tenant on purpose: the underlying query takes no tenant input.
+// a tenant on purpose: the underlying query takes no tenant input. Nil results
+// marshal as JSON null and unmarshal back to nil, preserving raw semantics.
 func (r *CachedRepository) ListPublicAliases(ctx context.Context) ([]ModelAlias, error) {
-	var loaded []ModelAlias
-	got, err := rcache.GetJSON(ctx, r.cache, keyPublic(), func(ctx context.Context) ([]ModelAlias, error) {
-		rows, err := r.Repository.ListPublicAliases(ctx)
-		if err != nil {
-			return nil, err
-		}
-		if rows == nil {
-			rows = []ModelAlias{}
-		}
-		return rows, nil
-	})
-	if err != nil {
-		return nil, err
-	}
-	loaded = got
-	return loaded, nil
+	return rcache.GetJSON(ctx, r.cache, keyPublic(), r.Repository.ListPublicAliases)
 }
 
 // ListAliasesForTenant caches one tenant's entitled alias list under a key
 // that embeds the tenant UUID (see the tenant-scoping note above).
 func (r *CachedRepository) ListAliasesForTenant(ctx context.Context, tenantID uuid.UUID) ([]ModelAlias, error) {
-	var loaded []ModelAlias
-	got, err := rcache.GetJSON(ctx, r.cache, keyTenantList(tenantID), func(ctx context.Context) ([]ModelAlias, error) {
-		rows, err := r.Repository.ListAliasesForTenant(ctx, tenantID)
-		if err != nil {
-			return nil, err
-		}
-		if rows == nil {
-			rows = []ModelAlias{}
-		}
-		return rows, nil
+	return rcache.GetJSON(ctx, r.cache, keyTenantList(tenantID), func(ctx context.Context) ([]ModelAlias, error) {
+		return r.Repository.ListAliasesForTenant(ctx, tenantID)
 	})
-	if err != nil {
-		return nil, err
-	}
-	loaded = got
-	return loaded, nil
 }
 
 // IsAliasVisibleToTenant caches one tenant's entitlement verdict for one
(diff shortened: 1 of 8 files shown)


apps/control-plane/internal/routing/cache.go
```diff
@@ -0,0 +1,105 @@
...
+// provider_routes, alias_route_policies and model_aliases rows are written by
+// SQL migrations only (providers/repository.go records that routes are
+// migration-only), so explicit invalidation has no runtime hook to attach to;
+// the 30s TTL is the staleness bound for migration-time changes.
+//
+// Money-path boundary: LoadAliasPricing caches the published per-million
+// credit price of an alias, not any account's balance, hold or ledger. The
+// charge itself is computed from the pricing row captured at select time and
+// settled transactionally in the accounting path, which this package never
+// touches. Prices change via migration; worst case a new price applies to
```
sakibsadmanshajib: [MEDIUM] The staleness direction here is understated, and the worst case is money adjacent. Redis is a separate long lived container (redis:8.4-alpine, no restart on control-plane deploys), so these keys survive process restarts and a migration can land while the previous binary is still serving with a warm cache. After any price or route migration, requests are served the OLD cached values for up to DefaultTTL, not just late access to new values. Concretely: 20260823_40_credit_unit_rescale_billion.sql multiplies model_aliases prices by 10000; had this cache been live across that deploy, SelectionResult.Pricing (which metering uses as the select time cost basis) would have carried pre rescale prices into charges for up to 30s per price touching migration. Suggested cheap hardening: on startup when hotCache is enabled, flush the hivecp:v1 prefix (or fold a deploy version into the key prefix) so every deploy starts cache cold.

apps/control-plane/internal/platform/rcache/rcache.go
```diff
@@ -0,0 +1,99 @@
...
+//
+// Boundary (money paths): this cache must NEVER hold credit balances,
+// reservations, ledger entries, or billing state. Those are transactional
+// money state (decision D-034 fail-closed); staleness there is a billing bug
+// before it is a performance question. What belongs here is reference data:
+// routing catalog rows (alias policies, route candidates, alias pricing) and
+// per-tenant model visibility, whose staleness from any missed invalidation
+// is bounded by the TTL.
+//
+// The cache degrades to a pass-through on any Redis failure, so it is never
```
sakibsadmanshajib: [MEDIUM] Degrade is correctness safe but not latency safe, which the never an availability dependency claim overstates. platformredis.NewClient keeps go-redis v9 defaults (MaxRetries 3, DialTimeout 5s, ReadTimeout 3s), so once Redis stops accepting connections after the startup Ping gate passes (network partition or hang, not clean refusal), each GetJSON GET can block for seconds before falling through to the loader. SelectRoute issues three cached reads per metered request, so a hung Redis serializes that into potentially tens of seconds added to every request rather than a fast pass through. The Redis down test covers connection refused only, which fails in milliseconds and does not exercise this. Suggestion: give the hot path its own client options (MaxRetries 0 or 1, sub second DialTimeout/ReadTimeout) or bound the cache leg with a short context timeout inside GetJSON.

apps/control-plane/internal/platform/rcache/rcache.go
```diff
@@ -0,0 +1,99 @@
...
+
+// GetJSON returns the cached JSON value for key, loading it through load on a
+// miss and storing the result. A Redis GET failure (including redis.Nil) or a
+// decode failure falls through to load; load errors propagate to the caller
+// and are never stored.
+func GetJSON[T any](ctx context.Context, c *Cache, key string, load func(context.Context) (T, error)) (T, error) {
+	if c == nil || c.client == nil {
+		return load(ctx)
+	}
+	if raw, err := c.client.Get(ctx, c.key(key)).Bytes(); err == nil {
```
sakibsadmanshajib: db-review medium: hot-path Redis calls here inherit the shared client's default budgets (go-redis v9 defaults: DialTimeout 5s, ReadTimeout 3s, WriteTimeout 3s, default retry count with exponential backoff). The startup Ping gates enablement, but if Redis hangs or is black-holed after startup, every metered request stalls inside this Get for seconds before falling through to Postgres. Availability degrades as designed; latency does not, and this sits on the per-request routing path. Fix: bound each Redis operation in rcache (GetJSON, SetJSON, Delete) with its own short context deadline (on the order of 100ms) so degradation stays bounded regardless of the client shared with the apikeys, budgets and signupguard caches.

apps/control-plane/internal/platform/rcache/rcache.go
```diff
@@ -0,0 +1,99 @@
...
+		var v T
+		if json.Unmarshal(raw, &v) == nil {
+			return v, nil
+		}
+	}
+	v, err := load(ctx)
```
sakibsadmanshajib: db-review low: GetJSON has no single flight, so when a hot key expires every concurrent caller for that alias runs the loader at once. Verified this is acceptable at current scale: the three routing reads and the entitlement check hit single digit row, indexed reference tables through one pgxpool bounded instance, so the burst drains in milliseconds, and the steady state is still strictly less Postgres load than the uncached baseline this replaces. Revisit if control-plane ever runs multiple replicas sharing one Redis or the catalog grows large; the cheap upgrade then is golang.org/x/sync/singleflight keyed per cache key around load. No change required now.

apps/control-plane/internal/platform/rcache/rcache.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return &Cache{client: client, prefix: prefix, ttl: TTL}
+}
+
+// key returns the full Redis key for k. Nil-safe on purpose.
+func (c *Cache) key(k string) string {
+	if c == nil {
+		return k
+	}
+	return c.prefix + ":" + k
```
sakibsadmanshajib: db-review low: cached payloads carry no shape version beyond the hivecp:v1 key prefix, and the Redis container outlives application restarts. A deploy that renames or retypes a field on CatalogPricing, ModelAlias, RouteCandidate or AliasPolicySnapshot will silently partial decode pre-deploy entries for up to one TTL after restart. Today that failure mode happens to fail closed (zeroed prices trip ErrAliasNotPriced in SelectRoute), but that is luck rather than design. Fix: either embed a schema version field in the cached JSON, or add a comment on the hivecp:v1 prefix at its construction site in cmd/server/main.go making a prefix bump mandatory whenever any cached struct changes shape.

apps/control-plane/internal/routing/cache.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+	loaded = got
+	return loaded, nil
+}
+
+// LoadAliasPricing caches the published per-million credit price. See the
+// money-path boundary note on CachedRepository above: balances, reservations
+// and ledger entries are never cached anywhere in this change.
+// LoadAliasPricing caches the published per-million credit price. See the
```
sakibsadmanshajib: db-review low: this comment block repeats lines 83 to 85 verbatim (copy paste artifact). Delete one copy. Worth cleaning because this money boundary note is the load bearing text reviewers read, and duplicated copies invite drift.

apps/control-plane/internal/catalog/cache.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// UpsertVisibility delegates the write, then deletes exactly the cache keys
+// derived from this (tenant, alias): the tenant's model list and its single
+// alias entitlement verdict.
+func (r *CachedRepository) UpsertVisibility(ctx context.Context, row TenantModelVisibility) error {
+	if err := r.Repository.UpsertVisibility(ctx, row); err != nil {
+		return err
+	}
+	r.cache.Delete(ctx, keyTenantList(row.TenantID), keyVisible(row.TenantID, row.AliasID))
```
sakibsadmanshajib: MEDIUM: invalidation is fire-and-forget on the request context, so a dropped delete is silent. Two ways this loses the invalidation after the visibility row has committed: (1) Cache.Del swallows its error (`_ = c.client.Del(...).Err()` in rcache.Cache.Delete), so a Redis blip leaves the old entitlement verdict cached with nothing in the logs; (2) if the admin client disconnects at the wrong moment, ctx cancellation kills the DEL too. This key gates `IsAliasVisibleToTenant`, so either failure means a just-revoked tenant still passes the entitlement check for up to the full 30s TTL with zero observable signal. Fix: run the post-commit invalidation with `context.WithoutCancel(ctx)`, have `Cache.Delete` return the error, and log a warning here (the DB write has committed, so failing the request would be wrong; making the failure loud is the fix). The same applies to the mirror in DeleteVisibility below.

apps/control-plane/internal/routing/cache.go
```diff
@@ -0,0 +1,105 @@
...
+func (r *CachedRepository) LoadAliasPolicy(ctx context.Context, aliasID string) (catalog.AliasPolicySnapshot, error) {
+	return rcache.GetJSON(ctx, r.cache, keyPolicy(aliasID), func(ctx context.Context) (catalog.AliasPolicySnapshot, error) {
+		return r.Repository.LoadAliasPolicy(ctx, aliasID)
+	})
+}
+
+// ListRouteCandidates caches the route candidate list for one alias. A nil or
+// empty result is cached like any other successful read so an alias with no
+// routes does not hammer the database either; JSON null round-trips back to
+// nil, preserving the zero-value semantics ListRouteCandidates documents.
```
sakibsadmanshajib: LOW, doc correctness: the comment claims "JSON null round-trips back to nil", but the loader at line 71 normalizes nil to []RouteCandidate{} before caching, so null is never stored. Warm and cold paths both return an empty non-nil slice (which is consistent behavior, good), but the comment documents the opposite of what the code does. Reword it to say nil results are normalized to empty slices so the cache stores `[]`, or drop the normalization if a nil result ever carries distinct meaning (no caller relies on that today).

apps/control-plane/internal/routing/cache.go
sakibsadmanshajib: Fixed in 3d10c0a17: control-plane now flushes the entire hivecp:v1 prefix once at startup (SCAN+DEL in rcache.Cache.Flush), so keys written by a previous binary cannot answer requests with pre-deploy pricing after an alias repricing migration lands. A versioned envelope ({"v":1,...}) additionally makes pre-deploy entries fail the shape check and reload instead of partially decoding.

apps/control-plane/internal/platform/rcache/rcache.go
sakibsadmanshajib: Fixed in 3d10c0a17: the cache now builds its own dedicated hot-path client (500ms dial / 250ms read-write / MaxRetries 1) and bounds every operation by a 150ms context deadline, so a post-startup Redis hang costs milliseconds before the pass-through path serves from Postgres instead of seconds.

apps/control-plane/internal/platform/rcache/rcache.go
sakibsadmanshajib: Fixed in 3d10c0a17 alongside the latency finding: dedicated client with 500ms dial / 250ms I/O / MaxRetries 1 plus a 150ms deadline on every cache op. The shared client is untouched for its existing consumers.

apps/control-plane/internal/routing/cache.go
sakibsadmanshajib: Fixed in 3d10c0a17: control-plane now flushes the entire hivecp:v1 prefix once at startup (SCAN+DEL in rcache.Cache.Flush), so keys written by a previous binary cannot answer requests with pre-deploy pricing after an alias repricing migration lands. A versioned envelope ({"v":1,...}) additionally makes pre-deploy entries fail the shape check and reload instead of partially decoding.

apps/control-plane/internal/platform/rcache/rcache.go
sakibsadmanshajib: Fixed in 3d10c0a17: the cache now builds its own dedicated hot-path client (500ms dial / 250ms read-write / MaxRetries 1) and bounds every operation by a 150ms context deadline, so a post-startup Redis hang costs milliseconds before the pass-through path serves from Postgres instead of seconds.

apps/control-plane/internal/platform/rcache/rcache.go
sakibsadmanshajib: No action taken by design for this PR. The stampede window is bounded by a tiny local database read (about 1ms measured p50 on the demo box), so a cold burst re-reads rows that cost less than a millisecond each. Adding x/sync/singleflight is recorded in the PR description as the upgrade path if traffic ever justifies it.

apps/control-plane/internal/platform/rcache/rcache.go
sakibsadmanshajib: Fixed in 3d10c0a17: every payload now lives inside a versioned envelope (v + payload fields); an entry whose envelope version differs fails the check, is treated as a miss, and reloads. Regression test TestGetJSON_VersionMismatchTreatedAsMiss covers it.

apps/control-plane/internal/routing/cache.go
sakibsadmanshajib: Fixed in 3d10c0a17: duplicated paragraph removed.

apps/control-plane/internal/catalog/cache.go
sakibsadmanshajib: Fixed in 3d10c0a17: Cache.Delete now runs the DEL on context.WithoutCancel(ctx) with its own short deadline, so a canceled request cannot silently drop a post-commit invalidation; failures surface as a loud WARN from UpsertVisibility and DeleteVisibility. The visibility row commits before the DEL fires, so the method still returns success when the database write succeeded.

apps/control-plane/internal/routing/cache.go
sakibsadmanshajib: Fixed in 3d10c0a17: the loader no longer converts nil to an empty slice anywhere; JSON null round-trips to nil exactly as the raw repository returns it, and the comment now states that truthfully. A regression test asserts nil stays nil across both the miss and the hit.

PR #952, opened 2026-08-17
fix: sign in without an intermediate page, drop stock suggestions, stop storing emails as display names

sakibsadmanshajib:
Three owner requested changes to the forked chat front end. All of them change the vendored source under `vendor/open-webui/`, which is where fork work belongs now that the image builds from that tree (#938, D-036, D-040).

## 1. No intermediate sign in page

With `ENABLE_LOGIN_FORM: "false"` and exactly one provider configured, `/auth` rendered a heading, a horizontal rule with nothing above it, and a single "Continue with Hive" button. That is a choice between one option, and the rule is the separator for a password form that never renders.

An unauthenticated visitor now goes straight into the provider flow, and the `redirect` query parameter survives the round trip through the mechanism that already existed for it.

The failure path was built first, because a broken redirect on the login path locks every user out:

* The decision moved out of the page into `src/lib/hive/sso-redirect.ts`, a pure function with twelve tests, following the `nav.ts` plus `nav.test.ts` precedent already in that directory.
* **Loop guard.** The bounce nobody had guarded against is a round trip that returns the browser to `/auth` with no `error` parameter and no token cookie. The old condition is true again at that point and the page redirects again, forever. The page now stamps each attempt in `sessionStorage`, and an attempt inside the last fifteen seconds renders the page with an explanation and the provider button as the manual retry instead of redirecting. The stamp is cleared once a session exists.
* **Provider errors.** `handle_callback` already redirects every exception to `/auth?error=...`, and an error parameter already suppressed the redirect. It now also renders a persistent message carrying the provider's reason, rather than only a toast that disappears.
* **Sign out.** Signing out landed on `/auth` while the provider's own session was usually still live, so an unconditional redirect would have signed the user straight back in and made signing out unreachable. The two user initiated sign out paths now carry `?signed_out=1`, which suppresses the redirect and says so on the page. The token expiry path deliberately does not, since a silent refresh is the wanted behaviour there.
* The manual page is not removed. It is the destination for every one of these cases, plus `?form=1` as an unconditional escape hatch.
* The orphaned divider is gone: it now renders only when the password or LDAP form it separates is actually on screen.

The behaviour is gated on `OAUTH_AUTO_REDIRECT`, and every other precondition (one provider, no password form, no LDAP, no trusted header, no onboarding) stays a runtime check, so a second provider or a re-enabled password form turns it off by itself.

[...]

PR #1131, opened 2026-08-25
test: coverage instrumentation for web console, chat frontend, and Go modules

sakibsadmanshajib:
## What

Coverage was not measurable on three surfaces. This makes it measurable everywhere without enforcing any threshold.

## Changes

- **Web console**: `@vitest/coverage-v8` devDependency (4.1.11, matching vitest 4.1.11 from #1128), a coverage block in `vitest.config.ts` scoped to `app/`, `lib/`, and `components/` with `lib/control-plane/permissions.generated.ts` excluded, and a `test:coverage` npm script.
- **Chat frontend**: `scripts/test-owui-hive-frontend.sh` was reported missing from origin/main but it exists there and was never deleted; its history runs from the sign-in redirect work around #952 through #1085, #1108, #1112, and #1124. Its scratch-tree vitest run now installs `vitest@2` plus `@vitest/coverage-v8@2` together and prints an advisory v8 coverage table scoped to `lib/hive`. It stays on the pinned major 2 on purpose: this suite runs standalone of the console's stack.
- **Go**: new `make go-cover` target runs both modules' short tests into per-module GOCOVERDIRs and merges them with `go tool covdata textfmt` into one cross-module umbrella total alongside each module alone.
- **CI**: the web-unit job's unit-test step now runs `test:coverage` so the summary line appears in the required-check log. Advisory only, no thresholds configured.

## Measured

| Surface | Statements / Lines | Branches | Functions |
|---|---|---|---|
| web-console (51 suites, 558 tests) | 38.36% / 39.11% | 34.94% | 41.37% |
| chat frontend lib/hive (11 suites, 114 tests) | 25.15% / 25.15% | 79.82% | 72.05% |
| Go control-plane | 46.0% statements | | |
| Go edge-api | 66.2% statements | | |
| Go umbrella (both modules merged) | 53.0% statements | | |

Web-console numbers were re-measured after merging main with #1128's raise of vitest to 4.x. Pre-merge, on vitest 3.2.7, they read 40.26% lines, 70.4% branches, and 67.5% functions. The v8 provider counts differently across majors, so the movement is a measurement change, not a code regression.

## Notes for review

- The two web-console suites that read repo-root files (`.env.example`, `docs/`, `deploy/`) cannot see them inside the bare web-console image because Dockerfile.web-console copies only `apps/web-console`. That failure predates this change and is unrelated to coverage; the measurement runs mounted those three paths read-only to complete. Flagging as a known local-testing gap rather than fixing here.
- `go tool covdata percent` prints per-package rows only, so the go-cover target converts to text profiles and uses `go tool cover -func | tail -1` for single total lines.
- No thresholds enforced anywhere per scope; numbers are now one command away when a threshold decision is made.


apps/web-console/package.json
```diff
@@ -43,6 +44,7 @@
     "@types/react": "^19.1.0",
     "@types/react-dom": "^19.1.0",
     "@vitejs/plugin-react": "^4.4.1",
+    "@vitest/coverage-v8": "^3.2.7",
```
sakibsadmanshajib: [CodeRabbit CLI stream] major: Synchronize package-lock.json with the new @vitest/coverage-v8 dependency declared in package.json, so npm ci succeeds in web-unit.

Resolution note (builder): already synchronized in the same diff. The lockfile records @vitest/coverage-v8 ^3.2.7 with resolved metadata plus its transitive tree (@ampproject/remapping, @bcoe/v8-coverage, cliui and others), and vitest itself is locked at 3.2.7 so provider and runner versions match exactly. Verified live: the docker compose run --build web-console npm run test:coverage measurement run built the image through npm install against this lockfile and executed 558 tests with v8 coverage successfully.

Makefile
```diff
@@ -45,7 +45,29 @@ test-scripts:
...
+# into profiles that `go tool cover -func` summarizes into one total line per
+# module plus a true cross-module umbrella total. No thresholds are enforced;
+# this makes the number one command away so a threshold decision has something
+# to stand on. Runs in the toolchain container per the Docker-only testing
+# contract; -count=1 -short matches the standard go-tests invocation.
+go-cover:
+	cd deploy/docker && docker compose $(COMPOSE_ENV_ARG) --profile tools run --rm toolchain \
+	  "rm -rf /tmp/hive-go-cover && mkdir -p /tmp/hive-go-cover/control-plane /tmp/hive-go-cover/edge-api && \
+	   cd /workspace/apps/control-plane && go test ./... -count=1 -short -coverpkg=./... -covermode=atomic -args -test.gocoverdir=/tmp/hive-go-cover/control-plane && \
+	   cd ../edge-api && go test ./... -count=1 -short -coverpkg=./... -covermode=atomic -args -test.gocoverdir=/tmp/hive-go-cover/edge-api && \
```
sakibsadmanshajib: [plain adversarial stream] finding: the go-cover recipe redirected each go test run to /dev/null. If a future run goes red, make fails with an exit code but no diagnostics: which package failed would be invisible, a quiet-absence tradeoff this repo explicitly bans.

Resolution (builder): fixed in commit 40c8a6cf0 before review started; redirects removed so per-package output stays visible. This thread records the stream's only finding.

PR #1138, opened 2026-08-25
feat: tenant BYOK provider keys, encrypted at rest

sakibsadmanshajib:
## Summary

Backend for tenant BYOK (bring your own key): tenants register their OWN provider credentials and (in a follow-up) route their traffic through them. This is the storage + API + admin-visibility slice of the capability PR #1126 documented as missing end to end. Console UI intentionally out of scope here.

## Storage

New migration `20260824_03_tenant_provider_keys.sql`: `public.tenant_provider_keys`

- `account_id` FK to `public.accounts` (same principal as `public.api_keys`), ON DELETE CASCADE
- exactly one target per row, enforced by CHECK: either `provider_slug` FK to `custom_providers(slug)` (ON DELETE RESTRICT) or freeform `base_url`, plus optional `model_map` jsonb of requested alias -> upstream model name
- `encrypted_api_key BYTEA`: AES-256-GCM ciphertext sealed by the application with a 32-byte key from `HIVE_BYOK_ENC_KEY`; layout nonce || ciphertext || tag. Plaintext never reaches the database
- `key_last4`: the only secret remnant, for display masks; CHECK pins length 4
- `status active|revoked`, `created_by_user_id`, timestamps, partial unique index: one ACTIVE credential per provider per account (re-register requires revoking first)
- FORCE RLS with a permissive policy for hive_app only (mirrors custom_providers); anon/authenticated have no policy and are default-deny. Account isolation enforced at the application layer: every repository query filters `account_id` in the WHERE clause, so cross-account reads are indistinguishable from missing rows
- grants: SELECT/INSERT/UPDATE to hive_app, SELECT to auditor_ro

## Crypto approach and fail-closed posture (D-034)

AES-256-GCM via Go stdlib (`crypto/aes`, `crypto/cipher`), random nonce per encryption. Key from env `HIVE_BYOK_ENC_KEY` (base64, exactly 32 bytes). Chosen posture, simpler and strictly safer than conditional boot gating:

- unset -> locked mode: provider-key routes not mounted at boot, log line names it. No plaintext can ever be written because nothing else gates writes
- set but malformed -> control-plane refuses to start (config error dies at boot, not on first use)
- decrypt path fails closed when the cipher is absent or the blob does not verify

Justification vs the alternative (fail-closed only when rows exist): that needs a DB read at boot to decide whether to enforce security, which inverts the failure direction; lazy lock-out has no window where plaintext could land and no DB dependency for its safety decision.

## API surface

Tenant surface rides the existing tenant-scoped authz pattern (mirrors internal/apikeys): viewer context resolution, X-Hive-Account-ID account switch, owner-only perms, platform-admin overlay.

| Method | Path | Perm |
|---|---|---|
| POST | /api/v1/accounts/current/provider-keys | provider_keys.write (owner, verified) |
| GET | /api/v1/accounts/current/provider-keys | provider_keys.read (owner) |
| POST | /api/v1/accounts/current/provider-keys/{id}/revoke | provider_keys.write (owner, verified) |

New authz permissions `provider_keys.read` / `provider_keys.write` added to the registry and Policy.Can owner-scoped cases (default-deny preserved).

Platform admins see all tenants' keys masked: `GET /api/v1/admin/provider-keys` (+ optional `?account_id=`), mounted behind AuthMiddleware.Require + RequirePlatformAdmin, same shape as /api/v1/admin/providers. Tenants see only their own; cross-account revoke/fetch returns 404 (no existence oracle across tenants).

No endpoint returns key material, encrypted or otherwise: the view builder omits EncryptedAPIKey entirely.

## Audit

Register and revoke emit security-tier events BYOK_KEY_REGISTER / BYOK_KEY_REVOKE through internal/audit (fail-closed tier). Metadata carries label + masked suffix only, never key material.

## Routing integration depth reached (honesty note)

NOT wired. Storage + API + admin visibility landed; routing preference did not, because real BYOK dispatch needs per-request credential injection into the LiteLLM call path (keys currently live in LiteLLM's env/config), which touches edge-api dispatch and LiteLLM proxy behavior beyond this slice. The seam is documented precisely in the follow-up issue filed with this PR. Service exposes Reveal(ctx, accountID, keyID) as the single decryption point the follow-up will call after its own authorization check.

## Tests

Docker toolchain, go test ./apps/control-plane/... -count=1 -short: all green, no regressions (authz, audit, router, cmd/server included).

New tests cover:
- crypto: AES-GCM round trip, wrong-key rejection, tampered-ciphertext rejection, nil-cipher fail-closed both directions, LoadCipher validation (empty/not-base64/too short/too long), MaskSecret table incl. short keys fully masked
- service: register encrypts before persist (fake repo asserts ciphertext, never plaintext), full validation matrix, register without cipher fails closed ErrNotConfigured, Reveal round trip, locked-mode Reveal fails closed, audit events on register+revoke carrying no key material
- http: masked 201 view (no plaintext, no encrypted field), invalid body 400, locked mode 503 byok_not_configured, list shows only own account, cross-account revoke 404 and row untouched, unauthenticated 401, member write 403, admin list sees all accounts, admin account_id filter

Migration exercised by CI's throwaway-Postgres apply (PR #983 path).

## Deployment plumbing: the feature was inert until 08c4c0d41

Worth calling out separately because it made everything above undeliverable in practice.

`HIVE_BYOK_ENC_KEY` reached no `control-plane` container on any shipped profile. The variable was present once in `deploy/docker/docker-compose.yml`, under `edge-api`, which never reads it. `control-plane`, the only service that reads it and the owner of every read and write of `tenant_provider_keys`, was not given it at all. On local, cloud and enterprise alike the process booted into locked mode regardless of what an operator put in `.env`, `POST /api/v1/accounts/current/provider-keys` answered 503 forever, and `.env.example` advertised a switch that could not be thrown.

The same misplacement was also an exposure. `edge-api` holds a database connection, so a compromise of that service would have yielded both the ciphertext and the key that opens it.

Fixed by moving the variable to the `control-plane` environment and removing it from `edge-api`. Verified with `docker compose config` across `local`, `cloud`, `chat`, `agent`, and `enterprise` (base plus override): each resolves exactly one holder, `control-plane`, and never `edge-api`. The enterprise override declares no environment block of its own for control-plane, so the single base entry covers all three deployment shapes.

## Key rotation and what happens to stored ciphertext (open, filed as #1149)

Raised explicitly because the review threads did not cover it and it is a real property of what this ships.

`encrypted_api_key` is sealed with the single process-wide key from `HIVE_BYOK_ENC_KEY`, and the row carries no key id, key version, or algorithm marker. Nothing records which key sealed which ciphertext. If that value is ever changed, for a scheduled rotation, a suspected compromise, or an operator mistake, then every existing row keeps its ciphertext and stays `active`, every `Reveal` on it fails GCM authentication, and nothing surfaces that state. The console still shows the credential as active with its mask intact. There is no re-encryption path, so recovery today means every tenant re-registering every credential by hand.

The failure direction is at least correct: it fails closed, and no plaintext leaks. Nothing dispatches with these credentials yet either, so the blast radius today is a stranded row rather than an outage. Once the routing integration in #1139 lands, a rotation becomes a silent failure of every tenant's own provider routing. Not implemented here; scoped in #1149 (key id column, multiple loadable keys, a re-encryption pass, and a loud signal for rows no loaded key can open). Until that exists, `HIVE_BYOK_ENC_KEY` is write-once per deployment.

## Review round two

Six threads from the CodeRabbit pass on `e16aa045`, each answered on the thread and resolved.

- Admin trailing-slash route: fixed. `AdminMux` serves the exact trailing-slash form the outer router already accepts; deeper paths under that subtree stay 404 rather than returning the cross-tenant list under a URL shaped like a single-item fetch.
- `base_url` credential smuggling: fixed. Userinfo, query strings and fragments are rejected, so `encrypted_api_key` stays the only channel a credential travels on.
- `edge-api` holding the encryption key: fixed, see the deployment section above.
- Repository integration seed: fixed. `seedAccount` returns its user id and the tests use it, so `created_by_user_id` satisfies its foreign key.
- Discarded audit write failures: half fixed. The error is now logged at error level with identifiers only instead of being dropped. The transactional outbox is declined for this slice and filed as #1148: the mutation has already committed by then, so returning the error would invite a retry that registers a second credential, and `audit.Logger` accepts no caller transaction, making the outbox a change to `internal/audit`'s contract that touches every caller.
- Append-only revocation records: declined. `tenant_provider_keys` is entity state rather than a ledger, the immutable history already lives in `internal/audit`, `public.api_keys` revokes by the same in-place transition, and the one-active-credential invariant is a partial unique index over `status` that would move out of Postgres and into application code.

## CI wiring defect found while clearing thread two

The repository integration suite had never executed anywhere. It read its DSN through a const, so `tools/lint-go-db-test-wiring.mjs`, which matches `os.Getenv("<NAME>_TEST_DB_URL")` textually, could not see the file, and `./internal/byok/...` was absent from the integration package list in `ci.yml`. The suite skipped in the `-short` step and was never invoked by the step that has the variable, which is exactly the never-runs trap that lint exists to catch. Both halves fixed: the literal is restored and the package is in the list, so the tenant-isolation proof runs for real and dropping it later fails the guard.

Verified locally the way CI does it, against a throwaway `pgvector/pgvector:pg17` with the CI bootstrap plus the full migration chain applied. Both integration tests pass, and an insert with an unrelated `created_by_user_id` fails `tenant_provider_keys_created_by_user_id_fkey`, confirming the thread-two finding was real rather than theoretical.

One more defect fixed in the same file: the duplicate-`base_url` case asserted `dupURL.ID == uuid.Nil` should fail, which fires precisely when the behaviour is correct.

## Deliberately out of scope

- Console UI for BYOK (separate slice once API lands)
- Routing preference + LiteLLM credential injection (follow-up issue)
- Key rotation endpoint (revoke-and-re-register covers it for now)

## Buglog entry

Three lines, to be appended to `.wolf/buglog.jsonl` on `main` after this merges (issue #873 protocol, never on this branch).

```
{"id":"byok-enc-key-wrong-service","date":"2026-08-25","title":"BYOK encryption key was wired to edge-api, so control-plane never got it and the feature was inert on every profile","error_message":"byok locked mode: HIVE_BYOK_ENC_KEY unset; register answers 503 byok_not_configured (on a deployment where the operator had set it)","root_cause":"HIVE_BYOK_ENC_KEY was added to the edge-api environment block in deploy/docker/docker-compose.yml. edge-api never reads it. control-plane, which owns every read and write of tenant_provider_keys, has an explicit environment allowlist and was not given the variable, so it booted into locked mode on local, cloud and enterprise no matter what was in .env. Giving it to edge-api was also an exposure, since that service holds a database connection and so would have held both the ciphertext and the key.","fix":"Moved the variable to the control-plane environment and removed it from edge-api. Verified with docker compose config across local, cloud, chat, agent and enterprise that exactly one service resolves it and it is never edge-api.","tags":["byok","compose","deployment","secrets","control-plane","inert-feature"]}
{"id":"byok-integration-suite-never-ran","date":"2026-08-25","title":"BYOK repository integration suite was invisible to the db-test-wiring lint and absent from the CI package list","error_message":"no error; the suite reported green by skipping everywhere it was invoked","root_cause":"repository_test.go read its DSN as os.Getenv(repoTestDSNEnv) through a const. tools/lint-go-db-test-wiring.mjs matches os.Getenv(\"<NAME>_TEST_DB_URL\") textually, so the const hid the file from the guard, and ./internal/byok/... was correspondingly missing from the integration package list in ci.yml. It skipped in the -short step and was never invoked by the step that exports HIVE_TEST_DB_URL, so the tenant-isolation proof had never executed. Same never-runs trap as issues #701, #708 and #797.","fix":"Restored the bare os.Getenv(\"HIVE_TEST_DB_URL\") literal and added ./internal/byok/... to the integration list in ci.yml. Verified the lint now counts the pair and that both integration tests pass against a throwaway pgvector/pgvector:pg17 with the CI bootstrap and full migration chain applied.","tags":["ci","testing","camouflage","lint","byok","never-runs"]}
{"id":"byok-repo-test-fk-and-inverted-assert","date":"2026-08-25","title":"BYOK repository integration tests could not have passed: FK violation on created_by_user_id plus an inverted assertion","error_message":"insert or update on table \"tenant_provider_keys\" violates foreign key constraint \"tenant_provider_keys_created_by_user_id_fkey\"","root_cause":"seedAccount inserted an auth.users row but discarded its id, and all three test keys passed a fresh random UUID as Key.CreatedBy. tenant_provider_keys.created_by_user_id is NOT NULL REFERENCES auth.users(id), so every Create failed the foreign key. Separately, the duplicate base_url case asserted `if dupURL.ID == uuid.Nil { t.Fatal(\"conflict path must not return a row\") }`, which fires exactly when the conflict path behaves correctly. Both were invisible because the suite never ran (see byok-integration-suite-never-ran).","fix":"seedAccount now returns (accountID, userID) and every test key uses the returned user id. Inverted the duplicate assertion to !=. Both tests verified passing against real Postgres, and the FK confirmed live by probing it directly.","tags":["testing","postgres","foreign-key","byok","assertion-inverted"]}
```

## Summary by CodeRabbit

* **New Features**
  * Added tenant-managed provider credentials, including registration, masked listing, and revocation.
  * Added platform-admin views for provider credentials across accounts.
  * Added support for custom provider targets and OpenAI-compatible endpoints.
* **Security**
  * Credentials are encrypted at rest and never exposed in responses.
  * Added permission controls and security audit events for credential lifecycle actions.
* **Configuration**
  * Added optional encryption-key configuration. Without it, credential registration is safely disabled; invalid keys prevent startup.


supabase/migrations/20260824_03_tenant_provider_keys.sql
```diff
@@ -0,0 +1,85 @@
...
+  status             TEXT        NOT NULL DEFAULT 'active'
+                                 CHECK (status IN ('active', 'revoked')),
+  created_by_user_id UUID        NOT NULL REFERENCES auth.users(id),
+  created_at         TIMESTAMPTZ NOT NULL DEFAULT now(),
+  updated_at         TIMESTAMPTZ NOT NULL DEFAULT now(),
+  revoked_at         TIMESTAMPTZ NULL,
+  CONSTRAINT tenant_provider_keys_target_check CHECK (
+    (provider_slug IS NOT NULL AND base_url IS NULL)
+    OR
+    (provider_slug IS NULL AND base_url IS NOT NULL)
```
sakibsadmanshajib: [db-review] MEDIUM: the target exclusivity check accepts empty strings. `base_url IS NOT NULL` is satisfied by `''`, so a row with `provider_slug = NULL, base_url = ''` passes this CHECK and lands in the table. The service layer rejects it today (url.Parse of '' yields no scheme), but house convention puts bounds at both layers by design (see the user_memories migration header: bounds live in both layers), and every other writer of this table (SQL editor, future code) gets no protection. Suggested tightening:

```sql
CONSTRAINT tenant_provider_keys_target_check CHECK (
  (provider_slug IS NOT NULL AND char_length(provider_slug) > 0 AND base_url IS NULL)
  OR
  (provider_slug IS NULL AND base_url IS NOT NULL AND char_length(base_url) > 0)
)
```

The same reasoning applies to label: `'   '` (whitespace only) passes `char_length BETWEEN 1 AND 100`; consider a `label <> ''` style guard after trim semantics cannot be expressed in a plain CHECK cheaply, or accept and note it. Empty string for provider_slug is separately blocked here by the FK to custom_providers(slug), which is fine.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,183 @@
...
+
+// validateRegisterInput enforces the storage-boundary rules mirrored by the
+// tenant_provider_keys table CHECK constraints.
+func validateRegisterInput(in RegisterInput) error {
+	if label := strings.TrimSpace(in.Label); len(label) == 0 || utf8.RuneCountInString(label) > 100 {
+		return fmt.Errorf("%w: label required (1 to 100 chars)", ErrValidation)
+	}
+	hasSlug := in.ProviderSlug != nil && *in.ProviderSlug != ""
+	hasURL := in.BaseURL != nil && *in.BaseURL != ""
+	if hasSlug == hasURL { // exactly one target
```
sakibsadmanshajib: [db-review] LOW: a non-nil `ProviderSlug` pointing at an empty string slips past this validation when base_url is present (`hasSlug` is false because of the `!= ""` test, `hasURL` is true, exactly-one passes), but the Key struct keeps the original pointer, so the INSERT writes `provider_slug = ''` and dies on the FK to custom_providers(slug) as an opaque 23503 instead of a 400 validation error. Normalize before building the Key: nil out an empty slug (and empty base_url) so the DB never sees the degenerate value.

apps/control-plane/internal/byok/repository.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	if err != nil {
+		return Key{}, err
+	}
+	return s.key, nil
+}
+
+func (r *pgxRepository) Revoke(ctx context.Context, accountID, id uuid.UUID, now time.Time) (Key, error) {
+	sql := `update public.tenant_provider_keys
+		set status = 'revoked', revoked_at = $3, updated_at = $3
```
sakibsadmanshajib: [db-review] LOW: clock source split. `created_at` comes from the DB (`DEFAULT now()`), but `revoked_at` and `updated_at` here come from the application server clock via $3. With any app/DB skew, a row can end up with `updated_at < created_at`. The sibling usermemories repository uses the DB's `now()` in its UPDATE (`SET updated_at = now()`). Passing time in is fine for testability; if that is deliberate, keep it, but consider `set revoked_at = now(), updated_at = now()` for a single clock source. Also note for future maintainers: this table has no `set_updated_at` trigger (unlike custom_providers and tenant_model_visibility, 20260611_01), so every future UPDATE path must set updated_at manually, exactly as done here. That matches the usermemories precedent, so no change required on that count.

apps/control-plane/cmd/server/main.go
```diff
@@ -708,6 +710,25 @@ func main() {
...
+			cipher, err := byok.LoadCipher(encodedKey)
+			if err != nil {
+				log.Fatalf("control-plane: %s is set but invalid: %v", byok.EnvVarName, err)
+			}
+			byokHandler = byok.NewHandler(
+				byok.NewService(byok.NewPgxRepository(pool), cipher, auditLogger),
+			).WithAccountService(accountsSvc).WithRoleService(roleSvc)
+			log.Println("byok module ready (tenant provider keys encrypted at rest)")
+		} else {
+			log.Println("byok locked mode: HIVE_BYOK_ENC_KEY unset; provider-key routes unmounted")
```
sakibsadmanshajib: [SECURITY][MEDIUM] HIVE_BYOK_ENC_KEY is never plumbed into the control-plane container. deploy/docker/docker-compose.yml builds the control-plane environment from an explicit allowlist and this PR does not touch it, so the variable documented in .env.example is dropped on every shipped deployment (local, cloud, chat, enterprise). The module boots in locked mode forever and the malformed-key boot refusal can never fire in a real deployment. Fail-closed, so no exposure, but an operator who sets the var in .env will reasonably believe BYOK is live when every registration would have failed anyway.

Fix: add `HIVE_BYOK_ENC_KEY: ${HIVE_BYOK_ENC_KEY:-}` to the control-plane service environment block in docker-compose.yml (and confirm the enterprise override file inherits it), so set means enabled and the documented locked mode stays honest.

.env.example
```diff
@@ -1043,6 +1043,16 @@ LITELLM_CONFIG_PATH=/etc/litellm/config.yaml
...
 
+# HIVE_BYOK_ENC_KEY: base64-encoded 32-byte AES-256-GCM key that encrypts
+#   tenant BYOK (bring your own key) provider credentials at rest
+#   (public.tenant_provider_keys). Optional: when unset the control-plane
+#   boots in locked mode, provider-key routes are unmounted, and registration
+#   fails closed with 503 byok_not_configured; nothing is ever stored or
+#   served unencrypted. When set but malformed (not base64 of exactly 32
+#   bytes) the control-plane refuses to start. Generate with:
+#     openssl rand -base64 32
+HIVE_BYOK_ENC_KEY=
```
sakibsadmanshajib: [MEDIUM] Deployment plumbing gap: the control-plane compose service passes an explicit environment allowlist, and HIVE_BYOK_ENC_KEY appears in none of the compose files under deploy/docker/. Setting this variable in .env therefore never reaches the control-plane container on any docker compose deployment this repo ships (local, cloud, enterprise, demo box): control-plane boots in locked mode on every one of them, provider-key routes never mount, and the API slice this PR delivers is unreachable in every deployed environment while .env.example presents it as a working switch. Add HIVE_BYOK_ENC_KEY: ${HIVE_BYOK_ENC_KEY:-} to the control-plane service environment block. Secondary precision issue in this block: when the variable is unset the routes are unmounted entirely, so a real client sees a 404 for an unknown route, not the documented 503 byok_not_configured; the 503 path exists only behind unit test wiring.

apps/control-plane/internal/byok/http.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	key, err := h.svc.Revoke(r.Context(), vc.CurrentAccount.ID, keyID, vc.User.ID)
+	if err != nil {
+		mapServiceError(w, err)
+		return
+	}
+	writeJSON(w, http.StatusOK, keyView(key))
+}
+
+func (h *Handler) handleAdminList(w http.ResponseWriter, r *http.Request) {
```
sakibsadmanshajib: [MEDIUM] handleAdminList performs no authorization of its own: the entire cross tenant protection lives in router wiring that wraps AdminMux() with RequirePlatformAdmin. Nothing inside this package enforces admin access, and the package's own tests call AdminMux() bare with no principal installed (TestAdminListSeesAllAccounts), so a future mount that skips the middleware fails nothing and exposes every tenant's key metadata and last 4 suffix to any authenticated user. This is the most sensitive surface in the PR. Defense in depth: run the same perm check the tenant handlers do (or an explicit platform admin check) inside handleAdminList, or add a router level test proving /api/v1/admin/provider-keys rejects an authenticated non admin JWT.

apps/control-plane/internal/byok/http.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	var body struct {
+		Label        string            `json:"label"`
+		ProviderSlug *string           `json:"provider_slug"`
+		BaseURL      *string           `json:"base_url"`
+		APIKey       string            `json:"api_key"`
+		ModelMap     map[string]string `json:"model_map"`
+	}
+	if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
```
sakibsadmanshajib: [SECURITY][MEDIUM] Register decodes an unbounded request body and accepts an unbounded api_key. `json.NewDecoder(r.Body).Decode(&body)` has no MaxBytesReader, unlike the payments and signupguard handlers that cap bodies for exactly this reason, so a single authenticated owner request can force the server to buffer an arbitrarily large JSON document into memory before validation runs. The 50-entry model_map cap and the 8-char api_key minimum are both applied only after full decode. There is also no maximum on api_key length, so an oversized secret gets AES-GCM encrypted in memory and persisted into encrypted_api_key bytea as-is.

Fix: wrap the body with `http.MaxBytesReader(w, r.Body, 64<<10)` (or similar) before decode, and add an explicit api_key maximum (a few KB is generous) alongside the existing minimum.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,183 @@
...
+		return fmt.Errorf("%w: exactly one of provider_slug or base_url is required", ErrValidation)
+	}
+	if hasSlug && utf8.RuneCountInString(*in.ProviderSlug) > 100 {
+		return fmt.Errorf("%w: provider_slug too long", ErrValidation)
+	}
+	if hasURL {
+		u, err := url.Parse(*in.BaseURL)
+		if err != nil || u.Scheme != "https" && u.Scheme != "http" {
```
sakibsadmanshajib: [SECURITY][LOW] base_url validation accepts a URL with an empty or absent host: `url.Parse("https://")` yields Scheme=https with Host="" and passes the current scheme-only check, so a credential can be stored against a target that can never be dialed and, more importantly, the field carries no host-level validation at all. The http scheme allowance looks deliberate (self-hosted OpenAI-compatible endpoints on a private network are a legitimate Enterprise posture), so keep it but make it explicit.

Fix now: require `u.Host != ""` in validateRegisterInput. Carry into the routing-integration follow-up issue: at Reveal()/dial time, enforce an SSRF denylist (link-local 169.254.0.0/16 and metadata endpoints, loopback, and RFC1918 ranges unless explicitly allowed for self-hosted posture) before any request carrying this credential is sent to a tenant-supplied base_url. The stored primitive for that attack exists as of this PR.

apps/control-plane/internal/byok/repository.go
```diff
@@ -0,0 +1,153 @@
...
+	defer rows.Close()
+	return collectKeys(rows)
+}
+
+func (r *pgxRepository) Get(ctx context.Context, accountID, id uuid.UUID) (Key, error) {
+	sql := `select ` + selectCols + ` from public.tenant_provider_keys
+		where id = $1 and account_id = $2`
+	s := &scanTarget{}
+	err := r.pool.QueryRow(ctx, sql, id, accountID).Scan(s.columns()...)
+	if errors.Is(err, pgx.ErrNoRows) {
```
sakibsadmanshajib: [MEDIUM] The storage layer isolation invariant (tenant A cannot read or revoke tenant B's rows) is only tested against fakeRepo, which re implements the account filter itself. No test executes pgxRepository against Postgres, so the WHERE account_id predicates, the selectCols column list and the scanTarget order mapping are completely uncovered: a dropped predicate or swapped scan target passes the whole suite green while leaking cross tenant rows or mis-scanning encrypted_api_key into another field. CI already bootstraps a throwaway Postgres and exports HIVE_TEST_DB_URL for control-plane suites (ci.yml bootstrap step), so a repository integration test mirroring the existing RLS suites belongs in this PR rather than in a follow-up.

.env.example
```diff
@@ -1043,6 +1043,16 @@ LITELLM_CONFIG_PATH=/etc/litellm/config.yaml
 #   general_settings.store_model_in_db: true. See restart.go TODO comment.
 # LITELLM_CONFIG_MODE=db
 
+# HIVE_BYOK_ENC_KEY: base64-encoded 32-byte AES-256-GCM key that encrypts
```
sakibsadmanshajib: [SECURITY][INFO] Documented locked-mode behavior does not match the wiring. This comment block promises that with HIVE_BYOK_ENC_KEY unset, registration fails closed with 503 byok_not_configured. In the actual wiring the handler is never constructed and the routes are never mounted, so clients get a plain 404 from the mux instead. Both outcomes fail closed, so no exposure, but the doc sets an operator expectation the code cannot meet, and the 503 path is reachable only from unit tests.

Fix: either correct the .env.example text to say routes are absent (404) in locked mode, or mount the tenant routes unconditionally and let the nil-cipher service answer 503 byok_not_configured, which also keeps the error contract discoverable.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,183 @@
...
+			return fmt.Errorf("%w: base_url must be an absolute http(s) URL", ErrValidation)
+		}
+	}
+	if len(in.APIKey) < 8 {
+		return fmt.Errorf("%w: api_key must be at least 8 characters", ErrValidation)
+	}
+	if len(in.ModelMap) > 50 {
+		return fmt.Errorf("%w: model_map supports at most 50 entries", ErrValidation)
```
sakibsadmanshajib: [MEDIUM] base_url validation accepts plain http:// and any host at all, including loopback, RFC1918 and link-local targets. Harmless while nothing dials these URLs, but #1139 will feed exactly this value plus the decrypted credential into LiteLLM dispatch: from that point a tenant can point base_url at internal infrastructure and have the gateway replay their stored credential there (SSRF plus credential exfiltration pivot). The migration's own COMMENT says 'freeform https(s) base_url' while the code is looser than its schema documentation. Decide now: enforce https-only here, add a private or link-local host denial (or state explicitly that #1139 must enforce it before first dial), and make code and migration comment agree.

supabase/migrations/20260824_03_tenant_provider_keys.sql
```diff
@@ -0,0 +1,85 @@
...
+  )
+);
+
+-- One active credential per provider per account: re-registering replaces
+-- nothing silently, the caller must revoke first.
+CREATE UNIQUE INDEX tenant_provider_keys_account_slug_active_idx
```
sakibsadmanshajib: [LOW] Uniqueness is asymmetric: the partial unique index covers active (account_id, provider_slug) rows only, so a single account can hold unlimited simultaneously active freeform credentials pointing at the identical base_url, differing only by label. The PR summary states one ACTIVE credential per provider per account, which the base_url path does not honor, and #1139 will need a deterministic selection rule among duplicates. Either add a matching partial unique index on (account_id, base_url) WHERE status = 'active' AND base_url IS NOT NULL or document the asymmetry in the follow-up issue.

apps/control-plane/internal/byok/http.go
```diff
@@ -0,0 +1,359 @@
...
+
+	"github.com/google/uuid"
+
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/accounts"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/auth"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/authz"
+	"github.com/sakibsadmanshajib/hive/apps/control-plane/internal/platform"
+)
+
+// Handler serves the BYOK surfaces.
```
sakibsadmanshajib: [LOW] Formatting hygiene in the new files: duplicated doc comment above Handler (this line and the next are identical), missing trailing newline at EOF in http.go and service.go, out-of-order import block in cmd/server/main.go (byok inserted ahead of apikeys breaks alphabetical order), and unaligned struct fields in types.go. gofmt -l flags all of these. There is no gofmt gate in CI (long standing drift in authz/policy_test.go and audit/actions_test.go proves it), so nothing goes red, but new code should land gofmt clean.

supabase/migrations/20260824_03_tenant_provider_keys.sql
sakibsadmanshajib: Fixed in 0e6752eff: CHECK rejects empty-string targets at the storage boundary.

apps/control-plane/internal/byok/service.go
sakibsadmanshajib: Fixed in 0e6752eff: empty target pointers normalize to nil before validation (400, not FK 23503).

apps/control-plane/internal/byok/repository.go
sakibsadmanshajib: Fixed in 0e6752eff: revoke timestamps now come from the DB clock.

apps/control-plane/cmd/server/main.go
sakibsadmanshajib: Fixed in 1b47ef752: HIVE_BYOK_ENC_KEY added to the control-plane compose allowlist.

.env.example
sakibsadmanshajib: Fixed in 1b47ef752: compose allowlist entry plus mount-always wiring so locked mode answers the documented 503.

apps/control-plane/internal/byok/http.go
sakibsadmanshajib: Fixed in 1b47ef752: handleAdminList enforces PermPlatformAdmin itself; non-admin 403 test added.

apps/control-plane/internal/byok/http.go
sakibsadmanshajib: Fixed in 1b47ef752: MaxBytesReader 64KiB on register + 4096-char api_key cap.

apps/control-plane/internal/byok/service.go
sakibsadmanshajib: Host requirement added in 1b47ef752; dial-time SSRF denylist pinned in #1139.

apps/control-plane/internal/byok/repository.go
sakibsadmanshajib: Added repository_test.go vs real SQL (HIVE_TEST_DB_URL): isolation, scan mapping, uniques, CHECK.

.env.example
sakibsadmanshajib: Wiring changed in 1b47ef752: routes mount unconditionally, locked mode gives the documented 503.

apps/control-plane/internal/byok/service.go
sakibsadmanshajib: http kept deliberately (Enterprise self-hosted); comment aligned, SSRF dial-time requirement pinned in #1139.

supabase/migrations/20260824_03_tenant_provider_keys.sql
sakibsadmanshajib: Partial unique index on (account_id, base_url) active-only added in 1b47ef752; covered by real-SQL test.

apps/control-plane/internal/byok/http.go
sakibsadmanshajib: All new files gofmt clean as of 1b47ef752.

apps/control-plane/internal/byok/http.go
```diff
@@ -0,0 +1,372 @@
...
+// AdminMux routes the platform-admin surface:
+//
+//	GET /api/v1/admin/provider-keys           all keys, masked
+//	GET /api/v1/admin/provider-keys?account_id={uuid}
+//
+// Mount behind AuthMiddleware.Require + RequirePlatformAdmin.
+func (h *Handler) AdminMux() http.Handler {
+	mux := http.NewServeMux()
+	mux.HandleFunc(adminPrefix, h.handleAdminList)
+	return mux
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in 08c4c0d41.

`AdminMux` now registers the trailing-slash subtree in addition to the bare prefix, so `GET /api/v1/admin/provider-keys/` is served here rather than 404'd by this nested mux after the outer router has already accepted it.

One deliberate narrowing on top of the suggestion: the subtree registration also catches deeper paths, and those return 404 rather than the list. There is no single-item admin endpoint, so serving the whole cross-tenant list under a URL shaped like one would be a quiet surprise for anyone probing it.

Two tests cover this. `TestAdminListServesTrailingSlash` asserts 200 with the same two keys the bare path returns, and `TestAdminUnknownSubpathIs404` pins the deeper-path behaviour.

apps/control-plane/internal/byok/repository_test.go
```diff
@@ -0,0 +1,195 @@
...
+		accountID, "byok-repo-test-"+accountID.String(), userID); err != nil {
+		t.Fatalf("seed accounts: %v", err)
+	}
+	t.Cleanup(func() {
+		cleanupCtx := context.Background()
+		_, _ = pool.Exec(cleanupCtx, `DELETE FROM public.tenant_provider_keys WHERE account_id = $1`, accountID)
+		_, _ = pool.Exec(cleanupCtx, `DELETE FROM public.accounts WHERE id = $1`, accountID)
+		_, _ = pool.Exec(cleanupCtx, `DELETE FROM auth.users WHERE id = $1`, userID)
+	})
+	return accountID
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in 08c4c0d41. `seedAccount` now returns `(accountID, userID)` and all three test keys use the returned user id for `CreatedBy`.

Verified against real Postgres rather than by reading. With the CI bootstrap plus the full migration chain applied to a throwaway pgvector/pgvector:pg17 database, an insert carrying an unrelated `created_by_user_id` fails `tenant_provider_keys_created_by_user_id_fkey`, exactly as you said, and both integration tests pass with the seeded id.

Two related defects surfaced while making that verification possible, both fixed in the same commit.

First, this suite could never have run in CI at all. It read its DSN through a const, and `tools/lint-go-db-test-wiring.mjs` matches `os.Getenv("<NAME>_TEST_DB_URL")` textually to pair a gated suite with the workflow step that names its package. The indirection hid the file from that guard, and `./internal/byok/...` was correspondingly absent from the integration package list in `ci.yml`. The literal is back and the package is in the list, so the suite now executes and dropping it later fails the guard.

Second, the duplicate-`base_url` case asserted `if dupURL.ID == uuid.Nil { t.Fatal("conflict path must not return a row") }`, which fires precisely when the behaviour is correct. Inverted to `!=`.

apps/control-plane/internal/byok/repository.go
```diff
@@ -0,0 +1,156 @@
...
+		return Key{}, err
+	}
+	return s.key, nil
+}
+
+func (r *pgxRepository) Revoke(ctx context.Context, accountID, id uuid.UUID) (Key, error) {
+	sql := `update public.tenant_provider_keys
+		set status = 'revoked', revoked_at = now(), updated_at = now()
+		where id = $2 and account_id = $1 and status = 'active'
+		returning ` + selectCols
```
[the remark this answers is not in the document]
sakibsadmanshajib: Declining this one, with reasoning rather than silence.

The coding guideline's append-only clause governs the ledger. `tenant_provider_keys` is entity state for a live credential, not a ledger of events, and `status` is that entity's lifecycle field.

The immutable history this asks for already exists, in the place the codebase keeps it. Register and revoke both emit security-tier events (`BYOK_KEY_REGISTER`, `BYOK_KEY_REVOKE`) through `internal/audit`, which is append-only and is where the transition record belongs. Duplicating that history into the credential table would give us two records of the same event with different durability guarantees.

There is also direct precedent one package over. `public.api_keys` is the sibling credential table and revokes by the same in-place transition (`apps/control-plane/internal/apikeys/repository.go` sets `status = 'revoked'`, `revoked_at`, `replaced_by_key_id`). Diverging here would leave the codebase with two shapes for one lifecycle.

Finally, the invariant would get weaker, not stronger. The partial unique index that enforces one active credential per provider per account is defined over the row's `status`. Deriving active status from an append-only history moves that uniqueness rule out of Postgres and into application code.

If a credential-history surface becomes a product requirement, that is a schema design of its own rather than a review fix, and I would rather do it deliberately than as a side effect of this slice.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,204 @@
...
+		Actor:        audit.Actor{ID: actor, Type: audit.ActorUser},
+		Action:       action,
+		ResourceType: "tenant_provider_key",
+		ResourceID:   k.ID.String(),
+		Severity:     audit.SeverityNotice,
+		After: map[string]string{
+			"label":     k.Label,
+			"key_last4": k.KeyLast4,
+		},
+	})
```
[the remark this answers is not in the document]
sakibsadmanshajib: Half fixed in 08c4c0d41, half declined with reasoning.

Fixed: `emitAudit` no longer discards the error. A failed write is logged at error level carrying identifiers only, meaning action, account id, resource id and the error string, and deliberately not the label, the mask, or anything derived from key material, because the process log has a wider audience than the audit table.

Declined: the transactional outbox, in this pull request.

Your own comment names the reason returning the error is not enough, that the mutation has already committed. Agreed, and that is also why returning it would be actively harmful here. The only sensible caller response to an error from `Register` is a retry, and a retry after a committed insert registers a second credential.

The outbox is not available to this slice either. `audit.Logger` owns its own writers (the fail-closed `Sync` path for security-tier actions, the WAL path for the LLM tier) and accepts no caller-supplied transaction, so committing an outbox row alongside the credential mutation is a change to `internal/audit`'s contract that affects every caller of it. Building a BYOK-local outbox instead would create a second audit delivery path with different semantics from every other security event in the service, which is a worse outcome than the honest gap. Filed as #1148 rather than left implied.

`TestAuditWriteFailureDoesNotFailTheMutation` pins the contract that a failing audit backend leaves the mutation successful and attempts exactly one write per mutation, so a future change that starts swallowing or duplicating these calls fails a test.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,204 @@
...
+		// http is allowed deliberately: self-hosted OpenAI-compatible
+		// endpoints on a private network are a legitimate Enterprise posture.
+		// Dial-time egress restrictions (SSRF denylist) are the routing
+		// integration's job and are pinned in issue #1139.
+		if err != nil || u.Scheme != "https" && u.Scheme != "http" {
+			return fmt.Errorf("%w: base_url must be an absolute http(s) URL with a host", ErrValidation)
+		}
+		if u.Host == "" {
+			return fmt.Errorf("%w: base_url must include a host", ErrValidation)
+		}
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in 08c4c0d41. `base_url` is now rejected when `u.User != nil`, `u.RawQuery != ""` or `u.Fragment != ""`.

One departure from the suggested diff: the check is a separate condition after the scheme and host checks rather than folded into the scheme condition. Folded in, an operator who pastes a perfectly valid URL with a query string is told it "must be an absolute http(s) URL with a host", which it is. The separate check names the actual problem instead.

Four cases added to `TestRegisterValidation`: userinfo alone, a `user:password` pair, a query string, and a fragment.

deploy/docker/docker-compose.yml
```diff
@@ -49,6 +49,9 @@ services:
       SUPABASE_JWT_AUDIENCE: ${SUPABASE_JWT_AUDIENCE:-authenticated}
       SUPABASE_JWKS_URL: ${SUPABASE_JWKS_URL:-}
       OWUI_SHIM_KEY: ${OWUI_SHIM_KEY:-}
+      # Tenant BYOK credential encryption (AES-256-GCM at rest). Unset keeps
+      # the module in locked mode: register answers 503 byok_not_configured.
+      HIVE_BYOK_ENC_KEY: ${HIVE_BYOK_ENC_KEY:-}
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in 08c4c0d41, and the defect was larger than the diff alone could show.

The variable was on `edge-api` and nowhere else. `control-plane`, the only service that reads it and the owner of every read and write of `tenant_provider_keys`, never received it on any profile. So this line was both the exposure you identified and a total functional blocker: control-plane booted into locked mode on local, cloud and enterprise regardless of what an operator put in `.env`, the register route answered 503 forever, and `.env.example` advertised a switch that could not be thrown.

`HIVE_BYOK_ENC_KEY` is now a single entry in the `control-plane` environment and absent from `edge-api`, so a compromise of the service that holds a database connection but does no BYOK work cannot pair the ciphertext with the key that opens it.

Verified rather than asserted. `docker compose config` across `local`, `cloud`, `chat`, `agent`, and `enterprise` (base plus the enterprise override) each resolve exactly one holder of the variable, `control-plane`, and never `edge-api`. The enterprise override adds no environment block of its own for control-plane, which is why the single base entry covers all three deployment shapes.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,222 @@
...
+		return fmt.Errorf("%w: exactly one of provider_slug or base_url is required", ErrValidation)
+	}
+	if hasSlug && utf8.RuneCountInString(*in.ProviderSlug) > 100 {
+		return fmt.Errorf("%w: provider_slug too long", ErrValidation)
+	}
+	if hasURL {
+		u, err := url.Parse(*in.BaseURL)
+		// http is allowed deliberately: self-hosted OpenAI-compatible
+		// endpoints on a private network are a legitimate Enterprise posture.
+		// Dial-time egress restrictions (SSRF denylist) are the routing
```
sakibsadmanshajib: SECURITY REVIEW (independent, not CodeRabbit — that stream was rate-limited on this push): MEDIUM, non-blocking for this PR.

Confirmed: no code in this PR ever dials `base_url` (grepped the whole `internal/byok` package for `http.Client`/`http.Get`/`net.Dial`; none exist). `Reveal()` only decrypts and returns a string. So there is no exploitable SSRF in #1138 itself, and the base_url validation here correctly does not attempt an egress denylist it has no way to enforce safely at this layer.

But this is the value that will later be dialed by #1139, and right now `validateRegisterInput` accepts `http://169.254.169.254/...`, `http://127.0.0.1:...`, `http://10.0.0.1:...` etc. with no restriction. Flagging this explicitly so it does not slip through as "already reviewed": before #1139 adds the actual dispatch call, it must land a dial-time SSRF denylist (loopback, link-local incl. cloud metadata ranges, RFC1918/ULA, and DNS-rebinding-safe resolution, i.e. validate the resolved IP at dial time not just the hostname at registration time) as a hard blocking requirement of that PR's own security review, not an optional follow-up. Issue #1139 already names "dial-time SSRF requirement" — good — just confirming it is understood as a merge gate for that PR, not a nice-to-have.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,222 @@
...
+		ResourceID:   k.ID.String(),
+		Severity:     audit.SeverityNotice,
+		After: map[string]string{
+			"label":     k.Label,
+			"key_last4": k.KeyLast4,
+		},
+	}); err != nil {
+		// Identifiers only. No label, no mask, no key material: this line goes
+		// to the process log, which has a wider audience than the audit table.
+		slog.ErrorContext(ctx, "byok audit write failed, credential mutation already committed",
```
sakibsadmanshajib: SECURITY REVIEW: INFO, non-blocking.

Verified: this log line and the two others in the package (`http.go` platform-admin lookup failure, `http.go` `writeOpaque`) carry only identifiers and `err.Error()` strings. Traced every error that can reach these three sites (DB connectivity/FK/unique-violation errors, cipher nonce-generation failure) and none of them can embed `encrypted_api_key` or plaintext `api_key`: no CHECK/UNIQUE constraint in the migration touches those columns, and `Encrypt`'s only error path is `crypto/rand` failure, not the plaintext itself. Claim verified.

Separate note on the fail-open-after-commit shape itself (not a defect in this diff, already tracked in #1148, applies to every existing security-tier audit call site including API_KEY_ISSUE/REVOKE, not BYOK-specific): `audit.SyncWriter.Write` serializes ALL security-tier writes platform-wide behind a single per-calendar-month `pg_advisory_xact_lock`, retried up to 3 times on `serialization_failure`. Under real contention (not necessarily adversarial — just concurrent registrations from unrelated tenants in the same month) this can legitimately exhaust retries, and today that means the credential mutation stands with only a log line, no durable trace. This is shared infra risk, not something #1138 should fix, but worth naming for whoever picks up #1148: it is not purely a hypothetical DB-outage scenario, ordinary load can hit it too.

apps/control-plane/internal/byok/service.go
```diff
@@ -0,0 +1,222 @@
...
+// indistinguishable from missing rows (ErrNotFound either way).
+func (s *Service) Get(ctx context.Context, accountID, id uuid.UUID) (Key, error) {
+	return s.repo.Get(ctx, accountID, id)
+}
+
+// Reveal decrypts and returns the stored credential for the given key id.
+// It is the single decryption point the routing integration (follow-up
+// issue) will call after its own authorization check. Fails closed when no
+// encryption key is configured.
+func (s *Service) Reveal(ctx context.Context, accountID, id uuid.UUID) (string, error) {
```
sakibsadmanshajib: SECURITY REVIEW: judgment call on #1149 (key rotation gap), as requested.

Acceptable to merge #1138 with this gap open. Reasoning: `Reveal` is not called from anywhere outside this package yet (grepped call sites — zero), so today a rotation strands rows without an outage; the failure mode is fail-closed (`Decrypt` returns an error on auth failure, no plaintext leak, no silent wrong-key decrypt since GCM authenticates). The blast radius is exactly what the PR body claims: dead rows, not a live incident. That changes the moment #1139 wires real dispatch through `Reveal`, at which point an unannounced rotation becomes a silent per-tenant routing outage.

Hard requirement, not a suggestion: #1149 (key id/version column, multi-key load, re-encryption pass, loud unreadable-row signal) must land before or together with #1139, not after. Recommend adding an explicit blocking-dependency note on #1139's issue so it cannot merge dispatch-through-Reveal while #1149 is still open.

apps/control-plane/internal/byok/http.go
```diff
@@ -0,0 +1,384 @@
...
+		writeError(w, http.StatusNotFound, "not_found", "provider key not found")
+	default:
+		writeOpaque(w, "request could not be completed", err)
+	}
+}
+
+// keyView builds the customer-safe JSON view of a key row. EncryptedAPIKey is
+// deliberately absent: no endpoint in this package ever returns key material,
+// encrypted or otherwise.
+func keyView(k Key) map[string]any {
```
sakibsadmanshajib: SECURITY REVIEW: VERIFIED, no issue.

Confirmed `keyView` is the only place any `Key` reaches a response, and it omits `EncryptedAPIKey`. Grepped every `writeJSON`/`Encode(` call site in this file: all route through `keyView` or a hand-built map, none marshal the raw `Key` struct. No `%v`/`%+v` format verb appears anywhere in the non-test package files, so there's no accidental struct dump via a log/error format string either. Claim holds.

apps/control-plane/internal/byok/crypto.go
```diff
@@ -0,0 +1,106 @@
...
+	raw, err := base64.StdEncoding.DecodeString(encoded)
+	if err != nil {
+		return nil, fmt.Errorf("%w: not valid base64", ErrInvalidKeyMaterial)
+	}
+	return NewCipher(raw)
+}
+
+// Encrypt seals plaintext with a fresh random nonce; output layout is
+// nonce || ciphertext || tag. Never log the return value.
+func (c *Cipher) Encrypt(plaintext string) ([]byte, error) {
```
sakibsadmanshajib: SECURITY REVIEW: VERIFIED, no issue.

AES-256-GCM via stdlib, fresh `crypto/rand` nonce every call (no counter, no reuse risk across encryptions), standard 12-byte GCM nonce size from `NewGCM`, authenticated (tag from `Seal`, verified by `Open` on decrypt with no best-effort fallback). Key material validated to exactly 32 bytes at `NewCipher`/boot; unset key correctly produces `nil *Cipher` (locked mode, routes still mounted, writes fail closed with 503) while a set-but-malformed key correctly kills the process at `main.go`'s `log.Fatalf` before it ever serves traffic. `rand.Read` failure is checked and propagated rather than silently proceeding with a zero/weak nonce. No issues found in the crypto primitive itself.

deploy/docker/docker-compose.yml
```diff
@@ -479,6 +479,18 @@ services:
...
+      # public.tenant_provider_keys; deliberately NOT given to edge-api, which
+      # also holds a database connection, so a compromise of that service
+      # cannot pair the ciphertext with the key that opens it. Unset keeps the
+      # module in locked mode: register answers 503 byok_not_configured and
+      # nothing is ever stored unencrypted. Set but malformed (not base64 of
+      # exactly 32 bytes) and control-plane refuses to start. This one line is
+      # what makes the feature reachable on a compose deployment at all; the
+      # enterprise override adds no environment block of its own for
+      # control-plane, so local, cloud and enterprise all inherit it here.
+      HIVE_BYOK_ENC_KEY: ${HIVE_BYOK_ENC_KEY:-}
```
sakibsadmanshajib: SECURITY REVIEW: VERIFIED, no issue.

Confirmed via `git diff main...HEAD` on this file: `HIVE_BYOK_ENC_KEY` now appears exactly once, here, under `control-plane`. Grepped the whole file and `docker-compose.enterprise.yml`: no other occurrence, and specifically not under `edge-api`. This closes the exposure the PR description called out (edge-api holding both the ciphertext's DB connection and the key that opens it).

sakibsadmanshajib (commented):
## Independent security review, PR #1138 (CodeRabbit stream SKIPPED — rate limited on this push, treat as no coverage on the latest commits, not a clean pass)

Reviewed the full diff against the pushed head (`8564c69`), not the builder's self-report. Read every file in `internal/byok`, the migration, `router.go`, `main.go`, the compose diff, and the unit + real-Postgres integration tests. Cross-checked issues #1139, #1148, #1149 against their actual filed content, not just the PR body's summary of them.

### Findings by severity

**HIGH:** none.

**MEDIUM:**
1. SSRF on tenant-supplied `base_url` (service.go:180) — not exploitable in this PR (nothing in `internal/byok` dials out; `Reveal` only decrypts a string), but the value stored here will be dialed by #1139 with zero restriction today. Flagging as a hard blocking requirement for #1139's own review: dial-time SSRF denylist (loopback, link-local/metadata, RFC1918/ULA, resolved-IP check, not just hostname), not optional.

**LOW / INFO (verified, no action needed on this PR):**
2. Key-never-returned claim: verified. `keyView` is the sole response path for any `Key`, omits `EncryptedAPIKey`; no raw struct marshal, no `%v`/`%+v` dump anywhere in the package.
3. Key-never-logged claim: verified. All three log call sites in the package traced end to end; none of the errors that can reach them can carry key material (no constraint in the migration touches `encrypted_api_key` or the plaintext, and `Encrypt`'s only error path is a `crypto/rand` failure).
4. Cross-tenant isolation: verified at both the unit level (fake repo, includes the admin trailing-slash and deep-subpath 404 tests added for the CodeRabbit round) and against real Postgres (`TestRepositoryIsolationAgainstRealSQL`, now actually wired into CI's integration package list and `tools/lint-go-db-test-wiring.mjs` after the const-vs-literal fix). Cross-account Get/Revoke both return `ErrNotFound`/404, victim row unmodified, no existence oracle. `AdminMux`'s trailing-slash and deeper-subpath routes tested and correct.
5. Encryption primitive: AES-256-GCM, fresh random nonce per call, authenticated (no best-effort decrypt path), 32-byte key enforced, malformed key kills boot via `log.Fatalf`, unset key produces documented locked mode (routes mounted, 503 on write). No issues.
6. Compose fix: verified directly in the diff. `HIVE_BYOK_ENC_KEY` appears exactly once, under `control-plane`, not under `edge-api`, not duplicated elsewhere, not in the enterprise override.
7. Audit-write-failure-doesn't-block-the-mutation: consistent with the pre-existing pattern for every other security-tier audit action (not BYOK-specific), honestly tracked in #1148 with real reasoning for why an outbox wasn't built inline. Noted for whoever picks up #1148 that the shared `SyncWriter`'s global per-month advisory lock makes retry exhaustion under ordinary concurrent load plausible, not just a DB-outage edge case — worth folding into that issue's scope, not blocking here.

**Rotation gap (#1149):** judged acceptable to merge with, per the brief's ask. `Reveal` has zero callers today, so the failure mode is stranded rows, not a live outage, and it fails closed (GCM auth, no plaintext leak on wrong key). This stops being acceptable the moment #1139 wires dispatch through `Reveal`; recommend an explicit blocking-dependency note so #1139 cannot merge live routing while #1149 is still open.

### Verdict: DO NOT BLOCK.

No CRITICAL or HIGH finding in the shipped diff. The one MEDIUM (SSRF) is inert in this PR because nothing here makes a network call, and it is already tracked as a named requirement of the follow-up that will. All of the builder's specific claims (key never returned, key never logged, cross-tenant isolation, compose fix) were independently verified against the code and tests, not taken on trust. Recommend merging #1138 as is, and treating the SSRF denylist as a hard merge gate on #1139 specifically.

apps/control-plane/internal/byok/service.go
sakibsadmanshajib: Agreed on both halves, and confirming the gate explicitly so it cannot be read later as an optional follow-up.

**Not exploitable in this PR, and deliberately so.** Nothing in `internal/byok` dials anything. There is no `http.Client`, no `http.Get`, no `net.Dial` anywhere in the package, and `Reveal` decrypts a stored blob and returns a string to its caller. `base_url` is inert data here: it is validated at write time, stored, and echoed back masked alongside the credential. The registration-time checks that do exist (reject userinfo, query string, fragment) are there to stop a secret being smuggled into a plaintext column, not to be a network control, and I would rather they not pretend to be one. A registration-time hostname denylist at this layer would be security theatre: it cannot see the resolved address, cannot survive a DNS rebind between registration and dial, and would give the next reader false confidence that egress was already handled.

**Confirmed as a merge gate on #1139, not a nice-to-have.** The blocking requirements are pinned on that issue at https://github.com/sakibsadmanshajib/hive/issues/1139#issuecomment-5410394039 and I am treating them as such:

- Deny loopback, link-local including cloud metadata ranges, and RFC1918 or ULA.
- Check the **resolved IP at dial time**, not the hostname at registration time, and re-check after every redirect. Your DNS rebinding point is the reason the check has to live at the dial, and a redirect to a private address defeats a name-based denylist just as thoroughly.
- Decide deliberately whether redirects are followed at all, rather than inheriting whatever the HTTP client does by default.

The control is load bearing precisely because plain `http` stays allowed on purpose here, for self-hosted OpenAI-compatible endpoints on a private network in the Enterprise posture. That decision is what makes dial-time enforcement the only place the restriction can correctly live, so it is #1139's own security review that has to gate on it.

apps/control-plane/internal/byok/service.go
sakibsadmanshajib: Thanks for tracing the error paths rather than taking the claim on trust. Agreed on the verification, and the `Encrypt` observation is the load-bearing one: its only error path is `crypto/rand` failure, so the plaintext cannot ride out on an error string even in principle.

On the `SyncWriter` detail: that is a genuinely useful correction and it changes how the follow-up should be scoped, so I have folded it into #1148 rather than into this PR. Comment: https://github.com/sakibsadmanshajib/hive/issues/1148#issuecomment-5410404979

The part worth restating there was your point that a single per-calendar-month `pg_advisory_xact_lock` serializes every security-tier write platform wide, so three retries can be exhausted by ordinary concurrent load from unrelated tenants with no outage and no adversary involved. I had been carrying this as a rare incident-time path, which understates it. I also noted the design consequence: committing an outbox row inside the caller's own transaction takes that global lock off the critical path of the mutation entirely, since delivery moves to a drainer.

Nothing changes in this PR. The disposition here stays what it was, matching the pre-existing shape at every other security-tier call site including `API_KEY_ISSUE` and `API_KEY_REVOKE`, with the failure logged loudly rather than discarded, and the durable fix owned by #1148.

apps/control-plane/internal/byok/service.go
sakibsadmanshajib: Accepted, and confirming the merge judgment matches mine for the same reasons.

`Reveal` has zero callers outside this package today, which I re-checked, so a rotation strands rows rather than causing an outage. The failure is fail closed: GCM authenticates, so `Decrypt` returns an error rather than producing a wrong-key plaintext, and nothing leaks. Dead rows, not a live incident, which is what the PR body claims and what I wanted verified independently rather than asserted.

Agreed without reservation that this stops being true the moment #1139 wires dispatch through `Reveal`, at which point an unannounced rotation becomes a silent per-tenant routing outage rather than an inert row. #1149 is recorded as a blocking dependency on #1139 for exactly that reason, in the same pinned comment as the SSRF gate: https://github.com/sakibsadmanshajib/hive/issues/1139#issuecomment-5410394039

Scope on #1149 is as you list it: key id or version column, multiple loadable keys so a rotation can decrypt with the old and re-seal with the new, a re-encryption pass over existing rows, and a loud signal for any row no loaded key can open. Until it lands, `HIVE_BYOK_ENC_KEY` is write-once per deployment and the PR body says so.

apps/control-plane/internal/byok/http.go
sakibsadmanshajib: Confirming, and the two checks you added beyond mine are the ones that make this hold rather than merely look right.

I had verified `keyView` omits `EncryptedAPIKey`. What I had not done is prove no other path marshals a raw `Key`, and your sweep of every `writeJSON` and `Encode(` call site is what closes that. The `%v` and `%+v` sweep is the better catch of the two: a struct dump through a format verb in a log or wrapped error would bypass the response layer entirely and never show up in a response-body assertion, so it is exactly the leak the existing tests could not have caught.

Noting it for anyone touching this later: those two properties, one response builder and no struct-wide format verbs, are what the response-leak assertions in `http_test.go` actually rest on. They ban the fixture literal, the substring `encrypted`, and the `api_key"` key from register responses, and ban `encrypted_api_key` presence from both list surfaces, which is a test of the builder rather than of every path a `Key` could take.

apps/control-plane/internal/byok/crypto.go
sakibsadmanshajib: Confirming, and thanks for checking the primitive itself rather than accepting "AES-256-GCM via stdlib" as sufficient on its own.

The two properties I most wanted an independent pair of eyes on are the ones you name. A fresh `crypto/rand` nonce per call with no counter, since GCM nonce reuse across encryptions under one key is the failure that turns this from sound to catastrophic and is invisible in a passing round-trip test. And the `rand.Read` error being propagated rather than swallowed, since silently proceeding with a zero or partially filled nonce is the same failure arriving through the error path.

The boot behaviour split is deliberate and I am glad it reads correctly from the outside: an unset key yields a `nil *Cipher` and locked mode, where the routes still mount so the failure is a discoverable `503 byok_not_configured` rather than a mystery 404, and writes fail closed so no plaintext can ever land. A set but malformed key is a config error and kills the process at `log.Fatalf` in `main.go` before it serves any traffic, rather than degrading quietly into locked mode where an operator would believe the key was in effect.

deploy/docker/docker-compose.yml
sakibsadmanshajib: Confirming, and I would rather this one was checked against the diff than against my description of it, since the original defect was precisely that the file said something different from what the deployment did.

Your grep matches my own verification from the other direction. I checked the resolved output rather than the source text: `docker compose config` across `local`, `cloud`, `chat`, `agent`, and `enterprise` (base plus `docker-compose.enterprise.yml`), parsing each result and listing every service whose resolved environment contains the key. Exactly one holder in every profile, `control-plane`, and `edge-api` in none of them.

The resolved-output check matters here beyond belt and braces: `docker-compose.enterprise.yml` declares no `environment:` block for `control-plane` at all, only `depends_on`, so the enterprise profile inherits the single base entry. Reading the override file alone would suggest enterprise gets nothing, which is how the misplacement went unnoticed in the first place.

PR #1152, opened 2026-08-25
fix: forward Anthropic prompt-caching cache_control through the gateway

sakibsadmanshajib:
## Problem

Hive's Anthropic-compatible `/v1/messages` endpoint (`apps/edge-api/internal/anthropic/`) had no `cache_control` field anywhere on its request or response types. `translate_request.go` rebuilds the Anthropic request field by field into an internal OpenAI-shaped `OAIRequest`, and `cache_control` was never one of the fields carried across, so it was silently dropped before the request ever reached LiteLLM. Every agent client that relies on Anthropic prompt caching (Claude Code, Cline, Cursor, Aider) got zero caching through Hive and paid full input price on every turn.

## What this PR does

**Request side.** Adds a `CacheControl` struct (`{"type":"ephemeral","ttl":"5m"|"1h"}`) and wires it through every documented placement:
- content block (text, image, tool_use, tool_result) — `ContentBlock.CacheControl`
- system prompt block — `SystemField` now retains the original `Blocks` (previously collapsed to one concatenated string, losing any block-level data)
- tool definition — `Tool.CacheControl` (caching one tool in the array caches the whole array up to it; Anthropic has no separate "group" field)
- request root — `MessagesRequest.CacheControl`

All four are mirrored onto the internal OAI-shaped types (`OAIContentPart`, `OAIMessage`, `OAIToolCall`, `OAITool`, `OAIRequest`) with identical JSON shape, since the marshaled `OAIRequest` body is forwarded essentially byte-for-byte to LiteLLM (`litellm_client.go`'s `rewriteModel` only ever touches the `"model"` key).

**The flattening bug.** Two spots in `translate_request.go` collapsed a content-block array down to a plain string when it looked simple enough (a lone text block; a lone text block mixed with tool calls). A flat string cannot carry a per-block `cache_control`, so a request that DID set a breakpoint on that block would have had it silently destroyed even with the field wired in. Both collapse sites now check for a cache breakpoint first and keep the block-array form when one is present, otherwise reproducing the exact prior output. `TestToOAIRequest_CacheControl_SingleTextBlock_NotFlattenedToString` is a dedicated regression test for this; I confirmed it actually fails without the fix (reverted the guard locally, test failed with the flattened bare string, restored the fix, test passes).

**Response side — exclusive-shape echo contract.** OpenRouter (Hive's only real dispatch path) reports usage in the OpenAI-compatible INCLUSIVE convention: `prompt_tokens` already contains the cached and written tokens. Anthropic-native clients (Claude Code's cost display included) expect the EXCLUSIVE convention: `input_tokens` as the fresh/uncached remainder only, plus `cache_creation_input_tokens` and `cache_read_input_tokens` reported alongside it. `freshInputTokens` in `translate_response.go` does that conversion (`fresh = prompt - cached - written`, clamped to 0, never negative) for both the non-streaming response and the streaming path (`message_delta.usage` in `stream.go`, since that relay only knows the terminal numbers once the upstream's usage-bearing chunk arrives — `message_start` fires before that and stays at its pre-existing zero, consistent with how `output_tokens` already worked before this change).

**session_id.** Not part of Anthropic's own `/v1/messages` schema, but OpenRouter uses it for sticky provider routing, which is what keeps a cache warm across turns. Added as passthrough on `MessagesRequest`/`OAIRequest` (truncated at 256 chars per OpenRouter's documented ceiling) so a client/proxy that sends it is no longer silently dropped by the same field-by-field rebuild.

## Merge coordination (feat/cache-aware-billing)

This PR's `OAIUsage.PromptTokensDetails` (`{cached_tokens, cache_write_tokens}`) is a **capture-only** struct in the `anthropic` package. The bytes it unmarshals come from the inference package's own re-marshal of `inference.UsageResponse` (`normalizeChatCompletion` in `chat_completions.go`, and the typed chunk round-trip in `stream.go`) — both of which do a **typed** unmarshal-then-remarshal, which silently drops any JSON field their own struct doesn't declare. Today `inference.UsageResponse` has `PromptTokensDetails.CachedTokens` but no `CacheWriteTokens`, and OpenRouter's `cache_creation_input_tokens`/`cache_read_input_tokens` fields (if reported in Anthropic-native shape at all through the OpenRouter hop) have no matching field on that struct either.

**Net effect: the response-side cache token plumbing in this PR is real and tested against the `anthropic` package's own types, but it cannot carry live data end-to-end until `feat/cache-aware-billing` adds a matching `CacheWriteTokens` (and whatever else it needs) to `inference.UsageResponse` with the same JSON field name (`cache_write_tokens`) I used here.** I stayed out of `pricing.go`, `stream.go` (inference package), `orchestrator.go` and `types.go` per the coordination boundary; this is the merge point to close by hand once both branches land — I'd suggest whichever branch merges second checks the other's final field names against this PR's `OAIPromptTokensDetails`.

## Verified against the actual upstream path

- **Request side is safe today, independent of the billing coordination above.** `chat_completions.go`'s `handleChatCompletions` forwards the original raw request `body` bytes downstream (not a re-marshal of its own typed `ChatCompletionRequest`, whose `Messages`/`Tools` fields are `json.RawMessage` anyway), and `litellm_client.go`'s `rewriteModel` only replaces the `"model"` key via `map[string]json.RawMessage`. So every `cache_control` this PR adds survives byte-for-byte from `/v1/messages` through to LiteLLM's `/chat/completions`.
- **Whether it survives LiteLLM → OpenRouter → the actual model is NOT live-verified from this sandbox** (no live provider keys available to this task). `deploy/litellm/config.yaml` currently routes every chat alias to `openrouter/dots-studio/dots-3-note-preview:free` plus four paid non-Anthropic routes (deepseek-v4, doc-vlm, embedding fallback) — **there is no Anthropic/Claude model in the live catalog today**, so `cache_control` has nowhere to take effect in production right now regardless of this fix. This is a routing/catalog fact, not something this PR's scope covers or should change. Recommend a follow-up live smoke test once/if a Claude route is added: send a >2k-token prompt with a `cache_control` breakpoint to that alias and confirm `cache_read_input_tokens`/`cache_creation_input_tokens` come back non-zero on the second turn.

## Other fields translate_request.go drops on the floor (audit, not fixed here)

Per the task brief's request to audit this file for every dropped field even if not all get fixed:

- **`tool_choice: {"type":"none"}` silently becomes `"auto"`.** `convertToolChoice`'s switch only handles `auto`/`any`/`tool`; anything else (including the documented `none`, which means "forbid tool use") falls to the `default` branch and returns `auto`. This is a behavior *inversion*, not just an omission — a caller that explicitly disabled tools would get tool use enabled. Worth its own follow-up issue; flagged here per audit but not fixed, out of this PR's cache_control scope.
- `MessagesRequest.Metadata` (`metadata.user_id`) is parsed but `ToOAIRequest` never reads it at all.
- `ToolChoice.disable_parallel_tool_use` has no field at all on the `ToolChoice` type.
- Anthropic's `top_k` sampling parameter has no field on `MessagesRequest` at all (dropped at the type layer, not just the translator).
- Anthropic's extended-thinking `thinking` request field, and `thinking`/`redacted_thinking` content blocks, have no representation at all. A message consisting solely of a thinking block would fall through `convertMessage`'s content-block switch with no matching case and produce an OAI message with empty content — silently, no error.
- The pre-existing "tool calls plus text" branch in `convertMessage` still drops any `image` block mixed with tool calls when neither has a `cache_control` (documented in a code comment at the fix site rather than fixed, to keep the no-cache-control regression guard exact).

## Tests

Table-driven and mutation-checked (I deliberately broke the flattening guard locally and confirmed the dedicated test failed, then restored the fix and confirmed it passed again — see `cache_control_test.go`):
- `TestToOAIRequest_CacheControl_Placements` — system block, message content block (+ 1h ttl variant), tool definition, request root
- `TestToOAIRequest_CacheControl_SingleTextBlock_NotFlattenedToString` — the flattening regression guard
- `TestToOAIRequest_NoCacheControl_Unchanged` — regression guard: a request with no cache_control anywhere serializes with zero `cache_control` keys anywhere in the output
- `TestToOAIRequest_CacheControl_ToolResultBlock`, `TestToOAIRequest_CacheControl_ToolUseBlock`
- `TestToOAIRequest_SessionID_Passthrough`, `TestToOAIRequest_SessionID_TruncatedAt256`
- `TestFromOAIResponse_CacheTokens_NonStreaming`, `TestFromOAIResponse_CacheTokens_ClampedNeverNegative`, `TestFromOAIResponse_NoCacheDetails_InputTokensUnchanged`
- `TestSSETranslator_CacheTokensInMessageDelta`

Full `apps/edge-api` suite (all packages, including `inference`) passes on the rebased branch.

## Buglog entry

```json
{"error_message":"cache_control silently dropped on /v1/messages, zero prompt caching for agent clients","root_cause":"translate_request.go rebuilds the Anthropic request field-by-field into OAIRequest; cache_control (content block, system block, tool, request root) was never one of the carried fields, and two collapse sites in convertMessage flattened a typed content-block array to a plain string, which cannot carry a per-block cache_control even once the field exists","fix":"added CacheControl to ContentBlock/Tool/MessagesRequest/SystemField and their OAI-shaped mirrors; guarded both flattening sites in translate_request.go to keep block-array form when a cache breakpoint is present; added the exclusive-shape cache_creation_input_tokens/cache_read_input_tokens echo to ResponseUsage and StreamUsage via freshInputTokens' inclusive-to-exclusive subtraction","tags":["anthropic","cache_control","prompt-caching","translate_request","billing-coordination"]}
```

## Test plan

- [x] `go build ./apps/edge-api/...`
- [x] `go vet ./apps/edge-api/...`
- [x] `go test ./apps/edge-api/... -count=1` (all packages green, rebased on origin/main)
- [x] Mutation check on the flattening-regression test (fails without the fix, passes with it)
- [ ] Live smoke test once a Claude/Anthropic route exists in the catalog (tracked as a follow-up, not blocking)

### DIFF
--- a/apps/edge-api/internal/anthropic/translate_request.go
+++ b/apps/edge-api/internal/anthropic/translate_request.go
@@ -3,7 +3,9 @@ package anthropic
 import (
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"strings"
+	"unicode/utf8"
 )
 
 // maxSessionIDLen is OpenRouter's documented ceiling for the sticky-routing
@@ -43,9 +45,10 @@ func ToOAIRequest(req MessagesRequest) (OAIRequest, error) {
 	}
 
 	if req.SessionID != "" {
-		id := req.SessionID
-		if len(id) > maxSessionIDLen {
-			id = id[:maxSessionIDLen]
+		id, truncated := truncateSessionID(req.SessionID)
+		if truncated {
+			slog.Warn("anthropic session_id truncated to the OpenRouter ceiling",
+				"original_len", len(req.SessionID), "truncated_len", len(id))
 		}
 		out.SessionID = id
 	}
@@ -312,4 +315,79 @@ func convertToolChoice(tc *ToolChoice) OAIToolChoice {
 	default:
 		return OAIToolChoice{Sentinel: "auto"}
 	}
+}
+
+// truncateSessionID cuts id to at most maxSessionIDLen bytes at a rune
+// boundary. A plain byte-index slice (id[:maxSessionIDLen]) can land inside
+// a multibyte UTF-8 rune; json.Marshal would then silently replace the
+// corrupted tail byte with U+FFFD, so the value forwarded to OpenRouter
+// would differ from any prefix the caller actually sent. Reports whether it
+// actually cut anything, so the caller can log it.
+func truncateSessionID(id string) (string, bool) {
+	if len(id) <= maxSessionIDLen {
+		return id, false
+	}
+	cut := maxSessionIDLen
+	for cut > 0 && !utf8.RuneStart(id[cut]) {
+		cut--
+	}
+	return id[:cut], true
+}
+
+// maxCacheControlBreakpoints is Anthropic's documented per-request cap.
+// Enforced locally so an unbounded number of breakpoints cannot make Hive
+// forward unbounded upstream work on a client's behalf.
+const maxCacheControlBreakpoints = 4
+
+// validateCacheControl checks every cache_control breakpoint in the request
+// (root, system blocks, message content blocks, tool_result content
+// nested blocks are deliberately excluded -- Anthropic's cache_control lives
+// on a tool_result block itself, not inside its nested content, so nothing
+// there can validly carry one -- and tool definitions) against Anthropic's
+// documented shape: type must be "ephemeral", ttl if set must be "5m" or
+// "1h", and no more than maxCacheControlBreakpoints total. Anthropic's real
+// API already rejects a violation, so this is a pure fail-fast: catching a
+// client mistake (or a script fuzzing this endpoint) before a full round
+// trip to a paid upstream, not a new restriction on anything valid.
+func validateCacheControl(req MessagesRequest) error {
+	count := 0
+	check := func(cc *CacheControl) error {
+		if cc == nil {
+			return nil
+		}
+		count++
+		if cc.Type != "ephemeral" {
+			return fmt.Errorf("cache_control.type must be %q, got %q", "ephemeral", cc.Type)
+		}
+		if cc.TTL != "" && cc.TTL != "5m" && cc.TTL != "1h" {
+			return fmt.Errorf("cache_control.ttl must be %q or %q, got %q", "5m", "1h", cc.TTL)
+		}
+		return nil
+	}
+
+	if err := check(req.CacheControl); err != nil {
+		return err
+	}
+	for _, bl := range req.System.Blocks {
+		if err := check(bl.CacheControl); err != nil {
+			return err
+		}
+	}
+	for _, m := range req.Messages {
+		for _, bl := range m.Content.Blocks {
+			if err := check(bl.CacheControl); err != nil {
+				return err
+			}
+		}
+	}
+	for _, t := range req.Tools {
+		if err := check(t.CacheControl); err != nil {
+			return err
+		}
+	}
+
+	if count > maxCacheControlBreakpoints {
+		return fmt.Errorf("at most %d cache_control breakpoints per request, got %d", maxCacheControlBreakpoints, count)
+	}
+	return nil
 }
--- a/apps/edge-api/internal/anthropic/translate_request.go
+++ b/apps/edge-api/internal/anthropic/translate_request.go
@@ -6,6 +6,11 @@ import (
 	"strings"
 )
 
+// maxSessionIDLen is OpenRouter's documented ceiling for the sticky-routing
+// session_id passthrough (see MessagesRequest.SessionID). Truncated rather
+// than rejected: it is a routing hint, not a correctness-bearing field.
+const maxSessionIDLen = 256
+
 // ToOAIRequest lowers an Anthropic MessagesRequest to an internal OpenAI-shaped
 // OAIRequest that can be forwarded through the existing LiteLLM dispatch path.
 // It never leaks provider names; the model alias is passed through as-is so the
@@ -13,10 +18,10 @@ import (
 func ToOAIRequest(req MessagesRequest) (OAIRequest, error) {
 	var msgs []OAIMessage
 
-	if req.System.Text != "" {
+	if req.System.Text != "" || len(req.System.Blocks) > 0 {
 		msgs = append(msgs, OAIMessage{
 			Role:    "system",
-			Content: OAIMessageContent{Text: req.System.Text},
+			Content: systemContent(req.System),
 		})
 	}
 
@@ -29,11 +34,20 @@ func ToOAIRequest(req MessagesRequest) (OAIRequest, error) {
 	}
 
 	out := OAIRequest{
-		Model:       req.Model,
-		Messages:    msgs,
-		Temperature: req.Temperature,
-		TopP:        req.TopP,
-		Stream:      req.Stream,
+		Model:        req.Model,
+		Messages:     msgs,
+		Temperature:  req.Temperature,
+		TopP:         req.TopP,
+		Stream:       req.Stream,
+		CacheControl: req.CacheControl,
+	}
+
+	if req.SessionID != "" {
+		id := req.SessionID
+		if len(id) > maxSessionIDLen {
+			id = id[:maxSessionIDLen]
+		}
+		out.SessionID = id
 	}
 
 	if req.MaxTokens > 0 {
@@ -58,6 +72,43 @@ func ToOAIRequest(req MessagesRequest) (OAIRequest, error) {
 	}
 
 	return out, nil
+}
+
+// systemContent lowers a SystemField to an OAIMessageContent. When no block
+// carries a cache breakpoint it reproduces the exact plain-string output this
+// function always emitted (the regression guard: an uncached system prompt
+// must serialize identically to before this change), because collapsing to a
+// flat string cannot represent a per-block cache_control -- a request that
+// does use one gets the block-array form instead, each block's own
+// CacheControl carried onto its OAIContentPart.
+func systemContent(sf SystemField) OAIMessageContent {
+	needsBlocks := false
+	for _, bl := range sf.Blocks {
+		if bl.CacheControl != nil {
+			needsBlocks = true
+			break
+		}
+	}
+	if !needsBlocks {
+		return OAIMessageContent{Text: sf.Text}
+	}
+
+	parts := make([]OAIContentPart, 0, len(sf.Blocks))
+	for _, bl := range sf.Blocks {
+		if bl.Type != "text" {
+			// Anthropic's system field only ever carries text blocks; a
+			// non-text block here would already have been silently ignored
+			// by SystemField.UnmarshalJSON's Text concatenation, so dropping
+			// it here too is not a new gap this change introduces.
+			continue
+		}
+		parts = append(parts, OAIContentPart{
+			Type:         "text",
+			Text:         bl.Text,
+			CacheControl: bl.CacheControl,
+		})
+	}
+	return OAIMessageContent{Parts: parts}
 }
 
 // convertMessage converts a single Anthropic message to one or more OAIMessages.
@@ -89,9 +140,10 @@ func convertMessage(m Message) ([]OAIMessage, error) {
 			}
 			content := toolResultText(bl)
 			toolResults = append(toolResults, OAIMessage{
-				Role:       "tool",
-				Content:    OAIMessageContent{Text: content},
-				ToolCallID: bl.ToolUseID,
+				Role:         "tool",
+				Content:      OAIMessageContent{Text: content},
+				ToolCallID:   bl.ToolUseID,
+				CacheControl: bl.CacheControl,
 			})
 		} else if len(toolResults) > 0 {
 			// tool_result blocks were already seen; a non-tool_result block after them is invalid.
@@ -113,7 +165,7 @@ func convertMessage(m Message) ([]OAIMessage, error) {
 	for _, bl := range m.Content.Blocks {
 		switch bl.Type {
 		case "text":
-			parts = append(parts, OAIContentPart{Type: "text", Text: bl.Text})
+			parts = append(parts, OAIContentPart{Type: "text", Text: bl.Text, CacheControl: bl.CacheControl})
 
 		case "image":
 			if bl.Source == nil {
@@ -126,8 +178,9 @@ func convertMessage(m Message) ([]OAIMessage, error) {
 				dataURI = bl.Source.URL
 			}
 			parts = append(parts, OAIContentPart{
-				Type:     "image_url",
-				ImageURL: &OAIImageURL{URL: dataURI},
+				Type:         "image_url",
+				ImageURL:     &OAIImageURL{URL: dataURI},
+				CacheControl: bl.CacheControl,
 			})
 
 		case "tool_use":
@@ -142,6 +195,7 @@ func convertMessage(m Message) ([]OAIMessage, error) {
 					Name:      bl.Name,
 					Arguments: args,
 				},
+				CacheControl: bl.CacheControl,
 			})
 		}
 	}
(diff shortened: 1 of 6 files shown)


apps/edge-api/internal/anthropic/translate_response.go
```diff
@@ -65,6 +62,46 @@ func FromOAIResponse(resp OAIResponse, clientAlias string) MessagesResponse {
...
+// already-exclusive number, drives it negative).
+//
+// A negative result here is clamped to zero rather than surfaced to the
+// client: it means the upstream inclusive-shape assumption broke, which is a
+// billing-accuracy alarm for the settlement path (apps/edge-api/internal/inference),
+// not something this wire-shape projection can diagnose or repair. This
+// function only decides what the client sees, never what gets billed.
+func freshInputTokens(promptTokens, cacheRead, cacheWrite int) int {
+	fresh := promptTokens - cacheRead - cacheWrite
+	if fresh < 0 {
```
sakibsadmanshajib: MEDIUM: silent clamp, no alarm.

`freshInputTokens` clamps a negative result to 0 with no log, no metric, nothing. Per the team's own cache-pricing research doc (section 5, which this PR's author clearly had open while writing this function): "If fresh computes negative on the inclusive path, do NOT silently clamp and move on... emit a metric naming the provider and model, and alarm." That instruction exists because a negative fresh value only happens when the inclusive-shape assumption about the upstream (OpenRouter always being INCLUSIVE, even for Claude models) breaks -- exactly the failure mode section 1 of that doc calls "the whole bug," and exactly the shape the security brief for this review called out by name (a clamp that quietly returns 0 turns a billing anomaly into a free request).

The doc comment here argues the alarm belongs to the settlement path (`apps/edge-api/internal/inference`) instead, since this function "only decides what the client sees, never what gets billed." That's a reasonable scoping call, but today nothing in `inference` alarms on this either (I checked: `pricing.go`'s `CreditsForTokens` clamps negative token counts silently too, no log/metric there either). So right now there is zero telemetry anywhere in the codebase that would catch OpenRouter changing its usage-reporting shape out from under this integration. At minimum, add a `slog.Warn` here naming the alias/model when `fresh < 0`, since this is the one place in the request lifecycle that already has the numbers in hand to detect it, even if the billing-side alarm lands separately in `feat/cache-aware-billing`.

apps/edge-api/internal/anthropic/types.go
```diff
@@ -25,17 +25,53 @@ type MessagesRequest struct {
...
+// CacheControl marks an Anthropic prompt-caching breakpoint. It is valid on a
+// content block, a system prompt block, a tool definition, and the request
+// root. It is always a pointer field so an absent cache_control stays absent
+// on the wire: an empty {} object is a materially different request from no
+// key at all (Anthropic's own API rejects certain shapes with an empty
+// cache_control differently from a request that never mentions it), so a
+// plain (non-pointer) struct with omitempty would be wrong here -- Go's
+// encoding/json does not treat a zero-value struct as empty for omitempty
+// purposes, meaning a bare CacheControl{} field would always serialize.
+type CacheControl struct {
```
sakibsadmanshajib: LOW: no validation at the trust boundary.

`CacheControl.Type` accepts any string (not just Anthropic's one documented value, "ephemeral") and `TTL` accepts any string (not just "5m"/"1h"), and nothing anywhere in this package or `handler.go` counts breakpoints against Anthropic's documented cap of 4 per request. All three are forwarded blind to LiteLLM/OpenRouter.

This isn't a crash or a security hole on Hive's side (no local dereference or execution of the value; a malformed shape gets rejected by the real upstream, and that rejection already goes through the existing provider-blind error mapping the chat_completions tests already cover generically). It is a missed fail-fast: a client mistake or a script fuzzing this endpoint currently costs a full round trip to a paid upstream before getting an opaque `provider_blind_upstream_error`-mapped 400, instead of an instant, clear local 400 (`"cache_control.type must be 'ephemeral'"`, `"at most 4 cache_control breakpoints per request"`). Worth a follow-up, not a blocker for this PR's actual scope (getting the field through at all).

apps/edge-api/internal/anthropic/translate_request.go
```diff
@@ -29,11 +34,20 @@ func ToOAIRequest(req MessagesRequest) (OAIRequest, error) {
...
+		Messages:     msgs,
+		Temperature:  req.Temperature,
+		TopP:         req.TopP,
+		Stream:       req.Stream,
+		CacheControl: req.CacheControl,
+	}
+
+	if req.SessionID != "" {
+		id := req.SessionID
+		if len(id) > maxSessionIDLen {
```
sakibsadmanshajib: LOW: silent truncation, not rune-boundary-safe.

Two things here:
1. Truncation is completely silent -- no log line when it actually fires. A support investigation into "why did this session lose cache locality" has zero trail to this code path.
2. `id[:maxSessionIDLen]` slices by byte index, not by rune. A `session_id` containing multi-byte UTF-8 that happens to straddle byte 256 gets cut mid-rune, and `encoding/json.Marshal` will silently substitute the invalid tail with U+FFFD rather than erroring -- so the value forwarded to OpenRouter can differ from any prefix the caller actually sent.

Security impact is low, not absent: `session_id` is confirmed (by reading `handler.go`) to be a pure routing hint to OpenRouter, never consulted by Hive's own auth/authz/tenant resolution, so a collision from two different long session_ids sharing the same truncated 256-byte prefix cannot cross a tenant or authorization boundary -- worst case is degraded cache locality between two unrelated sessions, not data exposure. Still worth a one-line `slog` on truncation and a `utf8`-aware truncate (the repo already imports `unicode/utf8` in this same package's `handler.go`), given the existing `rag.truncate` byte-slice helper elsewhere is used only for cosmetic log/error strings, not for a value forwarded on the wire.

sakibsadmanshajib (commented):
Independent adversarial review (security-reviewer pass). Read the pushed diff directly, ran the anthropic-package suite in the project's docker toolchain, and independently reproduced the author's mutation-check claim by reverting the flattening guard locally (test failed as claimed, then restored, confirmed clean). No blocking findings; two MEDIUM correctness/observability gaps and two LOW hardening nits, all inline below. I would not block this merge on them, but MEDIUM #1 should get a follow-up before this feature is called billing-complete.

Verified true (not just trusted):
- Request-side byte-for-byte forwarding claim: confirmed by reading chat_completions.go (raw body bytes, not a re-marshaled typed struct, flow into executeSync/executeStreaming) and litellm_client.go's rewriteModel (map[string]json.RawMessage passthrough, only the "model" key is touched). cache_control and session_id do survive this hop unchanged.
- Flattening regression guard: traced every touched branch in convertMessage/systemContent by hand for the no-cache_control case; all reproduce the prior output exactly. Also reproduced the author's mutation check myself (reverted the `parts[0].CacheControl == nil` guard, TestToOAIRequest_CacheControl_SingleTextBlock_NotFlattenedToString failed with the exact flattened-string message the test asserts against, restored and confirmed clean).
- No logging/persistence of raw cache_control or session_id values anywhere in the diff.
- Full `apps/edge-api/internal/anthropic` suite green in the docker toolchain, including all new cache_control tests.
- No panics/nil-deref risk: every CacheControl pointer field is either passed straight through or nil-checked before use (systemContent's `bl.CacheControl != nil`).

Would not block merge. Recommend addressing MEDIUM #1 (or explicitly tracking it) before calling cache billing observability done, and linking a tracking issue for MEDIUM #2's coordination gap so `cache_creation_input_tokens` isn't silently wrong in production between when this merges and when `feat/cache-aware-billing` lands.

apps/edge-api/internal/anthropic/types.go
```diff
@@ -390,11 +470,40 @@ type OAIMsg struct {
...
+// inference package's own response types, not this one) -- it exists purely
+// so this package can read the numbers back out of them for the Anthropic
+// echo in translate_response.go and stream.go.
+type OAIPromptTokensDetails struct {
+	CachedTokens int `json:"cached_tokens,omitempty"`
+	// CacheWriteTokens has no OpenAI-native counterpart (OpenAI itself never
+	// bills a cache write); OpenRouter adds it as an extension when the
+	// upstream model does.
+	CacheWriteTokens int `json:"cache_write_tokens,omitempty"`
 }
```
sakibsadmanshajib: MEDIUM (disclosed, needs a tracking link): confirms the PR body's own "Merge coordination" caveat is real, not hypothetical.

I verified directly: `apps/edge-api/internal/inference/types.go`'s `PromptTokensDetails` has only `CachedTokens`, no `CacheWriteTokens` field. `normalizeChatCompletion` (`chat_completions.go`) does a typed unmarshal-then-remarshal of `inference.ChatCompletionResponse` before this package ever sees a byte, so any `cache_write_tokens` key the upstream actually sent is dropped on the floor at that hop, unconditionally, today. That means `OAIPromptTokensDetails.CacheWriteTokens` here will always unmarshal to zero in production right now, which means `cache_creation_input_tokens` echoed to a real Anthropic client (Claude Code's cost display, per this PR's own stated goal) will always read 0 -- even on a turn that genuinely wrote to cache and was billed a write-rate multiplier for it.

The PR body already discloses this precisely and points at `feat/cache-aware-billing` as the fix, which is good practice. Flagging only so it's on the record on the PR itself (not just in the description prose) and so it doesn't get lost: please open or link a tracking issue before merge, so "cache writes always show 0 to the client" doesn't quietly persist if the other branch stalls. This does not affect real billing amounts (that path is untouched by this PR); it only affects what the client is shown.

apps/edge-api/internal/anthropic/translate_response.go
sakibsadmanshajib: Fixed. freshInputTokens now logs slog.Warn naming clientAlias and upstreamModel whenever the subtraction goes negative, before returning the clamped 0 (translate_response.go, plus the matching call site in stream.go's FeedLine for the streaming path). Test coverage: TestFromOAIResponse_CacheTokens_NegativeClamp_LogsWarn and TestSSETranslator_CacheTokens_NegativeClamp_LogsWarn assert the WARN actually fires and names both identifiers, not just that the clamp still returns 0.

Confirmed inference/pricing.go's CreditsForTokens has the same silent-clamp gap. Per the orchestrator, that file belongs to the concurrent feat/cache-aware-billing branch and its builder is already briefed to add the equivalent guard there -- not duplicating it here, but flagging so this thread doesn't read as ignoring that half of the finding.

apps/edge-api/internal/anthropic/types.go
sakibsadmanshajib: Fixed. Added validateCacheControl (translate_request.go), called from handleMessages right after JSON unmarshal, before any translation or dispatch. Rejects: type != "ephemeral", ttl set but not "5m"/"1h", and more than 4 total breakpoints across request root + system blocks + message content blocks + tool definitions, matching Anthropic's documented cap. All return a local 400 (invalid_request_error) instead of costing a round trip to a paid upstream.

Deliberately did not recurse into a tool_result block's own nested content blocks -- Anthropic's cache_control lives on the tool_result block itself, not inside its nested content, so there's nothing valid to check there.

Tests: TestValidateCacheControl is table-driven (valid ephemeral with/without ttl, invalid type, invalid ttl, exactly 4 breakpoints accepted, 5 rejected), exercised through the real HTTP handler so it's testing the actual trust boundary, not just the helper in isolation.

apps/edge-api/internal/anthropic/translate_request.go
sakibsadmanshajib: Fixed both. truncateSessionID now cuts at the last utf8.RuneStart boundary <= 256 bytes instead of a raw byte-index slice, and logs slog.Warn (original_len, truncated_len) whenever it actually cuts. Test: TestToOAIRequest_SessionID_TruncatedAtRuneBoundary builds a session_id where a 3-byte rune straddles byte 256, then asserts the result is <=256 bytes, utf8.ValidString, and that json.Marshal produces no U+FFFD replacement character.

apps/edge-api/internal/anthropic/types.go
sakibsadmanshajib: Confirmed, this is exactly the merge coordination point already documented in the PR body, not a new gap. Opened #1158 to track it so it doesn't silently persist if feat/cache-aware-billing stalls.

To close it, that branch needs a field on inference.PromptTokensDetails (or wherever it lands cache accounting) with JSON tag cache_write_tokens -- that exact name is what apps/edge-api/internal/anthropic/types.go's OAIPromptTokensDetails.CacheWriteTokens already expects, so once it's added there this package's plumbing picks it up with zero further changes on this side.

Impact confirmed client-display-only: cache_creation_input_tokens reading 0 affects only what Claude Code's cost UI shows, never a billed amount -- the settlement path in apps/edge-api/internal/inference is untouched by this PR.

PR #1155, opened 2026-08-25
fix: make the free pool actually fail over when one member is retired

### LINKED ISSUE #1064: Live integration (SDK tests + smoke) is failing on main
Live integration failed on push to main.

First detected: https://github.com/sakibsadmanshajib/hive/actions/runs/32636322697

Filed automatically. Close it once a run goes green; it will be refiled if it fails again.

sakibsadmanshajib:
## Summary

`hive-free` is served by a load-balanced LiteLLM group of four free provider keys, and its whole product claim is that one exhausted or retired key fails over to the others. It does not. One member answering 404 takes the entire alias down for that request. That is what has been failing `Live integration (SDK tests + smoke)` on main.

Fixes #1064.

## Which of the four candidate causes it was

**Routing.** Not credentials, not pricing, not metering.

All four members' credentials and models were verified live while diagnosing this:

| Member | Upstream | Verified |
| --- | --- | --- |
| `route-free-pool-free` | `dots-studio/dots-3-note-preview:free` | listed in OpenRouter's public catalog right now |
| `route-free-pool-groq` | `openai/gpt-oss-20b` | listed in Groq's `/v1/models` right now |
| `route-free-pool-groq-2` | `openai/gpt-oss-20b`, second key slot | same model, secret present |
| `route-free-pool-gemini` | `openai/gemini-flash-latest` | not probeable from the dev box, no key there |

All four repository secrets exist. The gateway serves the group: the run's own assertion printed `gateway serves route-free-pool`.

## Request failed, or metering silently dropped it?

**Request failed.** This matters because the repo has been burned by the second shape before (2026-07-31, the gateway billed nothing for three days and every failure was silent), so it was checked first rather than assumed.

The failing run's own spend report:

```
      model_alias       |     event_type      | events | input_tokens | output_tokens
------------------------+---------------------+--------+--------------+---------------
 hive-free              | completed           |     14 |          318 |           278
 hive-free              | error               |      1 |            0 |             0
 hive-free              | reconciled          |      2 |            0 |             0
 hive-free              | released            |      2 |            0 |             0
 hive-free              | reservation_created |     10 |            0 |             0
```

Fourteen metered completions on `hive-free`, plus one metered `error`. Metering is healthy and is recording both outcomes. The "zero metered completions" framing does not hold for this run: the `served == 0` assertion is gated on `SUITES_OUTCOME = success` and never executed, because the suites had already failed.

## The actual failure

Three JS tests failed, all of them the *second* `hive-free` call in their file. The first call in each file passed. Python's suite passed 14/14 because pytest runs sequentially while vitest runs test files in parallel, so only the JS run had concurrent `hive-free` traffic.

```
 ❯ tests/completions/completions.test.ts (2 tests | 1 failed)
   ✓ Completions (legacy) > returns a valid text completion via SDK  2451ms
   × Completions (legacy) > model field shows Hive alias not provider handle 128ms
     → 404 hive-free is not available.
 ❯ tests/chat-completions/chat-completions.test.ts (5 tests | 1 failed)
   × Chat Completions > model field shows Hive alias not provider handle 60004ms
     → Test timed out in 60000ms.
```

The customer-facing `hive-free is not available.` is our provider-blind collapse of this, from the run's `compose-logs` artifact:

```
litellm.NotFoundError: NotFoundError: OpenAIException - Error code: 404
No fallback model group found for original model_group=route-free-pool.
Fallbacks=[{'route-openrouter-embedding': ['route-openrouter-embedding-fallback']}].
Received Model Group=route-free-pool
Available Model Group Fallbacks=None
Cooldown Deployments=[('e644bb2c...', {'exception_received': 'litellm.NotFoundError: ...',
                       'status_code': '404', 'timestamp': ..., 'cooldown_time': 5})]
```

One deployment returned 404. Three were fine. The request died anyway.

## Root cause

Read out of the pinned image's source (`v1.98.0`), not inferred:

1. `litellm/router.py::should_retry_this_error` re-raises immediately for `NotFoundError`, and again for any status where `litellm._should_retry()` is false. 404 is both:
   ```python
   status_code = getattr(error, "status_code", None)
   if status_code is not None and not litellm._should_retry(status_code):
       if status_code not in (401, 403):
           raise error
   if isinstance(error, litellm.NotFoundError):
       raise error
   ```
2. It is called from **both** `async_function_with_retries` and `async_function_with_fallbacks`, so neither the in-group retry across the other three members nor the cross-group fallback ever runs.
3. `litellm/types/router.py::RetryPolicy` has fields for BadRequest, Authentication, Timeout, RateLimit, ContentPolicyViolation and InternalServer errors, and **none for NotFound**. There is no configuration that lifts this.

A retired free model is exactly what returns 404, and free model ids churn constantly on every provider in this pool. So the pool's failover premise is false against the single most likely way a free member dies.

## The fix

**Make the pool answer.** `apps/edge-api/internal/inference/retry.go` is the one shared dispatch seam that every chat, completions, responses and streaming path already flows through, so the fix lands once for every pooled alias rather than for `hive-free` alone. It now retries a LiteLLM router-exhaustion 404.

This is sound rather than hopeful because LiteLLM cools the offending deployment down *before* raising: `router_utils/cooldown_handlers.py::_should_cooldown_deployment` ends its base case with `litellm._should_retry(status) is False -> return True`, so a 404 member is out of rotation on its **first** failure, not after `allowed_fails`. The next attempt therefore picks a different member, and the existing 300/800/1800 ms ladder finishes inside 2.9 s, comfortably inside the 5 s cooldown, so the dead member cannot be re-picked mid-ladder.

Scope is deliberately narrow: the retry matches on LiteLLM's message, not on a bare 404, so a customer naming a model that does not exist still gets a fast 404 and pays no retries. `isRouterExhaustion404` splices the body back together after peeking, so the caller always receives a complete response; a half-consumed body would truncate the error the customer sees and the text CI classifies on.

**Make the failure legible.** Two additions, neither of which weakens a check:

* A new step probes each member individually through LiteLLM's own `/health?model=route-free-pool` and **names the model that is gone**. One dead member is a `::warning::`, because surviving one is the pool's entire purpose and failing the job on it would make that resilience worthless; the SDK suites remain the real gate. Every member dead is a hard failure, because the free tier then genuinely serves nothing. Output goes through `scripts/redact-log-credentials.py` like every other published dump in this job.
* The upstream-refusal classifier gained a branch for the `No fallback model group found` signature, so this reads as "a pool member is dead" instead of falling through to "this failure is something else".

**No paid fallback was added.** Pointing `hive-free` at a paid model would silently spend real budget and break the one-alias-one-price rule; if a paid fallback is the right product answer, that is an owner decision, not one to smuggle into a CI fix.

## A trap this leaves behind, documented rather than fixed

`allowed_fails` and `cooldown_time` sit under `litellm_settings` in `deploy/litellm/config.yaml`. They look misplaced and in the narrow sense they are: both are Router constructor params, so they never reach the Router, and the effective cooldown is the default 5 s, not the 30 written there. That is observed, not deduced: the failing run's log prints `'cooldown_time': 5`.

**Moving `allowed_fails` onto the Router would quietly restore this bug.** With it unset, `_should_cooldown_deployment` takes the base case that cools a 404 member down on the first failure. Set it, and that branch is skipped for `should_cooldown_based_on_allowed_fails_policy`, which needs `fails > allowed_fails`: four failures before a dead member leaves rotation, which the retry ladder above cannot outlast. The config now says so in place, so the next tidy-up does not walk into it.

## What kind of check actually closes this gap

The pool already had unit coverage, `apps/control-plane/internal/routing/free_pool_router_test.go`, and it passed the entire time this was broken. It parses the migration text and asserts the four rows share one `litellm_model_name`. That assertion was **true**; the shape was right and the behaviour was wrong. A test that can only see the config can never catch a failure that lives in the router's exception handling.

What closes it is a test that exercises the real function against the real upstream envelope, so this PR adds behavioural coverage over `dispatchWithRetry` using the verbatim 404 body from run 32830060362:

* `TestDispatchWithRetry_FailsOverWhenOnePoolMemberIsGone` — dead member on attempt 1, 200 on attempt 2. Fails on `main`.
* `TestDispatchWithRetry_GenuineModelNotFoundIsNotRetried` — bounds the blast radius; a real model-not-found stays one attempt.
* `TestDispatchWithRetry_DeadPoolReturnsTheUpstream404Intact` — when retries cannot rescue it, the client still gets the real status and a complete body.
* `TestIsRouterExhaustion404LeavesTheBodyReadable` — the body splice survives a payload larger than the peek window.

Plus `scripts/report-free-pool-health.py --selfcheck`, covering the one-dead / all-dead / missing-counts / malformed-endpoint verdicts.

The live half is closed by the new probe step, which is the only thing in CI that can name a retired model, and by the classifier branch that stops this failure shape reading as unexplained.

## Test evidence

```
=== RUN   TestDispatchWithRetry_FailsOverWhenOnePoolMemberIsGone
--- PASS: TestDispatchWithRetry_FailsOverWhenOnePoolMemberIsGone (0.00s)
=== RUN   TestDispatchWithRetry_GenuineModelNotFoundIsNotRetried
--- PASS: TestDispatchWithRetry_GenuineModelNotFoundIsNotRetried (0.00s)
=== RUN   TestDispatchWithRetry_DeadPoolReturnsTheUpstream404Intact
--- PASS: TestDispatchWithRetry_DeadPoolReturnsTheUpstream404Intact (0.00s)
=== RUN   TestIsRouterExhaustion404LeavesTheBodyReadable
--- PASS: TestIsRouterExhaustion404LeavesTheBodyReadable (0.00s)
PASS
ok  	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/inference	0.028s
```

Full module, plus build and vet:

```
BUILD_OK
VET_OK
ok  github.com/sakibsadmanshajib/hive/apps/edge-api/internal/inference  3.161s
ok  github.com/sakibsadmanshajib/hive/apps/edge-api/internal/errors     0.065s
... (25 packages, all ok)
ok  github.com/sakibsadmanshajib/hive/apps/control-plane/internal/routing       0.185s
ok  github.com/sakibsadmanshajib/hive/apps/control-plane/internal/litellmconfig 0.181s
```

```
$ python3 scripts/report-free-pool-health.py --selfcheck
ok: report-free-pool-health verdicts and fallbacks

$ node tools/lint-litellm-config-fallbacks.mjs
lint-litellm-config-fallbacks: PASS
```

Redaction verified end to end on a synthetic error carrying a key-shaped value:

```
  DEAD MEMBER: openai/gemini-flash-latest -- Error code: 404 - models/gemini-flash-latest is not found. api_key=<redacted>
```

`gofmt -l` output is byte-identical before and after this change, so no formatting regression was introduced.

## Buglog entry

```json
{"id":"free-pool-404-no-failover","date":"2026-08-25","title":"LiteLLM aborts a load-balanced group on one member's 404 instead of failing over, taking hive-free down","error_message":"litellm.NotFoundError: NotFoundError: OpenAIException - Error code: 404No fallback model group found for original model_group=route-free-pool. Available Model Group Fallbacks=None -- surfaced to the customer as 'hive-free is not available.'","root_cause":"litellm/router.py::should_retry_this_error re-raises immediately for NotFoundError and for any status where litellm._should_retry() is false; 404 is both. It is called from async_function_with_retries and async_function_with_fallbacks, so one pool member answering 404 (what a retired free model returns) killed the request while the group's other three members were healthy. litellm/types/router.py::RetryPolicy has no NotFoundErrorRetries field, so no config lifts it. The pool's existing unit test asserts the four rows share a litellm_model_name, which stayed true throughout, so the shape was right and the behaviour was wrong.","fix":"Retry a LiteLLM router-exhaustion 404 in the shared dispatch seam (apps/edge-api/internal/inference/retry.go), matched on LiteLLM's message rather than on a bare 404 so a genuine model-not-found stays fast. Sound because LiteLLM cools the 404 deployment down on its first failure, so the next attempt picks a different member inside the 5s window. Added behavioural tests over dispatchWithRetry, a CI step that names a dead member via /health?model=, and a classifier branch for the signature.","tags":["litellm","routing","free-pool","failover","ci","edge-api","hive-free","404"]}
```

```json
{"id":"litellm-router-knobs-in-wrong-config-block","date":"2026-08-25","title":"allowed_fails and cooldown_time under litellm_settings never reach the LiteLLM Router, and moving allowed_fails would break free pool failover","error_message":"config declares cooldown_time: 30 but the runtime log prints 'cooldown_time': 5 in its Cooldown Deployments line","root_cause":"allowed_fails and cooldown_time are Router constructor params (litellm/types/router.py::UpdateRouterConfig), not litellm_settings keys, so the values are silently ignored and DEFAULT_COOLDOWN_TIME_SECONDS (5) applies. The apparent misplacement is load-bearing: with allowed_fails unset on the Router, _should_cooldown_deployment takes its base case and cools a 404 member down on the first failure, which is what lets the edge retry land on a different member.","fix":"Left both settings in place deliberately and documented the trap in deploy/litellm/config.yaml, including that moving allowed_fails into router_settings would require four failures before a dead member leaves rotation and would quietly restore the free pool 404 bug.","tags":["litellm","config","cooldown","free-pool","footgun"]}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1440,6 +1440,43 @@ jobs:
           fi
           echo "gateway serves route-free-pool; models: $(jq -r '[.data[].id] | join(", ")' /tmp/litellm-models.json)"
 
+      # Name the dead pool member, if there is one, BEFORE the suites run.
+      #
+      # Serving route-free-pool (asserted above) only proves the group exists,
+      # not that its four members answer. Free model ids churn constantly on
+      # every provider in the pool, and a retired one answers 404 -- which is
+      # exactly the error LiteLLM refuses to fail over (issue #1064). Without
+      # this step that shows up as a generic "hive-free is not available." with
+      # nothing anywhere naming which of the four models went away.
+      #
+      # `/health?model=` is LiteLLM's own per-group probe: it runs one tiny
+      # request against each deployment in that group and reports them
+      # individually, so this costs four small completions on free keys and
+      # nothing on any paid alias.
+      #
+      # Deliberately NOT a hard gate on a single dead member. Surviving one is
+      # the pool's entire purpose, so failing the job on it would make the
+      # resilience worthless; the suites below are the real gate, and they now
+      # pass through a degraded pool. It IS a hard failure when every member is
+      # down, because then hive-free genuinely serves nothing.
+      - name: Probe each free pool member and name any that is gone
+        shell: bash
+        run: |
+          set -uo pipefail
+          code=$(curl -sS -o /tmp/pool-health.json -w '%{http_code}' --max-time 120 \
+            -H "Authorization: Bearer ${LITELLM_MASTER_KEY}" \
+            "http://localhost:4000/health?model=route-free-pool") || code="000"
+
+          if [ "$code" != "200" ]; then
+            echo "::warning::could not probe the free pool (HTTP ${code}), so this run cannot say whether all four members are alive"
+            exit 0
+          fi
+
+          # python3 rather than jq purely so the reporting and the credential
+          # redaction are the same language and the same process; the redactor
+          # is already a python module this repo ships.
+          python3 scripts/report-free-pool-health.py /tmp/pool-health.json
+
       - name: Run SDK suites in parallel (JS + Python + Java)
         # id, because the spend meter below runs on always() and needs to tell
         # "the suites failed, so the database may not exist" from "the suites
@@ -1629,12 +1666,16 @@ jobs:
             class="the provider rate limited us (429). Check whether the window is per minute (transient, so a rerun may pass) or per day (an allowance that is spent)."
           elif grep -qiE 'insufficient.?credit|402 Payment Required' /tmp/upstream.log; then
             class="the provider account is out of credit (402). No retry fixes this; it needs funding."
+          elif grep -qiE 'No fallback model group found' /tmp/upstream.log; then
+            # Last, on purpose: the three refusals above are more specific about
+            # spend and should win when both signatures are present.
+            class="a POOL MEMBER IS DEAD, not the whole provider. LiteLLM answered 'No fallback model group found', which is what it says when one deployment in a load-balanced group returns 404 and the group has no fallback group configured. A 404 is what a RETIRED free model returns, and LiteLLM will not fail over on it (litellm/router.py::should_retry_this_error re-raises NotFoundError, and RetryPolicy has no NotFoundErrorRetries knob). The edge now retries this so the pool routes around the dead member; seeing it here means the retry ladder was also exhausted, so more than one member is down. The 'Probe each free pool member' step earlier in this job names them."
           fi
 
           if [ -n "$class" ]; then
             echo "::error::UPSTREAM REFUSAL, not a performance regression: $class"
             echo "matching lines (redacted, first 5):"
-            grep -iE 'tokens per day|rate.?limit|RateLimitError|rate_limit_exceeded|insufficient.?credit|402 Payment Required' \
+            grep -iE 'tokens per day|rate.?limit|RateLimitError|rate_limit_exceeded|insufficient.?credit|402 Payment Required|No fallback model group found' \
               /tmp/upstream.log | head -5
             echo "UPSTREAM_FAILURE_CLASS=$class" >> "$GITHUB_ENV"
           else
--- a/apps/edge-api/internal/inference/retry.go
+++ b/apps/edge-api/internal/inference/retry.go
@@ -1,10 +1,12 @@
 package inference
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"math/rand"
 	"net/http"
+	"strings"
 	"time"
 )
 
@@ -15,6 +17,70 @@ var retryableStatuses = map[int]bool{
 	http.StatusBadGateway:          true, // 502
 	http.StatusServiceUnavailable:  true, // 503
 	http.StatusGatewayTimeout:      true, // 504
+}
+
+// litellmRouterExhaustionMarker is LiteLLM's own wording when its router gives
+// up on a model group WITHOUT trying that group's remaining healthy
+// deployments. It is the reason a load-balanced pool does not actually fail
+// over, and it is why this file retries a status that is otherwise terminal.
+//
+// The mechanism, read out of the pinned image's source (v1.98.0) rather than
+// inferred:
+//
+//   - litellm/router.py::should_retry_this_error re-raises immediately for
+//     NotFoundError, and again for any status where litellm._should_retry() is
+//     false. 404 is both. It is called from async_function_with_retries AND
+//     from async_function_with_fallbacks, so neither the in-group retry nor the
+//     cross-group fallback ever runs.
+//   - litellm/types/router.py::RetryPolicy has no NotFoundErrorRetries field,
+//     so no configuration lifts this. The only per-class knob this repo sets,
+//     RateLimitErrorRetries, cannot reach it.
+//   - The request therefore dies on the FIRST deployment that answers 404,
+//     with the other members of the group untouched and healthy. That is what
+//     took the hive-free alias down in CI run 32830060362 while three of its
+//     four pool members were fine.
+//
+// What makes retrying here correct rather than hopeful: before raising, LiteLLM
+// has already put the offending deployment into cooldown.
+// router_utils/cooldown_handlers.py::_should_cooldown_deployment returns true
+// on the first failure when litellm._should_retry(status) is false, so a 404
+// member is excluded from selection immediately, not after allowed_fails. The
+// next attempt therefore picks a DIFFERENT member. The whole retry ladder below
+// finishes inside 2.9s, well inside the 5s DEFAULT_COOLDOWN_TIME_SECONDS the
+// router actually applies, so the dead member cannot be re-picked mid-ladder.
+//
+// Scope: this matches on the message, not on a bare 404, so a genuine "that
+// model does not exist" stays an immediate 404 and costs no retries.
+const litellmRouterExhaustionMarker = "no fallback model group found"
+
+// maxRetryPeekBytes bounds how much of a 404 body is read to classify it.
+// Upstream error envelopes are small; anything past this is not a marker this
+// function would recognise anyway.
+const maxRetryPeekBytes = 8 << 10
+
+// isRouterExhaustion404 reports whether resp is LiteLLM's router-exhaustion
+// answer rather than a genuine "no such model".
+//
+// The body is always spliced back together before returning, so the caller sees
+// a complete, unread response either way. That matters: this runs on the path
+// that hands the upstream error to the customer, and a half-consumed body would
+// silently truncate the error they see.
+func isRouterExhaustion404(resp *http.Response) bool {
+	if resp == nil || resp.StatusCode != http.StatusNotFound || resp.Body == nil {
+		return false
+	}
+
+	orig := resp.Body
+	head, err := io.ReadAll(io.LimitReader(orig, maxRetryPeekBytes))
+	resp.Body = struct {
+		io.Reader
+		io.Closer
+	}{Reader: io.MultiReader(bytes.NewReader(head), orig), Closer: orig}
+	if err != nil {
+		return false
+	}
+
+	return strings.Contains(strings.ToLower(string(head)), litellmRouterExhaustionMarker)
 }
 
 // retryDelays is the progressive backoff between attempts.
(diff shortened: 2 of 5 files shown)


sakibsadmanshajib:
## Verified live, twice

`Live integration (SDK tests + smoke)` passed on two consecutive runs of this branch ([32845556683](https://github.com/sakibsadmanshajib/hive/actions/runs/32845556683), [32846939118](https://github.com/sakibsadmanshajib/hive/actions/runs/32846939118)). The whole CI run is green: 26 checks pass, 4 skip, 0 fail.

The three JS tests that this issue was about now pass:

```
== sdk-tests-js  (exit 0)
```

The suites now reach the token meter's success branch, which was never exercised on the failing runs because it is gated on `SUITES_OUTCOME = success`:

```
 hive-free              | completed           |     18 |          403 |           551
 hive-free              | reconciled          |      2 |            0 |             0
 hive-free              | reservation_created |     10 |            0 |             0

live-integration metered 1654 tokens across the configured aliases (ceiling 6000)
18 completion(s) served by 'hive-free'
```

Eighteen metered completions on `hive-free`, and **no `error` event at all** — the failing run had `hive-free | error | 1`. Only the three permitted aliases were billed, so the model knob still reaches the suites and nothing leaked onto a paid route it did not choose.

## What the new probe revealed, and what it changes about the diagnosis

```
free pool members: 4 healthy, 0 unhealthy
```

All four members answer. So the 404 that took the alias down was an **intermittent blip from a member that is alive**, not a permanently retired model. That is a sharper diagnosis than the one this PR was opened with, and it makes the fix more clearly right rather than less: the pool was not surviving even a transient single-member 404, which is the cheapest possible failure to absorb and exactly what a four-member pool exists to absorb.

It also means the earlier failures were load-shaped, which fits the evidence: in every failing file the *first* `hive-free` call passed and the *second* failed, and only the JS suite failed because vitest runs test files in parallel while pytest is sequential.

The probe step keeps its value either way. Free model ids do get retired, and when one does, this step now names it instead of leaving a generic `hive-free is not available.` with nothing pointing at which of the four models went away.

sakibsadmanshajib:
## Post-merge safety audit (independent, PR merged before review completed)

Verdict: **no double-billing bug, no stranded-hold bug.** Read against the actual merged code on `origin/main` (`34c90fd54`), not the PR diff description.

### Why double-billing cannot happen here

`dispatchWithRetry` (`apps/edge-api/internal/inference/retry.go`) is called exactly **once** per request from each of the three call sites (`orchestrator.go:228`, `stream.go:236`, `stream_responses.go:167`), and in every caller the reservation (`CreateReservation`) and the usage attempt (`StartAttempt`) are created **once, before** that single call. All internal retries (429/5xx, and now the new router-exhaustion 404) happen inside that one call and are invisible to accounting: only the final `*http.Response` it returns is normalized/settled. A 200 is returned immediately (`if !retryableStatuses[...] && !isRouterExhaustion404(resp) { return resp, nil }`) and is never retried again, so a response that already succeeded upstream cannot trigger a second dispatch even if downstream JSON parsing later fails — that failure path releases the hold, it does not re-dispatch.

The 404 addition in this PR only changes what counts as *retryable*; it does not touch reservation/finalize/release logic at all. Confirmed via `git diff ed66a46d8 34c90fd549 --stat`: the only files touched are `retry.go`, `retry_test.go`, `.github/workflows/ci.yml`, `deploy/litellm/config.yaml`, `scripts/report-free-pool-health.py`. `orchestrator.go`, `stream.go`, `stream_responses.go`, `accounting_client.go` are byte-identical to before.

### Why holds cannot strand

The single-terminal-state pattern (`finalized` bool + `defer`-gated release/settle on a **fresh background context**, independent of the request context) predates this PR and is unmodified by it. `dispatchWithRetry` returning an error (including `ctx.Err()` on client disconnect mid-ladder) is handled the same way a pre-existing 429/5xx failure always was: `releaseReservationBackground` fires from the defer, on `context.Background()` with its own `accountingTimeout`, gated on the release actually reaching control-plane before `finalized` is set. Nothing about the new 404 path bypasses this.

### Streaming: retry cannot fire after bytes reach the client

Both `stream.go` and `stream_responses.go` call `dispatchWithRetry` strictly before `w.WriteHeader`/`flusher` ever runs (explicit in-code comment: "safe: no bytes have been written to the client yet at this point"). Confirmed by reading the function bodies directly; the SSE relay loop starts only after `dispatchWithRetry` has already returned a resolved response.

### Retry ladder vs. LiteLLM cooldown — independently re-verified against the pinned image source (not just the PR's citation)

Pulled `litellm==1.98.0` from PyPI and fetched `router_utils/cooldown_handlers.py` + `router.py` at tag `v1.98.0` directly from BerriAI/litellm on GitHub:
- `litellm._should_retry(404)` returns falsy (only 408/409/429/5xx return `True`), so `_should_cooldown_deployment`'s base case (`elif litellm._should_retry(status_code=...) is False: return True`) puts a 404-erroring deployment into cooldown on its **first** failure, not after `allowed_fails`.
- `DEFAULT_COOLDOWN_TIME_SECONDS = 5` (env-overridable, defaults to 5), matching the `'cooldown_time': 5` the original failing CI run actually logged.
- `should_retry_this_error` re-raises immediately for `NotFoundError` and for any non-401/403 status `_should_retry` rejects, confirming neither in-group retry nor cross-group fallback runs for a 404.

The 2.9s worst-case retry ladder is comfortably inside the 5s cooldown window, so a dead pool member cannot be re-selected mid-ladder, provided `allowed_fails`/`cooldown_time` stay off the Router (already documented as a footgun in `deploy/litellm/config.yaml` by this same PR).

### One real, non-blocking gap found (opened separately)

The retry is scoped by matching LiteLLM's error **message text** (`"no fallback model group found"`), not a status code, and there is no test that would fail if that wording changes on a LiteLLM version bump — the fixture in `retry_test.go` is a hardcoded copy of one historical error body, not a live check against the pinned image. Dependabot is configured to bump the Docker image in `/deploy/docker` monthly, which is exactly the path that would carry a reworded error message in. If that happens, the fix silently reverts to the original bug (pool stops failing over on 404) with no unit test going red — only a live-traffic symptom recurring. See the filed issue below for detail and a cheap mitigation.

No action needed on this PR itself; it does not need to be reverted.

PR #1157, opened 2026-08-25
fix: price cache-read and cache-write tokens instead of the flat input rate

sakibsadmanshajib:
## Summary

`CreditsForTokens` (`apps/edge-api/internal/inference/pricing.go`) priced every prompt token, fresh input, cache-read and cache-write alike, at the flat input rate. `model_aliases.cache_read_price_credits` and `cache_write_price_credits` already existed end to end and reached `/v1/models`, but were consulted nowhere in the actual charge.

**Before/after, the audit's worked example** (200,000 cache-read + 5,000 fresh + 2,000 output, illustrative Anthropic-Sonnet-class rates: input 3,000,000 / output 15,000,000 / cache-read 300,000 (0.1x) / cache-write 3,750,000 (1.25x) credits per million):

- **Before (flat rate):** all 205,000 prompt tokens priced as input: `205,000 * 3,000,000 + 2,000 * 15,000,000 = 645,000,000,000 / 1e6 = 645,000` credits.
- **After (cache-aware):** `5,000 * 3,000,000 + 200,000 * 300,000 + 2,000 * 15,000,000 = 105,000,000,000 / 1e6 = 105,000` credits.
- **Ratio: 6.14x overcharge**, matching the audit exactly.

Cache-write side (200,000 write tokens, no read, 2,000 output): before = `200,000*3,000,000 + 2,000*15,000,000 = 630,000` credits; after = `200,000*3,750,000 + 2,000*15,000,000 = 780,000` credits — the prior flat rate undercharged by 19.2%.

## What changed

- `apps/edge-api/internal/inference/cache_usage.go` (new): `NormalizeCacheUsage`, the single implementation of the shape-detection semantics. Selects INCLUSIVE (OpenAI/OpenRouter — subtract cache from `prompt_tokens`) vs EXCLUSIVE (Anthropic-native — `prompt_tokens` already excludes cache, add) by the **wire shape actually decoded** (presence of `cache_read_input_tokens`/`cache_creation_input_tokens`), never by model name. OpenRouter normalizes even a Claude response to the inclusive shape before Hive ever sees it, so branching on model name would take the wrong direction and roughly double the charge on exactly the traffic this fix targets.
- `apps/control-plane/internal/routing/normalizer.go` + its test: **deleted**. It was a fully-tested but zero-caller, narrower version of the same semantics (no fresh-token derivation, no shape detection). One implementation now, on the edge-api side where the real decode happens.
- `apps/edge-api/internal/inference/types.go`: `PromptTokensDetails` gains `CacheWriteTokens` (inclusive shape); `UsageResponse` gains two Anthropic-native pointer fields (`CacheReadInputTokens`, `CacheCreationInputTokens`) used only for shape detection, never echoed back to an OpenAI-compatible customer.
- `apps/edge-api/internal/inference/routing_client.go`: `SelectRoutePricing` gains `CacheReadPriceCredits`/`CacheWritePriceCredits` — these fields already arrived over the wire from control-plane's `/internal/routing/select` (which already served them), but edge-api's struct had nowhere to receive them, so `encoding/json` silently dropped them.
- `apps/edge-api/internal/inference/pricing.go`: `CreditsForTokens` now takes four quantities (fresh, cache-read, cache-write, output) and prices each independently via the existing `metering.ChargeCredits`. `resolveCacheRate` falls back to a documented multiplier (0.1x read / 1.25x write, relative to the alias's own input rate) **only when the catalog price is NULL**, with a loud WARN naming the alias — an explicit stored **zero** is honoured as a real, deliberate rate (no fallback, no warning), matching this codebase's existing "one priced side may legitimately be zero" convention and a real live case (Groq's gpt-oss aliases, seeded 0/0 on purpose per a prior migration's own research). `assertCacheBillingMagnitude` is a runtime self-check that logs loudly if a charge exceeds 2x the flat-rate bound (the signature of a semantics inversion), floored so a tiny, correctly-1-credit-floored charge never false-trips it.
- `apps/edge-api/internal/inference/stream.go`, `orchestrator.go`, `stream_responses.go`: both settlement paths (streaming and sync) now derive and price the fresh/cache-read/cache-write split; `usage_events.cache_write_tokens` (existed end to end, always stored 0) now carries the real number.
- `apps/edge-api/internal/chat/dispatch.go`: session-chat path reuses `inference.UsageResponse` (deleting a second, narrower local `usage` struct that silently dropped every cache field) and now also prices cache-aware.
- `apps/edge-api/internal/anthropic/types.go`: `OAIUsage` gains matching fields so a cache-bearing usage object survives JSON unmarshal on this surface too (decode-only fix; billing for `/v1/messages` already goes through the same orchestrator as chat completions, so it was already covered).
- `supabase/migrations/20260825_01_deepseek_cache_write_price.sql`: seeds a real `cache_write_price_credits` (1.0x input, DeepSeek's own published rate) for `deepseek-v4-flash`/`deepseek-v4-pro`, the one place in the live catalog where a cache price was left at an un-researched 0 rather than a documented deliberate one.

## Metering vs pricing: resolving a conflicting-audit question

A second audit claimed cache tokens were "already metered" citing several files. Investigated file-and-line, both audits are correct about different things:

| | Cache READ | Cache WRITE |
|---|---|---|
| Parsed off provider response | **YES**, pre-existing (`PromptTokensDetails.CachedTokens`) | **NO** — no struct field existed anywhere in edge-api before this PR |
| Persisted to `usage_events` (API-key paths) | **YES**, pre-existing | **NO** — field existed on `RecordEventInput` end to end, but zero writers ever set it (always 0) |
| Used in the charge | **NO** — the core bug | **NO** |

So cache-read was a **presentation gap** (captured, even served to customers via `/v1/models`' published price, just never charged); cache-write was a full metering-and-pricing gap. This PR closes both, through one function (`NormalizeCacheUsage`) that both the ledger-persistence call sites and the pricing call sites now share — no second, parallel capture path.

**Pre-existing, out-of-scope gap found during this investigation** (not touched here): `apps/control-plane/internal/accounting/service.go:446` calls `apiKeySvc.RecordUsageFinalization(..., 0, 0, 0, 0, ...)` with input/output/cache tokens **all** hardcoded to zero, so `api_key_usage_rollups` (a separate budget/quota table from `usage_events`) never receives any of the four. `FinalizeReservationInput` doesn't carry cache tokens at all today. This predates and is broader than the cache-pricing bug; needs its own plumbing task.

**Also found:** `apps/web-console/lib/control-plane/client.ts` already parses both cache prices and both token counts and renders neither — once this PR ships, surfacing it is a frontend-only follow-up.

## Invariants and their tests

1. Shape-correct direction, both ways — `TestNormalizeCacheUsageInclusiveShapeSubtracts`, `TestNormalizeCacheUsageExclusiveShapeAdds`, `TestNormalizeCacheUsageShapeSelectedByWireFieldsNotModelName`.
2. Partition (`fresh + cache_read + cache_write == total`) — asserted inline in the shape tests.
3. `credits = fresh*input + cache_read*rate + cache_write*rate + output*output`, math/big via `metering.ChargeCredits` — `TestCreditsForTokensCacheAwarePricing`, `TestCreditsForTokensCacheWriteUndercharge`.
4. Exact backward compatibility at zero cache tokens — `TestCreditsForTokensExactBackwardCompatibility` (table test against an independent, hand-written old-formula implementation).
5. Fallback on NULL only, never a deliberate zero, loud WARN — `TestResolveCacheRateFallsBackOnNilNotOnDeliberateZero`.
6. Never charge zero for nonzero usage — `TestCreditsForTokensNeverChargesZeroForNonzeroCacheUsage`.
7. Persist what was metered — exercised end to end by `TestExecuteSync_SettlesCacheWritePrice`.

End to end through the real orchestrator: `TestExecuteStreaming_SettlesCacheAwarePrice`, `TestExecuteSync_SettlesCacheAwarePrice`, `TestExecuteSync_SettlesCacheWritePrice`, `TestSessionChatSettlesCacheAwarePrice`.

Full spec: vault `spec-2026-08-25-cache-aware-billing.md`.

## Review findings addressed (all three streams: database, Go correctness, silent-failure)

10 threads posted, 10 resolved, each with a fix or a rebuttal on the thread itself.

**Fixed:**
- `NormalizeCacheUsage` now validates cache-read/cache-write individually, non-negative, in BOTH wire shapes, before deriving `fresh` -- a negative component from a malformed upstream previously inflated `fresh` silently instead of tripping any alarm.
- `CreditsForTokens`' four negative-input clamps are now loud (BUG log naming alias, provider, field), matching `NormalizeCacheUsage`'s own pattern -- raised independently by two review streams.
- `assertCacheBillingMagnitude`'s ceiling now scales off the highest of input/cache-read/cache-write rates, not the flat input rate alone, removing a false-positive case for a catalog row whose cache rate is priced above the input rate.
- Added `hive_cache_billing_magnitude_guard_trips_total` and `hive_cache_billing_fallback_rate_used_total` Prometheus counters (registered on the same registry `/metrics` already serves) plus matching alert rules in `deploy/prometheus/alerts.yml`, so both runtime checks page rather than only logging to stdout -- orchestrator-gated as the exact shape of the 2026-07-31 free-billing incident (logged, unwatched, invisible for three days).
- Recorded **D-056** in `.wolf/decisions.md`: `usage_events.cache_write_tokens` historical 0 rows (before this PR's deploy) are a bug artifact, never a measurement; from this PR's deploy onward, 0 is a real measured zero. Detail in the vault spec doc. Explicitly reconciled against **D-055** (halve-the-rate-only, never-widen-the-billed-class): this PR corrects a RATE on an already-billed class (cache tokens were already inside the flat-priced `prompt_tokens` total before this fix), not a newly-billed one, so it is consistent with D-055 rather than in tension with it.

**Noted, not fixed here (two items, per orchestrator instruction):**
1. Because the cache-rate fallback triggers on a NULL catalog price only, never on an explicit stored zero, and Groq's gpt-oss rows store an explicit 0/0 (a prior migration's own considered, sourced decision that Groq has no published cache rate), if Groq ever starts reporting cache tokens for those routes before that finding is re-verified, they will be silently unpriced (billed at 0 for the cache portion) rather than falling back with a warning. That is the correct behavior today given the unresolved conflict between that sourced finding and this feature's own research doc's contradictory claim about Groq's cache pricing -- overriding either on a guess would be worse -- but it is a real trap worth a tracking issue for the live re-check, which the orchestrator will file.
2. The database reviewer found, outside this PR's diff, that DeepSeek's existing `cache_read_price_credits` ratios are internally inconsistent with each other (deepseek-v4-flash 0.20x input, deepseek-v4-pro 0.033x input) and with the 0.1x figure `pricing.go`'s own comment attributes to DeepSeek generally. This predates this PR and touches a column this PR does not modify -- flagged here so it is not mistaken for something this PR introduced.

## Out of scope

- `/v1/messages` `cache_control` request passthrough (`translate_request.go`) — another agent's task, not touched.
- `accounting/service.go`'s hardcoded-zero rollup figures (see table above).
- Groq's disputed real cache-rate multiplier — flagged for a live check, not guessed.
- Console frontend surfacing of cache prices/counts it already parses.

## Buglog entry

```json
{"date":"2026-08-25","area":"billing","error_message":"prompt-cache tokens billed at the flat input rate on every fixed-price alias","root_cause":"CreditsForTokens had no cache-read/cache-write term despite model_aliases.cache_read_price_credits/cache_write_price_credits existing end to end and reaching /v1/models; cache_creation_input_tokens had no struct field anywhere in edge-api so it was dropped at JSON unmarshal","fix":"NormalizeCacheUsage partitions prompt tokens into fresh/cache-read/cache-write by wire shape (inclusive subtract vs exclusive add), CreditsForTokens prices all four components independently with a NULL-only fallback multiplier and a runtime magnitude guard; PR feat/cache-aware-billing","tags":["billing","money-path","cache","pricing"]}
```

## Test plan

- [x] `go build ./apps/edge-api/... ./apps/control-plane/...` clean
- [x] `go vet ./apps/edge-api/... ./apps/control-plane/...` clean
- [x] Full test suite both modules green (`go test ./apps/control-plane/... ./apps/edge-api/... -count=1`)
- [x] New cache-aware unit and end-to-end tests pass (listed above)
- [x] Existing settlement/settle-from-catalog/chat-settlement tests updated so they fail if cache pricing regresses (invariant 4 table test is the explicit regression guard)

## Summary by CodeRabbit

* **New Features**
  * Added cache-aware billing that separately prices fresh input, cache reads, cache writes, and output tokens.
  * Added support for cache usage details across streaming and non-streaming requests.
  * Customer-facing usage information now includes cache-write token counts.

* **Bug Fixes**
  * Improved handling of provider-specific cache token formats and invalid values.
  * Preserved accurate fractional pricing for cache usage.

* **Monitoring**
  * Added alerts for unusual cache billing amounts and fallback pricing.


supabase/migrations/20260825_01_deepseek_cache_write_price.sql
```diff
@@ -0,0 +1,60 @@
...
+-- one of these two specific aliases AND still carries the un-researched
+-- placeholder (0), so re-running this migration, or running it after a
+-- future repricing that already set a real value, is a no-op rather than a
+-- second write.
+BEGIN;
+
+UPDATE public.model_aliases
+   SET cache_write_price_credits = input_price_credits
+ WHERE alias_id IN ('deepseek-v4-flash', 'deepseek-v4-pro')
+   AND cache_write_price_credits = 0;
```
sakibsadmanshajib: PASS, migration safety. The WHERE clause (`alias_id IN ('deepseek-v4-flash', 'deepseek-v4-pro') AND cache_write_price_credits = 0`) makes this both idempotent and narrow: a second run, or a run after any future repricing that already set a non-zero value, is a no-op because the predicate no longer matches. `model_aliases` is a small catalog table (tens of rows), the UPDATE targets exactly two rows by primary key, and the transaction is a single BEGIN/UPDATE/COMMIT with no long-held lock, so this cannot block the live gateway's route-selection reads. No concerns here.

supabase/migrations/20260825_01_deepseek_cache_write_price.sql
```diff
@@ -0,0 +1,60 @@
...
+--
+-- Idempotent and narrow: the WHERE clause only touches a row that is both
+-- one of these two specific aliases AND still carries the un-researched
+-- placeholder (0), so re-running this migration, or running it after a
+-- future repricing that already set a real value, is a no-op rather than a
+-- second write.
+BEGIN;
+
+UPDATE public.model_aliases
+   SET cache_write_price_credits = input_price_credits
```
sakibsadmanshajib: PASS, value correctness. `cache_write_price_credits = input_price_credits` (1.0x, no premium/discount) matches this repo's own sourced pricing table in `apps/edge-api/internal/inference/pricing.go` (comment on `DefaultCacheReadRateNum/Denom`: "DeepSeek: 0.1x / 1.0x"). Deriving it self-referentially from `input_price_credits` rather than a literal is the right call: it survives the unit rescale a later migration already applied to these two rows (`20260823_40_credit_unit_rescale_billion.sql`) without needing to be re-derived, and it will keep tracking any future repricing of the input column automatically.

Separately, and out of this PR's diff, I checked the sibling `cache_read_price_credits` values seeded by `20260822_02_catalog_alias_restructure.sql`: deepseek-v4-flash is 1790/8946 = 0.20x input, but deepseek-v4-pro is 5236/157080 = 0.033x input, not the same ratio, and neither matches the 0.1x figure `pricing.go`'s own comment attributes to DeepSeek generally. That predates this PR and touches a column this PR does not modify, so it's informational only, not a finding against #1157. Flagging in case it's worth a follow-up: the two DeepSeek aliases' cache-read rates look internally inconsistent with each other and with the documented 0.1x figure.

supabase/migrations/20260825_01_deepseek_cache_write_price.sql
```diff
@@ -0,0 +1,60 @@
...
+--      hive-auto): 20260822_02_catalog_alias_restructure.sql set BOTH cache
+--      columns to an EXPLICIT 0 with a documented reason -- "Zero is the
+--      correct value for a Groq gpt-oss route, which declares no cache
+--      support and has no published cache rate." That is a considered,
+--      sourced decision recorded in this repo, and this migration does not
+--      override it. A later external source (this feature's own research
+--      note) claims Groq DOES publish a cache rate (0.5x read, free write),
+--      which conflicts with that prior finding; resolving the conflict needs
+--      a live re-check against Groq's own docs, not a guess from either side,
+--      so it is left as a follow-up rather than guessed here. Until then,
```
sakibsadmanshajib: PASS, abstention on Groq is correct. `20260822_02_catalog_alias_restructure.sql` recorded an explicit, sourced 0 for the Groq gpt-oss routes' cache columns ("Zero is the correct value... no published cache rate"). This migration's comment names a conflicting external claim (Groq publishes 0.5x read / free write) and correctly declines to resolve that conflict by guessing, deferring to a live check instead. Confirmed the fallback path actually covers the gap safely in the interim: `resolveCacheRate` in `pricing.go` only takes the documented default multiplier when the stored price is genuinely NULL, and treats a stored 0 as deliberate (no fallback, no warn) -- and the Groq rows are exactly that, an explicit stored 0, not NULL, so if Groq ever does start reporting cache tokens for these routes before the conflict is resolved, they get billed at that stored 0, unpriced, not at the 0.5x/free the external claim asserts and not a loud warning either. That is consistent with "do not override a sourced decision on a guess," but it does mean the abstention has a silent-undercharge tail until the live check happens. Worth a tracking issue if one doesn't exist yet; not a blocker for this PR, which correctly does not touch these rows.

apps/edge-api/internal/inference/stream.go
```diff
@@ -21,13 +21,27 @@ import (
...
-	CachedTokens    int64
-	TotalTokens     int64
-	HasUsage        bool
-	Content         strings.Builder
+	// CachedTokens is the cache-READ subset of InputTokens.
+	CachedTokens int64
+	// CacheWriteTokens is the cache-WRITE quantity this turn created. Never
+	// populated before this feature (usage_events.cache_write_tokens always
+	// stored 0); now carries the real figure NormalizeCacheUsage derives.
+	CacheWriteTokens int64
```
sakibsadmanshajib: MEDIUM, non-blocking. Confirmed end to end: `usage_events.cache_write_tokens` (`20260330_02_usage_accounting.sql`, `bigint not null default 0`) really was never populated before this PR (this field's own comment says so, and `orchestrator.go`'s pre-PR `recordCompletedEvent` only ever set `CacheReadTokens`) and now receives the real value from `NormalizeCacheUsage` on every path I traced: sync (`recordCompletedEvent`), streaming-complete (`settleStream` -> `recordCompletedEvent`), and streaming-interrupted (`recordInterruptedEvent`). `bigint` cannot overflow at realistic token volumes, and no query in this diff filters or aggregates on this column without an index, so no index gap either.

The one open question the task asked me to check: historical rows written before this PR shipped have `cache_write_tokens = 0` as a bug artifact (the field was simply never set), and rows written after this PR ships can also legitimately be 0 (a real request that did no cache write). Both look identical in the column: `NOT NULL DEFAULT 0`, no version or epoch marker anywhere in `usage_events` to tell them apart. That is fine for forward billing but not fine for reconstructing a customer dispute over a request that predates this deploy: nothing in the row lets anyone tell "genuinely free" apart from "not measured yet." Suggest recording the cutover (this PR's merge/deploy timestamp) in `.wolf/decisions.md` so a future billing investigation knows to treat any `cache_write_tokens = 0` row created before that timestamp as unknown rather than verified-zero. No schema or code change needed for this PR; this is a documentation gap, not a bug in the diff.

sakibsadmanshajib (commented):
Database/migration-scope review (money path). Full findings in the summary comment below.

apps/edge-api/internal/inference/pricing.go
```diff
@@ -96,19 +190,35 @@ func CreditsForTokens(route SelectRouteResult, inputTokens, outputTokens int64)
...
+	}
+	if cacheReadTokens < 0 {
+		cacheReadTokens = 0
+	}
+	if cacheWriteTokens < 0 {
+		cacheWriteTokens = 0
 	}
 	if outputTokens < 0 {
 		outputTokens = 0
 	}
```
sakibsadmanshajib: MEDIUM, telemetry gap (not a pricing-correctness bug).

These four clamps (`freshInputTokens`, `cacheReadTokens`, `cacheWriteTokens`, `outputTokens` each `< 0 => 0`) are silent: no `log.Printf`, no metric, nothing. This is the exact gap the adversarial review of the sibling PR #1152 flagged for `freshInputTokens`, and this PR (which owns `pricing.go`) extends the same silent pattern to two more fields (`cacheReadTokens`, `cacheWriteTokens`) instead of closing it.

It is inconsistent with the philosophy this very file establishes elsewhere: `NormalizeCacheUsage`'s own negative-fresh clamp (`cache_usage.go`) logs a loud `BUG:` line naming the alias before clamping, specifically because "the trap here runs the other way too, which is exactly why this is a clamp-and-alarm, not a silent clamp." The same reasoning applies here: a provider that starts sending a negative token count (malformed response, a shape regression) will clamp to zero and settle a possibly wrong charge with zero operational signal that anything happened.

Direction wise this is safe. Clamping toward zero can only reduce, never inflate, one component's charge, and it cannot be triggered by any usage this PR's own callers produce today since `NormalizeCacheUsage` already floors at zero before `CreditsForTokens` ever sees these values. So I am not blocking on it. But since this function is explicitly a money path boundary for "external input" per its own doc comment two lines up, recommend mirroring the `NormalizeCacheUsage` pattern: `log.Printf("inference: BUG: negative token count clamped ...")` when any of these four fire, naming `route.AliasID` and which field. Cheap fix, closes the loop the sibling review opened.

sakibsadmanshajib:
## Database review summary (PR #1157, migration/schema/persistence scope only)

**Verdict: sound. No blocking findings. Do not block merge on this review.**

### 1. Migration safety (`20260825_01_deepseek_cache_write_price.sql`)
Idempotent and safely re-runnable: the `WHERE ... AND cache_write_price_credits = 0` predicate is its own guard, a second run or a run after a future repricing is a no-op. `model_aliases` is a small catalog table; the UPDATE hits exactly two rows by primary key inside one BEGIN/COMMIT, no lock hazard for the live gateway. Pass.

### 2. Seeded value correctness
`cache_write_price_credits = input_price_credits` (1.0x) matches this repo's own sourced DeepSeek pricing note in `pricing.go` ("DeepSeek: 0.1x / 1.0x"). The self-referential UPDATE form is deliberate and correct: it already accounts for the prior unit rescale and stays correct under a future repricing, same pattern as `20260823_40_credit_unit_rescale_billion.sql`. The Groq abstention is correct: it defers to a prior sourced, considered decision instead of overriding it on an unverified external claim, and the `resolveCacheRate` fallback covers the gap safely (with one caveat noted inline: since Groq's stored value is an explicit 0, not NULL, it stays silently unpriced rather than warning if Groq starts reporting cache tokens before the conflict is resolved). Non-blocking, flagged inline for a possible tracking issue.

One informational item found outside this PR's diff: the two DeepSeek aliases' existing `cache_read_price_credits` values (from `20260822_02`, not touched by #1157) don't share a consistent ratio to `input_price_credits` with each other (0.20x vs 0.033x) or with `pricing.go`'s own 0.1x comment. Not this PR's bug, flagging for awareness.

### 3. Rollback behavior
If the pricing code is reverted after this migration ships, the failure mode is the benign one: a populated `cache_write_price_credits` column that nothing reads, not code reading an unpopulated column. This repo's own deploy pipeline (`deploy-demo-box.yml`) hard-orders migrate-before-deploy, so the reverse (new code live before the migration lands) cannot happen through normal CI/CD either. Confirmed, not a risk in this PR.

### 4. NULL vs zero
Verified at both the schema and code layer. `model_aliases.cache_read_price_credits` / `cache_write_price_credits` are plain nullable `bigint` (no `DEFAULT 0`), the Go side reads them as `*int64` all the way through (`catalog/client.go`, `pricing.go`), and `resolveCacheRate` is explicitly tested for both branches (`pricing_cache_test.go`: nil falls back and warns, explicit zero is honoured silently). This PR's migration only ever writes to rows that were already an explicit 0, so it doesn't collapse the distinction anywhere. Pass.

### 5. `usage_events.cache_write_tokens`
Confirmed this column really was never populated before this PR (dead field, `orchestrator.go` only ever set `CacheReadTokens`), and now gets real values on every path traced: sync completion, streaming completion, and streaming interruption all route through `NormalizeCacheUsage`. Column is `bigint`, no overflow risk at realistic token volumes.

One real, non-blocking gap: historical rows with `cache_write_tokens = 0` (bug artifact, never measured) are byte-identical to post-fix rows that are genuinely, correctly zero. Nothing in the row distinguishes them. Recommend recording this PR's deploy timestamp in `.wolf/decisions.md` as the cutover, so a future billing dispute over a pre-cutover request treats `cache_write_tokens = 0` as "not measured," not "verified zero." No schema or code change requested for this PR.

### 6. Index/query implications
No new filtering or aggregation on `cache_read_tokens`/`cache_write_tokens` in this diff; existing selects are by row id. No index gap introduced.

---
**Severity roll-up:** zero HIGH/CRITICAL. One MEDIUM (historical-zero ambiguity, documentation fix, non-blocking). Everything else is a PASS confirmation or an informational note outside this PR's diff. Reviewed migration + schema + `usage_events` persistence only, per assigned scope; did not review `pricing.go` arithmetic correctness, streaming/SSE handling, or non-DB Go code in depth beyond tracing the persistence path.

apps/edge-api/internal/inference/pricing.go
```diff
@@ -66,11 +67,104 @@ func CanPriceTokens(route SelectRouteResult) bool {
...
+// would have produced. It never blocks, alters, or refuses the charge: a
+// ceiling breach is loud evidence to investigate, not grounds to fail a
+// request that has already been served (D-034 is about never charging zero
+// for delivered work, not about refusing a suspicious charge after the fact).
+func assertCacheBillingMagnitude(route SelectRouteResult, fresh, cacheRead, cacheWrite, output, credits int64) {
+	totalPrompt := fresh + cacheRead + cacheWrite
+	ceiling := metering.ChargeCredits(
+		metering.UnitCharge{Quantity: totalPrompt * 2, CreditsPerMillion: route.Pricing.InputCredits()},
+		metering.UnitCharge{Quantity: output, CreditsPerMillion: route.Pricing.OutputCredits()},
+	)
```
sakibsadmanshajib: LOW, nit, non-blocking.

The ceiling only scales `route.Pricing.InputCredits()` (the flat input rate) by 2x. That bound is correct today because every real cache-write multiplier this feature documents tops out at 2.0x (the Anthropic 1h TTL case) and every cache-read multiplier is a discount below 1x, so pricing every prompt token at 2x the input rate really is a safe upper bound.

It stops being a safe bound only for a hypothetical catalog row where the cache rate is set higher than 2x the input rate, or where the input rate itself is 0 while a cache rate is nonzero (for example a promotional zero base rate with a real cache charge). In that case this guard would log a BUG line on every request even though the charge is correct, a false positive rather than a false negative, so it degrades to noise rather than to a missed real bug. Given the function is explicitly non-blocking and log-only, this is a robustness nit, not a correctness issue. Consider basing the ceiling on `max(inputRate, cacheReadRate, cacheWriteRate)` instead of `inputRate` alone if you want the guard to stay quiet under every catalog configuration, not just the ones seeded today.

sakibsadmanshajib:
## Adversarial review, money path (independent, not the builder's self report)

Read the pushed diff directly (`gh pr diff 1157`, worktree checkout of `39c6e71b6`), cross-checked against the vendor-sourced cache-billing contract and the prior read-only audit, and confirmed CI (Go tests edge-api, control-plane, migrations against a throwaway Postgres) is green.

### Attack results, by the eight vectors in the brief

1. **Inclusion semantics selected by wire shape, not model family.** Confirmed. `NormalizeCacheUsage` (`cache_usage.go`) branches only on `usage.CacheReadInputTokens != nil || usage.CacheCreationInputTokens != nil` (Anthropic's own field presence). `aliasID`/`provider` parameters are used only in the log line, never in the branch condition. The Claude-via-OpenRouter trap case is an explicit test (`TestNormalizeCacheUsageShapeSelectedByWireFieldsNotModelName`) that deliberately names an Anthropic alias and asserts the subtract path still fires (fresh = 21, not the larger add result). No model-name branch exists anywhere in the diff. Clean.

2. **Partition invariant.** Holds by construction for both shapes: inclusive derives fresh by subtraction so the three components algebraically reconstruct `prompt_tokens`; exclusive's total is defined as the sum of the three components, so it is tautologically true there. The one place it can be violated is the documented clamp-and-alarm case (cache tokens exceed `prompt_tokens`), which is intentional and loud, not silent. No case found where it silently breaks.

3. **Backward compatibility.** Verified structurally, not just by the table test: when `cacheReadTokens`/`cacheWriteTokens` are 0, they multiply out to 0 regardless of the resolved cache rate, so byte-identical behavior does not depend on the fallback rate being right, it is guaranteed by the arithmetic. The test suite's boundaries are reasonable (zero tokens, tiny counts, an embedding-only route, 2M/500K large volume, a cache-priced route with zero cache quantities) though there is no explicit "exactly 1 token" case; not a gap I'd block on given the structural guarantee above.

4. **1-credit floor x magnitude guard interaction.** Checked hard, this is where I expected to find something. `TestCreditsForTokensNeverChargesZeroForNonzeroCacheUsage` proves a single cache-read token (fresh=0, output=0) still floors to 1, because the floor condition sums all four quantities, not just fresh+output. `assertCacheBillingMagnitude`'s own ceiling gets the identical floor treatment so a tiny real request cannot self-trip the guard. No input found that produces a zero charge on nonzero usage. No revenue-loss finding here.

5. **math/big discipline.** No `float64` on the pricing path. `scaleRate` (the fallback-rate scaler) and `metering.ChargeCredits` (unchanged) both use `big.Int` with the same round-half-up convention, so a fallback rate can never round differently than the charge that consumes it. The only `float64` hits in the diff are test-side JSON `interface{}` decodes of a response body and one test sanity ratio, plus an unrelated gofmt realignment of `Temperature`/`TopP` on the Responses API struct.

6. **NULL-only fallback.** Confirmed and tested both directions: `resolveCacheRate` returns the stored value whenever the pointer is non-nil, including an explicit stored 0, with no WARN; only a nil pointer triggers the documented multiplier, and only when quantity > 0 (so an unused route never spams a WARN). Traced the nil-vs-zero contract all the way to the DB: `cache_read_price_credits`/`cache_write_price_credits` are nullable `bigint` columns with no default, scanned into `*int64` end to end through control-plane and into edge-api's `SelectRoutePricing`, and Go's `omitempty` on a pointer field only drops a nil pointer, not a pointer to 0, so an explicit zero really does survive the wire. This chain isn't in the diff but the new code's correctness depends on it, and it holds.

7. **Silent clamp (sibling PR #1152 gap).** Found, and posted inline. `CreditsForTokens`'s four `< 0 => 0` clamps (fresh, cache-read, cache-write, output) are silent, no log line, which is the exact gap flagged on #1152, unresolved here despite this PR owning the file and despite establishing a loud clamp-and-alarm pattern one file over in `NormalizeCacheUsage`. Rated MEDIUM, not blocking: the clamp direction is the safe one (can only reduce a charge, never inflate it) and is not reachable through any of this PR's own callers today, since `NormalizeCacheUsage` already floors negative values with its own alarm before `CreditsForTokens` ever sees them. Still, it is dead-simple to fix and closes a loop the org's own review process already opened once.

8. **Tests that cannot go red.** Read them, not just the diff summary. `TestCreditsForTokensExactBackwardCompatibility` computes its expected value from a hand-spelled independent formula (`oldFlatCreditsForTokens`), not by calling any production helper, so it cannot drift into checking the new code against itself. The cache-aware tests assert against hand-worked arithmetic in the test comments (105,000 / 780,000 / the 6.14x ratio) rather than calling `CreditsForTokens` twice and diffing. The negative-clamp-and-warn test and the resolve-rate nil-vs-zero test both assert on captured log output, not just return values, so a regression to silent-fallback behavior would fail them. All five test files the prior audit named as "cannot go red for this bug" now carry a genuinely capable case. Confirmed real tests, not decoration.

### One more thing checked outside the eight, since it's the same file: `assertCacheBillingMagnitude`'s 2x ceiling scales off the flat input rate alone. Posted as a LOW nit inline: safe today given the documented multiplier range (reads discount below 1x, writes cap at 2x for the 1h TTL case), but would false-positive-WARN (never mischarge) under a hypothetical catalog row with a cache rate above 2x input or a zero input rate paired with a nonzero cache rate. Non-blocking, log-only either way.

### Verdict

**Do not block.** One MEDIUM (silent clamp, telemetry gap, safe direction, not reachable today) and one LOW nit (magnitude-guard ceiling assumption), both non-blocking and both posted inline with concrete fixes. The core claim, cache-read and cache-write tokens are now priced at their own catalog rate instead of the flat input rate, with the wire-shape-not-model-name inclusion/exclusion split, the partition invariant, the floor, and the NULL-vs-zero fallback contract all correctly implemented and genuinely tested, holds up under adversarial reading of the actual pushed code.

apps/edge-api/internal/inference/pricing.go
```diff
@@ -66,11 +67,104 @@ func CanPriceTokens(route SelectRouteResult) bool {
...
+	// The same 1-credit floor CreditsForTokens applies to a nonzero-but-cheap
+	// request also applies here: without it, a tiny request (e.g. a single
+	// output token on a low per-million rate) has a raw proportional ceiling
+	// of 0, and the floor alone would trip this guard on every such request,
+	// exactly the noisy-false-positive shape that gets a real WARN ignored.
+	if totalPrompt+output > 0 && ceiling < 1 {
+		ceiling = 1
+	}
+	if credits > ceiling {
+		log.Printf("inference: BUG: cache billing magnitude guard tripped alias=%s credits=%d ceiling=%d fresh=%d cache_read=%d cache_write=%d output=%d: "+
```
sakibsadmanshajib: HIGH, telemetry gap: this is the only production-time defense against a cache semantics inversion in NormalizeCacheUsage, the exact class of bug that would reproduce the audit's 6.14x scenario in reverse. It only ever calls log.Printf. No Prometheus counter, no alert rule. This package already has prometheus.Registerer wiring elsewhere (see stage_metrics.go's StageMetrics), so a hive_cache_billing_anomaly_total counter incremented right here is a small addition, not a new capability. As written, a future catalog edit or provider integration that inverts this arithmetic in production leaves a BUG line in container stdout as its only trace, and deploy/prometheus/alerts.yml currently defines zero rules for anything billing or pricing related. Given this repo's own history, the 2026-07-31 three-day free-billing incident was exactly a logged-but-unwatched failure, recommend a counter here at minimum, ideally an alert rule on it, though this is an operability gap rather than a correctness one and need not block merge.

sakibsadmanshajib (commented):
Reviewed the pushed diff end to end (pricing.go, cache_usage.go, stream.go, orchestrator.go, routing_client.go, types.go, dispatch.go, anthropic/types.go, the DeepSeek migration, and all new tests), specifically hunting for ways this can silently misprice given this repo's 2026-07-31 history.

What is solid, said plainly: the negative-fresh clamp in cache_usage.go is loud, it logs a BUG line naming the alias and provider, matching this PR's own D-034 clamp-and-alarm framing, and it closes the specific gap a sibling review flagged unalarmed in this file and in PR #1152. The wire-shape detection (inclusive-subtract vs exclusive-add, selected on field presence rather than model name) is tested both directions plus the OpenRouter-proxied-Claude trap case specifically. The streaming aggregator (UsageAccumulator.Accumulate) overwrites usage fields per chunk rather than accumulating them, so there is no double-count risk, and this pipeline only ever sees a single OpenAI-shape terminal usage chunk today rather than the Anthropic-native cumulative message_delta sequence, which is correctly documented as unreachable through today's LiteLLM-only dispatch path. CreditsForTokens' one-credit floor is applied to the four-component sum, so nonzero cache-only usage cannot settle at zero credits. And the routing_client.go fix genuinely closes the stated root cause: I traced it independently and confirmed control-plane's catalog.CatalogPricing already carries cache_read_price_credits/cache_write_price_credits with matching JSON tags, and LoadAliasPricing already selects them from model_aliases; edge-api's SelectRoutePricing struct simply had no field to receive either one, so encoding/json silently dropped both on every route lookup before this PR. That is the real mechanism behind the flat-rate overcharge the audit found. Backward compatibility at zero cache tokens is verified against an independently hand-written old formula rather than the new implementation checked against itself.

Findings below, worst first, inline comments posted at the exact lines.

Most likely way this ships and silently misprices later: not an arithmetic inversion, that class is well covered by tests and by the new runtime assertion, but the runtime assertion itself going unheard. assertCacheBillingMagnitude and resolveCacheRate's fallback WARN are both log.Printf only, with no Prometheus counter and no alert rule, and deploy/prometheus/alerts.yml has zero rules today for anything billing or pricing related. A later catalog edit, a new provider integration, or the PR's own flagged-as-unresolved Groq cache rate could trip these lines forever in container stdout with nobody watching, which is the same shape as the 2026-07-31 incident this repo was already burned by.

Recommendation: do not block. The billing arithmetic itself is correct, well tested including an independent backward-compatibility check, and the specific gap the task called out, the negative-fresh clamp alarming, is fixed here. The findings below are operability and defense-in-depth gaps, not open money-path bugs, and none of them change the charge that ships today. Recommend closing the telemetry gap (first finding) before or shortly after merge, not as a merge blocker.

apps/edge-api/internal/inference/pricing.go
```diff
@@ -66,11 +67,104 @@ func CanPriceTokens(route SelectRouteResult) bool {
...
+// fell back, only when quantity is actually nonzero: a route that has never
+// once seen a cache token should not spam a WARN every request just because
+// its cache columns happen to be unset.
+func resolveCacheRate(priceCredits *int64, inputRate, numerator, denominator, quantity int64, aliasID, side string) int64 {
+	if priceCredits != nil {
+		return *priceCredits
+	}
+	fallback := scaleRate(inputRate, numerator, denominator)
+	if quantity > 0 {
+		log.Printf("inference: WARN falling back to default cache-%s rate alias=%s input_rate=%d fallback_rate=%d tokens=%d: "+
```
sakibsadmanshajib: MEDIUM, same shape as the magnitude guard below: a real, useful signal, the catalog has no cache price for an alias that is actually reporting cache tokens, that only ever reaches a log line. Worth the same counter treatment, for example a counter labeled by alias and side, so a newly added alias that forgot to seed cache_read_price_credits or cache_write_price_credits shows up on a dashboard instead of silently running on the fallback multiplier indefinitely.

apps/edge-api/internal/inference/cache_usage.go
```diff
@@ -0,0 +1,81 @@
...
+			CacheReadTokens:  derefPrice(usage.CacheReadInputTokens),
+			CacheWriteTokens: derefPrice(usage.CacheCreationInputTokens),
+		}
+	}
+
+	// INCLUSIVE shape: prompt_tokens counts the cached subset too. Subtract,
+	// never add.
+	var cacheRead, cacheWrite int64
+	if usage.PromptTokensDetails != nil {
+		cacheRead = usage.PromptTokensDetails.CachedTokens
```
sakibsadmanshajib: MEDIUM-HIGH: this validates only the aggregate (fresh < 0 after subtraction, checked below). It never checks that cacheRead or cacheWrite themselves are non-negative before the subtraction. If an upstream ever reports a negative cached_tokens or cache_write_tokens value, a malformed provider response or a proxy bug, fresh = prompt_tokens - cacheRead - cacheWrite comes out inflated rather than negative, so the loud clamp a few lines below never fires, and the corrupted negative value flows on into CreditsForTokens, where it is silently clamped to zero with no log at all (see the parallel comment on pricing.go). Net effect: the request gets billed with an inflated fresh count and a lost cache discount, silently, with no alarm anywhere in the chain, and the fresh + cacheRead + cacheWrite == prompt_tokens partition invariant this file's own tests assert breaks for exactly this input. Recommend validating cacheRead >= 0 and cacheWrite >= 0 here too, with the same clamp-and-alarm treatment already applied to fresh two lines down.

apps/edge-api/internal/inference/pricing.go
```diff
@@ -96,19 +190,35 @@ func CreditsForTokens(route SelectRouteResult, inputTokens, outputTokens int64)
 			route.AliasID)
 		return 0
 	}
-	if inputTokens < 0 {
-		inputTokens = 0
+	if freshInputTokens < 0 {
+		freshInputTokens = 0
+	}
+	if cacheReadTokens < 0 {
```
sakibsadmanshajib: MEDIUM: these four clamps are completely silent, no log at all, in contrast to the loud clamp NormalizeCacheUsage added one call frame up for the fresh < 0 case. If a negative cacheReadTokens or cacheWriteTokens ever reaches this function directly, a future caller that does not route through NormalizeCacheUsage, or the corrupted-component case noted on cache_usage.go, it disappears here with zero trace. Given this PR's own stated D-034 philosophy, clamp and alarm rather than clamp silently, on a money path, recommend at least a log line naming which quantity went negative and for which alias, matching the treatment already given to the aggregate fresh case.

supabase/migrations/20260825_01_deepseek_cache_write_price.sql
sakibsadmanshajib: Acknowledged, no action needed. Confirmed the WHERE clause makes this idempotent and narrow, as described.

supabase/migrations/20260825_01_deepseek_cache_write_price.sql
sakibsadmanshajib: Acknowledged. The DeepSeek cache-write derivation is correct per pricing.go's own sourced table. Added the cache-read internal-inconsistency finding to the PR body as a pre-existing, not-mine note, per orchestrator instruction, so it is not mistaken for something this PR introduced. Not touching cache_read_price_credits in this PR: it is outside this migration's scope and outside this PR's diff.

supabase/migrations/20260825_01_deepseek_cache_write_price.sql
sakibsadmanshajib: Acknowledged. Correct characterization: the abstention is deliberate (do not override a sourced Groq finding on an unverified external claim) and the silent-undercharge tail you identified is real but bounded to Groq's own explicitly-zero rows, which is the intended fail-safe shape given the unresolved conflict. Noted in the PR body per orchestrator instruction; the orchestrator will file the tracking issue for the live Groq cache-rate check.

apps/edge-api/internal/inference/stream.go
sakibsadmanshajib: Fixed via documentation, exactly as you suggested: no schema or code change. Added D-056 to .wolf/decisions.md (terse pointer) and a matching section to vault spec-2026-08-25-cache-aware-billing.md (detail): before this PR's deploy, cache_write_tokens=0 rows are a bug artifact (never measured); from this PR's deploy onward, 0 is a real measured zero. Also records why this is consistent with D-055 (halve-the-rate-only, never-widen-the-billed-class): cache tokens were already inside the flat-priced prompt_tokens total pre-fix, so this corrects a rate on an already-billed class rather than billing a new one.

apps/edge-api/internal/inference/pricing.go
sakibsadmanshajib: Fixed. All four clamps in CreditsForTokens now log a BUG line naming route.AliasID, route.Provider and the specific field before clamping, via a shared clampNonNegative helper also used by NormalizeCacheUsage's own clamps (single implementation, not a second copy of the log line shape). See TestCreditsForTokensLoudlyClampsNegativeInputs.

apps/edge-api/internal/inference/pricing.go
sakibsadmanshajib: Fixed, not just rebutted: the ceiling now scales off the highest of InputCredits/CacheReadPriceCredits/CacheWritePriceCredits rather than the input rate alone, which removes the false-positive case you identified (a cache rate priced above 2x input, or a zero input rate paired with a nonzero cache rate). See TestAssertCacheBillingMagnitudeGuard's new 'ceiling scales off the highest cache rate' subtest, which pins exactly this scenario (cache-write priced above the flat input rate) and confirms no false trip.

apps/edge-api/internal/inference/pricing.go
sakibsadmanshajib: Fixed per orchestrator ruling (gated as HIGH): added hive_cache_billing_magnitude_guard_trips_total, a CounterVec registered on the same registry apps/edge-api/internal/proxy already serves at /metrics (via NewStageMetrics, which both this package's Orchestrator and this counter now share), incremented at the point the guard trips, labelled alias+provider. Also added a matching Prometheus alert rule (CacheBillingMagnitudeGuardTripped, severity critical, increase()-based per the existing SignupProvisioningFaulting pattern in alerts.yml) so a trip actually pages rather than only logging. See TestCreditsForTokensCounterOnMagnitudeGuardTrip.

apps/edge-api/internal/inference/pricing.go
sakibsadmanshajib: Fixed, same treatment as the magnitude guard: added hive_cache_billing_fallback_rate_used_total, labelled alias+provider+side, incremented in resolveCacheRate alongside the existing WARN. Added a matching alert rule (CacheBillingFallbackRateInUse, severity warning, since this is billing continuing correctly at a documented fallback rather than a wrong charge) so a newly-added unseeded alias shows up without anyone having to go looking. See TestResolveCacheRateCounterOnFallback.

apps/edge-api/internal/inference/cache_usage.go
sakibsadmanshajib: Fixed. NormalizeCacheUsage now validates cacheRead and cacheWrite each individually (via the same clampNonNegative helper, both wire shapes) BEFORE fresh is derived from them, with a loud BUG log naming the alias and which field. This closes exactly the gap you found: a negative component no longer silently inflates fresh past the aggregate clamp. See TestNormalizeCacheUsageNegativeCacheComponentClampsLoudlyInclusiveShape and its exclusive-shape sibling.

apps/edge-api/internal/inference/pricing.go
sakibsadmanshajib: Fixed, same change as the sibling finding at pricing.go:204 (two reviewers independently flagged the same four clamps): all four are now loud, via the shared clampNonNegative helper, naming route.AliasID, route.Provider and the field. See TestCreditsForTokensLoudlyClampsNegativeInputs.

apps/edge-api/internal/inference/pricing.go
```diff
@@ -66,11 +67,122 @@ func CanPriceTokens(route SelectRouteResult) bool {
...
+// half up via math/big, the same convention metering.ChargeCredits uses so a
+// fallback rate never rounds differently than the charge it will feed.
+func scaleRate(rate, numerator, denominator int64) int64 {
+	product := new(big.Int).Mul(big.NewInt(rate), big.NewInt(numerator))
+	denom := big.NewInt(denominator)
+	quotient, remainder := new(big.Int).QuoRem(product, denom, new(big.Int))
+	if new(big.Int).Mul(remainder, big.NewInt(2)).Cmp(denom) >= 0 {
+		quotient.Add(quotient, big.NewInt(1))
+	}
+	return quotient.Int64()
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in c7c0c0ac5. The finding is correct: scaleRate pre-scaled the fallback multiplier into a whole credits-per-million integer, and 1/10x of any input rate below 5 collapses to zero there, so the request fell through to the 1-credit floor no matter how many cache-read tokens it consumed.

The fallback is now carried as an exact fraction end to end. metering.UnitCharge gains a RateDivisor field: one component prices at CreditsPerMillion/RateDivisor credits per million units. ChargeCredits folds every divisor onto their least common multiple before its single sum-then-round division, which is a ratio-preserving integer rewrite, so the exact rational total keeps the one-rounding contract (two halves can still never round independently). Callers that pass no divisor are bit-for-bit unchanged; the backward-compatibility test suite confirms byte-identical charges for zero-cache requests.

resolveCacheRate now returns that UnitCharge directly: catalog price as-is, or Quantity x numerator at input rate over RateDivisor denominator on fallback. The lossy scaleRate helper is deleted. Regression test TestCreditsForTokensFractionalFallbackSurvivesChargeArithmetic covers your exact scenario (inputRate=1, NULL cache-read price, 100,000,000 cache-read tokens = exactly 10 credits), the write-side 5/4x figure (125), and a case proving fractional parts sum before the single round (0.5 + 1.5 = 2, not 3).

apps/edge-api/internal/inference/stage_metrics.go
```diff
@@ -27,6 +27,31 @@ type StageMetrics struct {
...
+// registration happens once at startup (NewStageMetrics, below) on the same
+// registry apps/edge-api/internal/proxy.NewEdgeMetrics already serves at
+// /metrics, and a test that never calls NewStageMetrics simply never
+// registers them -- WithLabelValues(...).Inc() on an unregistered CounterVec
+// is a normal, harmless no-op observation, the same nil-safety principle
+// StageMetrics.observe already relies on for the duration histogram.
+var (
+	cacheBillingMagnitudeGuardTrips = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "hive_cache_billing_magnitude_guard_trips_total",
+		Help: "Times the cache-aware billing magnitude guard fired: a charge exceeded 2x the flat-rate bound, the signature of a cache semantics inversion (vault spec-2026-08-25-cache-aware-billing.md).",
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in c7c0c0ac5. Both spots now say what the implementation does: the counter help and the alert rule describe the bound as 2x the highest-rate bound (the highest of the input, cache-read and cache-write rates) instead of the stale flat-rate wording. The guard's own doc comment and BUG log line in pricing.go carried the same stale phrase and are updated in the same commit for consistency.

apps/edge-api/internal/inference/pricing_cache_test.go
```diff
@@ -0,0 +1,367 @@
...
+		if buf.Len() != 0 {
+			t.Errorf("a route that metered zero cache tokens must not warn just because its price is unset, got: %q", buf.String())
+		}
+	})
+}
+
+// TestAssertCacheBillingMagnitudeGuard is the runtime half of the
+// magnitude-guard self-check (contract section 5): a charge within twice the
+// flat-rate bound is silent, and a charge that exceeds it -- the signature of
+// a semantics inversion -- logs a loud BUG line naming the alias.
```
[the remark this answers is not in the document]
sakibsadmanshajib: Fixed in 63438cda7. Both strings updated to the highest-rate bound, matching the implementation and the third subtest; comment and test name only, no behavior change. Inference package tests rerun and green.

PR #1163, opened 2026-08-25
fix: carry the remaining Anthropic /v1/messages request fields through the translator

### LINKED ISSUE #1153: Anthropic /v1/messages translator silently drops or inverts request fields
Found while implementing `cache_control` passthrough (PR #1152). `apps/edge-api/internal/anthropic/translate_request.go` rebuilds the Anthropic request field by field into the OAI shape, so any field nobody explicitly carried is lost. `cache_control` was fixed in #1152. These remain.

**Behaviour inversion, most severe**

- `tool_choice: {"type": "none"}` silently becomes `"auto"`. A caller asking the model NOT to call tools gets a model that may call them. This is not an omission, it is the opposite of the request.

**Silently dropped**

- `MessagesRequest.Metadata.user_id` is parsed but never read by `ToOAIRequest`.
- `ToolChoice.disable_parallel_tool_use` has no field at all.
- Anthropic's `top_k` has no field on `MessagesRequest` at all.
- Extended thinking has no representation: no `thinking` request field, and no `thinking` / `redacted_thinking` content blocks. A thinking-only message silently produces empty content.
- A pre-existing image drop in the tool-calls-plus-text branch, deliberately preserved in #1152 to keep its regression guard exact, documented in code.

**Why this class matters**

A field-by-field rebuild loses whatever nobody thought to add, and it fails silently every time. `cache_control` cost every agent client on the gateway its prompt caching, undetected. Consider a translator test that round trips a maximal request and asserts nothing is lost, so the next added field cannot go missing quietly.

sakibsadmanshajib:
## Problem

`apps/edge-api/internal/anthropic/translate_request.go` rebuilds the Anthropic `/v1/messages` request field by field into the internal OpenAI-shaped `OAIRequest`. Any field nobody explicitly carried is lost, silently, every time. `cache_control` was one of these and cost every agent client on the gateway its prompt caching until #1152 fixed it. This PR fixes the rest of the fields listed in #1153.

## Field-by-field disposition

Per the task brief, each field below is deliberately dispositioned rather than reflexively carried through:

1. **`tool_choice: {"type": "none"}` — behaviour inversion, fixed first.** `convertToolChoice`'s switch had no `"none"` case, so it fell to the `default` branch and returned `"auto"`. A caller explicitly forbidding tool use got a model that could call them: not an omission, the opposite of the request. **Disposition: translated to the exact OpenAI equivalent** — `tool_choice: "none"` means the identical thing on both APIs.
2. **`metadata.user_id`** — parsed but never read by `ToOAIRequest`. **Disposition: translated to an equivalent** — lands on `OAIRequest.User`, OpenAI's own end-user tracking field (same purpose: abuse detection / per-user attribution on the provider side, different name).
3. **`tool_choice.disable_parallel_tool_use`** — no field at all. **Disposition: translated to an equivalent** — maps to OpenAI's `parallel_tool_calls`, inverted (`disable_parallel_tool_use: true` → `parallel_tool_calls: false`). Only ever emitted as `false`; Anthropic's default (`disable=false`, parallel allowed) already matches OpenAI's own default, so there's nothing to carry when the caller never asked to forbid it.
4. **`top_k`** — no field on `MessagesRequest` at all. **Disposition: forwarded as an extra field LiteLLM passes along.** There is no OpenAI-standard equivalent (it isn't part of the chat/completions surface), so this is carried through unchanged on `OAIRequest.TopK`. Whether the eventual provider honours it is a LiteLLM/provider concern; this translator must not be the one to silently drop it.
5. **Extended thinking** — no representation at all: no `thinking` request field, no `thinking`/`redacted_thinking` content blocks. A thinking-only message silently produced empty content (no case in `convertMessage`'s block switch matched, so both `parts` and `toolCalls` stayed empty).
   - `MessagesRequest.Thinking` (**disposition: carried through as-is** — identical field name and shape to Anthropic's own definition; LiteLLM's own Anthropic adapter interprets it on the way out, Hive has no local reasoning-budget concept to translate it into).
   - `ContentBlock.Thinking` / `.Signature` / `.Data` for `thinking` / `redacted_thinking` blocks in conversation history, surfaced on `OAIMessage.ThinkingBlocks` (**disposition: carried through via LiteLLM's own documented `thinking_blocks` convention** for round-tripping Anthropic extended-thinking content through an OpenAI-shaped assistant message — the signature must survive verbatim or Anthropic rejects the next turn, so this is a carry-through, not a reconstruction). Attached to every return path in `convertMessage`'s mixed-content section (alone, with tool_use, with text) so a thinking block can no longer produce a silently empty message in any combination.
6. **Pre-existing image drop in the tool-calls-plus-text branch** (preserved by #1152 to keep its regression guard exact) — **fixed here.** No existing test locked in the image-drop behaviour specifically (checked before touching it), so `partsNeedArrayForm` now also triggers on any non-text part, not just a cache breakpoint. An image mixed with a tool call no longer silently flattens away; the plain-text-only regression guard (#1152's actual concern) is untouched since it only ever exercised text parts.

## Structural fix

Per the brief: fixing six instances of a recurring bug isn't fixing the bug. `TestToOAIRequest_MaximalRequest_NoFieldLost` (new file `translate_request_maximal_test.go`) builds one Anthropic request with every documented `MessagesRequest` field populated — including nested `ContentBlock`, `Tool`, and `ToolChoice` sub-fields — translates it once, and asserts each field survives with an explicit per-field check. `TestToOAIRequest_MaximalRequest_ToolChoiceNone_NotInverted` covers the one combination (`tool_choice: "none"`) that can't share the same request object as `disable_parallel_tool_use`. The next field added to `MessagesRequest` has a structural guard now, not just a hope someone remembers.

## Tests

Table-driven, one test per fix plus the maximal round-trip:
- `TestToOAIRequest_ToolChoice_None`
- `TestToOAIRequest_Metadata_UserID`, `TestToOAIRequest_Metadata_Nil_UserOmitted`
- `TestToOAIRequest_ToolChoice_DisableParallelToolUse`, `TestToOAIRequest_ToolChoice_ParallelToolUseAllowed_OmitsField`
- `TestToOAIRequest_TopK`
- `TestToOAIRequest_ThinkingConfig_Passthrough`, `TestToOAIRequest_ThinkingOnlyMessage_NotEmptyContent`, `TestToOAIRequest_RedactedThinkingBlock`, `TestToOAIRequest_ThinkingBlock_MixedWithToolUse`
- `TestToOAIRequest_ImageBlock_MixedWithToolUse_NotDropped`
- `TestToOAIRequest_MaximalRequest_NoFieldLost`, `TestToOAIRequest_MaximalRequest_ToolChoiceNone_NotInverted`

**Mutation check (per the task brief):** temporarily reverted the `tool_choice: "none"` case in `convertToolChoice`, reran the suite — `TestToOAIRequest_ToolChoice_None` and `TestToOAIRequest_MaximalRequest_ToolChoiceNone_NotInverted` both failed with `want "none" got "auto"`, exactly the pre-fix inversion — then restored the fix and confirmed both pass again.

```
=== RUN   TestToOAIRequest_MaximalRequest_ToolChoiceNone_NotInverted
    translate_request_maximal_test.go:241: tool_choice: want "none" got "auto"
--- FAIL: TestToOAIRequest_MaximalRequest_ToolChoiceNone_NotInverted (0.00s)
=== RUN   TestToOAIRequest_ToolChoice_None
    translate_request_test.go:407: tool_choice json: want "none" got "auto" (an inversion back to "auto" means tool use is silently re-enabled)
--- FAIL: TestToOAIRequest_ToolChoice_None (0.00s)
FAIL	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/anthropic	0.011s
```
Restored, reran:
```
ok  	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/anthropic	0.179s
```

Full `apps/edge-api` suite (all packages) green after the fix:
```
ok  	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/anthropic	0.187s
ok  	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/inference	3.054s
... (all other apps/edge-api packages ok)
```
`go vet ./apps/edge-api/...` clean. `gofmt -l` clean on every touched file.

## Scope

Strictly inside `apps/edge-api/internal/anthropic/`. Did not touch `.github/workflows/ci.yml`, any `go.mod`, `apps/control-plane/internal/payments/stripe/rail_test.go`, or `apps/edge-api/internal/inference/` (owned elsewhere).

## Buglog entry

```json
{"error_message":"tool_choice:{\"type\":\"none\"} silently inverted to auto on /v1/messages, plus five other Anthropic request fields (metadata.user_id, disable_parallel_tool_use, top_k, thinking config/blocks, image-drop in tool-calls branch) silently lost by the same field-by-field rebuild","root_cause":"translate_request.go's convertToolChoice switch had no case for Anthropic's documented \"none\" tool_choice type and fell through to a default that returns \"auto\"; the same field-by-field rebuild pattern that caused the cache_control loss fixed in #1152 dropped five more fields with no representation on MessagesRequest/OAIRequest at all, and convertMessage's block switch had no case for thinking/redacted_thinking blocks so a thinking-only message produced an OAIMessage with empty content and no error","fix":"added an explicit \"none\" case to convertToolChoice mapping to OpenAI's own \"none\" sentinel; added MessagesRequest.TopK/Thinking, OAIRequest.TopK/Thinking/User/ParallelToolCalls, ToolChoice.DisableParallelToolUse, ContentBlock.Thinking/Signature/Data, and OAIMessage.ThinkingBlocks; wired Metadata.UserID to OAIRequest.User and disable_parallel_tool_use to the inverse parallel_tool_calls=false; widened partsNeedArrayForm to also trigger on any non-text part so an image mixed with a tool call keeps the block-array form instead of flattening away; added a maximal round-trip test asserting every documented MessagesRequest field survives translation as a structural guard against the next field going missing the same way","tags":["anthropic","translate_request","tool_choice","thinking","top_k","metadata","parallel_tool_calls","translator-field-loss"]}
```

Fixes #1153

## Summary by CodeRabbit

* **New Features**
  * Added support for extended thinking configuration and preservation of thinking content, including redacted blocks.
  * Added support for `top_k`, user metadata, and parallel tool-use controls.
  * Preserved images and other content when combined with tool calls.
* **Bug Fixes**
  * Correctly handles `tool_choice: none`.
  * Rejects unsupported tool-choice values instead of silently changing behavior.
  * Prevents thinking-only messages and mixed content from being dropped.

### DIFF
--- a/apps/edge-api/internal/anthropic/translate_request_maximal_test.go
+++ b/apps/edge-api/internal/anthropic/translate_request_maximal_test.go
@@ -1,0 +1,243 @@
+package anthropic_test
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/sakibsadmanshajib/hive/apps/edge-api/internal/anthropic"
+)
+
+// TestToOAIRequest_MaximalRequest_NoFieldLost is the structural guard issue
+// #1153 asks for: a single Anthropic request that populates every documented
+// MessagesRequest field (plus every documented sub-field on the types that
+// hang off it -- ContentBlock, Tool, ToolChoice), translated once, with an
+// explicit assertion per field. This is what stops the next field silently
+// dropped or inverted by translate_request.go from going unnoticed the way
+// cache_control did (PR #1152) and the way the six fields fixed by this
+// change did (issue #1153): a translator that rebuilds its output field by
+// field has no other structural defense against "nobody remembered to carry
+// this one across."
+//
+// When a new field is added to MessagesRequest, the correct response is to
+// add it here too, not just in a field-specific test elsewhere in this
+// package.
+func TestToOAIRequest_MaximalRequest_NoFieldLost(t *testing.T) {
+	raw := `{
+		"model": "claude-3-5-sonnet",
+		"max_tokens": 4096,
+		"system": "You are a helpful assistant.",
+		"messages": [
+			{"role": "user", "content": "What's the weather in Dhaka, and analyze this chart?"},
+			{"role": "assistant", "content": [
+				{"type": "thinking", "thinking": "I should call the weather tool.", "signature": "sig_001"},
+				{"type": "text", "text": "Let me check.", "cache_control": {"type": "ephemeral"}},
+				{"type": "image", "source": {"type": "base64", "media_type": "image/png", "data": "abc123"}},
+				{"type": "tool_use", "id": "tu_01", "name": "get_weather", "input": {"city": "Dhaka"}, "cache_control": {"type": "ephemeral"}}
+			]},
+			{"role": "user", "content": [
+				{"type": "tool_result", "tool_use_id": "tu_01", "content": "Sunny, 32C", "cache_control": {"type": "ephemeral"}}
+			]}
+		],
+		"tools": [
+			{"name": "get_weather", "description": "Get current weather", "input_schema": {"type": "object", "properties": {"city": {"type": "string"}}}, "cache_control": {"type": "ephemeral"}}
+		],
+		"tool_choice": {"type": "auto", "disable_parallel_tool_use": true},
+		"temperature": 0.65,
+		"top_p": 0.92,
+		"top_k": 40,
+		"stop_sequences": ["STOP", "\n\n"],
+		"stream": true,
+		"metadata": {"user_id": "end-user-789"},
+		"cache_control": {"type": "ephemeral", "ttl": "1h"},
+		"session_id": "sess-abc-123",
+		"thinking": {"type": "enabled", "budget_tokens": 8192}
+	}`
+
+	var req anthropic.MessagesRequest
+	if err := json.Unmarshal([]byte(raw), &req); err != nil {
+		t.Fatalf("unmarshal: %v", err)
+	}
+
+	got, err := anthropic.ToOAIRequest(req)
+	if err != nil {
+		t.Fatalf("ToOAIRequest: %v", err)
+	}
+
+	// --- Request-root fields ---
+
+	if got.Model != "claude-3-5-sonnet" {
+		t.Errorf("model: got %q", got.Model)
+	}
+	if got.MaxTokens == nil || *got.MaxTokens != 4096 {
+		t.Errorf("max_tokens: got %v", got.MaxTokens)
+	}
+	if got.Temperature == nil || *got.Temperature != 0.65 {
+		t.Errorf("temperature: got %v", got.Temperature)
+	}
+	if got.TopP == nil || *got.TopP != 0.92 {
+		t.Errorf("top_p: got %v", got.TopP)
+	}
+	if got.TopK == nil || *got.TopK != 40 {
+		t.Errorf("top_k: got %v (issue #1153 item 4)", got.TopK)
+	}
+	if len(got.Stop) != 2 || got.Stop[0] != "STOP" || got.Stop[1] != "\n\n" {
+		t.Errorf("stop_sequences: got %v", got.Stop)
+	}
+	if got.Stream != true {
+		t.Errorf("stream: got %v", got.Stream)
+	}
+	if got.User != "end-user-789" {
+		t.Errorf("user (from metadata.user_id): got %q (issue #1153 item 2)", got.User)
+	}
+	if got.CacheControl == nil || got.CacheControl.Type != "ephemeral" || got.CacheControl.TTL != "1h" {
+		t.Errorf("cache_control (root): got %+v", got.CacheControl)
+	}
+	if got.SessionID != "sess-abc-123" {
+		t.Errorf("session_id: got %q", got.SessionID)
+	}
+	if got.Thinking == nil || got.Thinking.Type != "enabled" || got.Thinking.BudgetTokens != 8192 {
+		t.Errorf("thinking (request field): got %+v (issue #1153 item 5a)", got.Thinking)
+	}
+
+	// --- tool_choice + disable_parallel_tool_use ---
+
+	if got.ToolChoice == nil {
+		t.Fatal("tool_choice: nil")
+	}
+	tcJSON, _ := json.Marshal(got.ToolChoice)
+	if string(tcJSON) != `"auto"` {
+		t.Errorf("tool_choice: got %s", tcJSON)
+	}
+	if got.ParallelToolCalls == nil || *got.ParallelToolCalls != false {
+		t.Errorf("parallel_tool_calls (from disable_parallel_tool_use): got %v (issue #1153 item 3)", got.ParallelToolCalls)
+	}
+
+	// --- tools + tool cache_control ---
+
+	if len(got.Tools) != 1 {
+		t.Fatalf("tools len: got %d", len(got.Tools))
+	}
+	tool := got.Tools[0]
+	if tool.Function.Name != "get_weather" || tool.Function.Description != "Get current weather" {
+		t.Errorf("tool: got %+v", tool.Function)
+	}
+	if tool.CacheControl == nil || tool.CacheControl.Type != "ephemeral" {
+		t.Errorf("tool cache_control: got %+v", tool.CacheControl)
+	}
+
+	// --- messages ---
+
+	if len(got.Messages) != 4 {
+		t.Fatalf("messages len: want 4 got %d", len(got.Messages))
+	}
+
+	sys := got.Messages[0]
+	if sys.Role != "system" {
+		t.Errorf("messages[0].role: got %q", sys.Role)
+	}
+	sysJSON, _ := json.Marshal(sys.Content)
+	if string(sysJSON) != `"You are a helpful assistant."` {
+		t.Errorf("messages[0].content: got %s", sysJSON)
+	}
+
+	userText := got.Messages[1]
+	if userText.Role != "user" {
+		t.Errorf("messages[1].role: got %q", userText.Role)
+	}
+	userTextJSON, _ := json.Marshal(userText.Content)
+	if string(userTextJSON) != `"What's the weather in Dhaka, and analyze this chart?"` {
+		t.Errorf("messages[1].content: got %s", userTextJSON)
+	}
+
+	mixed := got.Messages[2]
+	if mixed.Role != "assistant" {
+		t.Errorf("messages[2].role: got %q", mixed.Role)
+	}
+	if len(mixed.ThinkingBlocks) != 1 {
+		t.Fatalf("messages[2].thinking_blocks len: want 1 got %d (issue #1153 item 5b)", len(mixed.ThinkingBlocks))
+	}
+	tb := mixed.ThinkingBlocks[0]
+	if tb.Type != "thinking" || tb.Thinking != "I should call the weather tool." || tb.Signature != "sig_001" {
+		t.Errorf("messages[2].thinking_blocks[0]: got %+v", tb)
+	}
+	if len(mixed.ToolCalls) != 1 {
+		t.Fatalf("messages[2].tool_calls len: want 1 got %d", len(mixed.ToolCalls))
+	}
+	tc := mixed.ToolCalls[0]
+	if tc.ID != "tu_01" || tc.Function.Name != "get_weather" {
+		t.Errorf("messages[2].tool_calls[0]: got %+v", tc)
+	}
+	if tc.CacheControl == nil || tc.CacheControl.Type != "ephemeral" {
+		t.Errorf("messages[2].tool_calls[0].cache_control: got %+v", tc.CacheControl)
+	}
+	mixedContentJSON, _ := json.Marshal(mixed.Content)
+	var mixedParts []map[string]interface{}
+	if err := json.Unmarshal(mixedContentJSON, &mixedParts); err != nil {
+		t.Fatalf("messages[2].content is not an array (image or cache_control text part lost): %s", mixedContentJSON)
+	}
+	var sawText, sawImage bool
+	for _, p := range mixedParts {
+		switch p["type"] {
+		case "text":
+			if p["text"] == "Let me check." {
+				sawText = true
+			}
+			cc, _ := p["cache_control"].(map[string]interface{})
+			if cc == nil || cc["type"] != "ephemeral" {
+				t.Errorf("messages[2].content text part cache_control: got %+v", p["cache_control"])
+			}
+		case "image_url":
+			iu, _ := p["image_url"].(map[string]interface{})
+			if iu != nil && iu["url"] == "data:image/png;base64,abc123" {
+				sawImage = true
+			}
+		}
+	}
+	if !sawText {
+		t.Errorf("messages[2].content: text part not found: %s", mixedContentJSON)
+	}
+	if !sawImage {
+		t.Errorf("messages[2].content: image_url part not found, dropped by the tool-calls-plus-parts flatten branch (issue #1153 item 6): %s", mixedContentJSON)
+	}
+
+	toolResult := got.Messages[3]
+	if toolResult.Role != "tool" {
+		t.Errorf("messages[3].role: got %q", toolResult.Role)
+	}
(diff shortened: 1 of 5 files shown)


sakibsadmanshajib (commented):
test1

apps/edge-api/internal/anthropic/types.go
```diff
@@ -45,6 +58,12 @@ type RequestMetadata struct {
 	UserID string `json:"user_id,omitempty"`
 }
 
+// ThinkingConfig requests Anthropic extended thinking on the model turn.
+type ThinkingConfig struct {
+	Type         string `json:"type"` // "enabled" | "disabled"
+	BudgetTokens int    `json:"budget_tokens,omitempty"`
```
sakibsadmanshajib: MEDIUM: ThinkingConfig is missing Anthropic's documented `display` sub-field on the enabled variant (`ThinkingConfigEnabledParam.display`, values `"summarized"` or `"omitted"`, controls whether thinking content is redacted in the response while still returning a signature for multi-turn continuity; verified against anthropic-sdk-python's current type definitions, not from memory).

This is the exact bug class this PR exists to eliminate: a documented Anthropic request field silently dropped by a field-by-field struct rebuild, now reproduced in the very struct this PR introduces to fix that class of bug. The new maximal round-trip test does not include or assert `display` either, so this specific gap would not be caught by the new guard.

Suggest either carrying it through as `Display *string \`json:"display,omitempty"\`` (same disposition as the rest of ThinkingConfig, an unchanged carry-through) or documenting explicitly why it is out of scope, the way `top_k`'s disposition is documented above it. Not blocking on its own, but it should be a deliberate call rather than a silent omission, given the standard this PR itself sets.

apps/edge-api/internal/anthropic/translate_request_maximal_test.go
```diff
@@ -0,0 +1,243 @@
...
+// dropped or inverted by translate_request.go from going unnoticed the way
+// cache_control did (PR #1152) and the way the six fields fixed by this
+// change did (issue #1153): a translator that rebuilds its output field by
+// field has no other structural defense against "nobody remembered to carry
+// this one across."
+//
+// When a new field is added to MessagesRequest, the correct response is to
+// add it here too, not just in a field-specific test elsewhere in this
+// package.
+func TestToOAIRequest_MaximalRequest_NoFieldLost(t *testing.T) {
```
sakibsadmanshajib: MEDIUM, this is the important question the review brief asked to settle: does this test actually catch a newly added field being dropped? Verified: no.

This is a hand-written JSON literal plus hand-written per-field assertions, with no reflection and no mechanism that enumerates MessagesRequest's actual field set against what got asserted. If a developer adds a new field to MessagesRequest tomorrow and forgets to add it to the raw JSON literal here and add an assertion for it, this test stays green.

It is a genuinely thorough regression test for the six fields fixed today, and a good template for the next author (the PR body itself says the right move is to add the new field here too), but that is still manual discipline, the same failure mode this PR exists to eliminate, just moved one file over rather than removed. The PR description's framing ("a structural guard now, not just a hope someone remembers") overstates what this test mechanically does; recommend correcting that claim, and optionally filing a follow-up for a real reflection-based guard (walk MessagesRequest's fields via reflect, assert every field set to a non-zero value in the input surfaces as non-zero somewhere in the translated output) so a missed field actually fails the build. Not blocking merge on its own, since the six real bugs are genuinely fixed and well covered by the field-specific tests, but the team should not carry the stronger claim forward as fact.

sakibsadmanshajib (commented):
Independent adversarial review, read from the pushed diff directly (gh pr diff 1163), not from the PR description. Full verdict and the two verified-sound items are in the follow-up summary comment.

sakibsadmanshajib:
## Independent adversarial review (PR #1163, issue #1153)

Read from the pushed diff directly (`gh pr diff 1163`) and the checked-out branch, not from the PR description. Two inline comments posted at the exact lines; everything else, including one item that fell outside any diff hunk and could not carry an inline comment, is here.

### Findings

**MEDIUM: `ThinkingConfig` drops Anthropic's documented `display` sub-field.** See inline comment on `types.go`. `ThinkingConfigEnabledParam.display` (`"summarized"` | `"omitted"`) is a real, current field per anthropic-sdk-python's type definitions, and it is not carried through, not documented as an intentional non-goal, and not covered by the new maximal test. Same bug class this PR fixes, reproduced in the struct it introduces. Not blocking.

**MEDIUM: the round-trip test is not a structural guard against future field loss.** See inline comment on `translate_request_maximal_test.go`. Verified directly: no reflection, no field-set comparison, purely hand-written JSON plus hand-written assertions. A field added to `MessagesRequest` tomorrow and not also added here stays green. Good regression coverage for the six fields fixed today; the PR description's "structural guard, not just a hope someone remembers" claim overstates what it mechanically does and should be corrected before anyone relies on it as an automatic backstop. Not blocking.

**LOW / informational, no inline anchor (line is unchanged, outside any diff hunk):** `convertToolChoice`'s `default` branch (`translate_request.go`, line 353 on the PR head) still silently maps any unrecognized `tool_choice.type` to `"auto"`. All four currently documented Anthropic values (`auto`, `any`, `tool`, `none`) are now explicit, so nothing live hits this today, but a future fifth value (or a malformed client value) would reproduce the exact inversion this PR just fixed. Worth a follow-up to fail loud instead of silently defaulting to "auto" on the unrecognized case. Not blocking.

### Verified sound, not manufactured findings

- **tool_choice enum completeness.** Cross-checked against Anthropic's real four documented values. Fixed correctly, no other missing case for any value the live API sends.
- **`disable_parallel_tool_use` polarity and three-state mapping.** Anthropic absent and Anthropic `false` both correctly collapse to OpenAI omitted (both mean "parallel allowed," matching OpenAI's own default); only Anthropic `true` maps to explicit `false`. No inversion.
- **`thinking_blocks` field shape.** Checked byte-for-byte against LiteLLM's own `ChatCompletionThinkingBlock` / `ChatCompletionRedactedThinkingBlock` TypedDicts in `litellm/types/llms/openai.py`. Field names and shape (`type`, `thinking`, `signature` / `type`, `data`) match exactly. Not a near-miss.
- **`top_k` and `thinking` actually reach LiteLLM.** Traced the full pipeline: `/v1/messages` handler marshals `OAIRequest` to JSON and hands it to the in-process `/v1/chat/completions` handler as raw bytes; that handler unmarshals into a narrower `ChatCompletionRequest` struct only for validation and routing, but dispatches the original raw `body` bytes, not a re-marshal of that narrower struct. The two byte-level rewrites in the path (`clampCompletionLimit`, `rewriteModel`) both decode to `map[string]json.RawMessage` and preserve every other key byte for byte. So `top_k` and the request-level `thinking` config genuinely survive to LiteLLM, not just to the next hop. `drop_params: true` is set in `deploy/litellm/config.yaml`, and Groq's and OpenRouter's LiteLLM chat transformations both inherit the generic drop-params-aware param mapping with no `top_k` override, so an unsupported `top_k` gets dropped by LiteLLM rather than 400ing the client.
- **Image-drop fix scope.** `partsNeedArrayForm`'s widening is only reachable from the tool-calls-plus-parts branch; the pure-image and pure-text paths are untouched. The #1152 regression guard it claims not to disturb (`TestToOAIRequest_NoCacheControl_Unchanged`) uses a plain-string message with no content blocks at all, so it never exercises this branch and cannot regress.
- **No nil-pointer or crash surface** on any new pointer field. `DisableParallelToolUse` and `Metadata.UserID` are nil-checked before dereference; `TopK` and `Thinking` are passed through as pointers with no local dereference.

### Verdict

No CRITICAL or HIGH issues. Two MEDIUM findings (both about honesty of claims/coverage, not correctness of the six shipped fixes) and one LOW follow-up suggestion. The six fixed bugs are real, correctly fixed, and well tested. Do not block merge on this review; address the two MEDIUM items (or explicitly accept them) before or after merge at the team's discretion.

sakibsadmanshajib:
Both threads addressed in 18feecce (pushed).

**ThinkingConfig.display**: added as `Display *bool \`json:"display,omitempty"\`` (carried-through, same disposition as Type/BudgetTokens). I could not independently re-verify the exact value domain (string enum "summarized"/"omitted" per your comment vs a plain boolean toggle) without live access to anthropic-sdk-python from this sandbox, so I want to flag that explicitly rather than assert false confidence: if it is in fact a string enum, the field type here needs to change from `*bool` to `*string`. Since this struct is a pure unchanged carry-through end to end (unmarshal MessagesRequest -> marshal OAIRequest, same JSON key, same shape), a wrong Go type would surface immediately and loudly as a JSON unmarshal error on any real request that sets it, not a silent drop -- which is the failure mode this PR is actually trying to eliminate. Flagging for a follow-up type check against the live SDK rather than blocking on it here.

**Structural test claim**: agreed in full, this was a fair catch and the PR body's "structural guard" language was overclaiming. Added `assertEveryFieldDispositioned` (translate_request_structural_test.go): reflects over a struct's field list and fails, naming the field, for anything missing an explicit disposition entry (carried/translated/forwarded/rejected). Applied to MessagesRequest, ContentBlock, Tool, ToolChoice. `TestStructuralGuard_CatchesAnUndispositionedField` proves the guard mechanically works via `reflect.StructOf` injecting an extra field at runtime. I also did the literal manual demonstration you'd want to see before trusting this: added a real dummy field to `MessagesRequest` in types.go, ran `TestMessagesRequest_EveryFieldHasADisposition`, confirmed it failed naming the field, then reverted:

```
=== RUN   TestMessagesRequest_EveryFieldHasADisposition
    translate_request_structural_test.go:84: MessagesRequest.DummyUnhandledField (json "dummy_unhandled_field") has no disposition entry in this test -- a field was added to the struct without deciding what ToOAIRequest does with it. Add an entry to the dispositions map for this type stating whether it is carried through unchanged, translated to an equivalent, forwarded as an opaque passthrough, or deliberately rejected, and wire it up in translate_request.go to match.
--- FAIL: TestMessagesRequest_EveryFieldHasADisposition (0.00s)
FAIL	github.com/sakibsadmanshajib/hive/apps/edge-api/internal/anthropic	0.010s
```

Explicit limitation, called out in code comments too: the reflection pass covers each type's own top-level fields, not sub-fields of a nested struct (ThinkingConfig, ImageSource, RequestMetadata, CacheControl). That's exactly how the `display` gap above got past the original round-trip test -- the guard would need to be recursive to catch that automatically, which I did not build (kept to a flat per-type pass over MessagesRequest + the three shallow nested types you asked about: ContentBlock, Tool, ToolChoice). Noting this rather than overclaiming a second time.

Also fixed the LOW item from the coordinator's pass (not a review thread here, but addressed in the same commit): `convertToolChoice`'s default branch used to silently map any unrecognized `tool_choice.type` to `"auto"`, the identical inversion shape as the `"none"` bug this PR fixes. It now returns an error, surfaced as the existing plain 400 in handler.go's translation-failure path, rather than silently re-enabling tool use. Test: `TestToOAIRequest_ToolChoice_UnknownType_Rejected`.

Full apps/edge-api suite green, go vet clean, gofmt clean on every touched file.

sakibsadmanshajib:
CodeRabbit's finding is correct and fixed in af92b349: \`ThinkingConfig.Display\` changed from \`*bool\` to \`*string\` (Anthropic documents \`"summarized"\`/\`"omitted"\`, not a boolean). Also did the second thing asked: added a deeper reflection pass, \`TestThinkingConfig_EveryFieldHasADisposition\`, covering ThinkingConfig's own three sub-fields (\`type\`, \`budget_tokens\`, \`display\`), since the MessagesRequest-level guard treats \`thinking\` as one opaque field and cannot see inside it -- which is exactly how the wrong type got past every other test in the first review round. Extended the maximal round-trip test to set \`display:"summarized"\` and assert it survives.

Every other nested struct (ImageSource, RequestMetadata, CacheControl) is still uncovered by reflection, noted explicitly in the new test's comment rather than implied otherwise.

Full apps/edge-api suite green, go vet clean, gofmt clean.

PR #1175, opened 2026-08-25
feat: add a model detail page at /console/catalog/[id]

sakibsadmanshajib:
## What and why

The developer console had no page for a single model. `/console/catalog` listed rows, the rows were not links, no `[id]` route existed, and the cache read and cache write prices that the control plane has carried and metered end to end were rendered nowhere. The OpenRouter similarity scorecard dated 2026-08-25 scores this row `missing` with a credit of 0.00 out of a weight of 6, and calls it one of the two heaviest total absences an evaluator hits first.

This adds `/console/catalog/[id]` and makes each catalog row navigate to it.

## What the page renders

Everything on it comes from what the control plane can answer today. Nothing is computed from a guess.

- Identity: alias id, lifecycle badge, capability badges, summary.
- Header tiles: in and out price, cache read and cache write price, context, pricing mode.
- A pricing table covering all four rate dimensions in credits per one million metered tokens, the same unit the catalog list already prints.
- The account's own usage of that alias over the last thirty days, from `/analytics/usage?group_by=model`, with a link into `/console/logs?model=<alias>`.

## What it refuses to invent

Per provider price, latency, throughput, uptime, benchmarks and the context window are held in the routing snapshot or not held at all, and no customer facing endpoint returns any of them. Each is rendered as a named absence with the reason rather than omitted or filled with a zero. A zero in those slots would read as a measurement.

## The three price outcomes

On a billing surface these mean three different things and the page keeps them apart:

- A published rate, zero included, prints as a number. `hive-default` publishes a cache write rate of exactly 0, and that zero is a decision that the dimension is not charged.
- A missing price on an `upstream_actual` alias prints as `Variable`, which is the design.
- A missing price on a `fixed` alias prints as `Unknown`, because that is a lookup that came back empty, and calling it `Variable` would present a broken decode as a pricing model.

`apps/web-console/lib/format/model-pricing.test.ts` pins each of those three against the other two, so collapsing any pair fails the suite.

No USD figure appears anywhere on the page. Credits are the console's unit on every other surface, and relaxing the existing no-USD presentation is a separate product decision, not something to slip in here.

## Scope discipline

Another branch is editing `components/catalog/model-catalog-table.tsx` to add cache price columns to the list. This PR touches that file in exactly one place: the model name becomes a `Link`. The price formatter it needs was added as a new module (`lib/format/model-pricing.ts`) rather than extracted out of that component, deliberately, so the two diffs do not collide. That leaves a five line twin of `formatPrice` in the table, marked with a `ponytail:` comment; folding the two together is a follow up once both have landed.

## Checks

```
docker compose run --rm --build web-console npm run build      -> compiled, route ƒ /console/catalog/[id] present
docker compose run --rm --build web-console npm run test:unit  -> 679 passed
docker compose run --rm web-console npx vitest run lib/format/model-pricing.test.ts -> 6 passed
```

`tests/unit/ci-web-e2e-secret-free.test.ts` fails inside that image on this branch and on `main` alike: it reads `.github/workflows/ci.yml`, which `Dockerfile.web-console` does not copy into the image. Pre-existing, unrelated.

## Visual proof

Captured against a console built from this branch and attached to a running self-hosted stack with a real control plane over a real Postgres. Full setup, the data caveat, the click assertion that a screenshot cannot make, and the browser console transcript are committed in `docs/proof/console-model-detail-2026-08-25/README.md`. Images are posted as a comment on this PR.

## Not in this PR: the docs surface (scorecard row 7)

Row 7 asked for a hosted docs and API reference generated from an OpenAPI document. That document does exist: `packages/openai-contract/generated/hive-openapi.yaml`, plus a machine readable `packages/openai-contract/matrix/support-matrix.json` with 164 endpoints across four support statuses. It is good material and a docs page over it would be honest and cheap.

It is blocked on one thing outside this PR's scope: `deploy/docker/Dockerfile.web-console` copies `apps/web-console`, `deploy/`, `.env.example` and `supabase/migrations` and nothing else, so neither the spec nor the matrix exists inside the console image. Importing either from `packages/` compiles locally and breaks the Docker build. The fix is one `COPY packages/openai-contract/ /app/packages/openai-contract/` line in that Dockerfile, and then a docs route can import the matrix and serve the YAML. Vendoring a second copy of the matrix into `apps/web-console` was rejected: a duplicated spec is exactly the rot that generating from a spec is supposed to avoid.

## Buglog entry

Nothing to log. This is a missing feature, not a fixed defect.

### DIFF
--- a/apps/web-console/app/console/catalog/[id]/page.tsx
+++ b/apps/web-console/app/console/catalog/[id]/page.tsx
@@ -1,0 +1,107 @@
+import type { ReactElement } from "react";
+import Link from "next/link";
+import { notFound, redirect } from "next/navigation";
+
+import {
+  getAccountProfile,
+  getAnalyticsUsage,
+  getCatalogModels,
+  getViewer,
+  type UsageSummaryRow,
+} from "@/lib/control-plane/client";
+import { ConsoleShell } from "@/components/app-shell/console-shell";
+import { ModelDetail } from "@/components/catalog/model-detail";
+import { PageHeader } from "@/components/ui/page-header";
+import { buttonVariants } from "@/components/ui/button";
+import { cn } from "@/lib/cn";
+
+interface ModelDetailPageProps {
+  params: Promise<{ id: string }>;
+}
+
+// The analytics window this page reports. 30d is the longest preset the
+// control plane accepts without an explicit from/to pair, and a model page is
+// a "what has this cost me" surface rather than a live one.
+const USAGE_WINDOW = "30d";
+const USAGE_WINDOW_LABEL = "last 30 days";
+
+export default async function ModelDetailPage(
+  props: ModelDetailPageProps,
+): Promise<ReactElement> {
+  const { id } = await props.params;
+  const viewer = await getViewer();
+  if (viewer.user.email_verified === false) {
+    redirect("/console/settings/profile");
+  }
+
+  // There is no single-model endpoint: /api/v1/catalog/models returns the whole
+  // tenant-filtered list and nothing narrower exists. Reading the list and
+  // selecting from it also means an alias this tenant may not see is a 404
+  // here for free, rather than a page that renders a model they cannot call.
+  const [models, profile] = await Promise.all([
+    getCatalogModels(),
+    getAccountProfile().catch(
+      (): { owner_name: string } => ({ owner_name: "" }),
+    ),
+  ]);
+
+  const model = models.find((row) => row.id === id);
+  if (!model) {
+    notFound();
+  }
+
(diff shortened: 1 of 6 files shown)


sakibsadmanshajib:
## Visual proof

Model detail page at /console/catalog/[id], captured against a console built from this branch on a running self-hosted stack. 01 catalog rows now link. 02 deepseek-v4-flash: cache read 1,790 and cache write 0 rendered. 03 hive-embedding-default: NULL cache prices render as Unknown, visibly distinct from a published 0. 04 hive-default: both cache rates a published 0. Account address masked in pixels.

PR #1187, opened 2026-08-25
feat: generate the console's API reference from the OpenAPI contract

### LINKED ISSUE #1179: Console Documentation link points off-product to hivegpt.io
Found while building the model detail page (#1175).

`apps/web-console/components/app-shell/console-shell.tsx` around line 206 renders a Documentation link pointing at `https://hivegpt.io`. That is off-product, it appears in the shell on EVERY console page, and a developer evaluating the gateway will click it. On a demo it is the kind of detail that undercuts everything else on the screen.

**Context that makes this cheap to fix properly.** The repo already generates an OpenAPI contract: `packages/openai-contract/generated/hive-openapi.yaml` (OpenAPI 3.0, server `/v1`) and `packages/openai-contract/matrix/support-matrix.json` (164 endpoints, 39 supported now, 72 explicitly unsupported at launch, 51 out of scope, 2 planned). So a real docs surface can be generated from the spec rather than hand written, which is what stops it rotting.

This is also scorecard row 7 in `session-2026-08-25-openrouter-console-similarity-scorecard.md`, currently verdict MISSING, and one of the two surfaces an evaluator opens first when judging a gateway.

**Known blocker, and the fix for it.** `deploy/docker/Dockerfile.web-console` copies only `apps/web-console/`, `deploy/`, `.env.example` and `supabase/migrations/`. A docs route importing from `packages/` compiles locally and then breaks the Docker build. Vendoring a second copy of the matrix into the app is the wrong answer, since duplicating it reintroduces exactly the drift that generating from a spec avoids.

Suggested shape:
1. Add `COPY packages/openai-contract/ /app/packages/openai-contract/` to `Dockerfile.web-console`.
2. `app/console/docs/page.tsx` with a quickstart plus an endpoint table generated from the support matrix, printing the matrix's own `generated` date so staleness is visible rather than hidden.
3. A route handler serving the YAML.
4. Repoint the shell's Documentation link at it.

Related: #1175 model detail page, and the console similarity scorecard in the vault.

sakibsadmanshajib:
Closes #1179.

## What this changes

The console shell rendered a Documentation link pointing at `https://hivegpt.io` on **every** console page, and there was no in-product API reference behind it. Hosted docs is row 7 of the console similarity scorecard, verdict MISSING, and one of the two surfaces a developer opens first when judging a gateway.

This adds `/console/docs`, generated at request time from the two artefacts the repo already produces, and repoints both in-app documentation links at it.

| File | What it does |
| --- | --- |
| `apps/web-console/lib/api-contract.ts` | Reads `packages/openai-contract/`, parses the support matrix, extracts the spec's operations and their `x-hive-status`, diffs the two, groups the result by status and resource family. |
| `apps/web-console/app/console/docs/page.tsx` | The page: quickstart, machine-readable spec pointer, disagreement report, full endpoint table. |
| `apps/web-console/app/api/openapi.yaml/route.ts` | Serves the raw 1.1 MB YAML, unauthenticated. |
| `apps/web-console/components/app-shell/console-shell.tsx` | Shell Documentation link now points at `/console/docs`. This is the single highest-visibility line in the PR. |
| `apps/web-console/app/console/page.tsx` | Second `hivegpt.io` link, on the overview page's "Need the API reference?" footer, repointed the same way. |
| `deploy/docker/Dockerfile.web-console`, `.prod` | `COPY packages/openai-contract/` so the route can read it inside the image. |
| `apps/web-console/tests/unit/console-docs-contract.test.ts` | 12 tests. |

Nothing on the page is typed in by hand. Every endpoint, status, count and date comes from the spec or the matrix.

## Honesty properties

- **The unsupported endpoints are on the page.** 72 explicitly unsupported at launch and 51 out of scope, listed with their notes behind a native disclosure, with counts in the section header. Docs that list only the happy path let an integrator build against something that will never answer.
- **The matrix's own `generated` date is printed** next to its version, with a line saying plainly that if the date is old, the classifications are that old too.
- **Disagreements are reported, not resolved.** See below.
- **The quickstart names a model alias read from the live catalog** (`getCatalogModels()`), falling back to `hive-default`, the alias seeded in `supabase/migrations/`, if the catalog cannot be read. No invented model name.

## Spec versus matrix: what the diff actually found

97 operations in the generated spec, 164 in the support matrix, **89 disagreements**, all rendered on the page:

| Kind | Count | Reading |
| --- | --- | --- |
| Classified in the matrix, absent from the spec | 67 | 51 are `out_of_scope`, which `scripts/sync_hive_contract.py` drops from the generated spec deliberately. The other **16 are `supported_now` Hive-native endpoints** the spec has never described: `/v1/rag/*` (6), `/v1/agent/tasks*` (4), `/v1/artifacts*` (3), `/v1/featuregate`, and the Anthropic-compatible `/v1/messages` and `/v1/messages/count_tokens`. The generated spec is built from OpenAI's upstream document, so nothing Hive added itself is in it. A developer generating a client from the YAML gets no schema at all for 16 live endpoints. |
| Different status in each file | 22 | Every one is the same direction: the spec annotates it `planned_for_launch`, the matrix classifies it `supported_now`. Including `POST /v1/chat/completions`. |
| Declared in the spec, unclassified in the matrix | 0 | |

**The finding worth carrying out of this PR: the stale artefact is the spec, not the matrix.** The matrix says `generated: 2026-03-28` and is the more current of the two; `generated/hive-openapi.yaml` was produced from an older matrix and still tells any tool that reads it that chat completions, embeddings, files, batches, audio, images, responses and uploads are merely planned. Regenerating it is a `packages/openai-contract/` change and out of scope here, so it is reported rather than papered over. Worth its own issue.

## The Docker blocker, and proof the fix works

A route importing from `packages/` compiles locally and then breaks the image, because neither web-console Dockerfile copied that tree. Both now do. `Dockerfile.web-console.prod` is included deliberately: it is what actually builds `console-hive.scubed.co`, so fixing only the dev image would have shipped a docs page that 500s in production while looking green everywhere else.

The package is read from disk rather than vendored into `apps/web-console/`, since a second copy of a generated artefact reintroduces exactly the drift that generating from a spec avoids.

Verified in the image, with `--build`, which is the point:

- `docker compose run --rm --build web-console npm run build` — exit 0. `/console/docs` and `/api/openapi.yaml` both appear in the route manifest, so the route resolves and typechecks inside the image.
- `docker compose run --rm --build web-console npm run test:unit` — **691 passed**, 12 of them new. The new tests call `loadSupportMatrix()` and `loadSpecOperations()` against the real files, so they would ENOENT inside the image without the COPY.
- The one failing file is `tests/unit/ci-web-e2e-secret-free.test.ts`, which reads `.github/workflows/ci.yml` that the Dockerfile never copies in. Known, pre-existing, container-only, passes in CI.

## What the tests guard

The three ways a generated docs page goes quietly wrong:

1. **An empty table that still returns 200.** The spec is read with a line scan, not a YAML parse (1.1 MB, and one fact needed). If a regeneration changes the formatting, the scan silently matches nothing. The extracted operation count is pinned against the spec's own `x-hive-status:` annotation count, which is an independent count of the same set, so that goes red instead of rendering an empty page.
2. **A quickstart pointing at nothing.** `API_BASE_URL`'s host is asserted against `deploy/cloudflare/tunnel-ingress.json`, the file that calls itself the single source of truth for which hive hostnames are public, and its path against the spec's own `servers[0].url`. Both have moved before.
3. **The link going off-product again.** Both files are asserted to contain no `href="https://hivegpt.io"` and to link `/console/docs`.

Plus a shape test per disagreement kind, an unknown-status test proving an unrecognised classification gets its own section rather than vanishing from the counts, and a parser test proving a malformed matrix throws rather than rendering an empty table.

The real-contract test pins the *direction* of the status mismatches rather than their count, so regenerating the spec makes it pass rather than cry wolf, but a mismatch in a new direction (say, the spec claiming supported for something the matrix calls unsupported) fails loudly.

## Buglog entry

```json
{"id":"console-docs-hivegpt-dead-link","date":"2026-08-25","title":"Console Documentation link pointed off-product at hivegpt.io on every page, and no in-product API reference existed","error_message":"Shell header link href=\"https://hivegpt.io\" rendered on every /console/* page; second copy on the overview page footer","root_cause":"No hosted docs surface was ever built. The repo generates an OpenAPI contract and a 164-endpoint support matrix under packages/openai-contract/, but deploy/docker/Dockerfile.web-console and .prod copy only apps/web-console/, deploy/, .env.example and supabase/migrations/, so any console route importing from packages/ compiles locally and then breaks the Docker build. That blocker is why the cheap fix was never taken.","fix":"Added /console/docs generated from the spec and matrix at request time, an unauthenticated /api/openapi.yaml route serving the raw spec, COPY packages/openai-contract/ in both web-console Dockerfiles, and repointed both hivegpt.io links. 12 unit tests guard the extraction count, the base URL against tunnel-ingress.json, and the absence of the off-product link.","tags":["web-console","docs","openapi","docker","dead-link","issue-1179"]}
```

## Visual proof

Posted as a comment on this PR, images on the permanent `visual-proof-assets` release. Captured against the image built from **this branch's `Dockerfile.web-console.prod`** (`next build` + `next start`, the same Dockerfile that builds `console-hive.scubed.co`), running against a live self-hosted Supabase and a live control-plane, signed in through `tests/e2e/support/live-auth.mjs`. No password was set or rotated.

Recorded in `docs/proof/console-hosted-docs-2026-08-25/capture.log`:

- `shell header Documentation link href = /console/docs`, and clicking it lands on `/console/docs`.
- `endpoint rows rendered = 164`, all four status sections present.
- `GET /api/openapi.yaml (no session) -> 200 application/yaml; charset=utf-8`, `spec bytes served = 1165460`. Served from the prod image, which is the COPY working end to end.
- `/console`, `/console/catalog`, `/console/docs` all load clean with no page errors.

The signed-in address is covered with Playwright's screenshot mask, which paints over the region without touching the DOM, so it reaches neither the pixels nor the log. `npm run lint:proof-tokens` passes over the committed log.

### DIFF
--- a/apps/web-console/app/api/openapi.yaml/route.ts
+++ b/apps/web-console/app/api/openapi.yaml/route.ts
@@ -1,0 +1,41 @@
+// Serves the generated OpenAPI contract so an integrator can point their own
+// tooling at it instead of reading a rendered table.
+//
+// Deliberately outside /console: middleware.ts redirects unauthenticated
+// requests under /console to sign-in, and a spec a code generator cannot fetch
+// without a session cookie is not a spec anyone will use.
+//
+// The file is read from packages/openai-contract/ rather than bundled: it is
+// 1.1 MB of YAML and no loader in this app turns that into a module. That
+// makes the read a real runtime dependency on the image containing the
+// package, which is why both deploy/docker/Dockerfile.web-console and
+// deploy/docker/Dockerfile.web-console.prod COPY it in. A missing file answers
+// 500 with a plain message rather than an empty 200, because an empty spec
+// silently generates an empty client.
+//
+// Node runtime only. `npm run build:cf` (the retired Workers path) has no
+// filesystem and this route would fail there.
+import { readFile } from "node:fs/promises";
+
+import { OPENAPI_SPEC_PATH } from "@/lib/api-contract";
+
+export async function GET(): Promise<Response> {
+  let spec: string;
+  try {
+    spec = await readFile(OPENAPI_SPEC_PATH, "utf8");
+  } catch (error) {
+    console.error("openapi spec unreadable", { path: OPENAPI_SPEC_PATH, error });
+    return new Response(
+      "The OpenAPI specification is not available in this deployment.\n",
+      { status: 500, headers: { "content-type": "text/plain; charset=utf-8" } },
+    );
+  }
+
+  return new Response(spec, {
+    headers: {
+      "content-type": "application/yaml; charset=utf-8",
+      "content-disposition": 'inline; filename="hive-openapi.yaml"',
+      "cache-control": "public, max-age=300",
+    },
+  });
+}
--- a/apps/web-console/app/console/docs/page.tsx
+++ b/apps/web-console/app/console/docs/page.tsx
@@ -1,0 +1,331 @@
+import { redirect } from "next/navigation";
+
+import {
+  getAccountProfile,
+  getCatalogModels,
+  getViewer,
+  type CatalogModel,
+} from "@/lib/control-plane/client";
+import {
+  API_BASE_URL,
+  OPENAPI_ROUTE,
+  type EndpointSection,
+  buildEndpointSections,
+  diffSpecAgainstMatrix,
+  loadSpecOperations,
+  loadSupportMatrix,
+} from "@/lib/api-contract";
+import { ConsoleShell } from "@/components/app-shell/console-shell";
+import { Badge } from "@/components/ui/badge";
+import {
+  Card,
+  CardContent,
+  CardDescription,
+  CardHeader,
+  CardTitle,
+} from "@/components/ui/card";
+import { PageHeader } from "@/components/ui/page-header";
+
+/**
+ * The console's own API reference.
+ *
+ * Every endpoint, status and count below is read from
+ * `packages/openai-contract/` at request time, never typed in here. That is
+ * the whole point: a hand written endpoint list is stale the first time a
+ * route changes, and a docs page that is quietly wrong is worse than no docs
+ * page. The unsupported and out-of-scope endpoints are listed too, because an
+ * integrator needs to know what will not work before building against it.
+ */
+
+const BADGE_TONE: Record<string, "success" | "accent" | "outline" | "neutral"> = {
+  supported_now: "success",
+  planned_for_launch: "accent",
+  explicitly_unsupported_at_launch: "outline",
+  out_of_scope: "neutral",
+};
+
+function CodeBlock({ children }: { children: string }) {
+  return (
+    <pre className="overflow-x-auto rounded-md border border-[var(--color-border)] bg-[var(--color-surface-inset)] px-4 py-3 text-xs leading-relaxed text-[var(--color-ink-2)]">
+      <code className="font-mono">{children}</code>
+    </pre>
+  );
+}
+
+export default async function DocsPage() {
+  const viewer = await getViewer();
+  if (viewer.user.email_verified === false) {
+    redirect("/console/settings/profile");
+  }
+
+  const [profile, models] = await Promise.all([
+    getAccountProfile().catch((): { owner_name: string } => ({ owner_name: "" })),
+    // The quickstart names a model that actually exists on this deployment.
+    // If the catalog cannot be read the snippets still have to be runnable, so
+    // they fall back to the alias seeded by supabase/migrations.
+    getCatalogModels().catch((): CatalogModel[] => []),
+  ]);
+
+  const alias = models[0]?.id ?? "hive-default";
+  const matrix = loadSupportMatrix();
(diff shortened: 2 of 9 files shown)


sakibsadmanshajib:
## Visual proof

Captured against the image built from this branch's Dockerfile.web-console.prod (next build + next start, the same Dockerfile that builds console-hive.scubed.co), signed in against a live self-hosted Supabase and a live control-plane. 1: the shell Documentation link now reads /console/docs, and clicking it lands there. 2: the generated quickstart, naming the hive-auto alias read from the live catalog. 3: all 89 spec-versus-matrix disagreements, reported rather than hidden. 4: the unsupported and out-of-scope endpoints, listed with counts. Log: docs/proof/console-hosted-docs-2026-08-25/capture.log

Issue #1148, opened 2026-08-25
Security-tier audit events can be lost after the mutation they describe has committed


sakibsadmanshajib:
Raised out of review on PR #1138 (tenant BYOK provider keys), but the gap is not specific to BYOK.

## The shape of it

`internal/audit`'s `Logger.Log` routes security-tier actions to the fail-closed `Sync` writer, which returns an error when Postgres is unreachable. Callers emit their audit event after the state change has already been committed, so at that point there is nothing useful to do with the error:

* Returning it to the caller reports a failed mutation that actually succeeded. For a create path the caller's natural response is a retry, which produces a second row. On PR #1138 that would mean a second registered provider credential.
* Discarding it loses a required security audit record silently.

Today most call sites discard it (`_ = ...Log(...)` in `internal/identity/finalize.go` and elsewhere). PR #1138 changed BYOK to log the failure at error level with identifiers only, which makes the loss visible on the process log but does not make the record durable.

## What would actually fix it

An outbox row written in the same transaction as the mutation, with a drainer that delivers it to the audit store and retries. The reason this was not done inside PR #1138 is that `audit.Logger` owns its own writers and accepts no caller-supplied transaction, so the change belongs to `internal/audit`'s contract and affects every caller. A module-local outbox in one package would create a second delivery path with different semantics from every other security event in the service, which is worse than the honest gap.

Note that the existing WAL path is not the answer as it stands: it is the LLM tier's fallback, and the security tier is deliberately fail-closed rather than WAL-backed.

## Scope

* Decide the seam: caller-supplied transaction on `Logger.Log`, or an outbox table plus drainer owned by `internal/audit`.
* Migrate security-tier call sites onto it.
* A test that proves an event survives the audit backend being down across the mutation.

sakibsadmanshajib:
Detail carried over from the independent security review of PR #1138, worth folding in before anyone scopes this.

The failure this issue describes is not only a database-outage scenario. `audit.SyncWriter.Write` serializes every security-tier write platform wide behind a single per-calendar-month `pg_advisory_xact_lock`, retried at most three times on `serialization_failure`. That is one global lock for all tenants and all security actions in a given month, so ordinary concurrent load can exhaust the retries on its own. Unrelated tenants registering credentials or issuing API keys at the same time is enough; no outage and no adversary required.

Two consequences for how this gets fixed.

[...]

PR #1188, opened 2026-08-25
fix: acquire audit writer advisory lock before SERIALIZABLE snapshot forms

### LINKED ISSUE #1182: audit.SyncWriter loses CRITICAL audit rows under concurrency; SERIALIZABLE snapshot precedes the advisory lock
Found while adding tenancy coverage (PR #1181). The shipped symptom in `tenants` is fixed there; this issue is the shared root cause, which lives in `apps/control-plane/internal/audit/sync.go` and affects every caller, not just tenants.

**Diagnosed live**, on a throwaway database: 10 fully simultaneous `audit.SyncWriter.Write` calls resulted in **6 of 10 failing with SQLSTATE 40001**, serialization failure, even after the writer's own three-attempt retry.

**Root cause.** The SERIALIZABLE snapshot is taken at `BeginTx`, which happens BEFORE the advisory lock is waited on. So Postgres's SSI still sees genuine concurrency regardless of the lock being held, and the lock does not serialize what the isolation level is checking. Three retries are not enough under ordinary concurrent load; this needs no outage and no adversary.

**Why it matters beyond a dropped row.** Audit rows here include CRITICAL severity events such as `CROSS_TENANT_ATTEMPT`. On a product sold on data sovereignty to regulated buyers, an audit trail that silently drops entries under load is worse than a slow one: the absence is indistinguishable from the event never happening. PR #1181 changed the `tenants` call sites from silent discard to logged failure, which converts silent loss into observable loss, but it does not stop the loss.

Related and compounding: #1148 records that the audit write fails open after the mutation has already committed, and notes the same global per-month advisory lock makes retry exhaustion plausible under ordinary load. This issue is the mechanism behind that observation.

**Options worth weighing, in rough order of preference.**
1. Take the advisory lock BEFORE `BeginTx`, so the snapshot is taken while already serialized. Likely the smallest correct fix.
2. Drop to `READ COMMITTED` for this writer if the integrity property it needs is provided by the lock rather than by SSI. Requires establishing what invariant the SERIALIZABLE level is actually buying here.
3. An outbox: write the audit row in the caller's own transaction, drain asynchronously. This is what #1148 already proposes, and it removes the global lock from the mutation's critical path entirely.
4. More retries with backoff. Weakest option, it lowers the failure rate without changing the shape, and it makes the remaining failures rarer and therefore harder to notice.

Whatever is chosen, the acceptance test already exists in spirit in PR #1181: N fully simultaneous writes must produce N rows, not "at least one". That PR deliberately asserts the honest current contract of at least one and at most N, because the writer does not today guarantee N. Closing this issue should let that assertion be tightened.

Related: #1148, #1181, and the `tenants` fix within it.

sakibsadmanshajib:
## Summary

Fixes #1182. Related: #1148, #1181.

`audit.SyncWriter.Write` took its `pg_advisory_xact_lock` as the first statement inside a SERIALIZABLE transaction. Postgres takes a SERIALIZABLE transaction's snapshot at the first statement inside it, not at BEGIN, so every writer blocked waiting on that lock had already formed its snapshot before it actually got its turn. Each one then committed against a `MAX(seq)` read that was already stale by the time it ran, and Postgres SSI aborted the COMMIT with `serialization_failure` (40001). Diagnosed live on a throwaway database: 10 fully-simultaneous `Write` calls produced 5-6 failures even after the writer's own 3-attempt retry. Audit rows lost this way include CRITICAL severity events such as `CROSS_TENANT_ATTEMPT`.

## Option chosen and why

Of the four options weighed in #1182, option 1: take the advisory lock before `BeginTx`. This is the smallest correct fix and it directly targets the diagnosed mechanism (snapshot timing vs lock timing), rather than working around the symptom.

- Option 2 (drop to READ COMMITTED) was rejected: establishing what invariant SERIALIZABLE actually buys here versus the lock is a separate investigation this issue didn't require, and downgrading isolation without that analysis is exactly the kind of thing the issue warned against.
- Option 3 (outbox) is the right end state per #1148 and removes the global lock from the mutation's critical path entirely, but it's a caller-visible contract change (`Logger.Log` would need a caller-supplied transaction, or a new drain path) touching every security-tier call site. Out of scope for this issue; #1148 stays open to track it.
- Option 4 (more retries) was explicitly rejected per the issue: it lowers the failure rate without changing the shape and makes the residual failures rarer and harder to notice.

## The fix

`writeOnce` now acquires a dedicated connection (`pool.Acquire`), takes a **session-scoped** `pg_advisory_lock` on that connection *before* `BeginTx`, then opens the SERIALIZABLE transaction, does the read/insert/commit, and releases the lock with an explicit `pg_advisory_unlock` in a deferred call (background context, so it still runs if the caller's context was cancelled). The transaction's own first statement (the `MAX(seq)` read) can then only execute once this writer already has exclusive possession of the month, so nothing downstream of `BEGIN` can observe a stale snapshot.

Session- not transaction-scoped deliberately: it has to survive from before `BeginTx` through `Commit`. If the connection drops mid-hold, Postgres releases the session lock when the backend terminates, so the failure mode is one orphaned physical connection (which the pool detects and evicts), never a permanently stuck lock.

Same acquire/exec/defer-release shape as the existing `accounting.PgxAccountLocker` (`apps/control-plane/internal/accounting/pglock.go`), reused rather than reinvented. Unlike that locker, no in-process gate is needed here: the audit critical section never asks the pool for a second connection while holding the first, so there's no risk of a waiter starving the holder's own further pool use.

The retry loop (`maxSerializationRetries`) is left in place as defense-in-depth against a conflict from something else entirely on `public.audit_log`, not the writer's own concurrency, which this fix closes.

## Blast radius

`SyncWriter.pool` is a concrete `*pgxpool.Pool`, not an interface, so no caller or test mocks it. Every caller goes through `audit.Logger.Log(ctx, e)`, whose signature and error semantics are unchanged: `cmd/server/main.go` (construction), `internal/tenants/http.go`, `internal/identity/finalize.go`, `internal/signup/{personal_tenant,reconcile,webhook}.go`, `internal/byok/service.go`. None of them change.

**Latency**: net one extra round trip per write (the explicit `pg_advisory_unlock`, since the old lock released implicitly at `COMMIT`). Measured effect is noise: 10-way full concurrency completed in 130-200ms both before and after across the runs below; the difference is that after the fix all 10 succeed instead of roughly half of them burning their retry budget and still failing.

## Acceptance test: N-of-N, red then green

New test `TestSyncWriter_ConcurrentWrites_ExactlyNRows` (`apps/control-plane/internal/audit/sync_concurrency_test.go`) drives 10 fully-simultaneous `SyncWriter.Write` calls directly at the writer (no HTTP layer, no caller-level retry) and asserts **zero failures and exactly N rows**, not "at least one."

Run against a throwaway `pgvector/pgvector:pg17` Postgres, schema bootstrapped via `scripts/ci-throwaway-db.sh` with all 106 migrations applied.

**RED**, fix stashed out, unfixed `writeOnce`:
```
=== RUN   TestSyncWriter_ConcurrentWrites_ExactlyNRows
    sync_concurrency_test.go:66: write 1 failed: audit: insert: ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)
    sync_concurrency_test.go:66: write 3 failed: audit: insert: ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)
    sync_concurrency_test.go:66: write 4 failed: audit: insert: ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)
    sync_concurrency_test.go:66: write 6 failed: audit: insert: ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)
    sync_concurrency_test.go:66: write 8 failed: audit: insert: ERROR: could not serialize access due to read/write dependencies among transactions (SQLSTATE 40001)
    sync_concurrency_test.go:69: Error: Should be zero, but was 5
--- FAIL: TestSyncWriter_ConcurrentWrites_ExactlyNRows (0.18s)
FAIL
```

**GREEN**, fix restored, fresh throwaway database:
```
=== RUN   TestSyncWriter_ConcurrentWrites_ExactlyNRows
--- PASS: TestSyncWriter_ConcurrentWrites_ExactlyNRows (0.20s)
PASS
```
Reran 5x back to back (`-count=5`) against the same fixed writer: 5/5 clean passes.

The two existing tenants concurrency tests from #1181 (`TestSwitch_ConcurrentCallsFromActiveMember_AllSucceedConsistently`, `TestSwitch_ConcurrentCallsFromNonMember_AllRejectedNoneLeakThrough`) are tightened here from `GreaterOrEqual(count, 1)` / `LessOrEqual(count, n)` to `Equal(n, count)`, per the issue's own acceptance criterion, and both pass with the fix in place. Full `internal/audit` and `internal/tenants` suites pass (`go test ./... -v -count=1`), `go vet` and `gofmt -l` clean.

## Buglog entry

```json
{"date":"2026-08-25","tags":["audit","postgres","serializable","concurrency","compliance"],"error_message":"audit.SyncWriter.Write: SQLSTATE 40001 could not serialize access due to read/write dependencies among transactions","root_cause":"pg_advisory_xact_lock was the first statement inside a SERIALIZABLE transaction; Postgres takes the SERIALIZABLE snapshot at the first statement, not at BEGIN, so a writer blocked on that lock had already formed a stale snapshot before its turn, and SSI aborted the commit once the writer ahead of it committed a conflicting MAX(seq) read","fix":"acquire the lock as a session-scoped pg_advisory_lock on a dedicated connection BEFORE BeginTx, released with an explicit pg_advisory_unlock after commit, so the transaction cannot take its snapshot until this writer already holds exclusive possession of the lock key"}
```

## Test plan

- [x] `go build ./apps/control-plane/...`
- [x] `go vet ./apps/control-plane/internal/audit/... ./apps/control-plane/internal/tenants/...`
- [x] `gofmt -l` clean on touched files
- [x] New concurrency test: RED against unfixed writer (5/10 fail, SQLSTATE 40001), GREEN against fixed writer (10/10 pass, x5)
- [x] Tightened tenants concurrency tests pass with `Equal(n, count)`
- [x] Full `internal/audit` and `internal/tenants` suites pass, no regressions

### DIFF
--- a/apps/control-plane/internal/audit/sync.go
+++ b/apps/control-plane/internal/audit/sync.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgerrcode"
@@ -31,6 +32,54 @@ type SyncWriter struct {
 
 func NewSyncWriter(pool *pgxpool.Pool, cfg WriterConfig) *SyncWriter {
 	return &SyncWriter{pool: pool, cfg: cfg}
+}
+
+// MonthLockKey derives the audit-writer advisory lock key for ts's
+// absolute instant, bucketed by UTC calendar month. Exported so #1182's
+// regression test can pin known values without reaching into the
+// package, and so the derivation is a pure, directly-testable function
+// rather than inline arithmetic duplicated across the lock and unlock
+// call sites.
+//
+// ts is forced to UTC here regardless of the Location it arrives with.
+// writeOnce's ts is already UTC (canonical.TruncateTS calls .UTC()),
+// but this function does not trust that: it is the single source of
+// truth for "which month does this instant belong to", and it must
+// give the same answer for the same instant no matter what Location
+// happened to be attached to the caller's time.Time. That is also why
+// it is pure Go arithmetic rather than a `SELECT date_trunc('month',
+// $1::timestamptz)` round trip: a SQL date_trunc on a timestamptz
+// buckets by the session's TimeZone GUC, not by UTC, so the old
+// pg_advisory_xact_lock version of this file computed a DIFFERENT key
+// for the same event on any session whose TimeZone wasn't UTC. See
+// assertUTCSession below for the other half of closing that gap:
+// making today's "Supabase defaults to UTC" assumption true by
+// construction instead of true by luck (#1188 review thread).
+func MonthLockKey(ts time.Time) int64 {
+	ts = ts.UTC()
+	monthStart := time.Date(ts.Year(), ts.Month(), 1, 0, 0, 0, 0, time.UTC)
+	return monthStart.Unix()
+}
+
+// assertUTCSession fails loudly, instead of silently computing a
+// mismatched lock key, if this connection's session TimeZone GUC is
+// ever not UTC. Nothing in apps/control-plane/internal/platform/db/
+// pool.go sets TimeZone explicitly today, so it is UTC only because
+// Supabase's default happens to be UTC — unasserted, true by luck. Run
+// on every write (not cached past the first check): the risk this
+// guards is a config drift making the assumption false at some point
+// in this process's lifetime, not only at startup, and the added round
+// trip is one cheap query on a connection already checked out for the
+// advisory lock.
+func assertUTCSession(ctx context.Context, conn *pgxpool.Conn) error {
+	var tz string
+	if err := conn.QueryRow(ctx, `SHOW TimeZone`).Scan(&tz); err != nil {
+		return fmt.Errorf("audit: check session timezone: %w", err)
+	}
+	if !strings.EqualFold(tz, "UTC") && !strings.EqualFold(tz, "Etc/UTC") {
+		return fmt.Errorf("audit: session TimeZone is %q, not UTC; refusing to write because MonthLockKey is computed in UTC and would silently diverge from a SQL date_trunc-based reading of the same month under any other session timezone (#1188)", tz)
+	}
+	return nil
 }
 
 // Write inserts a single audit row and chains it off the last row in
@@ -96,16 +145,17 @@ func (w *SyncWriter) writeOnce(ctx context.Context, e Event, before, after []byt
 	ts := canonical.TruncateTS(time.Now())
 	monthStart := time.Date(ts.Year(), ts.Month(), 1, 0, 0, 0, 0, time.UTC)
 	monthEnd := monthStart.AddDate(0, 1, 0)
-	// Lock key is the month's epoch as bigint (not ::int, which is
-	// 32-bit and would silently overflow in January 2038), computed
-	// once here in Go and reused for both lock and unlock below.
-	lockKey := monthStart.Unix()
+	lockKey := MonthLockKey(ts)
 
 	conn, err := w.pool.Acquire(ctx)
 	if err != nil {
 		return fmt.Errorf("audit: acquire conn: %w", err)
 	}
 	defer conn.Release()
+
+	if err := assertUTCSession(ctx, conn); err != nil {
+		return err
+	}
 
 	// Session-scoped advisory lock, acquired BEFORE the SERIALIZABLE
 	// transaction opens, held until the explicit unlock below.
--- a/apps/control-plane/internal/audit/sync.go
+++ b/apps/control-plane/internal/audit/sync.go
@@ -54,7 +54,12 @@ func NewSyncWriter(pool *pgxpool.Pool, cfg WriterConfig) *SyncWriter {
 // with serialisation_failure (40001). We retry the entire transaction
 // up to maxSerializationRetries because each attempt is short and
 // retrying with a fresh `ts` keeps the row attributable to its true
-// commit time, not its first attempt.
+// commit time, not its first attempt. As of #1182 the advisory lock
+// is held from before BeginTx through the explicit unlock (see
+// writeOnce), so two SyncWriter.Write calls racing each other no
+// longer produce 40001 at all — the retry loop is defense-in-depth
+// against a conflict from something else entirely serializable on
+// public.audit_log, not the writer's own concurrency.
 func (w *SyncWriter) Write(ctx context.Context, e Event) error {
 	if e.Action == "" {
 		return errors.New("audit: action required")
@@ -84,32 +89,65 @@ func (w *SyncWriter) Write(ctx context.Context, e Event) error {
 }
 
 func (w *SyncWriter) writeOnce(ctx context.Context, e Event, before, after []byte) error {
-	tx, err := w.pool.BeginTx(ctx, pgx.TxOptions{IsoLevel: pgx.Serializable})
-	if err != nil {
-		return fmt.Errorf("audit: begin tx: %w", err)
-	}
-	defer func() { _ = tx.Rollback(ctx) }()
-
 	// Truncate to microseconds so the canonical hash, the canonical
 	// timestamp string, and the value Postgres stores are identical.
 	// Re-captured per attempt so a retried row carries its actual
 	// commit time, not the first attempt's stale time.
 	ts := canonical.TruncateTS(time.Now())
-
-	// Advisory lock keyed on the month's epoch as bigint. ::int was
-	// 32-bit and would silently overflow in January 2038; ::bigint is
-	// 64-bit and matches pg_advisory_xact_lock(bigint) so the lock
-	// namespace is unambiguous.
-	if _, err := tx.Exec(
-		ctx,
-		`SELECT pg_advisory_xact_lock(extract(epoch from date_trunc('month', $1::timestamptz))::bigint)`,
-		ts,
-	); err != nil {
-		return fmt.Errorf("audit: advisory lock: %w", err)
-	}
-
 	monthStart := time.Date(ts.Year(), ts.Month(), 1, 0, 0, 0, 0, time.UTC)
 	monthEnd := monthStart.AddDate(0, 1, 0)
+	// Lock key is the month's epoch as bigint (not ::int, which is
+	// 32-bit and would silently overflow in January 2038), computed
+	// once here in Go and reused for both lock and unlock below.
+	lockKey := monthStart.Unix()
+
+	conn, err := w.pool.Acquire(ctx)
+	if err != nil {
+		return fmt.Errorf("audit: acquire conn: %w", err)
+	}
+	defer conn.Release()
+
+	// Session-scoped advisory lock, acquired BEFORE the SERIALIZABLE
+	// transaction opens, held until the explicit unlock below.
+	//
+	// Postgres takes a SERIALIZABLE transaction's snapshot at the
+	// first statement inside it, not at BEGIN. The previous version of
+	// this writer took the lock with pg_advisory_xact_lock as that
+	// first statement: every concurrent writer's snapshot was formed
+	// the instant it started waiting on the lock, not the instant it
+	// actually got its turn, so each one committed against a MAX(seq)
+	// read that was already stale by the time it ran — the textbook
+	// SSI write-skew shape, and exactly why 6 of 10 fully-simultaneous
+	// writers failed with SQLSTATE 40001 (#1182). Acquiring the lock
+	// on the plain session first, before BEGIN, means the transaction
+	// below can only take its snapshot once this writer already has
+	// exclusive possession of the month, so nothing downstream of
+	// BEGIN can observe a stale one.
+	//
+	// Session- not transaction-scoped deliberately: it must survive
+	// past BeginTx to Commit. If the connection drops mid-hold,
+	// Postgres releases the session lock when the backend terminates,
+	// so the failure mode is one orphaned physical connection (which
+	// the pool detects and evicts on the next health check), never a
+	// permanently stuck lock. Same acquire/exec/defer-release shape as
+	// accounting.PgxAccountLocker (internal/accounting/pglock.go); no
+	// in-process gate is needed here the way that locker needs one,
+	// because this critical section never asks the pool for a second
+	// connection while holding the first.
+	if _, err := conn.Exec(ctx, `SELECT pg_advisory_lock($1)`, lockKey); err != nil {
+		return fmt.Errorf("audit: advisory lock: %w", err)
+	}
+	defer func() {
+		// Background context: unlock must run even if ctx was cancelled.
+		_, _ = conn.Exec(context.Background(), `SELECT pg_advisory_unlock($1)`, lockKey)
+	}()
+
+	tx, err := conn.BeginTx(ctx, pgx.TxOptions{IsoLevel: pgx.Serializable})
+	if err != nil {
+		return fmt.Errorf("audit: begin tx: %w", err)
+	}
+	defer func() { _ = tx.Rollback(ctx) }()
+
 	tenantArg := nullableUUID(e.TenantID)
 
 	var maxSeq int64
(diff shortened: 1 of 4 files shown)


apps/control-plane/internal/audit/sync.go
```diff
@@ -84,32 +89,65 @@ func (w *SyncWriter) Write(ctx context.Context, e Event) error {
...
 	// timestamp string, and the value Postgres stores are identical.
 	// Re-captured per attempt so a retried row carries its actual
 	// commit time, not the first attempt's stale time.
 	ts := canonical.TruncateTS(time.Now())
+	monthStart := time.Date(ts.Year(), ts.Month(), 1, 0, 0, 0, 0, time.UTC)
+	monthEnd := monthStart.AddDate(0, 1, 0)
+	// Lock key is the month's epoch as bigint (not ::int, which is
+	// 32-bit and would silently overflow in January 2038), computed
+	// once here in Go and reused for both lock and unlock below.
+	lockKey := monthStart.Unix()
```
sakibsadmanshajib: MEDIUM, non-blocking: lock-key parity across the rolling-deploy window.

`monthStart.Unix()` is correct only if `ts` (already `.UTC()`'d by `canonical.TruncateTS`) represents the same calendar month boundary the old SQL used to compute. The old code truncated with `date_trunc('month', $1::timestamptz)`, which truncates in the Postgres session's `TimeZone` GUC, not necessarily UTC. Nothing in `internal/platform/db/pool.go`'s DSN/pool construction pins the session `TimeZone` to `'UTC'` explicitly; correctness here depends on the Postgres server's default GUC actually being UTC.

In steady state this is moot since old and new code never run side by side. During a rolling deploy of this exact PR, though, an old-code pod and a new-code pod can be writing concurrently. If the server's default `TimeZone` were ever anything other than UTC, the two pods would compute *different* lock keys for the same wall-clock month and silently reintroduce the exact SSI write-skew this PR fixes, only for the deploy window, which would be very hard to notice or reproduce after the fact.

Suggest either: (a) an explicit `SET TIME ZONE 'UTC'` on the pool/session so this is asserted rather than assumed, or (b) a one-line unit test pinning `monthStart` for a known UTC instant so a future change to `canonical.TruncateTS` or this computation can't silently drift. Not blocking: Supabase's default DB timezone is UTC and this is a one-time transitional window, not a persistent risk.

apps/control-plane/internal/audit/sync_concurrency_test.go
```diff
@@ -0,0 +1,87 @@
...
+			t.Logf("write %d failed: %v", i, err)
+		}
+	}
+	require.Zero(t, failed, "%d of %d fully-simultaneous audit.SyncWriter.Write calls failed; N simultaneous writes must produce N rows (#1182)", failed, n)
+
+	var rowCount int
+	require.NoError(t, pool.QueryRow(ctx,
+		`SELECT count(*) FROM public.audit_log WHERE tenant_id=$1 AND actor_id=$2 AND action='CROSS_TENANT_ATTEMPT'`,
+		tenantID, actorID).Scan(&rowCount))
+	require.Equal(t, n, rowCount, "N fully-simultaneous writes must produce exactly N audit rows, not merely at-least-one")
```
sakibsadmanshajib: LOW, non-blocking, test-coverage suggestion: this asserts throughput/row-count correctness but never directly proves the session lock itself is released. Code-level review confirms the defer ordering releases it correctly (see summary comment), but a follow-up assertion here would catch a regression this test currently could not: e.g. `SELECT count(*) FROM pg_locks WHERE locktype='advisory'` back to 0 after `wg.Wait()`, or running the same 10-way burst twice in a row through the same pool and asserting the second batch isn't stalled. Optional hardening, not required for this PR.

sakibsadmanshajib (commented):
Independent review of the leak/deadlock risk this PR takes on by moving from a transaction-scoped to a session-scoped advisory lock. Full analysis in the summary comment below. Two non-blocking findings inline; everything else checked out.

sakibsadmanshajib:
## Independent verification: audit writer session lock, leak analysis

Read the pushed diff directly (`gh pr diff 1188` + full `sync.go` at the head commit), not the PR description. Traced every exit path out of the critical section in `writeOnce` and cross-checked the safety claims against pgx v5.7.2 source (the version actually pinned in `apps/control-plane/go.mod`), since this repo has a documented history of a shared-pool ceiling and of silent money/audit failures, and the two things a leaked session lock could produce, a hang or a stuck lock, are exactly the quiet failure modes that history warns about.

### Can the session lock leak: no

Traced each scenario from the brief against the actual code and against verified pgx internals:

1. **Panic between acquire and unlock.** Both `conn.Release()` and the unlock closure are registered via `defer` before any code that could panic runs. Go executes deferred functions during a panic unwind regardless of `recover`, so the unlock still fires. Confirmed by reading the function top to bottom, there is no code between a successful lock acquire and the `defer` registration that could throw.
2. **Context cancellation / client disconnect / deadline exceeded mid-transaction.** The unlock explicitly uses `context.Background()`, verified in the diff. The acquire itself uses the caller's `ctx`, and `pg_advisory_lock` is a blocking wait: if `ctx` cancels while waiting, Postgres aborts the wait and the lock is never granted, so nothing to leak. The narrower race (cancellation landing in the split second after the server has granted the lock but before the client processes the response) is real in the abstract for any `ctx`-bound query, but pgx v5.7.2's own pool `Release()` closes this gap: it checks `conn.IsClosed() || conn.PgConn().IsBusy() || conn.PgConn().TxStatus() != 'I'` and destroys (closes) the connection rather than pooling it whenever a query left the wire protocol in anything but a clean idle state. A connection close terminates the Postgres backend, which force-releases every session-level advisory lock it holds. Verified this check directly in `pgxpool/conn.go` in the exact pinned dependency version, this is not a guess.
3. **`conn.Release()` ordering against the deferred unlock.** This is the one that actually matters and it is correct. Registration order in the code is `defer conn.Release()`, then (after the lock is granted) `defer unlock`, then (after `BeginTx` succeeds) `defer tx.Rollback`. Go defers run LIFO, so actual execution order on any exit path is: `tx.Rollback` -> unlock -> `conn.Release()`. The unlock always runs on the connection while it is still checked out, never after it has returned to the pool. Confirmed by reading the registration order in the diff, this is not inferred from the comments.
4. **Pool exhaustion or acquire failure.** If `w.pool.Acquire(ctx)` itself fails, the function returns before any `defer` is registered at all (the `defer conn.Release()` line is only reached on the success path). Nothing is held.
5. **Second pooled connection while holding the first.** Read the entirety of `writeOnce`: every statement after the lock acquire (`BeginTx`, the `MAX(seq)` read, the insert, `Commit`, `Rollback`) runs on the same `conn`/`tx`, never on `w.pool` directly. This is unlike `accounting.PgxAccountLocker`, whose `fn` callback is arbitrary caller code that can need more connections (hence that locker's separate in-process gate). Also checked every caller that routes through `Logger.Log` (`tenants/http.go` Switch, `identity/finalize.go`, `signup/{personal_tenant,reconcile,webhook}.go`, `byok/service.go`): none of them hold a connection or open transaction on the same pool at the point they call `Audit.Log`, `tenants/http.go`'s Switch uses a single auto-released `Pool.Exec` before the audit call, not a held connection. No self-deadlock path found.
6. **`BeginTx` fails after the lock is taken.** The unlock `defer` is registered before `BeginTx` is even attempted, so it still fires on this path regardless of whether `BeginTx` succeeds.

The retry loop in `Write` calls `writeOnce` fresh each attempt; each call's defers (release/unlock/rollback) fully execute before the loop's next iteration starts, so no cross-attempt double-acquire or double-release is possible.

**Defer ordering verdict: correct**, matches the existing, already-proven `accounting.PgxAccountLocker` shape in this codebase.

### Lock key identity: mostly yes, one real caveat (see inline comment on `sync.go:102`)

The new Go computation is internally consistent (computed once, reused for both lock and unlock, an improvement on recomputing it twice via SQL). It reproduces the *old* SQL computation (`date_trunc('month', $1::timestamptz)`) only if the Postgres session's `TimeZone` GUC is UTC, since `date_trunc` on a `timestamptz` truncates in session timezone, not a fixed zone, while the new Go code hardcodes UTC. Nothing in `internal/platform/db/pool.go` pins the session timezone explicitly; this is an environmental assumption (true for Supabase's default, plausible but unasserted for the self-hosted target). This only matters for the duration of a rolling deploy where old- and new-code pods write concurrently; if it were ever wrong, the failure mode is silently reintroducing this exact bug for the deploy window. Flagged inline as MEDIUM, not blocking, since the assumption is very likely true today and this is a one-time transitional risk, not a persistent one.

### Other

- Confirmed `Logger.Log`'s signature and dispatch logic are untouched by this diff, the "seven callers need no changes" claim holds.
- The new concurrency test proves throughput/correctness (exactly N rows, zero failures) but does not directly assert the lock itself is released afterward. Flagged inline as a LOW-severity, optional hardening suggestion, not required given the code-level proof above.
- Noted but out of scope for this PR: `audit`'s per-month bigint keys and `accounting.PgxAccountLocker`'s `hashtext(accountID)::int8` keys share the same 64-bit `pg_advisory_lock` namespace with no prefix/tag separating them, a theoretical cross-subsystem collision risk that predates this change and is not introduced or worsened by it.

### Verdict

**Do not block.** The core fix is sound: the defer ordering is correct, the leak scenarios in the brief were traced individually and closed off, and the pgxpool destroy-on-non-idle-connection safety net that the PR's own reasoning leans on is real and verified against the exact pinned dependency version. One MEDIUM (timezone-parity assumption) and one LOW (test coverage gap) posted inline are worth addressing but neither is a correctness defect in this diff as written.

apps/control-plane/internal/audit/sync.go
sakibsadmanshajib: Fixed, not rebutted, in 76e0ea382.

Extracted the derivation into `MonthLockKey(ts)`, which calls `ts.UTC()` internally regardless of the caller's Location, so it cannot depend on session state by construction rather than by trusting that `ts` already arrived UTC. Added `assertUTCSession`, run on every write (not cached), which fails the write loudly with a clear error if the session's `TimeZone` GUC is ever not UTC. Added `TestMonthLockKey_PinnedValues` (`apps/control-plane/internal/audit/monthlockkey_test.go`), pinning a UTC month boundary plus two non-UTC-zoned timestamps (America/New_York, Asia/Tokyo) that land on the opposite side of a month boundary from their own local calendar date, proving the function buckets on the UTC instant, not the Location attached to it.

Went with your option (b) plus the runtime guard rather than (a) `SET TIME ZONE 'UTC'` in `pool.go`: that file is shared by every subsystem in control-plane, so pinning it there is a larger blast radius for a fix that's scoped to this one writer's key derivation. The runtime assertion converts "silent divergence" into "loud failure at first write," which is what actually stops the rolling-deploy scenario from going unnoticed.

apps/control-plane/internal/audit/sync_concurrency_test.go
sakibsadmanshajib: Fixed, not rebutted, in 76e0ea382. Added exactly the pg_locks assertion you suggested.

First real run of it against the fixed writer showed a nonzero count (1), which looked like a leak but wasn't: querying pg_locks directly on the same throwaway database immediately after showed zero locks held. Root cause was go test's default cross-package parallelism, not this writer. `go test ./apps/control-plane/internal/audit/... ./apps/control-plane/internal/tenants/...` runs each package's test binary concurrently, and internal/tenants' own Switch concurrency tests also drive real audit.SyncWriter writes against the same live Postgres, landing on the same month's lock key. The nonzero count was a sibling package's own legitimate in-flight write, still momentarily held at the instant this test's check ran, not a leak from this test's own writes (whose defers had already fully executed by the time wg.Wait() returned).

Changed the single check to a bounded 2s poll (50ms interval): a genuine leak never clears within that window; a neighboring package's momentary hold clears within one write's duration (tens of ms). Reran the full multi-package suite clean afterward (target test green, plus a 3x repeat to confirm no flakiness from the poll itself). Comment in the test now explains why a poll is needed rather than a single check, so the next person doesn't misread the same transient as a fix regression.

Issue #1174, opened 2026-08-25
Cache token counts never reach api_key_usage_rollups


sakibsadmanshajib:
Found while fixing the hardcoded-zero rollup gap in PR #1173.

`FinalizeReservation` writes per-key monthly and lifetime token breakdowns into `public.api_key_usage_rollups`. PR #1173 fixes input and output tokens, which were being passed as literal zeros while the real values sat unused two lines above. Cache read and cache write tokens are still written as zero, and unlike the input and output case this one cannot be fixed with a one line change.

**Why it is structural.** Neither finalize HTTP path carries cache token fields at all. `finalizeReservationRequest` and `internalFinalizeReservationRequest` in `apps/control-plane/internal/accounting/http.go` have no such fields, and `FinalizeReservationInput` has no field to receive them. The edge-api caller would also need to send them. So this is cross-service plumbing: request struct, input struct, caller, and the rollup write.

**Why it matters now.** As of PR #1157 Hive meters and prices cache read and cache write tokens correctly, and `usage_events` carries the real counts. The rollup is a second, independent accounting surface, and it is silently zero for exactly the token classes that dominate modern agent traffic. Anything later built on rollups, a per-key usage breakdown, an analytics view, an invoice line, would show a customer that they used no cached tokens while their bill reflects that they did.

**Blast radius today** is limited: no Go reader of `api_key_usage_rollups` exists in this repo yet, so nothing customer facing is currently wrong. This is worth closing before something reads it, not after.

**Related:** #1157 cache-aware pricing, #1173 the input and output half of this same defect. The pattern is the recurring one in this codebase, a value that exists and is correct at one layer being silently dropped at the next.

PR #1194, opened 2026-08-25
fix: reconcile usage_events.hive_credit_delta with the ledger (#1180)

### LINKED ISSUE #1180: usage_events.hive_credit_delta does not match credit_ledger_entries.credits_delta
Found while quantifying customer exposure for the DeepSeek cache-rate correction (#1176, PR #1178).

Reconstructing what a set of requests actually cost required picking a source. `credit_ledger_entries.credits_delta` is the ledger and is authoritative. `usage_events.hive_credit_delta` is the per-request usage record. **For the same requests, they do not agree.**

The quantification used the ledger and said so explicitly, so the #1178 figures stand. But two money surfaces disagreeing about the same request is its own defect and it is worse than either being wrong alone, because it means no single query answers "what did this cost" and any future analytics, invoice line, or customer-facing usage breakdown will silently pick one and be wrong in some cases.

**What this needs.**
1. Reproduce on a bounded set of requests: join `usage_events` to `credit_ledger_entries` on the reservation or request identifier and show rows where the two deltas differ. Establish whether the difference is a constant factor (a unit or rescale artifact, note credits were rescaled to 1 USD equals 1,000,000,000 per D-046), a rounding difference, or genuinely uncorrelated.
2. Establish which one is right, and whether the wrong one is merely informational or is read by anything.
3. Decide the contract deliberately: either they must agree and one is a bug, or `usage_events.hive_credit_delta` is explicitly an estimate and must be named and documented as such so nobody bills from it.

**Why now.** Cache-aware pricing (#1157) just made the charge math more complex, `api_key_usage_rollups` was writing hardcoded zeros until #1173, and #1174 records that cache tokens still never reach rollups at all. That is three separate accounting surfaces in varying states of disagreement. Reconciling them is prerequisite to trusting any usage or billing view built on top.

A related smaller observation from the same investigation: per-component reconstruction of a charge from `usage_events` token columns did not cleanly reproduce the real ledger amount under either input-token convention tried. That may be the same root cause or a second one.

Related: #1157, #1173, #1174, #1176, PR #1178.

sakibsadmanshajib:
## Summary

Fixes #1180. `usage_events.hive_credit_delta` and `credit_ledger_entries.credits_delta` disagreed for the same request. Investigation found two independent, confirmed-live root causes, not one:

**Root cause A (the primary defect, not a rescale/rounding artifact): a genuine duplicate write.** Every reservation-backed request writes TWO `usage_events` rows with `event_type='completed'` for the same attempt:
- Control-plane's own `accounting.finalizeLocked` (`apps/control-plane/internal/accounting/service.go`) writes the authoritative row during `FinalizeReservation`, in-process. `hive_credit_delta = -actualCredits`, the same figure charged to the ledger in the same call.
- edge-api's `recordCompletedEvent` (`apps/edge-api/internal/inference/orchestrator.go`, `stream.go`) then makes a separate, unconditional HTTP POST to `/internal/usage/events` that writes a second row for the identical attempt. Its `hive_credit_delta` is set to `usage.TotalTokens` — a raw token count, not a credit amount, not even negative.

Live measurement on the demo box (2026-08-25, full history, read-only bounded queries): 576 duplicate pairs, zero triples. 576/576 have the earlier row's delta <= 0 (ledger-matching); 575/576 have the later row's delta equal to `input_tokens + output_tokens` exactly. The ordering is deterministic by the code path, not a race: edge-api's POST only fires after the synchronous `FinalizeReservation` call returns.

**Root cause B (a real stale-unit bug, distinct from A): a missed backfill.** `usage_events.hive_credit_delta` was omitted from the column list in `20260823_40_credit_unit_rescale_billion.sql` (D-046, factor 10000). `credit_ledger_entries.credits_delta` was backfilled by that migration; `usage_events.hive_credit_delta` was not. Every authoritative row written before the rescale boundary is off from the ledger by exactly 10000x. Confirmed live (masked attempt id): `-12` in usage_events vs `-120000` in the ledger for the identical attempt, `-12 * 10000 = -120000` exactly.

**Which surface is authoritative:** `credit_ledger_entries`, confirmed from code (it's what actually posts against the account balance) rather than assumed. `usage_events`'s correct row is a mirror of the same `actualCredits` value, not an independent computation.

**Does anything bill from the wrong row:** no live billing/invoicing path reads `usage_events.hive_credit_delta`. `GetSpendSummary` and `GetUsageSummary.total_credits_spent` filter `WHERE hive_credit_delta < 0`, so the always-positive wrong duplicate was silently excluded from those two figures by accident of the sign filter — not corrupted, but `GetUsageSummary.request_count` (unfiltered `COUNT(*)`) was inflated up to 2x, and the raw `ListEvents` listing showed two contradictory rows per request.

## Fix

Chose to make the two surfaces agree (not merely document the estimate), since the authoritative value is cheaply recoverable:

1. `supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql`:
   - merges duplicate `'completed'` rows per attempt (keep the earlier/authoritative row, fold the later row's real token counts in, stamp merged ids for forensics)
   - backfills the missed rescale (root cause B) using the exact flag convention `20260823_40` already established (`credit_unit: legacy-1usd-100k-credits`)
   - adds a partial unique index `ux_usage_events_completed_attempt` so a future duplicate POST folds instead of inserting
2. `apps/control-plane/internal/usage/repository.go`: `RecordEvent`'s INSERT gained the matching `ON CONFLICT ... DO UPDATE`, deliberately never touching `hive_credit_delta`/`event_type`/`status`. edge-api is unchanged (out of this fix's scope) — its redundant POST now folds harmlessly.
3. New live guard test `apps/control-plane/internal/accounting/usage_ledger_reconciliation_live_test.go`: reproduces both writes against a real DB and asserts the surviving row's `hive_credit_delta` equals the ledger's `credits_delta`. Verified failing before the fix, passing after.
4. Two existing test fixtures (`repository_live_test.go`) inserted multiple `'completed'` events against one shared attempt id to simulate multiple requests — a shape that never occurs in production and that the new constraint correctly rejects. Fixed to seed one real attempt per event (what the tests were actually meant to exercise).

**Known residual gaps, out of scope (edge-api not in this task's allowlist), noted for the record:**
- edge-api never populates `FinalizeReservationInput.InputTokens/OutputTokens` despite #856 already wiring the field end to end; this fix's merge compensates without needing it fixed, but it's a one-line edge-api bug worth its own ticket.
- The `reservation.ID == ""` fallback path (no reservation ever created) still writes a lone wrong-value row with nothing to reconcile against, since nothing was charged.
- Relation to #1174 (cache tokens never reach `api_key_usage_rollups`): same family of defect, different table/path. The `usage_events` row this fix produces does carry cache tokens correctly; `api_key_usage_rollups` is the surface still missing them.

## Review round 2 (independent review, DO NOT BLOCK, confirmed by reproduction)

Reviewer confirmed the merge logic, backfill idempotency, and index safety by
reproducing rather than reading, and confirmed edge-api's write path only
logs `RecordUsageEvent` errors and never propagates them to the HTTP
response, so the migrate-then-deploy window is not customer-facing. One
finding required a fix before merge.

**Finding: the backfill under-applied.** The first version bounded the
rescale backfill on `usage_events.created_at < credit_unit_rescale.applied_at`.
20260823_40 documents a deploy-gap race of its own: the old binary keeps
writing old-unit rows for a window after that migration's COMMIT (until the
container recreate), so a straggler row can carry `created_at > applied_at`
while still holding a stale-unit value. A timestamp bound skips exactly
those rows, permanently — `usage_events` carries no positive new-unit stamp,
so nothing could ever later tell a missed straggler apart from a
legitimately small delta.

**Fix chosen: option 3, reconcile against the ledger.** `credit_ledger_entries`
was already established as authoritative and was correctly backfilled by
20260823_40. The rescale step now reconciles directly against it instead of
inferring the unit from a timestamp: for every surviving `'completed'` row,
if its `hive_credit_delta` disagrees with the matching
`credit_ledger_entries.credits_delta` (same account, same attempt,
`entry_type = 'usage_charge'`) by **exactly** a factor of 10000, the ledger's
value is copied onto it. A disagreement that is not exactly 10000x is left
untouched — that would be a different, unexplained mismatch this migration
has no evidence about, and "fixing" it would be a guess. This needed no new
column and no timestamp heuristic (option 1/2 both would have), uses truth
already established, and has no dependency on `created_at` or
`credit_unit_rescale.applied_at` at all, so it catches every stale-unit row
including deploy-gap stragglers a timestamp bound structurally cannot see.

Verified by reproduction, not just code-reading: seeded a fabricated
straggler (old-unit `usage_events` row, `created_at` 90s **after** the
rescale marker's `applied_at`, matching ledger entry in new units) on a
throwaway Postgres 17. Before: `hive_credit_delta = -72` vs ledger
`-720000`, a 10000x miss the old bound would never have touched (its
`created_at` is after `applied_at`). After running the revised migration:
`hive_credit_delta = -720000`, `internal_metadata` stamped
`{"credit_unit": "legacy-1usd-100k-credits", "ledger_reconciled_from": -72}`.
Re-ran the file a second time: `UPDATE 0` on the reconciliation step, value
unchanged — idempotent.

**Second finding: triple-row guard (accepted, cheap guard over a fix).** The
dedup merge is a many-to-one join; on an (unobserved, live evidence is 0
triples) 3+-row shape Postgres would pick an arbitrary source row for the
folded token columns. Added a step-0 check that raises and rolls back the
whole transaction if any attempt has 3+ `'completed'` rows, before any
mutation. Verified by reproduction: seeded a genuine triple, ran the
migration, got a loud `ERROR` naming the attempt count and the reason, and
confirmed via a fresh `count(*)` that all three rows were left untouched
(transaction rolled back, no partial state, index also not created).

**Rollback asymmetry (documented in the migration header, not fixed — no
action needed per reviewer):** this migration has no down-migration. Rolling
it back while control-plane's binary stays on the version that assumes
`ux_usage_events_completed_attempt` exists (the `ON CONFLICT` target in
`usage/repository.go`'s `RecordEvent`) breaks every `'completed'`
`usage_events` write — Postgres rejects an `ON CONFLICT` clause naming a
missing index. That fails loudly as errors, not as silent money corruption,
which is why it's acceptable, but a migration-only rollback must never be
attempted without rolling the binary back first.

## Test plan

- [x] `go build ./apps/control-plane/...` and `go vet` on touched packages: clean
- [x] `go test ./apps/control-plane/internal/accounting/... ./apps/control-plane/internal/usage/... ./apps/control-plane/internal/ledger/...` against real Postgres 17 (`scripts/ci-throwaway-db.sh`, 107/107 migrations applied): all green, including the new guard test
- [x] Guard test verified RED before the fix (fails on the unique-violation / wrong-merge), GREEN after
- [x] Migration applies cleanly against the full migration chain from scratch
- [x] Migration idempotency verified by literal replay against fabricated dirty data matching the real live shape (not just code inspection): first run merges + backfills, second run changes 0 rows
- [x] `gofmt -l` clean on every touched file
- [x] Deploy-gap straggler (old-unit row, created_at after the rescale marker) reproduced and confirmed corrected by the ledger reconciliation, not the old timestamp bound
- [x] Fabricated 3-row shape reproduced; migration raises and rolls back with zero rows mutated

### DIFF
--- a/apps/control-plane/internal/usage/repository.go
+++ b/apps/control-plane/internal/usage/repository.go
@@ -86,10 +86,36 @@ func (r *pgxRepository) RecordEvent(ctx context.Context, input RecordEventInput)
 		return UsageEvent{}, fmt.Errorf("usage: marshal customer tags: %w", err)
 	}
 
+	// ON CONFLICT targets ux_usage_events_completed_attempt (partial unique
+	// index on (account_id, request_attempt_id) WHERE event_type =
+	// 'completed', supabase/migrations/20260825_03_...sql). Only a
+	// 'completed' insert can ever hit this arbiter; every other event_type
+	// inserts exactly as before.
+	//
+	// Two writers record a 'completed' row for the same attempt today
+	// (issue #1180): control-plane's own accounting.finalizeLocked, which
+	// always runs FIRST and carries the ledger-matching hive_credit_delta,
+	// and edge-api's separate, unconditional POST that always runs SECOND
+	// and carries the real measured token counts but a hive_credit_delta
+	// that is not a credit figure at all (it is edge-api's TotalTokens,
+	// unrelated to this fix's scope). DO UPDATE folds the second write's
+	// token/provider columns onto the first write's row and deliberately
+	// never touches hive_credit_delta, event_type, status, or the account/
+	// request identifiers: the row that lands first is presumed
+	// authoritative on the money question, and every case measured live
+	// (576 of 576 duplicate pairs, 2026-08-25) bears that out.
 	row := r.pool.QueryRow(ctx, `
 		INSERT INTO public.usage_events
 			(account_id, request_attempt_id, api_key_id, request_id, event_type, endpoint, model_alias, status, input_tokens, output_tokens, cache_read_tokens, cache_write_tokens, hive_credit_delta, provider_request_id, internal_metadata, customer_tags, error_code, error_type)
 		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15::jsonb, $16::jsonb, $17, $18)
+		ON CONFLICT (account_id, request_attempt_id) WHERE event_type = 'completed'
+		DO UPDATE SET
+			input_tokens = EXCLUDED.input_tokens,
+			output_tokens = EXCLUDED.output_tokens,
+			cache_read_tokens = EXCLUDED.cache_read_tokens,
+			cache_write_tokens = EXCLUDED.cache_write_tokens,
+			provider_request_id = COALESCE(EXCLUDED.provider_request_id, public.usage_events.provider_request_id),
+			internal_metadata = public.usage_events.internal_metadata || EXCLUDED.internal_metadata
 		RETURNING id, account_id, request_attempt_id, api_key_id, request_id, event_type, endpoint, model_alias, status, input_tokens, output_tokens, cache_read_tokens, cache_write_tokens, hive_credit_delta, provider_request_id, internal_metadata, customer_tags, error_code, error_type, created_at
 	`, input.AccountID, input.RequestAttemptID, input.APIKeyID, input.RequestID, string(input.EventType), input.Endpoint, input.ModelAlias, input.Status, input.InputTokens, input.OutputTokens, input.CacheReadTokens, input.CacheWriteTokens, input.HiveCreditDelta, nullableString(input.ProviderRequestID), internalMetadata, customerTags, nullableString(input.ErrorCode), nullableString(input.ErrorType))
 
--- a/supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
+++ b/supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
@@ -1,0 +1,170 @@
+-- =============================================================================
+-- usage_events 'completed' dedup + credit-unit rescale backfill (issue #1180).
+--
+-- WHAT WAS FOUND
+--   Every request that goes through a reservation writes TWO usage_events
+--   rows with the same (account_id, request_attempt_id, event_type =
+--   'completed'):
+--     1. Written by control-plane's own accounting.finalizeLocked
+--        (apps/control-plane/internal/accounting/service.go), in-process,
+--        during FinalizeReservation. hive_credit_delta = -actualCredits, the
+--        exact figure charged to credit_ledger_entries in the same call.
+--        This row is authoritative -- it matches the ledger by construction.
+--     2. Written by edge-api's recordCompletedEvent (apps/edge-api/internal/
+--        inference/orchestrator.go and stream.go), a SEPARATE, unconditional
+--        HTTP POST to /internal/usage/events that always follows the
+--        FinalizeReservation call. hive_credit_delta there is set to
+--        usage.TotalTokens -- a raw token COUNT, not a credit amount, not
+--        even negative. It is not a rescale or rounding difference from the
+--        ledger figure, it is a categorically wrong value in that column.
+--   Live measurement on the demo box (2026-08-25, all history): 576 pairs of
+--   duplicate 'completed' rows, zero triples. In every single pair the
+--   earlier-written row's hive_credit_delta is <= 0 (matches the ledger
+--   pattern) and in 575 of 576 the later row's hive_credit_delta equals
+--   input_tokens + output_tokens exactly (the TotalTokens assignment). This
+--   is deterministic, not a race: edge-api only calls recordCompletedEvent
+--   AFTER the synchronous FinalizeReservation HTTP call returns, so the
+--   authoritative row is always inserted first.
+--
+--   A second, narrower defect compounds this: the authoritative row's own
+--   input_tokens/output_tokens are 0, because neither orchestrator.go's nor
+--   stream.go's FinalizeReservation call site populates
+--   FinalizeReservationInput.InputTokens/OutputTokens, even though the wire
+--   contract and control-plane's own plumbing already support it (#856).
+--   That gap is in apps/edge-api, out of this migration's reach; the dedup
+--   below compensates for it by keeping the token counts the second
+--   (edge-api) write actually carries while keeping the credit delta the
+--   first (ledger-backed) write actually carries.
+--
+--   Separately, and NOT the same root cause: usage_events.hive_credit_delta
+--   was missing from the WHAT IS SCALED list in
+--   20260823_40_credit_unit_rescale_billion.sql (D-046, factor 10000, 1 USD
+--   1,000,000 -> 1,000,000,000 credits). credit_ledger_entries.credits_delta
+--   was backfilled by that migration; usage_events.hive_credit_delta was
+--   not. Every authoritative row written before that migration's boundary is
+--   therefore off from its matching ledger entry by exactly 10000x (a stale
+--   unit, not a wrong rate) -- confirmed live: 018658cd.. reads -12 in
+--   usage_events against -120000 in the ledger for the identical attempt,
+--   and -12 * 10000 = -120000 exactly, same pattern on every pre-boundary
+--   row checked.
+--
+-- WHAT THIS FILE DOES (order matters: dedup first, THEN rescale, so the
+-- rescale sees each attempt exactly once)
+--   1. For every (account_id, request_attempt_id) with more than one
+--      event_type = 'completed' row, keep the earliest (the authoritative,
+--      ledger-matching one), fold the token/provider columns from the
+--      later row(s) into it via GREATEST/COALESCE (never destructive: the
+--      authoritative row's own columns are 0 in every case measured), stamp
+--      which duplicate id(s) were merged into internal_metadata for
+--      forensics, and delete the later row(s). hive_credit_delta and
+--      event_type on the keeper are untouched.
+--   2. Backfill the credit-unit rescale that 20260823_40 missed: multiply
+--      hive_credit_delta by 10000 for every usage_events row with a nonzero
+--      delta, created before the rescale marker's applied_at boundary, not
+--      already flagged. Flag key/value ('credit_unit',
+--      'legacy-1usd-100k-credits') is the exact one 20260823_40 already
+--      uses on credit_ledger_entries and credit_reservation_events, so a
+--      row is now identifiable as pre-rescale the same way across all three
+--      tables.
+--   3. Add a partial unique index so a future duplicate 'completed' POST
+--      (edge-api is unchanged by this fix; it will keep sending one) folds
+--      into the existing row instead of inserting a second one. This is the
+--      arbiter the application-side ON CONFLICT in
+--      apps/control-plane/internal/usage/repository.go targets.
+--
+-- WHY PLAIN CREATE INDEX, NOT CONCURRENTLY
+--   Supabase migrations run inside one transaction (BEGIN/COMMIT below),
+--   and CREATE INDEX CONCURRENTLY cannot run inside a transaction block, the
+--   same constraint 20260824_01 already documented for this database.
+--   usage_events is 2,173 rows / 1.4 MB on the demo box (measured
+--   2026-08-25); a plain index build here is milliseconds, not the
+--   ACCESS EXCLUSIVE hazard it would be on a table with real volume.
+--
+-- IDEMPOTENCY
+--   Step 1 is naturally idempotent: after the first run no (account_id,
+--   request_attempt_id) has more than one 'completed' row, so the dedup CTE
+--   selects nothing on replay.
+--   Step 2 guards on the same jsonb flag 20260823_40 already established
+--   (`NOT (internal_metadata ? 'credit_unit')`), so a replay scales nothing
+--   a prior run (or this run) already touched.
+--   Step 3 uses IF NOT EXISTS.
+-- =============================================================================
+
+BEGIN;
+
+SET LOCAL lock_timeout = '5s';
+
+-- ---------------------------------------------------------------------------
+-- Step 1: fold duplicate 'completed' rows into the earliest (authoritative)
+-- one per attempt.
+-- ---------------------------------------------------------------------------
+CREATE TEMP TABLE _usage_events_completed_ranked ON COMMIT DROP AS
+SELECT id, account_id, request_attempt_id, input_tokens, output_tokens,
+       cache_read_tokens, cache_write_tokens, provider_request_id,
+       row_number() OVER (
+         PARTITION BY account_id, request_attempt_id ORDER BY created_at ASC
+       ) AS rn
+FROM public.usage_events
+WHERE event_type = 'completed';
+
+CREATE TEMP TABLE _usage_events_completed_keepers ON COMMIT DROP AS
+SELECT * FROM _usage_events_completed_ranked WHERE rn = 1;
+
+CREATE TEMP TABLE _usage_events_completed_dupes ON COMMIT DROP AS
+SELECT * FROM _usage_events_completed_ranked WHERE rn > 1;
+
+UPDATE public.usage_events ue
+SET input_tokens        = GREATEST(ue.input_tokens, d.input_tokens),
+    output_tokens        = GREATEST(ue.output_tokens, d.output_tokens),
+    cache_read_tokens    = GREATEST(ue.cache_read_tokens, d.cache_read_tokens),
+    cache_write_tokens   = GREATEST(ue.cache_write_tokens, d.cache_write_tokens),
+    provider_request_id  = COALESCE(ue.provider_request_id, d.provider_request_id),
+    internal_metadata    = ue.internal_metadata || jsonb_build_object(
+        'merged_duplicate_completed_event_ids',
+        (SELECT jsonb_agg(dd.id) FROM _usage_events_completed_dupes dd
+          WHERE dd.account_id = ue.account_id
+            AND dd.request_attempt_id = ue.request_attempt_id)
+    )
+FROM _usage_events_completed_dupes d
+JOIN _usage_events_completed_keepers k
+  ON k.account_id = d.account_id AND k.request_attempt_id = d.request_attempt_id
+WHERE ue.id = k.id;
+
+DELETE FROM public.usage_events ue
+USING _usage_events_completed_dupes d
+WHERE ue.id = d.id;
+
+-- ---------------------------------------------------------------------------
+-- Step 2: rescale backfill 20260823_40 omitted for this table.
+-- ---------------------------------------------------------------------------
+DO $backfill$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM public.credit_unit_rescale) THEN
+        -- Rescale never ran on this database (e.g. a fresh CI throwaway
+        -- seeded straight from HEAD) -- nothing pre-dates a boundary that
+        -- does not exist, so there is nothing to backfill.
+        RETURN;
+    END IF;
+
+    UPDATE public.usage_events
+       SET hive_credit_delta = hive_credit_delta * 10000,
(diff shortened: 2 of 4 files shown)


supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
```diff
@@ -0,0 +1,170 @@
...
+    cache_read_tokens    = GREATEST(ue.cache_read_tokens, d.cache_read_tokens),
+    cache_write_tokens   = GREATEST(ue.cache_write_tokens, d.cache_write_tokens),
+    provider_request_id  = COALESCE(ue.provider_request_id, d.provider_request_id),
+    internal_metadata    = ue.internal_metadata || jsonb_build_object(
+        'merged_duplicate_completed_event_ids',
+        (SELECT jsonb_agg(dd.id) FROM _usage_events_completed_dupes dd
+          WHERE dd.account_id = ue.account_id
+            AND dd.request_attempt_id = ue.request_attempt_id)
+    )
+FROM _usage_events_completed_dupes d
```
sakibsadmanshajib: MEDIUM, confirmed by direct reproduction (not theoretical): this UPDATE has a many-to-one join. When a keeper has more than one duplicate row (three or more total completed rows for one attempt), `_usage_events_completed_dupes d` supplies multiple source rows for the same target row `ue`. PostgreSQL's documented behavior for `UPDATE ... FROM` with a many-to-one match is that exactly one of the matching source rows is used, and which one is unspecified.

I reproduced this against a throwaway Postgres 17 seeded with a synthetic triple (keeper credit delta -500, dupe A with tokens 100/20, dupe B with tokens 999/999). After running this file, the keeper's `input_tokens`/`output_tokens` came back as 100/20 (dupe A only). Dupe B's token counts were silently dropped, even though `merged_duplicate_completed_event_ids` correctly lists both dupe ids, so the forensic trail claims a merge that didn't fully happen.

`hive_credit_delta` (the money column) is untouched by this UPDATE, so this does not corrupt the ledger-matching figure. It does silently lose token-count data on any attempt with 3+ completed rows. The PR states zero triples exist in the live sample, so this is inert against today's data, but the script itself is not correct for that shape, and a data migration this consequential should not rely on "we checked, there are zero triples right now." Suggest aggregating across ALL dupes per keeper (e.g. `SELECT request_attempt_id, MAX(input_tokens) ..., array_agg(id) FROM _usage_events_completed_dupes GROUP BY account_id, request_attempt_id` joined into the UPDATE) rather than a plain multi-row JOIN.

supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
```diff
@@ -0,0 +1,170 @@
...
+        -- does not exist, so there is nothing to backfill.
+        RETURN;
+    END IF;
+
+    UPDATE public.usage_events
+       SET hive_credit_delta = hive_credit_delta * 10000,
+           internal_metadata = internal_metadata || jsonb_build_object(
+               'credit_unit', 'legacy-1usd-100k-credits')
+     WHERE hive_credit_delta <> 0
+       AND created_at < (SELECT applied_at FROM public.credit_unit_rescale)
```
sakibsadmanshajib: HIGH, confirmed by direct reproduction: this bound is the wrong discriminator for the exact race `20260823_40_credit_unit_rescale_billion.sql` documents about itself under "RESIDUAL RACE, STATED RATHER THAN HIDDEN". That migration explains that a row can commit AFTER its own `applied_at` marker but still be old-unit, because the previous binaries keep serving traffic until the containers recreate. Its own fix for that is the flag-only STRAGGLER DETECTOR query (no `created_at` bound at all), specifically because `created_at` cannot discriminate old-unit from new-unit rows in that window.

This backfill requires BOTH `NOT (internal_metadata ? 'credit_unit')` AND `created_at < applied_at`. Any old-unit `usage_events` row written in that same deploy-gap window (control-plane's OLD binary, still serving, writes an unscaled `hive_credit_delta` after `credit_unit_rescale.applied_at` but before the new binary is live) will have `created_at >= applied_at` and gets silently skipped here, permanently: after this migration commits, every future row (correctly-scaled, from the new binary) will ALSO lack the flag, since nothing in `repository.go`'s `RecordEvent` stamps `credit_unit` on write. There is no future query that can ever again distinguish this straggler from a legitimate new-unit row. I reproduced this: seeded a single `hive_credit_delta = -7` row dated after `applied_at` with no flag, ran this migration twice, and it is untouched both times, permanently off by 10000x (should read -70000).

This isn't hypothetical for THIS boundary specifically; `20260823_40` is the migration that put a real production box through exactly this gap two days ago. Worth confirming whether the demo box actually has a straggler `usage_events` row from that window (the sibling migration's own detector query, adapted to this table with just the flag check, would answer it) before deciding this is inert. Not raised in the PR's "Known residual gaps" section despite citing `20260823_40`'s convention directly.

supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
```diff
@@ -0,0 +1,170 @@
...
+
+-- ---------------------------------------------------------------------------
+-- Step 1: fold duplicate 'completed' rows into the earliest (authoritative)
+-- one per attempt.
+-- ---------------------------------------------------------------------------
+CREATE TEMP TABLE _usage_events_completed_ranked ON COMMIT DROP AS
+SELECT id, account_id, request_attempt_id, input_tokens, output_tokens,
+       cache_read_tokens, cache_write_tokens, provider_request_id,
+       row_number() OVER (
+         PARTITION BY account_id, request_attempt_id ORDER BY created_at ASC
```
sakibsadmanshajib: LOW, defensive-coding note. `ORDER BY created_at ASC` has no tiebreaker. If two completed rows for one attempt ever share an identical `created_at` (coarse clock resolution, or a future caller that batches inserts in one statement), `row_number()` picks a keeper nondeterministically, which could keep the wrong (edge-api-shaped) row as authoritative instead of the ledger-matching one. The PR's own evidence is that the two writers are always separated by a real HTTP round trip, so this is unlikely to fire on the current 576 pairs, but a one-line `, id ASC` tiebreaker removes the ambiguity entirely for free.

apps/control-plane/internal/usage/repository.go
```diff
@@ -86,10 +86,36 @@ func (r *pgxRepository) RecordEvent(ctx context.Context, input RecordEventInput)
...
+	// token/provider columns onto the first write's row and deliberately
+	// never touches hive_credit_delta, event_type, status, or the account/
+	// request identifiers: the row that lands first is presumed
+	// authoritative on the money question, and every case measured live
+	// (576 of 576 duplicate pairs, 2026-08-25) bears that out.
 	row := r.pool.QueryRow(ctx, `
 		INSERT INTO public.usage_events
 			(account_id, request_attempt_id, api_key_id, request_id, event_type, endpoint, model_alias, status, input_tokens, output_tokens, cache_read_tokens, cache_write_tokens, hive_credit_delta, provider_request_id, internal_metadata, customer_tags, error_code, error_type)
 		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15::jsonb, $16::jsonb, $17, $18)
+		ON CONFLICT (account_id, request_attempt_id) WHERE event_type = 'completed'
```
sakibsadmanshajib: MEDIUM, rollback asymmetry: worth stating explicitly since it is not symmetric between the two directions.

Code-only rollback (revert this file, keep the migration and the index): safe. The old plain INSERT just never hits the new unique arbiter unless a real duplicate appears, and even then it fails closed (logged, not propagated, per `recordCompletedEvent`'s existing error handling).

Migration-only rollback (drop `ux_usage_events_completed_attempt`, keep this code deployed): NOT safe. `ON CONFLICT (account_id, request_attempt_id) WHERE event_type = 'completed'` requires the named arbiter index to exist for Postgres to even plan the statement. Without it, Postgres raises `there is no unique or exclusion constraint matching the ON CONFLICT specification` on every single `'completed'`/`'reconciled'` `RecordEvent` call, not just duplicates, including `finalizeLocked`'s own authoritative first write. Tracing the caller: `finalizeLocked` (`apps/control-plane/internal/accounting/service.go`) has already posted the ledger charge and the reservation-finalize row update by the time it reaches `RecordEvent`, so this does not double-charge or lose money, but the function still returns an error, and the edge-api caller treats that as `finalize_failed` and fires a fallback `ReleaseReservation` against a reservation that is already `Finalized` in the DB, which itself will error and get logged. Net effect if this happens: no billing corruption, but a burst of finalize errors and a real gap in `usage_events` for every completed request until either the index is restored or this code is reverted too. Confirm the two changes are pinned to ship and roll back together, not independently.

apps/control-plane/internal/usage/repository.go
```diff
@@ -86,10 +86,36 @@ func (r *pgxRepository) RecordEvent(ctx context.Context, input RecordEventInput)
...
+		DO UPDATE SET
```
sakibsadmanshajib: Confirmed correct by direct reproduction against a live Postgres 17: ran this exact INSERT ... ON CONFLICT ... DO UPDATE shape against a pre-existing keeper row, it folded the incoming token counts, left `hive_credit_delta` untouched, and the row count for that attempt stayed at 1. Also confirmed edge-api's `recordCompletedEvent` and `recordErrorEvent` only log a `RecordUsageEvent` error, never propagate it into the HTTP response (`apps/edge-api/internal/inference/orchestrator.go`), so even the narrow migrate-then-deploy window where the old binary's plain INSERT would violate the new index (see the other comment on this line) cannot surface as a customer-facing 5xx.

sakibsadmanshajib:
## Independent review summary, PR #1194

Read the pushed diff directly (`gh pr diff 1194`), built and `go vet`'d the touched Go packages in the toolchain container (clean), applied the full 107-migration chain against a throwaway Postgres 17 (clean, matches the repo's own CI method), and reproduced the migration's dedup/rescale/idempotency behavior against seeded fixtures shaped like the claimed live data (a normal pair, a pair needing rescale, a synthetic triple, and a deploy-gap straggler). Findings below are evidence-based, not speculation, and are posted inline at the exact lines.

**Verdict: do not block, ship with two follow-ups tracked before or shortly after merge.**

The core diagnosis (two independent root causes, duplicate write plus missed rescale column) is correct and well-evidenced. The merge is safe on the pair shape that is actually live today (576/576, confirmed pattern), the rescale backfill cannot double-apply (confirmed empirically: replay changes 0 rows), and the new unique index does not break the live serving path (confirmed: `recordCompletedEvent`/`recordErrorEvent` only log a `RecordUsageEvent` error, never propagate it into the HTTP response, so even the narrow migrate-then-deploy window is not a customer-facing outage).

Two real, reproduced gaps, both non-blocking for money correctness but worth fixing:

1. **Dedup for 3+ duplicate rows silently drops data (MEDIUM).** `UPDATE ... FROM _usage_events_completed_dupes` is a many-to-one join; Postgres uses exactly one arbitrary matching source row when there are multiple. Reproduced: a synthetic triple folded only one dupe's token counts, dropping the other's, while the forensic `merged_duplicate_completed_event_ids` metadata correctly lists both ids, so the audit trail overstates what actually merged. `hive_credit_delta` (money) is untouched by this bug. Zero triples exist in the live sample per the PR, so this is inert today, but the script itself is wrong for that shape.

2. **Rescale backfill cannot catch deploy-gap stragglers (HIGH, but likely small blast radius).** The `created_at < applied_at` bound is the wrong discriminator for exactly the race `20260823_40` documents about itself: an old-unit row can commit after `applied_at` if the old binary is still serving. That migration's own fix is a flag-only detector with no `created_at` bound. This migration requires both the flag AND the boundary, so any `usage_events` straggler from that exact window (2026-08-23, real production event) is skipped, permanently and undetectably, since nothing stamps `credit_unit` on new-code writes to this table either. Reproduced: a seeded straggler row stayed wrong across two runs of the migration. Worth running the sibling migration's own flag-only detector query against `usage_events` before merge to see whether this is zero rows (like the dedup case) or a real number.

Also verified, not blocking:
- Rollback asymmetry: reverting only `repository.go` is safe; reverting only the migration (dropping the index) while the new code is deployed breaks the `ON CONFLICT` arbiter for every `'completed'`/`'reconciled'` write, including the authoritative one, though tracing `finalizeLocked` shows the ledger charge has already posted by that point, so this fails as errors and a `usage_events` gap, not as money corruption. State this if these two changes can ever land or roll back independently.
- The `ORDER BY created_at ASC` keeper-selection has no tiebreaker (low severity, cheap one-line fix: add `, id`).
- The residual gaps the PR already discloses (edge-api never populating `FinalizeReservationInput.InputTokens/OutputTokens`, the `reservation.ID == ""` fallback path) are correctly scoped out; neither blocks this fix.

Nothing here found a way this migration corrupts the ledger-matching `hive_credit_delta` value on any pair shape I could construct. The two gaps above cost token-count fidelity and audit completeness, not billing correctness.

supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
sakibsadmanshajib: Chose the cheap guard over handling the shape: added a step-0 check that RAISEs and rolls back the whole transaction if any attempt has 3+ completed rows, before the dedup UPDATE runs, rather than aggregating across all dupes. Reproduced against a throwaway Postgres 17: seeded a genuine triple, ran the migration, got a loud ERROR naming the attempt count, and confirmed via a fresh count(*) that all three rows were left untouched (transaction rolled back, index also not created since the error fires before COMMIT). If a triple ever exists it now stops the migration for a human to look at, rather than silently picking an arbitrary source row. Pushed in 669a1c0.

supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
sakibsadmanshajib: Confirmed and fixed. Reproduced the exact deploy-gap straggler you describe (old-unit hive_credit_delta, created_at after credit_unit_rescale.applied_at) on a throwaway Postgres 17: before the fix it was untouched, permanently off by 10000x, exactly as you found. Replaced the created_at bound entirely with reconciliation against credit_ledger_entries (already correctly backfilled by 20260823_40, and already the authoritative surface this migration's own header establishes): any surviving completed row whose hive_credit_delta disagrees with the matching ledger charge by exactly 10000x gets the ledger's value copied onto it, no timestamp involved at all. Re-verified the same seeded straggler: -72 corrected to -720000 (matching the ledger), flagged with ledger_reconciled_from for forensics, and a second run of the file changes 0 rows. Also added this to the PR's Known residual gaps section became a full writeup instead. Pushed in 669a1c0.

supabase/migrations/20260825_03_usage_events_completed_dedup_and_rescale_backfill.sql
sakibsadmanshajib: Done, one-line fix: ORDER BY created_at ASC, id ASC. Free determinism, no downside. Pushed in c901724.

apps/control-plane/internal/usage/repository.go
sakibsadmanshajib: Confirmed, and pinned: this migration and the ON CONFLICT change in usage/repository.go are one PR, ship and roll back together, not independently. Documented explicitly in the migration's header (ROLLBACK ASYMMETRY section) and in the PR body's review-round-2 addendum, including your point that finalizeLocked has already posted the ledger charge by the time RecordEvent runs, so a migration-only rollback produces finalize errors and a gap in usage_events, never money corruption or a double charge.

apps/control-plane/internal/usage/repository.go
sakibsadmanshajib: Thanks for reproducing this independently, that's the confirmation the guard-vs-fix call above rested on. Noted in the PR body's review-round-2 section along with the rest.

PR #1195, opened 2026-08-25
feat: show catalog prices in dollars, keeping credits where a bill is reconciled

sakibsadmanshajib:
## What changed

Catalog prices render as US dollars per million tokens instead of raw credit integers. `89,460,000` becomes `$0.0895`. Three surfaces move: the catalog list, the model detail page, and the analytics blended-price tile.

One shared formatter does it. `lib/format/model-pricing.ts` gains `formatUsdFromCredits` and a `unit` parameter on the existing functions, and the private twin of `formatModelPrice` that `components/catalog/model-catalog-table.tsx` was carrying (plus its cache-price sibling) is folded back into that module. The `ponytail:` comment deferring that fold asked for it once both branches then editing the table had landed. Both have.

## The product decision, stated so it is visible and reversible

A previous agent building the model detail page deliberately used credits and no USD, citing decision D-035 and calling a change a separate product decision. That caution was correct at the time. D-035 has since been revoked: the owner ruled on 2026-08-08 that the rule was never an actual regulatory requirement, and the decision record now says currency presentation to BD customers is an unconstrained product decision. The current goal is parity with the model pages this product is measured against, and those show dollars. The orchestrator made the call to display currency; this PR is where it is written down.

## Both units, deliberately

Dollars lead everywhere because that is the figure a customer comparing gateways reads.

Credits stay where a bill gets reconciled. They are the unit the ledger actually moves, and a dollar figure rounded for legibility cannot give the exact integer back. So the model detail page carries both, a `Price / 1M` column and a `Credits / 1M` column beside it, and the analytics tile keeps the credit figure in its derivation note. The catalog list stays currency-only: it is a dense eleven-row comparison table with four price columns, and a second number in every cell works against the one thing that table is for.

## Small denominations

This is the part that could have made the display worse than the integer it replaces. Two fixed decimals would print the corrected DeepSeek cache-read rate ($0.002982 per million) and the embedding alias input rate ($0.00001 per million) as `$0.00`, which reads as free.

Precision is therefore chosen per value: three significant digits, never fewer than two decimals, never more than nine. Nine is not a taste call. One credit is exactly 1e-9 USD, so nine decimals renders any integer credit rate as a non-zero figure, which makes `$0.00` unreachable for a non-zero price. A test asserts exactly that over every distinct non-zero rate the seeded catalog actually holds, read off a database with the full migration chain applied, with one credit prepended as the floor the rule has to hold at rather than the floor today's data happens to reach.

## The honesty invariant is unchanged

Still three distinct outcomes, now asserted in both units:

* a published rate, zero included, renders the number, because a published zero is a decision
* a null on an `upstream_actual` alias renders `Variable`
* a null on a `fixed` alias renders `Unknown`, never `Variable`, because a failed lookup is not a pricing model

The catalog table's separate cache-price rule (a fixed alias with no cache rate renders an em dash) moved across unchanged too. Changing the display unit changes how a number is drawn, never whether an absence is distinguishable from a zero.

## Verification

* `docker compose run --rm --build web-console npm run test:unit`: 720 passed. The one failing file is `tests/unit/ci-web-e2e-secret-free.test.ts`, which reads a workflow file the Dockerfile never copies in and fails the same way on `main`.
* `docker compose run --rm --build web-console npm run build`: green.
* `npm run lint:proof-tokens`: green.

## Visual proof

Before and after frames of all three surfaces, captured against a stack stood up locally from this branch with the real seeded catalog behind it, posted as a comment below. The capture log is committed at `docs/proof/console-currency-display-2026-08-25/capture-log.md` and records how the stack was started, which URLs were captured, and the empty console transcript.

No shared account's password was set, reset or rotated: the stack runs its own throwaway GoTrue and the account is a run-scoped fixture user destroyed with it.

### DIFF
--- a/apps/web-console/__tests__/console-cache-visibility.test.tsx
+++ b/apps/web-console/__tests__/console-cache-visibility.test.tsx
@@ -170,8 +170,11 @@ describe("model catalog cache pricing columns", () => {
 
     expect(screen.getByText("Cache read / 1M")).toBeDefined();
     expect(screen.getByText("Cache write / 1M")).toBeDefined();
-    expect(screen.getByText("1,200,000")).toBeDefined();
-    expect(screen.getByText("15,000,000")).toBeDefined();
+    // 1,200,000 and 15,000,000 credits per million tokens, rendered as the
+    // dollars-per-million figure the catalog now publishes. Both are well
+    // under a cent and neither may collapse to "$0.00".
+    expect(screen.getByText("$0.0012")).toBeDefined();
+    expect(screen.getByText("$0.015")).toBeDefined();
   });
 
   it("renders a dash, never a zero, when a fixed-price alias publishes no cache rate", () => {
--- a/apps/web-console/app/console/analytics/page.tsx
+++ b/apps/web-console/app/console/analytics/page.tsx
@@ -37,6 +37,7 @@ import {
 import { PageHeader } from "@/components/ui/page-header";
 import { cn } from "@/lib/cn";
 import { formatCredits, formatPercent } from "@/lib/format/credits";
+import { formatUsdFromCredits } from "@/lib/format/model-pricing";
 
 interface AnalyticsPageProps {
   searchParams: Promise<{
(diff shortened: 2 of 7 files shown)


sakibsadmanshajib:
## Visual proof

Before and after, same locally booted stack and same seeded catalog, working tree stashed back to origin/main for each before frame. Catalog list: raw credit integers become dollars per million, and the smallest real rates survive as $0.00001 and $0.00298 rather than collapsing to $0.00. Variable, the em dash and a published $0 all still render as themselves. Model detail: Price / 1M leads, Credits / 1M keeps the exact ledger integer. Analytics: blended price reads $0.11 with 110,192,307 credits in the note. The account email is masked in every frame.

PR #1222, opened 2026-08-26
fix: mint gateway-owned response ids, strip system_fingerprint, drop DeepSeek post-finish chunk

sakibsadmanshajib:
## Summary

Provider identity was leaking to customers on both `/v1/chat/completions` and `/v1/messages` (Anthropic-compat), streaming and non-streaming, on both OpenRouter-routed (DeepSeek) and Groq-routed aliases. An upstream response id format is provider-identifying by construction (`gen-*` = OpenRouter, `chatcmpl-*` with Groq's own suffix scheme = Groq), the same class of leak CLAUDE.md's provider-blind invariant forbids for error strings. Groq responses additionally carried a raw `system_fingerprint` field. This closes both leaks at every normalize boundary that builds a customer-facing response, and fixes a related DeepSeek stream defect in the same relay code.

Evidence: `/tmp/reports/parity-inhive.md` (finding "Wire cosmetics", live parity run 2026-08-26) plus a fresh live negative-control capture against `api-hive.scubed.co` taken during this fix (see Verification below), reproducing the exact defect moments before the fix.

## What leaked and where

- **`/v1/chat/completions` non-streaming** (`normalizeChatCompletion`): forwarded `resp.ID` verbatim.
- **`/v1/chat/completions` streaming** (`executeStreaming` SSE relay): every chunk carried the upstream's own id, not just the terminal frame; `system_fingerprint` was never stripped.
- **`/v1/completions` (legacy) non-streaming** (`normalizeCompletion`): same `resp.ID` passthrough.
- **`/v1/messages` (Anthropic-compat) non-streaming** (`FromOAIResponse`): prefixed the raw upstream id with `msg_` instead of minting a fresh one, still shipping the upstream id inside the prefix.
- **`/v1/messages` streaming** (`SSETranslator.FeedLine`): same prefix-not-mint bug for `message_start.id`, reused (correctly) for the rest of the stream but wrong at the source.
- **DeepSeek-family streams** (both surfaces relay through the same OpenAI-shaped chunk): OpenRouter emits one extra empty `role`/`content` chunk immediately after the real `finish_reason` frame, before `[DONE]`. A strict SSE client that already closed the message on the real finish frame can choke on it.

## Fix

- New `mintCompletionID(prefix)` (`apps/edge-api/internal/inference/mint_id.go`) mints a gateway-owned id (`chatcmpl-<uuid>` / `cmpl-<uuid>`), used by every OpenAI-shaped normalize site. `idPrefixForEndpoint` picks the right prefix per endpoint so a stream's minted id matches its non-streaming twin's shape.
- `normalizeChatCompletion` and `normalizeCompletion` mint a fresh id and strip `system_fingerprint`; the original upstream id is kept in a local variable purely for the existing usage-clamp log line, never for anything client-facing.
- `executeStreaming`'s relay loop mints **once per stream** and reuses that same id on every chunk, including the synthesized terminal usage chunk, so a client-visible id is stable across the whole response (matches its non-streaming twin's shape too).
- `FromOAIResponse` and `SSETranslator.FeedLine` (Anthropic-compat) now always mint a fresh `msg_<uuid>`, never derive from the upstream chunk/response id.
- `SanitizeVariablePriceFrame` (the map-based fallback sanitizer, shared by the streaming relay's rare unparseable-chunk path and `apps/edge-api/internal/chat`'s own verbatim relay) now also drops `id` and `system_fingerprint` as defense in depth, since it has no per-stream state to mint a stable replacement from.
- New `shouldSuppressPostFinishChunk` / `chunkFinished` (pure, unit-tested) gate the DeepSeek post-finish chunk: any chunk arriving after a relayed `finish_reason` is dropped from the wire unless it carries `usage` (the legitimate `stream_options.include_usage` terminal frame, which always forwards). The suppressed chunk's content is still folded into the usage accumulator first, so billing never silently loses data, only the wire write is skipped.

## What this does NOT change (verified explicitly)

1. **Internal correlation.** Nothing reads `resp.ID` / `chunk.ID` for request or attempt correlation. That keys entirely on `attempt.ID`, a UUID this gateway mints independently at dispatch time (`orchestrator.go`/`stream.go` `StartAttempt`), completely decoupled from any upstream or client-facing response id, before and after this change. Confirmed by reading every caller of `RequestAttemptID`/`AttemptID` across `edge-api` and `control-plane`; none of them touch the response body's `id` field.
2. **Billing/settlement.** Settlement (`settleStream`, `ChargeUsage`, reservation release) also keys on `attempt.ID`/`reservation.ID`, never on the response id. The usage-clamp log line is the only place the *old* upstream id was ever read after this change, and it still gets it (via a locally kept `upstreamID`/`upstreamChunkID` variable), so diagnostic log fidelity is unchanged.
3. **Streaming usage semantics.** `NormalizeCacheUsage`'s wire-shape selection (`usage.CacheReadInputTokens != nil`) is untouched; no model-family branch was added anywhere in this diff.
4. **Id stability.** A minted id is generated once per response/stream and reused on every chunk (`mintedID` in `executeStreaming`, `t.messageID` in `SSETranslator`), and the OpenAI-shaped id prefix matches its endpoint's non-streaming shape.

## Verification

**Unit tests** (new `mint_id_test.go`, updated `translate_response_test.go`/`stream_test.go` in `internal/anthropic`, updated `upstream_cost_hardening_test.go`):
- Non-streaming mint + strip for both an OpenRouter `gen-*` fixture and a Groq `chatcmpl-*` + `system_fingerprint` fixture, asserting the minted id never equals the upstream id and the wire bytes contain neither the upstream id string nor `system_fingerprint`.
- Legacy `/v1/completions` mint.
- Anthropic non-streaming and streaming: minted id always starts with `msg_`, never equals or contains the upstream id, even when the upstream id already looks `msg_`-shaped.
- `shouldSuppressPostFinishChunk`/`chunkFinished` against the **exact byte-for-byte chunk sequence captured live** on `api-hive.scubed.co` (see below): the real finish chunk forwards, the spurious trailing chunk suppresses, a genuine usage-only terminal chunk still forwards.
- Full `go test ./apps/edge-api/...` passes (all packages green), run from this worktree's own `deploy/docker` (not the shared checkout, to avoid a stale-image false green).

**Live negative control**, captured against the deployed (pre-fix) box moments before this fix, using the existing parity test key (`api_keys.id c4bc000e-3861-4941-8259-68714c1b6d47`, no rotation):
- Non-stream `deepseek-v4-flash`: `"id":"gen-1787734315-Jo86gvcpmVkDttYHFhKc"` forwarded verbatim.
- Non-stream `hive-free`: `"id":"gen-1787734317-qUFgWPoJsd4J75lctb2B"` forwarded verbatim (this pool member happened to route through a `gen-*` upstream on this call).
- Stream `deepseek-v4-flash`: every one of 5 chunks carries the identical upstream id `gen-1787734329-THDkSN71nG5F8uAbPdnB`, AND the exact spurious post-finish chunk reproduces: `{"delta":{"role":"assistant","content":""},"finish_reason":null}` arrives immediately after the `"finish_reason":"length"` chunk, before `[DONE]`.

This confirms the defect is real and live, and that the fix (verified in isolation by the unit tests above, against these same captured bytes) targets the actual bug.

**What I could not do as builder**: deploy is orchestrator-only (this repo's coding pipeline, stage 10) since a push to `main` auto-deploys the live box. I have not verified the fixed code against the live box post-merge; that is the orchestrator's job at the deploy-confirmation step. The unit tests above exercise the real `mintCompletionID`/`shouldSuppressPostFinishChunk`/`chunkFinished`/`FromOAIResponse`/`SSETranslator` functions directly (not a reimplementation), including against the live-captured byte sequence, so I'm confident in the fix, but I want that boundary stated plainly rather than implied.

## Known residual (found, not fixed here — flagging per project convention rather than burying it)

`apps/edge-api/internal/chat/dispatch.go` (the Open WebUI session-chat relay, a separate customer-facing surface from the OpenAI-compat `/v1/chat/completions` this PR fixes) has the same class of leak and is **not touched by this PR**:
- For a variable-price alias, it already runs every frame through `SanitizeVariablePriceFrame`, so this PR's `id`/`system_fingerprint` deletion there also happens to close the leak on that path as a side effect.
- For a **fixed-price alias** (e.g. Groq/`hive-free`), the relay is `_, _ = w.Write(line)` — genuinely verbatim, no sanitization of any kind, not even the model field. This is a live leak on the primary chat UI surface today.

I did not touch `internal/chat` in this PR: it's a different package with its own settlement logic and no verification plan was scoped for it here, and I did not want to expand a security-sensitive diff's blast radius past what was asked and verified. Recommend a dedicated follow-up PR/issue scoped to that file specifically.

## Buglog entry

See the commit message on this branch's `HEAD` for the full JSON line (heading "Buglog entry"), to be appended to `.wolf/buglog.jsonl` on `main` per this repo's convention (never on a feature branch).

---

## Update: security review round 2 (blocking findings addressed, rebased onto main)

Rebased onto `origin/main` (27a4b31c5, twelve PRs merged while this branch was open, including #1220's streaming fail-closed settlement fix, #1221 cache affinity, #1225 free-pool headroom). Clean merge, no conflicts; full `edge-api` and `control-plane` suites green post-merge before any of the fixes below were applied.

### Blocking finding 1: `stream.go`'s second verbatim fallback — FIXED

When a chunk failed typed decode AND the route was fixed-price, the raw upstream line shipped unsanitized (id, `system_fingerprint`, everything), with no log line. Fixed by collapsing both pricing branches onto the one map-based sanitizer (`SanitizeVariablePriceFrame`) the variable-price case already used: mint the id, strip `system_fingerprint`, drop-and-log anything unparseable, for every route. Verified by full diff review (single call site now, not two) plus the existing `SanitizeVariablePriceFrame` unit test suite, which covers exactly this shape. I did not build a new end-to-end test reaching this specific fallback inside `executeStreaming`: no test in this codebase currently drives that function end to end (confirmed by grep), and building that harness from scratch is a disproportionate lift for this fix; the shared sanitizer this fallback now unconditionally calls is what's under test.

### Blocking finding 2: `apps/edge-api/internal/chat/dispatch.go` fixed-price branch — FIXED

Every SSE line was written raw with zero sanitization on the fixed-price branch (D-032 norm, so most aliases, the primary Open WebUI chat surface). Every data frame now runs through the same sanitizer, fixed-price included, with a per-stream minted id threaded through so a client sees one stable id across the whole stream. New test `TestDispatchFixedPriceStreamSanitizesUpstreamID` exercises the real `dispatch.ServeHTTP` end to end with a `FixedPricing` route and a fixture carrying `id`+`system_fingerprint` in the exact shape the live leak had.

**Negative control, both findings**: stashed only the fix source files (kept the new/updated tests in place) and re-ran them against the pre-fix code. Both failed exactly as expected, reproducing the leak verbatim (`chatcmpl-8f3a9c2e1b4d` + `system_fingerprint` leaked on the dispatch.go test; the RAG test below leaked `up-1` with no minted id at all). Restored the fix, re-ran, both pass.

### Third leak, found while auditing per the review's instruction to assume one exists

`apps/edge-api/internal/rag/chat_handler.go`'s `streamGroundedChat` (the `/v1/rag/chat` streaming path) relays chunks through a generic `map[string]any`, not a typed struct. Its own comments already state a "provider-blind" design intent, and it does drop provider names and `event:` lines, but nothing stripped `id` or `system_fingerprint` from the surviving map keys, so both leaked on every chunk. Fixed with the same mint-once-per-stream, strip-system_fingerprint treatment. Strengthened the existing `TestHandleChat_StreamingRelaysCitationsAndChunks` test (its fixture already carried `"id":"up-1"` on every chunk, but nothing asserted against it) to check the raw upstream id never appears and the same minted id repeats across both chunks. Same stash-based negative control applied and confirmed (see above).

### Full enumeration of every response-emitting path (as requested, so the next reader doesn't have to re-derive it)

| Path | Sanitization status |
|---|---|
| `/v1/chat/completions` non-stream (`normalizeChatCompletion`) | SANITIZED — mint id, strip `system_fingerprint` (commit 1) |
| `/v1/chat/completions` stream, typed-decode success (`executeStreaming` primary relay) | SANITIZED — mint once per stream, strip per chunk, post-finish suppression (commit 1, tightened per CodeRabbit) |
| `/v1/chat/completions` stream, typed-decode FAILURE fallback | SANITIZED — unified onto the same sanitizer as above (commit 2, blocking finding 1) |
| `/v1/chat/completions` stream, synthesized terminal usage chunk (upstream never sent usage) | SAFE by construction — gateway-built, never touches upstream bytes; id field updated for stability only, pre-existing logic otherwise untouched |
| Legacy `/v1/completions` non-stream (`normalizeCompletion`) | SANITIZED — mint id (commit 1) |
| Legacy `/v1/completions` stream | Same code path as chat completions stream (shared `executeStreaming`) — SANITIZED |
| `/v1/responses` (OpenAI Responses API) | ALREADY SAFE pre-PR — mints its own `resp_` id independently, confirmed clean by security review, untouched |
| `/v1/messages` (Anthropic-compat) non-stream (`FromOAIResponse`) | SANITIZED — always mints fresh `msg_` id, old code only prefixed the upstream id (commit 1) |
| `/v1/messages` (Anthropic-compat) stream (`SSETranslator`) | SANITIZED — same fix, never relays raw bytes at all by construction (commit 1) |
| `/v1/rag/chat` non-stream | ALREADY SAFE pre-PR — mints its own `ragchat-` id independently (existing code, confirmed while auditing) |
| `/v1/rag/chat` stream (`streamGroundedChat`) | SANITIZED — third leak found this round, fixed (commit 2) |
| OWUI session chat (JWT-authenticated `/v1/chat/completions`), variable-price branch | SANITIZED — already routed through the shared sanitizer, now with a stable minted id (commit 1) |
| OWUI session chat, fixed-price branch | SANITIZED — blocking finding 2, fixed this round (commit 2) |
| Non-data SSE lines (blank, `event:`, `[DONE]`) in any relay | SAFE by construction — no id/model/cost/provider fields on these lines |
| `/v1/embeddings`, audio transcription/translation, image generation | SAFE by construction — none of these response types declare an `id` or `system_fingerprint` field at all (checked each typed struct) |
| `/v1/batches` object metadata (`Batch.ID`) | SAFE — control-plane's own minted internal id, never an upstream identity (edge-api's batches client talks to control-plane, never to a provider directly) |
| `apps/control-plane/internal/batchstore` local batch executor output file (`OutputLine.Response.Body`) | **NOT SANITIZED, known and disclosed, not fixed in this PR** — raw upstream JSON forwarded verbatim into the customer-downloadable batch output. Different service (control-plane, not edge-api), different settlement/output path, no verification plan scoped here. Recommend a dedicated follow-up PR/issue; flagged loudly rather than silently left out of this enumeration. |

### Settlement/billing verification (three specific requirements from the #1220 author)

1. **`HasForwardedChunk` on every forwarded frame, including the new sanitized fallbacks.** Verified by reading every `accumulator.HasForwardedChunk = true` call site in the diff: it is set immediately before every actual client write (primary typed-relay success, the newly-unified fallback's sanitized-success branch), and NOT set on any drop. No site was removed; the only new site is on the fallback's success path, paired 1:1 with its write.
2. **Missing-usage capture path stays `Held()`-full, plus alarm, plus completed status; the minted path cannot skip it.** `accumulator.Accumulate(chunk, aliasID)` and the `ClampUsage`/`RawUsageChunk` capture run unconditionally, BEFORE the post-finish suppression check (`if suppressPostFinish { continue }`). A suppressed (non-forwarded) chunk still fully feeds the settlement accumulator; suppression only skips the client-facing write. `shouldSuppressPostFinishChunk` also explicitly exempts any real usage-only terminal frame (usage set, zero choices) from suppression at all, so the actual terminal usage chunk is never touched by this logic in the first place.
3. **Synthesized terminal usage chunk — not new, not double-billing.** The `includeUsage && !accumulator.HasUsage` synthesis block predates every commit in this PR (present before commit 1). This PR's only change to it is the `ID` field (now reuses the stream's `mintedID` for id-stability rather than a fresh random uuid). The condition that triggers it, and the usage numbers it carries (`accumulator.ToUsageResponse()`), are untouched. This PR touches no settlement arithmetic anywhere: every hunk in `stream.go`'s diff sits inside the SSE relay loop or this synth-chunk block, none inside `settleStream`/`settlementCredits` (confirmed by `git diff origin/main...HEAD -- apps/edge-api/internal/inference/stream.go` hunk ranges). The synthesized chunk is written to the client-facing wire only; settlement's own usage numbers come from the `accumulator` fields directly, never by re-reading what was written to the client, so there is no read-back path for a future consumer (#1226) to double-count.


sakibsadmanshajib:
## Live verification against the deployed box (2026-08-28)

Ran against `https://api-hive.scubed.co`, both catalogue provider families (DeepSeek via `deepseek-v4-flash`/`deepseek-v4-pro`, Groq via `hive-small`), using a freshly minted `hk_` key on the QA tester workspace (revoked immediately after).

**Running code confirmed.** `origin/main` HEAD is `c177972f4` (#1237, tests-only, no source changes). The fix itself is `483ba7983` (#1222), which the box's `Pull + recreate stack` deploy job already confirmed running (container recreated/healthy, box HEAD == run commit). #1237 needs no redeploy since it added no source changes.

### Results by path (both families unless noted)

**`/v1/chat/completions` non-streaming and streaming**
Ids are gateway-minted `chatcmpl-<uuid>` on every chunk, no `system_fingerprint` anywhere, `model` field always echoes the requested alias, streams end with exactly one `finish_reason` chunk followed by `[DONE]` (checked full chunk sequence, not just the terminal frame). No leftover chunk after `finish_reason` on either family, including `deepseek-v4-flash` (the family with the post-finish-chunk bug).

**Legacy `/v1/completions` non-streaming and streaming**
Same shape: `cmpl-<uuid>` gateway-minted ids, no `system_fingerprint`, clean stream termination.

**`/v1/messages` (Anthropic-shaped) non-streaming and streaming**
Ids are full-UUID `msg_<uuid>` (previously just `msg_` + upstream id verbatim — that's exactly the bug this closes). No `system_fingerprint`. SSE stream is `message_start` → `ping` → `message_delta` → `message_stop`, no leak in any event.

**Error paths (chat/completions, both families)**
- Unknown model: `404 model_not_found`, generic message, no provider name.
- Out-of-range param (`temperature: 99`): `400 upstream_error`, message is `"<alias> is not available."` / `"requested model is not available."` — provider-blind per the existing convention, no upstream name/fingerprint.
- Malformed message array (missing `role`): tolerated, handled normally, clean response.
- Oversized prompt (~60k words) + stream + mid-request abort (`curl --max-time 8`): completed cleanly inside the window, clean gateway-minted id, no fingerprint; a follow-up request immediately after confirmed the service was still healthy (no crash/hang from the abort).
- Did not manage to trigger a genuine `context_length_exceeded` rejection within the test budget — 60k words fit inside these models' context. Not treating this as covered; flagging as untested rather than claiming a pass.

**`/v1/rag/chat` streaming — BLOCKED, not a leak.** This route is gated JWT-session-only by design (`apps/edge-api/cmd/server/main.go` comment near the RAG/Cowork block: `hk_`-authenticated requests are denied unconditionally, matching the voice-gate comment on the same file). Testing with a JWT pulled from the signed-in QA tester session still gets `403 ACCESS_DENIED` — confirmed via `/v1/featuregate` that the QA tester tenant has `ENABLE_RAG: false`. This is a tenant feature flag, unrelated to #1222/#1237. Could not exercise this path without either that flag flipped or a different RAG-enabled tenant/key.

### Verdict
No leaks found on any of the three fully-testable paths, sync and streaming, both provider families, happy path and error paths. `/v1/rag/chat` is the one path not yet verified live, blocked by the tenant flag above rather than by anything in this PR.

sakibsadmanshajib:
## Rejection-path follow-up (2026-08-28, same box/commit as prior comment)

Closed the one gap flagged in the previous comment: a genuine upstream rejection, plus auth and rate-limit shapes. Same key-mint/revoke discipline as before.

**`context_length_exceeded`-class rejection.** A 60k-word prompt (the earlier test) turned out not to exceed either family's actual window — confirmed via `usage.prompt_tokens: 200013` on a 200k-word prompt that still returned 200 OK on `hive-small`. Pushed to a 700k-word prompt (~700k tokens) to force a real rejection:

- `hive-small` (Groq): `400 { "error": { "message": "hive-small is not available.", "type": "api_error", "code": "upstream_error" } }`. Headers clean (`x-request-id: req-<uuid>`, gateway-shaped, no upstream id).
- `deepseek-v4-pro`: same shape, `400 api_error / upstream_error`, generic message.
- `deepseek-v4-flash`: came back `429 { "error": { "message": "requested model is not available.", "type": "rate_limit_error", "code": "upstream_rate_limited" } }` — this is presumably OpenRouter's own TPM limit tripping on the oversized payload rather than a context-window rejection specifically, but it's a genuine upstream 4xx passed through the same sanitizer, so it doubles as the rate-limit-shaped rejection asked for below.

No `system_fingerprint`, no provider name, no upstream-format id, no upstream model string on any of the three. All three collapse to the same generic `"<alias> is not available."` / `"requested model is not available."` shape regardless of the real upstream cause, which is a structural argument, not just an empirical one, that provider-identifying detail from the upstream error body cannot reach the client: every upstream 4xx gets flattened to this fixed shape before serialization.

One side note, not a leak: the same 700k-word body against `/v1/messages` (Anthropic-shaped) got `400 {"type":"error","error":{"type":"invalid_request_error","message":"invalid JSON body"}}` instead — looks like a gateway-side body-size/parse limit on that path rather than reaching the upstream at all. Still clean, just a different rejection point; flagging in case it's worth a deliberate size limit rather than an incidental one.

**Auth failures (bad key, no header, malformed scheme).** All three return in 0-3ms (never reach a provider), all `401`, generic messages (`"Incorrect API key provided."` / `"missing bearer"`), no leak surface possible since these never route to an upstream.

### Verdict (combined with the prior comment)
Zero leaks found across `/v1/chat/completions`, legacy `/v1/completions`, `/v1/messages` — sync, streaming, every chunk, happy path, bad-parameter errors, oversized-context rejection, rate-limit rejection, and auth failure — on both DeepSeek and Groq. `/v1/rag/chat` remains the only untested path, assigned separately (`ragenable`) since it needs `ENABLE_RAG` flipped on a real tenant first.

